PR #153, opened 2018-11-10
Make sure the benchmarks run on all OSes

### LINKED ISSUE #102: Make sure all benchmarks runs on non-Windows OSes too
Follow up of https://github.com/dotnet/performance/pull/100#issuecomment-410770218

After I am done with the port I will run all of the Benchmarks on MacOS and Ubuntu and fix the issues if there are any.

/cc @jorive

adamsitnik:
I tried to run all the benchmarks we have on MacOS and Ubuntu for 2.1 vs 2.2.

On Ubuntu it failed with too many files opened exception for the Console tests. This is why I have changed the benchmark to also dispose the stream. Previously I was hoping that single iteration will create not so many finalizable streams  and GC.Collect x2 with WaitForPendingFinalizers that we do in BDN after every iteration will clean them up. I was wrong ;)

The other fix is trivial: don't hardcode Windows path separator in the code. use `Path.Combine`.

Fixes #102

### DIFF
--- a/src/benchmarks/micro/corefx/System.Console/Perf.Console.cs
+++ b/src/benchmarks/micro/corefx/System.Console/Perf.Console.cs
@@ -4,6 +4,7 @@
 
 using System.IO;
 using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Engines;
 
 namespace System.ConsoleTests
 {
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/corefx/System.Console/Perf.Console.cs
```diff
@@ -13,17 +14,36 @@ namespace System.ConsoleTests
...
     {
+        private readonly Consumer consumer = new Consumer();
+
         [Benchmark]
-        public Stream OpenStandardInput() => Console.OpenStandardInput();
+        public void OpenStandardInput()
+        {
+            using (var input = Console.OpenStandardInput())
+            {
+                consumer.Consume(input);
```
billwert: I don't think consuming the variable is necessary here. the JIT can't know that OpenStandardInput and the destructor you're asking to be called don't have side effects, so they'll get called regardless. It's not a ton of extra work but if we don't need it why have it?

adamsitnik: I prefer to add it in defensive way to not rely on JIT implementation details (just always return the result to be consumed by the harness or consume it on your own)

PR #155, opened 2018-11-16
Add Arcade SDK files

adiaaida:
This change adds the arcade sdk files (the eng directory). It does not yet set up the performance repo to use arcade for building. This is the first step in moving the performance repo jobs from Jenkins to Azure DevOps.


NuGet.config
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="arcade" value="https://dotnetfeed.blob.core.windows.net/dotnet-tools-internal/index.json" />
+    <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
```
adamsitnik: the micro-benchmarks project uses latest bits from BenchmarkDotNet CI feed. I am not sure how Arcade works, but in case you can't restore the packages you gonna need to add this CI feed here as well

https://github.com/dotnet/performance/blob/0a033c272b3f5099c3588ed0b5156f3c9bcbbd80/src/benchmarks/micro/NuGet.Config#L11

adamsitnik:
BTW I am really happy to see progress on getting the perf repo work with our CI infrastructure!

adiaaida:
@adamsitnik @maririos @DrewScoggins PTAL

PR #156, opened 2018-11-27
do not include heavy arcade dependencies in MicroBenchmarks projects

adamsitnik:
The `.props` and `.targets` files added to the root folder directory in #155 affect projects with benchmarks

@janvorli got following error when running dotnet restore in `src\benchmarks\micro\`

```log
C:\Projects\performance\Directory.Build.props(7,3): error : Failed to resolve SDK 'Microsoft.DotNet.Arcade.Sdk'. Package restore was successful but a package with the ID of "Microsoft.DotNet.Arcade.Sdk" was not installed.
```

this is because the `src\benchmarks\micro\NuGet.Config` is missing address of Arcade NuGet feed.

However, I don't think that our lightweight benchmarks projects should rely on Arcade, so my changes ignores the `.props` and `.build` files

### DIFF
--- a/src/benchmarks/micro/Directory.Build.props
+++ b/src/benchmarks/micro/Directory.Build.props
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <ImportDirectoryBuildProps>false</ImportDirectoryBuildProps>
+    <ImportDirectoryBuildTargets>false</ImportDirectoryBuildTargets>
+  </PropertyGroup>
+</Project>


adiaaida:
This is going  to become a problem when coreclr/corefx move to arcade (corefx has already started this process) and we pull this repo into those repos, as there will be a Directory.Build.Props and Directory.Build.targets files in the directory structure above the repo. This should be dealt with by moving the performance repo to also use the arcade way of building, not just trying to work around it.

jorive:
Can we revert https://github.com/dotnet/performance/pull/155? It buys nothing on this repo and it is not even used. I feel that `because other repos need/use it` is not enough reason to add the dependency.

adamsitnik:
Fixed with #160

PR #158, opened 2018-11-27
improve the docs based on Jan V experience

adamsitnik:
@janvorli PTAL

### DIFF
--- a/docs/benchmarking-workflow.md
+++ b/docs/benchmarking-workflow.md
@@ -40,9 +40,22 @@ To build the benchmarks you need to have the right `dotnet cli`. This repository
 
 If you don't want to install all of them and just run the benchmarks for selected runtime(s), you need to manually edit the [MicroBenchmarks.csproj](../src/benchmarks/micro/MicroBenchmarks.csproj) file.
 
-```dif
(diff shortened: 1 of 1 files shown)


docs/benchmarking-workflow.md
```diff
@@ -206,13 +219,15 @@ Pass the path to CoreRun using `--coreRun` argument. In both CoreCLR and CoreFX
 * "C:\Projects\coreclr\bin\tests\Windows_NT.x64.Release\Tests\Core_Root\CoreRun.exe"
 * "C:\Projects\corefx\bin\runtime\netcoreapp-Windows_NT-Release-x64\CoreRun.exe"
 
+**Note:** BenchmarkDotNet expects a path to `CoreRun` file, not to `Core_Root` folder.
```
janvorli: On unix, the name is case sensitive and the file is named `corerun`. It may be worth mentioning. E.g.:
BenchmarkDotNet expects a path to `CoreRun.exe` file (`corerun` on Unix) ...

adamsitnik: @janvorli fixed, thanks!

PR #160, opened 2018-11-27
Revert "Add Arcade SDK files"

adamsitnik:
Reverts dotnet/performance#155 because it broke the benchmarks

Will send an email soon, but to tell the long story short we don't need to have a dependency on Arcade, just reuse the scripts Jose wrote in #150

This repo does not rely on CoreFX/CLR and the only thing we need is dotnet restore, build and run.


adamsitnik:
cc @adiaaida

adiaaida:
We need arcade so that we can move the jobs to Azure DevOps. So, like, yes, we do have a dependency on arcade.

PR #163, opened 2018-11-30
Add Arcade SDK files

adiaaida:
This change adds the Arcade files without modifying anything about how the performance repo builds. It adds everything that is necessary for creating the Azure DevOps jobs and onboarding to Dependency Flow, but nothing else.

Notes:
* Adds Directory.Build.props and Directory.Build.targets that are empty, which will harden us against cloning into repos/directories that have those files in the levels above
* adds global.json with only the msbuild-sdks list (so that Arcade can be updated automatically using dependency flow)
* adds the eng directory from arcade, which will be used for writing the AzDO jobs.

I have tested this and confirmed that we appear to be building in the same way as we did before, ie the MicroBenchmarks build into `performance\src\benchmarks\micro\bin\Release\netcoreapp3.0\publish\`


brianrob:
@adiaaida this change looks OK to me.  Per our conversation, let's discuss over e-mail tomorrow before merging.

jorive:
Do the steps below still work?
```cmd
rem Clear whatever you might have in cache.
nuget locals all -clear

rem Assuming you have dotnet.exe on PATH, from the root of the repo run:
pushd ".\src\benchmarks\micro"

dotnet restore MicroBenchmarks.csproj --packages .\packages\
dotnet build MicroBenchmarks.csproj --configuration Release --framework netcoreapp3.0 --no-restore
dotnet run --project MicroBenchmarks.csproj --configuration Release --framework netcoreapp3.0 --no-restore --no-build -- --allCategories coreclr --maxIterationCount 5 --minIterationCount 3 --filter *Adams*
```

PR #166, opened 2018-12-04
 Adding perf tests for the new System.Math/MathF APIs added in netcoreapp2.1/3.0

tannergooding:
This adds perf tests for the `Acosh`, `Asinh`, `Atanh`, and `Cbrt` functions added to `System.Math` and `System.MathF` in netcoreapp2.1 and the `FusedMultiplyAdd`, `ILogB`, `Log2`, and `ScaleB` functions added to `System.Math` and `System.MathF` in netcoreapp3.0

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -65,6 +65,14 @@
     <Compile Remove="corefx\System.Runtime\Perf.HashCode.cs" />
     <Compile Remove="corefx\System.Security.Cryptography.Primitives\**" />
     <Compile Remove="corefx\System.Security.Cryptography\Perf.Rfc2898DeriveBytes.cs" />
+    <Compile Remove="coreclr\Math\Functions\Double\Acosh.cs" />
+    <Compile Remove="coreclr\Math\Functions\Double\Asinh.cs" />
+    <Compile Remove="coreclr\Math\Functions\Double\Atanh.cs" />
+    <Compile Remove="coreclr\Math\Functions\Double\Cbrt.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\Acosh.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\Asinh.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\Atanh.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\Cbrt.cs" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -75,6 +75,17 @@
     <Compile Remove="coreclr\Math\Functions\Single\Cbrt.cs" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0')">
+    <Compile Remove="coreclr\Math\Functions\Double\FusedMultiplyAdd.cs" />
+    <Compile Remove="coreclr\Math\Functions\Double\ILogB.cs" />
+    <Compile Remove="coreclr\Math\Functions\Double\Log2.cs" />
+    <Compile Remove="coreclr\Math\Functions\Double\ScaleB.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\FusedMultiplyAdd.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\ILogB.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\Log2.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\ScaleB.cs" />
+  </ItemGroup>
+
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
     <Compile Remove="corefx\System.Drawing\Perf_Graphics*.cs" />
   </ItemGroup>
--- a/src/benchmarks/micro/coreclr/Math/Functions/Double/Acosh.cs
+++ b/src/benchmarks/micro/coreclr/Math/Functions/Double/Acosh.cs
@@ -1,0 +1,38 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using BenchmarkDotNet.Attributes;
+
+namespace System.MathBenchmarks
+{
+    public partial class Double
+    {
+        // Tests Math.Acosh(double) over 5000 iterations for the domain +1, +3
+
+        private const double acoshDelta = 0.0004;
+        private const double acoshExpectedResult = 6148.648751739127;
+
+        [Benchmark]
+        public void Acosh() => AcoshTest();
+
+        public static void AcoshTest()
+        {
+            var result = 0.0; var value = 1.0;
+
+            for (var iteration = 0; iteration < MathTests.Iterations; iteration++)
+            {
+                result += Math.Acosh(value);
+                value += acoshDelta;
+            }
+
+            var diff = Math.Abs(acoshExpectedResult - result);
+
+            if (double.IsNaN(result) || (diff > MathTests.DoubleEpsilon))
+            {
+                throw new Exception($"Expected Result {acoshExpectedResult,20:g17}; Actual Result {result,20:g17}");
+            }
+        }
+    }
+}
--- a/src/benchmarks/micro/coreclr/Math/Functions/Double/Asinh.cs
+++ b/src/benchmarks/micro/coreclr/Math/Functions/Double/Asinh.cs
@@ -1,0 +1,38 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using BenchmarkDotNet.Attributes;
+
+namespace System.MathBenchmarks
+{
+    public partial class Double
+    {
+        // Tests Math.Asinh(double) over 5000 iterations for the domain -1, +1
+
+        private const double asinhDelta = 0.0004;
+        private const double asinhExpectedResult = -0.88137358721605752;
+
+        [Benchmark]
+        public void Asinh() => AsinhTest();
+
+        public static void AsinhTest()
+        {
+            var result = 0.0; var value = -1.0;
+
+            for (var iteration = 0; iteration < MathTests.Iterations; iteration++)
+            {
+                result += Math.Asinh(value);
+                value += asinhDelta;
+            }
+
+            var diff = Math.Abs(asinhExpectedResult - result);
+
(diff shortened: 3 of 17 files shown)


tannergooding:
There are some more tests I wanted to add for new tests in netcoreapp3.0; but it looks like the consumed CoreCLR/CoreFX bits aren't quite in sync....

I would also like to fixup the existing tests that were ported, but @adamsitnik indicated that might need approval first.

src/benchmarks/micro/ValuesGenerator.cs
```diff
@@ -87,12 +94,34 @@ private static T GenerateValue<T>(Random random)
...
                 return (T)(object)random.NextDouble();
+            if (typeof(T) == typeof(float))
+                return (T)(object)(float)random.NextDouble();
             if (typeof(T) == typeof(string))
                 return (T) (object) GenerateRandomString(random, 1, 50);
             
             throw new NotImplementedException($"{typeof(T).Name} is not implemented");
         }
 
+        private static T GenerateValue<T>(Random random, T minValue, T maxValue)
```
tannergooding: Many of the `System.Math` functions have special handling in the JIT to either cache computed results or to perform constant folding. So we want to generate a unique input per iteration that is within the allowed domain.

jorive: @tannergooding Could you please add this as a function header/comment?

src/benchmarks/micro/coreclr/Math/Functions/Double/AcoshDouble.cs
```diff
@@ -0,0 +1,28 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace Functions
+{
+    [BenchmarkCategory(Categories.CoreCLR)]
+    public class AcoshDouble
```
jorive: The naming convention here should be:
class name: `public partial class MathTests`
benchmark method name: `AcoshDoubleBenchmark`

for reference please take a look at [AcosDouble.cs](https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/coreclr/Math/Functions/Double/AcosDouble.cs)

tannergooding: How would I modify `IterationSetup` to only apply to a given test (given that we need a different one for each `[Benchmark]` to generate the input value from the given domain)?

AndreyAkinshin: @tannergooding Setup/Clean attributes have `Target` property which allows specifying target benchmark for each Setup/Cleanup methods. Thus, you can write separate IterationSetup methods for each benchmark (a usage example: https://benchmarkdotnet.org/articles/samples/IntroSetupCleanupTarget.html)

brianrob:
cc: @billwert

src/benchmarks/micro/ValuesGenerator.cs
tannergooding: Fixed.

tannergooding:
Currently getting:
```
// * Warnings *
MinIterationTime
  MathTests.AcoshDoubleTest: InvocationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, UnrollFactor=1, WarmupCount=1 -> MinIterationTime = 400.0000 ns which is very small. It's recommended to increase it.
  MathTests.AsinhDoubleTest: InvocationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, UnrollFactor=1, WarmupCount=1 -> MinIterationTime = 500.0000 ns which is very small. It's recommended to increase it.
  MathTests.AtanhDoubleTest: InvocationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, UnrollFactor=1, WarmupCount=1 -> MinIterationTime = 700.0000 ns which is very small. It's recommended to increase it.
  MathTests.AtanhSingleTest: InvocationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, UnrollFactor=1, WarmupCount=1 -> MinIterationTime = 800.0000 ns which is very small. It's recommended to increase it.
MultimodalDistribution
  MathTests.AcoshDoubleTest: InvocationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, UnrollFactor=1, WarmupCount=1 -> It seems that the distribution is multimodal (mValue = 4.25)
  MathTests.AtanhDoubleTest: InvocationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, UnrollFactor=1, WarmupCount=1 -> It seems that the distribution is bimodal (mValue = 3.75)
  MathTests.CbrtDoubleTest: InvocationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, UnrollFactor=1, WarmupCount=1  -> It seems that the distribution is bimodal (mValue = 3.56)
  MathTests.AcoshSingleTest: InvocationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, UnrollFactor=1, WarmupCount=1 -> It seems that the distribution can have several modes (mValue = 3)
  MathTests.TanSingleBenchmark: IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, WarmupCount=1                                 -> It seems that the distribution can have several modes (mValue = 3.14)
```

What is the recommended "MinIterationTime" and how can I determine the correct value to scale these by?
* @adamsitnik mentioned that I can use `OperationsPerInvoke` and invoke the call multiple times in the benchmark body; so I just need to know how to determine how many times I should be invoking the call per iteration

src/benchmarks/micro/ValuesGenerator.cs
```diff
@@ -87,12 +94,40 @@ private static T GenerateValue<T>(Random random)
...
+            // This is currently used for many of the System.Math and System.MathF perf tests.
+            //
+            // These functions often have special handling in the JIT to either cache the computed
+            // result or to perform constant folding. This allows us to generate a unique input
+            // per iteration that is within the allowed input domain for a given method.
+
+            if (typeof(T) == typeof(int))
+                return (T)(object)random.Next((int)(object)minValue, (int)(object)maxValue);
+            if (typeof(T) == typeof(double))
+                return (T)(object)BitConverter.Int64BitsToDouble(
```
adamsitnik: Maybe we could use a simpler approach? From [SO](https://stackoverflow.com/a/1064907/5852046)

`random.NextDouble() * (maximum - minimum) + minimum;`

src/benchmarks/micro/ValuesGenerator.cs
```diff
@@ -87,12 +94,40 @@ private static T GenerateValue<T>(Random random)
...
+                    (long)random.Next(
+                        (int)(BitConverter.DoubleToInt64Bits((double)(object)minValue) >> 32),
+                        (int)(BitConverter.DoubleToInt64Bits((double)(object)maxValue) >> 32)
+                    ) << 32
+                );
+            if (typeof(T) == typeof(float))
+            {
+                int result = random.Next(Unsafe.As<T, int>(ref minValue), Unsafe.As<T, int>(ref maxValue));
```
adamsitnik: maybe we could take the double result and cast it to float?  `(T)(object)(float)GenerateValue<double>()`

src/benchmarks/micro/ValuesGenerator.cs
```diff
@@ -21,6 +22,12 @@ public static T GetNonDefaultValue<T>()
                 return ArrayOfUniqueValues<T>(2).First(value => !value.Equals(default));
         }
 
+        public static T GetValue<T>(T minValue, T maxValue)
+        {
+            var random = new Random(Seed);
```
adamsitnik: this is very tricky: this method will always return the same result because we use a const seed.

```cs
for (int i = 0; i < 10; i++)
    Console.WriteLine(ValuesGenerator.GetValue(1.0, double.PositiveInfinity));
```

```
3,98113421984948E+20
3,98113421984948E+20
3,98113421984948E+20
3,98113421984948E+20
3,98113421984948E+20
3,98113421984948E+20
3,98113421984948E+20
3,98113421984948E+20
3,98113421984948E+20
3,98113421984948E+20
```

In general, this is the desired approach. However, in our case, we want to have a different number per iteration to make sure the JIT does not cache the result?

src/benchmarks/micro/ValuesGenerator.cs
tannergooding: I believe the current approach gives the best distribution of results, over the desired range, without falling into a pit of failure due to IEEE rounding and result representation issues.

src/benchmarks/micro/ValuesGenerator.cs
tannergooding: Same comment as above.

src/benchmarks/micro/ValuesGenerator.cs
tannergooding: Ah, right. We would need to have it keep the same random that was initially seeded, which would also require the tests to run in a deterministic order.

tannergooding:
> I can see that the time it takes to execute most of the tested methods is dependent on the input. For example time(Math.Acosh(1.0)) != time(Math.Acosh(100.0)).

Yes, many of these methods special-case certain inputs or ranges of inputs for speed purposes.

src/benchmarks/micro/ValuesGenerator.cs
adamsitnik: I am sure it is. However, to understand this code one needs an expert knowledge. You have it, others (including me) don't ;)

Could you add a comment explaining how it works and some unit tests to make sure it's correct?

src/benchmarks/micro/ValuesGenerator.cs
adamsitnik: @gfoidl we should rather create it inside a method and pass to other methods like here:

https://github.com/dotnet/performance/blob/c4efbbb55e0a5d8f4408c9806bdb51b300a261fc/src/benchmarks/micro/ValuesGenerator.cs#L29-L35

so anytime the method is invoked it produces the same results

danmosemsft: > JIT does not cache the result?

By what mechanism would the JIT do that? It doesn't know enough to know it's the same result each time does it?

tannergooding: A number of the `System.Math` functions are intrinsicly recognized by the JIT and will under go constant folding since they should be deterministic.

tannergooding:
I'll try to get around to rebasing this and resolving remaining feedback sometime today.

billwert:
@tannergooding no worries. I was wondering if I could trick it into running anyway. :)

jorive:
@tannergooding Could you please follow the naming convention in the recently renamed Math benchmarks here https://github.com/dotnet/performance/tree/master/src/benchmarks/micro/coreclr/Math/Functions ?

tannergooding:
I've fixed up the new tests to match the existing namespace/layout and reverted the behavior back to be closer to what I had in https://github.com/dotnet/coreclr/pull/21289.

The reasoning being that they not only more closely mirror the behavior of the other Math/MathF tests, but it also ensures that we aren't falling subject to issues with the cost of the functions being too small or to the JIT performing constant folding on them and causing incorrect numbers.

tannergooding:
I think, if we want to fix up the tests to do some different behavior, we should instead do that all at once and in a separate PR.

tannergooding:
I think this is ready for a final review pass.

billwert:
LGTM. @tannergooding agreed to open an issue to capture recommendations for what the future of this entire suite looks like.

tannergooding:
I logged https://github.com/dotnet/performance/issues/423 which captures the general discussion had above and some of the issues to look out for.

PR #169, opened 2018-12-05
BenchviewCommitName was only set on private/PR runs.

jorive:


jorive:
@dotnet-bot test ci please

jorive:
@dotnet-bot test ci please

jorive:
```groovy
// It's unclear if the environment variables below will expand as desired
// Are there groovy variable I should be using instead? ${BranchName} ${sha1}?
rolling = "${benchviewCommitNamePrefix} %GIT_BRANCH% %GIT_COMMIT%"
```
@adiaaida @mmitche

adiaaida:
To be honest, I don't know. Where does this get used? It's something that I'll have to consider in the port to AzureDevOps.

jorive:
We use those variables in CoreClr perf.groovy, but I cannot tell where they come from.

adiaaida:
They are set by Jenkins as parameters to the job

PR #171, opened 2018-12-06
Make MicroBenchmarks.sln work for VS 2017

adamsitnik:
Today I was preparing for my tomorrow's demo and when I opened MicroBenchmarks.sln with most recent VS 2017 I run into following problems:

* VS offered me to update Unit Test project to "newer project system". It kept doing that... Adding `<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">` instead of `<Project Sdk="Microsoft.NET.Sdk">` made it happy
* VS did not recognize the dependency between `MicroBenchmarks.csproj` and `Tests.csproj` so I had to define it in explicit way in the solution file to get it working with pressing "Build" just once
* The old `<RuntimeFrameworkVersion Condition="'$(TargetFramework)' == 'netcoreapp2.2'">2.2.0-*</RuntimeFrameworkVersion>` was a workaround (https://github.com/dotnet/cli/issues/10074), which is not needed anymore since 2.2 was shipped

btw if you want to use preview of .NET Core 3.0 in VS 2017 you need to enable following setting in VS:
image

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.sln
+++ b/src/benchmarks/micro/MicroBenchmarks.sln
@@ -6,6 +6,9 @@ MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MicroBenchmarks", "MicroBenchmarks.csproj", "{D99F63AE-3154-4F13-9424-FA5F9D032D1D}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Tests", "Tests\Tests.csproj", "{A40E51E2-99EB-4F64-ADCC-94FEA3FF1399}"
+	ProjectSection(ProjectDependencies) = postProject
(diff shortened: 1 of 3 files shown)


jorive:
Do we need `*.sln` files? Can we just add `*.sln` to `.gitignore`? They are tide to VS and will always need to be updated. What happen when someone does not have VS[latest]? I think VS will create a `*.sln` file when a `*.csproj` is opened and there wasn't already one solution.

adamsitnik:
I think that once you have more than 1 project you have to have a solution file. Here we have 2: MicroBenchmarks.csproj and Tests.csproj

jorive:
Yes, I've noticed that if you open `MicroBenchmarks.csproj` it creates a `MicroBenchmarks.sln` (it does not need to load `Tests.csproj`), BUT if you open `Tests.csproj` it will create a `Tests.sln` without loading its dependent `MicroBenchmarks.csproj` project.

jorive:
I have mixed feelings about `*.sln` files.

PR #174, opened 2018-12-07
initial cut at guarded devirt perf tests

AndyAyersMS:
I can see these are getting compiled but they don't show up in listing or filtering.

So not sure what I'm doing wrong.

@adamsitnik can you take a look?

### DIFF
--- a/src/benchmarks/micro/coreclr/Devirtualization/GuardedThreeClassVirtual.cs
+++ b/src/benchmarks/micro/coreclr/Devirtualization/GuardedThreeClassVirtual.cs
@@ -73,12 +73,12 @@ public class TestInput
     }
 }
 
+[BenchmarkCategory(Categories.CoreCLR, Categories.Virtual)]
 public class Virtual
 {
-    [BenchmarkCategory(Categories.CoreCLR, Categories.Virtual)]
     [Benchmark(OperationsPerInvoke=TestInput.N)]
     [ArgumentsSource(nameof(GetInput))]
-    public static long Call(TestInput testInput)
+    public long Call3(TestInput testInput)
     {
         long sum = 0;
         B[] input = testInput.Array;
@@ -89,10 +89,10 @@ public class Virtual
         return sum;
     }
 
-    static int S = 10;
+    static int S = 3;
     static double delta = 1.0 / (double) S;
 
-    public IEnumerable<TestInput> GetInput()
+    public static IEnumerable<TestInput> GetInput()
     {
         double pB = 0;
 
--- a/src/benchmarks/micro/coreclr/Devirtualization/GuardedThreeClassVirtual.cs
+++ b/src/benchmarks/micro/coreclr/Devirtualization/GuardedThreeClassVirtual.cs
@@ -1,0 +1,113 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+// Performance test for virtual call dispatch with three
+// possible target classes mixed in varying proportions.
+
+namespace GuardedDevirtualizationThreeClass
+{
+
+public class B
+{
+    public virtual int F() => 33;
+}
+
+public class D : B
+{
+    public override int F() => 44;
+}
+
+public class E : B
+{
+    public override int F() => 55;
+}
+
+public class TestInput
+{
+    public const int N = 1000;
+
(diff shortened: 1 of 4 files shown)


src/benchmarks/micro/coreclr/Devirtualization/GuardedThreeClassVirtual.cs
```diff
@@ -0,0 +1,113 @@
...
+        return result;
+    }
+}
+
+public class Virtual
+{
+    [BenchmarkCategory(Categories.CoreCLR, Categories.Virtual)]
+    [Benchmark(OperationsPerInvoke=TestInput.N)]
+    [ArgumentsSource(nameof(GetInput))]
+    public static long Call(TestInput testInput)
```
adamsitnik: BDN requires it to be an instance method.

I just created an isse to give a good warning about that: https://github.com/dotnet/BenchmarkDotNet/issues/983

```suggestion
    public long Call(TestInput testInput)
```

src/benchmarks/micro/coreclr/Devirtualization/GuardedTwoClassVirtual.cs
```diff
@@ -0,0 +1,95 @@
...
+        return result;
+    }
+}
+
+public class Virtual
+{
+    [BenchmarkCategory(Categories.CoreCLR, Categories.Virtual)]
+    [Benchmark(OperationsPerInvoke=TestInput.N)]
+    [ArgumentsSource(nameof(GetInput))]
+    public static long Call(TestInput testInput)
```
adamsitnik:
```suggestion
    public long Call(TestInput testInput)
```

src/benchmarks/micro/coreclr/Devirtualization/GuardedTwoClassVirtual.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            sum += input[i].F();
+        }
+        return sum;
+    }
+
+    static int S = 10;
+    static double delta = 1.0 / (double) S;
+
+    public IEnumerable<TestInput> GetInput()
```
adamsitnik: when I was porting the old CoreFX benchmarks to the perf repo I found few benchmarks with hundreds of test cases (600, 400, 200). It was clearly a human error, but responsible for half of the execution time of the CoreFX suite ;)

To prevent users from adding too many test cases I have added a `TooManyTestCasesValidator` which does not allow for mor than 16 test cases per benchmark.

If you really want to test more than 16 test cases locally you need to comment this line:

https://github.com/dotnet/performance/blob/2a7df6527d2deec682ae3acae18b39bd8901ea9a/src/benchmarks/micro/Program.cs#L38

src/benchmarks/micro/coreclr/Devirtualization/GuardedThreeClassVirtual.cs
AndyAyersMS: Thanks -- I was sure this was working when run normally under BDN. Anyways, easy fix.

src/benchmarks/micro/coreclr/Devirtualization/GuardedTwoClassVirtual.cs
AndyAyersMS: I'll cut down the number of sample points for now, but it would be nice to be able to bypass this somehow.

Or I could split it up as multiple tests.

AndyAyersMS:
Added the interface versions. Simpler than I thought as BDN seems to run each case in its own process so there is no VSD carryover (one can tell by looking at the no GDV two class results, they will not be symmetric about p=0.5 if VSD is carrying state over from the previous input array).

Interface results:

 Method | testInput |  noGDV   |   GDV     |
------- |---------- |---------:|----------:|
  Call2 | pB = 0.00 | 2.480 ns | 2.8002 ns |
  Call2 | pB = 0.10 | 4.602 ns | 3.2773 ns |
  Call2 | pB = 0.20 | 6.162 ns | 2.9830 ns |
  Call2 | pB = 0.30 | 7.262 ns | 2.8617 ns |
  Call2 | pB = 0.40 | 8.396 ns | 2.7793 ns |
  Call2 | pB = 0.50 | 8.705 ns | 2.3494 ns |
  Call2 | pB = 0.60 | 8.726 ns | 2.1261 ns |
  Call2 | pB = 0.70 | 7.557 ns | 1.9230 ns |
  Call2 | pB = 0.80 | 5.930 ns | 1.7155 ns |
  Call2 | pB = 0.90 | 4.482 ns | 1.5456 ns |
  Call2 | pB = 1.00 | 2.466 ns | 0.7914 ns |

AndyAyersMS:
@adamsitnik think these are ready

PR #177, opened 2018-12-09
Addition:  Initial Benchmarks for System.Reflection: Attributes

NickCraver:
This benchmarks specifically `GetCustomAttributes()` and `IsDefined()` paths (used in https://github.com/dotnet/coreclr/pull/20779 and https://github.com/dotnet/coreclr/pull/20795)

Note: there are many code paths through the attribute pipeline which vary drastically in terms of performance and allocations - so there's what seems to be an excessive number of benchmarks here, but they are all pretty unique.

Here's a sample run of the whole category on `netcoreapp2.1`:
image

The issue I can see here is naming. I tried to match conventions, but the descriptions are long because of the differentiation needed - is this okay? Suggestions for shortening them?

### DIFF
--- a/src/benchmarks/micro/coreclr/System.Reflection/Attributes/AttributeTests.cs
+++ b/src/benchmarks/micro/coreclr/System.Reflection/Attributes/AttributeTests.cs
@@ -1,0 +1,43 @@
+﻿using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+using System;
+using System.Reflection;
+
+namespace Attributes
+{
+    [BenchmarkCategory(Categories.CoreCLR, Categories.Reflection)]
+    public partial class AttributeTests
+    {
+        private static readonly MethodInfo AttributedOverride = typeof(AttributedClass).GetMethod(nameof(AttributedClass.FooAttributed));
+        private static readonly MethodInfo NonAttributedOverride = typeof(NonAttributedClass).GetMethod(nameof(NonAttributedClass.FooAttributed));
+        private static readonly MethodInfo AttributedBase = typeof(BaseClass).GetMethod(nameof(BaseClass.FooAttributed));
+        private static readonly MethodInfo NonAttributedBase = typeof(BaseClass).GetMethod(nameof(BaseClass.FooUnattributed));
+
+        [My("Test")]
+        [Serializable]
+        public class AttributedClass : BaseClass
+        {
+            [My("FooAttributed")]
+            public override string FooAttributed() => null;
+        }
+
+        public class NonAttributedClass : BaseClass
+        {
+            public override string FooUnattributed() => null;
+        }
+
+        public class BaseClass
+        {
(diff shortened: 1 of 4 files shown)


src/benchmarks/micro/coreclr/System.Reflection/Attributes/AttributeTests.cs
```diff
@@ -0,0 +1,43 @@
+﻿using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+using System;
+using System.Reflection;
+
+namespace Attributes
```
adamsitnik: could you please call this namespace `System.Reflection`? it's going to allow us for very easy filtering in the future: `--filter System.Reflection*` and also we use `namespace.typeName.methodName` as benchmark ids in our internal reporting tools

src/benchmarks/micro/coreclr/System.Reflection/Attributes/AttributeTests.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    [BenchmarkCategory(Categories.CoreCLR, Categories.Reflection)]
+    public partial class AttributeTests
```
adamsitnik:
```suggestion
    public partial class Attributes
```

src/benchmarks/micro/coreclr/System.Reflection/Attributes/GetCustomAttributes.cs
```diff
@@ -0,0 +1,37 @@
+﻿using BenchmarkDotNet.Attributes;
+
+namespace Attributes
```
adamsitnik:
```suggestion
namespace System.Reflection
```

src/benchmarks/micro/coreclr/System.Reflection/Attributes/GetCustomAttributes.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public partial class AttributeTests
```
adamsitnik:
```suggestion
    public partial class Attributes
```

src/benchmarks/micro/coreclr/System.Reflection/Attributes/IsDefined.cs
```diff
@@ -0,0 +1,37 @@
+﻿using BenchmarkDotNet.Attributes;
+
+namespace Attributes
```
adamsitnik:
```suggestion
namespace System.Reflection
```

src/benchmarks/micro/coreclr/System.Reflection/Attributes/IsDefined.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public partial class AttributeTests
```
adamsitnik:
```suggestion
    public partial class Attributes
```

PR #180, opened 2018-12-10
Fixing BenchmarkDotNet restore timeout in the lab

jorive:
In the lab we are seeing getting this `The configured timeout 00:01:00 was reached!` while restoring the benchmarks. This change attempts to alleviate these errors.


jorive:
@dotnet-bot test ci please

PR #183, opened 2018-12-10
update BDN to version with consistent file names across all OSes

adamsitnik:
https://github.com/dotnet/BenchmarkDotNet/issues/981

So far BDN used `Path.GetInvalidFileNameChars()` to determine invalid path characters and `<` & `>` are valid on Unix, but not on Windows. The side effect was that when we tried to copy Unix results to Windows, we could not do it without changing the files names first.

This simple change replaces `>` and `<` with `_` for every OS.

/cc @AndyAyersMS


jorive:
:shipit:

PR #186, opened 2018-12-10
Use arcade to build performance artifacts

adiaaida:
This change updates various arcade files that we need to start building into the artifacts directory. It also updates a few of the ci scripts to handle the new build location.


jorive:
Do we need to add the `artifacts/` folder to `.gitignore`?

jorive:
After these changes, when a developer runs the following commands:
```cmd
pushd .\src\benchmarks\micro\
dotnet restore MicroBenchmarks.csproj --packages ..\..\..\packages\
dotnet build MicroBenchmarks.csproj --configuration Release --framework netcoreapp3.0 --no-restore
```
where do the `bin` and `obj` folders are created?

scripts/micro_benchmarks.py
```diff
@@ -221,14 +222,15 @@ def build(
...
     '''Restores and builds the benchmarks'''
 
-    packages = path.join(get_repo_root_path(), 'packages')
+    packages = path.join(get_artifacts_directory(), 'packages')
 
     if incremental == 'no':
         __log_script_header("Removing packages, bin and obj folders.")
         binary_folders = [
             packages,
             path.join(BENCHMARKS_CSPROJ.working_directory, 'bin'),
```
jorive: After these changes, do we need this line: `path.join(BENCHMARKS_CSPROJ.working_directory, 'bin'),`?

scripts/dotnet.py
```diff
@@ -74,7 +75,7 @@ def csproj_file(self) -> str:
     @property
     def bin_path(self) -> str:
```
jorive: Is this always the case when running the arcade/build or dotnet commands?

scripts/micro_benchmarks.py
adiaaida: probably we can delete it.

scripts/dotnet.py
adiaaida: Yes

adiaaida:
@chcosta

adiaaida:
@jorive the bin and obj directories will be in `artifacts\bin` and `artifacts\obj`

DrewScoggins:
The artifact upload step is failing in AzDO, is that something that we understand?

scripts/dotnet.py
jorive: Then references to `path.join(BENCHMARKS_CSPROJ.working_directory, 'obj')` should be updated.

adiaaida: Ok thanks

Directory.Build.props
```diff
@@ -1,3 +1,17 @@
...
+  <PropertyGroup>
+    <ImportNetSdkFromRepoToolset>false</ImportNetSdkFromRepoToolset>
+  </PropertyGroup>
+
+  <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+  <PropertyGroup>
+    <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true'">
+      https://ci.appveyor.com/nuget/benchmarkdotnet;
+      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
+      https://dotnetfeed.blob.core.windows.net/dotnet-tools-internal/index.json;
```
chcosta: what packages are being restored from dotnet.myget.org and dotnet-tools-internal?  Are those sources needed?

NuGet.config
```diff
@@ -0,0 +1,14 @@
...
+  <solution>
+    <add key="disableSourceControlIntegration" value="true" />
+  </solution>
+  <packageSources>
+    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
+    <clear />
+    <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+    <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="benchmarkdotnet-ci-feed" value="https://ci.appveyor.com/nuget/benchmarkdotnet" />
+    <add key="dotnet-myget-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
```
chcosta: is the myget source needed?

chcosta:
Consider moving the `build\stylecop.json` file into the eng folder and moving `scripts` under `eng`

chcosta:
You may want to replace https://github.com/dotnet/performance/blob/29c287318da091c9ebfb9be4dc2ffae84fccdeac/azure-pipelines.yml#L3 with the "helixRepo" parameter, https://github.com/dotnet/arcade/blob/master/azure-pipelines.yml#L17, which will auto-derive helixSource.

scripts/dotnet.py
adiaaida: There's only the one reference to the obj directory, and it's in the delete binary folders code.

Directory.Build.props
adiaaida: @adamsitnik would know better for the dotnet.myget.org, I just copied that from the nuget.config in the micro benchmark directory. dotnet-tools-internal is probably unnecessary. Removing.

NuGet.config
adiaaida: I believe it is used by something in the benchmarks. I'm not sure if it's needed in the nuget.config if it's in the Directory.Build.props.

adamsitnik: we need the feed to daily CoreFX and CoreCLR packages. Were they moved from MyGet to `dotnetfeed.blob.core.windows.net` ?

PR #187, opened 2018-12-12
Minor fixes

adamsitnik:
I did remove `[Benchmark(Baseline = true)]` from the code because it's confusing when running the benchmarks for more than 1 runtime.

Explanation:

When we have:

```cs
[Benchmark(Baseline = true)]
public void A() {} 

[Benchmark]
public void B() {}
```

and run it for 1 runtime, the baseline is A and the ratio for B is scaled to A.

However, when we run it for few runtimes, let's say `--runtimes netcoreapp2.1 netcoreapp3.0`

the Baseline is `A` for .NET Core 2.1 and all the ratios as scaled to it. This is very confusing when reading the results. I think we should remove it. I am also the person who added it so nobody will be missing it ;)


adamsitnik:
@danmosemsft thank you!

PR #190, opened 2018-12-12
Add a few Timer perf tests

stephentoub:
- Overhead of creating and disposing timers
- Multithreaded contention when the same threads create and then dispose timers.
- Multithreaded contention when likely different threads are creating and disposing the timers, or at least when there's queueing between those operations.

We don't currently have a way to benchmark / measure how much time is spent in a particular function, so this doesn't add any tests related to firing timers where we'd want to measure the CPU impact of FireNextTimers.

cc: @kouvel, @adamsitnik


adamsitnik:
@stephentoub thank you!

PR #195, opened 2018-12-14
Add perf tests for .NET Core 2.1 perf blog post

stephentoub:
Some of the benchmarks from https://blogs.msdn.microsoft.com/dotnet/2018/04/18/performance-improvements-in-net-core-2-1/ are already covered, but a bunch aren't.  This adds the missing ones. (The only one I didn't add is Enum.HasFlags, as it ends up conflicting with the 2.0 PR I put up earlier today; that can be added later.)

### DIFF
--- a/src/benchmarks/micro/corefx/System.Runtime/Perf.Int32.cs
+++ b/src/benchmarks/micro/corefx/System.Runtime/Perf.Int32.cs
@@ -26,11 +26,15 @@ namespace System.Tests
         [Benchmark]
         [ArgumentsSource(nameof(Int32Values))]
         public string ToString(int value) => value.ToString();
-        
+
+        [Benchmark]
+        [ArgumentsSource(nameof(StringValues))]
+        public int Parse(string value) => int.Parse(value);
+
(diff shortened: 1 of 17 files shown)


src/benchmarks/micro/corefx/System.Runtime/Perf.Int32.cs
```diff
@@ -26,11 +26,15 @@ public class Perf_Int32
...
+
+        [Benchmark]
+        [ArgumentsSource(nameof(StringValues))]
+        public int Parse(string value) => int.Parse(value);
+
 #if !NETFRAMEWORK && !NETCOREAPP2_0 // API added in .NET Core 2.1
         [Benchmark]
         [ArgumentsSource(nameof(StringValues))]
-        public int Parse(string value) => int.Parse(value.AsSpan());
+        public int ParseSpan(string value) => int.Parse(value.AsSpan());
```
adamsitnik: this is a breaking change (benchmark rename) but totally needed  and reasonable 👍

PR #206, opened 2018-12-19
add AsSpan and AsMemory benchmarks

adamsitnik:
Benchmarks for https://github.com/dotnet/coreclr/issues/21594

### DIFF
--- a/src/benchmarks/micro/corefx/System.Memory/Constructors.cs
+++ b/src/benchmarks/micro/corefx/System.Memory/Constructors.cs
@@ -78,9 +78,15 @@ namespace System.Memory
 
         [Benchmark]
         public System.Span<T> ArrayAsSpan() => _nonEmptyArray.AsSpan();
-        
+
+        [Benchmark]
+        public System.Span<T> ArrayAsSpanStartLength() => _nonEmptyArray.AsSpan(start: 0, length: Size);
+
         [Benchmark]
         public System.Memory<T> ArrayAsMemory() => _nonEmptyArray.AsMemory();
+
+        [Benchmark]
+        public System.Memory<T> ArrayAsMemoryStartLength() => _nonEmptyArray.AsMemory(start: 0, length: Size);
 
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/corefx/System.Memory/Constructors.cs
```diff
@@ -76,6 +76,12 @@ public Constructors()
         [Benchmark]
         public System.ReadOnlyMemory<T> ReadOnlyMemoryFromArrayStartLength() => new System.ReadOnlyMemory<T>(_nonEmptyArray, start: 0, length: Size);
 
+        [Benchmark]
+        public System.Span<T> ArrayAsSpan() => _nonEmptyArray.AsSpan();
+        
+        [Benchmark]
+        public System.Memory<T> ArrayAsMemory() => _nonEmptyArray.AsMemory();
```
stephentoub: What about with arguments?

src/benchmarks/micro/corefx/System.Memory/Constructors.cs
```diff
@@ -76,6 +76,12 @@ public Constructors()
         [Benchmark]
         public System.ReadOnlyMemory<T> ReadOnlyMemoryFromArrayStartLength() => new System.ReadOnlyMemory<T>(_nonEmptyArray, start: 0, length: Size);
 
+        [Benchmark]
+        public System.Span<T> ArrayAsSpan() => _nonEmptyArray.AsSpan();
```
jkotas: This does not cover the case that regressed in linked issue. The case that regressed was `AsSpan` overload that takes the index and length arguments.

danmosemsft: Also this is an array of 10 elements. @Anipik 's case was 100_000 elements. They may have different perf characteristics.

adamsitnik:
@stephentoub I did not have the time to rebuild and re-run the benchmarks with and without the change. However, for such a simple method the regression should be detected for sure.

danmosemsft:
>  the perf characteristics of this method are size-independent

If weuse whatever the most commonly used size is (don't know what ) then the test will tell us if that accidentally changes.

Issue #209, opened 2018-12-30
[Question]Run with --cli


MarcoRossignoli:
I'm trying to run benchmark with `--cli` options as outlined here https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/README.md#dotnet-cli
However if I don't have installed correct machine wide sdk I get error:
```
D:\git\performance\src\benchmarks\micro (master -> origin)
λ dotnet run -c Release -f netcoreapp3.0 -- --join --filter System.Collections*.Dictionary* *.Perf_Dictionary.* --cli ..\..\..\..\netcorepreviewsdk\dotnet.exe
C:\Program Files\dotnet\sdk\2.2.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(137,5): error NETSDK1045: The current .NET SDK does not support targeting .NET Core 3.0.  Either target .NET Core 2.2 or lower, or use a version of the .NET SDK that supports .NET Core 3.0. [D:\git\performance\src\benchmarks\micro\MicroBenchmarks.csproj]

The build failed. Please fix the build errors and run again.
```
But it works if I use downloaded(non machine wide) sdk:
```
D:\git\performance\src\benchmarks\micro (master -> origin)
λ ..\..\..\..\netcorepreviewsdk\dotnet.exe run -c Release -f netcoreapp3.0 -- --join --filter System.Collections*.Dictionary* *.Perf_Dictionary.* --cli ..\..\..\..\netcorepreviewsdk\dotnet.exe
// Validating benchmarks:
// ***** BenchmarkRunner: Start   *****
// ***** Building 1 exe(s) in Parallel: Start   *****
```
Is it expected?If so I propose to update guide with this scenario(I don't like install all preview machine wide), let me know if I can do a PR(if it makes sense to you).

/cc @adamsitnik @danmosemsft

adamsitnik:
Hi @MarcoRossignoli 

We have a nice python script that can download the right dotnet cli and run the benchmarks for you. I am gonna add some docs for it soon.

As of now you can use it in following way: 

```cmd
.\scripts\benchmarks_ci.py --frameworks netcoreapp3.0 --filter System.Collections*.Dictionary* *.Perf_Dictionary.* --bdn-arguments "--join"
```

/cc @jorive

MarcoRossignoli:
for `--CoreRun` list can we append after `--join`?

adamsitnik:
[...]

PR #210, opened 2019-01-01
Allow more than one CoreRun benchmarks_ci.py script

MarcoRossignoli:
Contributes to https://github.com/dotnet/performance/issues/209
This PR allows to have more than one `--corerun` runner, for instance:
```
.\scripts\benchmarks_ci.py --frameworks netcoreapp3.0 --filter *.Perf_Dictionary.* --corerun D:\git\corefxupstream\artifacts\bin\testhost\netcoreapp-Windows_NT-Release-x64\shared\Microsoft.NETCore.App\9.9.9\CoreRun.exe D:\git\corefx\artifacts\bin\testhost\netcoreapp-Windows_NT-Release-x64\shared\Microsoft.NETCore.App\9.9.9\CoreRun.exe --bdn-arguments \"--join\"
```

/cc @adamsitnik @jorive

### DIFF
--- a/scripts/micro_benchmarks.py
+++ b/scripts/micro_benchmarks.py
@@ -173,6 +173,7 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         '--corerun',
         dest='corerun',
         required=False,
+        nargs='*',
         type=__valid_file_path,
         help='Full path to CoreRun.exe (corerun on Unix)',
     )
--- a/scripts/micro_benchmarks.py
+++ b/scripts/micro_benchmarks.py
@@ -173,7 +173,7 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         '--corerun',
         dest='corerun',
         required=False,
-        nargs='*',
+        nargs='+',
         type=__valid_file_path,
         help='Full path to CoreRun.exe (corerun on Unix)',
     )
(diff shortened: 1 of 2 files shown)


MarcoRossignoli:
Aside questione, why python and not powershell core?

scripts/micro_benchmarks.py
```diff
@@ -173,6 +173,7 @@ def __valid_file_path(file_path: str) -> str:
         '--corerun',
         dest='corerun',
         required=False,
+        nargs='*',
```
jorive: I think this should be `nargs='+'`

MarcoRossignoli: You're right

jorive:
@MarcoRossignoli When we started writing automation scripts for .NET Core, Powershell Core hadn't shipped, and we needed a cross-platform scripts (instead of having batch and bash versions of the scripts).

scripts/micro_benchmarks.py
MarcoRossignoli: done

jorive:
You need to update this line too: https://github.com/dotnet/performance/blob/master/scripts/micro_benchmarks.py#L301

MarcoRossignoli:
> You need to update this line too: https://github.com/dotnet/performance/blob/master/scripts/micro_benchmarks.py#L301

Thank's done!(I didn't found it in my tests/debug path)

jorive:
@MarcoRossignoli Thanks for the contribution :)

PR #214, opened 2019-01-04
Add Helix SDK, tools, and native-tools to global.json

adiaaida:
I keep having to rebase my PR when we get changes to global.json from Arcade updates because part of it adds various things to the global.json. These can be added outside of that PR, so I would like to get them checked in now so it is no longer part of the larger PR.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,12 @@
 {
+  "tools": {
+      "dotnet" : "2.1.403"
+  },
   "msbuild-sdks": {
-    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19053.2"
-  }
+    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19053.2",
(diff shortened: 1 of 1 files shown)


global.json
```diff
@@ -1,5 +1,12 @@
...
+  },
   "msbuild-sdks": {
-    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19053.2"
-  }
+    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19053.2",
+    "Microsoft.DotNet.Helix.Sdk": "1.0.0-beta.18629.1"
+  },
+  "native-tools": {
+    "python3": "3.7.1"
+  }	  }
```
adamsitnik: One '}' too much?

```suggestion
  }
```

adiaaida: Crap.

adiaaida: new PR, one sec.

PR #217, opened 2019-01-06
Remove --no-build for run verb in dotnet.py

MarcoRossignoli:
Check comment https://github.com/dotnet/performance/issues/209#issuecomment-451759735

From guide https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-run?tabs=netcore21
```
--no-build

Doesn't build the project before running. It also implicit sets the --no-restore flag.
```

/cc @adamsitnik @jorive

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -140,7 +140,7 @@ class CSharpProject:
             '--project', self.csproj_file,
             '--configuration', configuration,
             '--framework', framework,
-            '--no-restore', '--no-build',
+            # '--no-build' temporary disabled due to BDN issue https://github.com/dotnet/performance/issues/209#issuecomment-451759735
         ]
 
         if args:


adamsitnik:
@MarcoRossignoli big thanks for the fix!

@jorive I know that this solution is not perfect but if it makes it work I think that we should merge this PR

scripts/dotnet.py
```diff
@@ -140,7 +140,7 @@ def run(self,
             '--project', self.csproj_file,
             '--configuration', configuration,
             '--framework', framework,
-            '--no-restore', '--no-build',
+            # '--no-build' temporary disabled due to BDN issue https://github.com/dotnet/performance/issues/209#issuecomment-451759735
```
jorive: If we are going to remove this line, then I feel we should open an issue on BDN and link it here.
```python
#  '--no-restore', '--no-build',  # https://github.com/dotnet/BenchmarkDotNet/issues/<Issue_Id>
```
How we execute the harness should not affect how BDN builds the microbenchmarks that is going to execute.

MarcoRossignoli: I'm not sure that this will be fixed in future after read @adamsitnik answer https://github.com/dotnet/performance/issues/209#issuecomment-451777570 or is there a chance?

jorive: This still seems to be a bug on BDN and/or dotnet-cli that can be worth fixing, and I feel it's still worth opening an issue regardless on how it may be resolved.

MarcoRossignoli: Ah sorry I read late after https://github.com/dotnet/performance/issues/209#issuecomment-452056191
Do you think we should close this now?The issue will not raise anymore...with new fix we wait some attempts but at the end it works, I think it's a temporary workaround, I'll open an issue on BDN to track the strange behaviour.

MarcoRossignoli:
Fixed in https://github.com/dotnet/BenchmarkDotNet/pull/1001

scripts/dotnet.py
MarcoRossignoli: https://github.com/dotnet/BenchmarkDotNet/issues/1002

PR #218, opened 2019-01-07
CoreRT support

adamsitnik:
There is a throughput difference for PlainText TechEmpower benchmark for CoreRT vs .NET Core 3.0 on Linux. It's big enough that the benchmarks should show where the problem is.

This PR adds CoreRT support. BenchmarkDotNet supported CoreRT for a while, but I needed to fix some minor issues (https://github.com/dotnet/BenchmarkDotNet/pull/1001) and update our scripts to support it.

@jorive CoreRT does not support full reflection, so to run the benchmarks for CoreRT we need to run the host process as .NET Core process let is use reflection etc to build a CoreRT benchmark process. This is why I needed to differentiate framework and target framework moniker in the Python scripts. When we want to run .NET Core benchmarks, framework = moniker. When we want to run the CoreRT ones, framework = CoreRT, but moniker is netcoreapp3.0.

@jkotas to run all our benchmarks for CoreRT we now just need to run following command:

```cmd
.\scripts\benchmarks_ci.py --frameworks corert --filter *
```

this script downloads .NET Core 3.0 cli from Azure, setups everything and run the benchmarks using BenchmarkDotNet. I am going to run them for Linux and share the results soon

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -58,7 +58,7 @@ if sys.platform == 'linux' and "linux_distribution" not in dir(platform):
 
 def init_tools(
         architecture: str,
-        frameworks: str,
+        target_framework_monikers: list,
         verbose: bool) -> None:
     '''
     Install tools used by this repository into the tools folder.
@@ -67,8 +67,8 @@ def init_tools(
     '''
     getLogger().info('Installing tools.')
     channels = [
-        micro_benchmarks.TargetFrameworkAction.get_channel(framework) or 'LTS'
-        for framework in frameworks
+        micro_benchmarks.FrameworkAction.get_channel(target_framework_moniker) or 'LTS'
+        for target_framework_moniker in target_framework_monikers
     ]
     dotnet.install(
         architecture=architecture,
@@ -259,7 +259,7 @@ def __get_corefx_os_name():
         return platform.platform()
 
 
-def __get_build_info(args, framework: str) -> benchview.BuildInfo:
+def __get_build_info(args, target_framework_moniker: str) -> benchview.BuildInfo:
     # TODO: Improve complex scenarios.
     #   Could the --cli-* arguments take multiple build info objects from the
     #   command line interface?
@@ -275,11 +275,11 @@ def __get_build_info(args, framework: str) -> benchview.BuildInfo:
         source_timestamp = dotnet.get_commit_date(commit_sha, repository)
     elif args.cli_source_info == 'init-tools':
         # Retrieve data from the installed dotnet tools.
-        branch = micro_benchmarks.TargetFrameworkAction.get_channel(
-            framework
+        branch = micro_benchmarks.FrameworkAction.get_channel(
+            target_framework_moniker
         )
         if not branch:
-            err_msg = 'Cannot determine build information for "%s"' % framework
+            err_msg = 'Cannot determine build information for "%s"' % target_framework_moniker
             getLogger().error(err_msg)
             getLogger().error(
                 "Build information can be provided using the --cli-* options."
@@ -342,7 +342,8 @@ def __run_benchview_scripts(args: list, verbose: bool) -> None:
     benchviewpy.machinedata(working_directory=bin_directory)
 
     for framework in args.frameworks:
-        buildinfo = __get_build_info(args, framework)
+        target_framework_moniker = FrameworkAction.get_target_framework_moniker(framework)
+        buildinfo = __get_build_info(args, )
 
         # BenchView build.py
         benchviewpy.build(
@@ -358,7 +359,7 @@ def __run_benchview_scripts(args: list, verbose: bool) -> None:
         working_directory = dotnet.get_build_directory(
             bin_directory=bin_directory,
             configuration=args.configuration,
-            framework=framework,
+            target_framework_moniker=target_framework_moniker,
         )
 
         # BenchView measurement.py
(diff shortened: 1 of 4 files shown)


MarcoRossignoli:
@adamsitnik this https://github.com/dotnet/BenchmarkDotNet/pull/1001/files#diff-83177b9866652943fa5d5bb8f80d847aR91 could fix other issue?
However should be slower 2 fail(build/publish) before actual run.

scripts/micro_benchmarks.py
```diff
@@ -70,13 +70,43 @@ def __get_framework_channel_map() -> dict:
         }
 
     @staticmethod
-    def get_channel(framework: str) -> str:
+    def __get_framework_target_framework_moniker_map() -> dict:
```
jorive: nit: rename function to something more meaningful (`__get_corert_target_framework_mapping`)

scripts/micro_benchmarks.py
```diff
@@ -70,13 +70,43 @@ def __get_framework_channel_map() -> dict:
...
+        return {
+            # to run CoreRT benchmarks we need to run the host BDN process as latest .NET Core
```
jorive: nit: move this comment to the line above, as a function comment.

scripts/micro_benchmarks.py
```diff
@@ -220,10 +250,40 @@ def __process_arguments(args: list) -> Tuple[list, bool]:
     parser = add_arguments(parser)
     return parser.parse_args(args)
 
+def __getBenchmarkDotNetArguments(
```
jorive: nit: two lines between global function definitions.

scripts/micro_benchmarks.py
```diff
@@ -220,10 +250,40 @@ def __process_arguments(args: list) -> Tuple[list, bool]:
...
+        args: list,
+        framework: str
+        ) -> list:
```
jorive: Thanks for removing the duplicated code :)

scripts/benchmarks_ci.py
```diff
@@ -342,7 +342,8 @@ def __run_benchview_scripts(args: list, verbose: bool) -> None:
     benchviewpy.machinedata(working_directory=bin_directory)
 
     for framework in args.frameworks:
-        buildinfo = __get_build_info(args, framework)
+        target_framework_moniker = FrameworkAction.get_target_framework_moniker(framework)
+        buildinfo = __get_build_info(args, )
```
jorive: Missing second argument.

scripts/micro_benchmarks.py
```diff
@@ -220,10 +250,40 @@ def __process_arguments(args: list) -> Tuple[list, bool]:
...
+    if args.bdn_arguments:
+        run_args += args.bdn_arguments
+        
+    # we need to tell BenchmarkDotNet where to restore the packages
+    # if we don't it's gonna restore to default global folder
+    run_args += ['--packages', get_packages_directory()]
+    # required for CoreRT where host process framework != benchmark process framework
+    run_args += ['--runtimes', framework]
+        
+    return run_args
```
jorive: nit: add an extra line after function definition.

scripts/micro_benchmarks.py
```diff
@@ -220,10 +250,40 @@ def __process_arguments(args: list) -> Tuple[list, bool]:
     parser = add_arguments(parser)
     return parser.parse_args(args)
 
+def __getBenchmarkDotNetArguments(
+        args: list,
```
jorive: nit: make *args the last parameter.
```python
def __getBenchmarkDotNetArguments(framework: str, *args)
```

scripts/benchmarks_ci.py
adamsitnik: good catch! and a noobish bug on my side

scripts/micro_benchmarks.py
adamsitnik: this convention is crazy, but ok ;p

scripts/micro_benchmarks.py
adamsitnik: I decided to remove it and replace entire method with one if statement.

scripts/benchmarks_ci.py
jorive: I recommend using `pyflakes` when modifying the python scripts, and, ideally, open an issue to create a CI that run linters on this code.

scripts/micro_benchmarks.py
jorive: [PEP 8 -- Style Guide for Python Code](https://www.python.org/dev/peps/pep-0008/)

jorive: You can use linters to help on this too ;)

scripts/micro_benchmarks.py
```diff
@@ -251,9 +243,10 @@ def __process_arguments(args: list) -> Tuple[list, bool]:
     return parser.parse_args(args)
 
 def __getBenchmarkDotNetArguments(
-        args: list,
-        framework: str
+        framework: str,
+        args: list
```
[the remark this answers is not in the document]
adamsitnik: @jorive If I do so I get following error:

```log
AttributeError: 'tuple' object has no attribute 'category'
```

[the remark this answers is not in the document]
jorive: I think that in the caller site, you need to pass `args` instead of `*args`

adamsitnik:
@jorive I updated BDN version to include https://github.com/dotnet/BenchmarkDotNet/pull/1013

thanks to @wojtpl2 BDN is now smarter, and does same thing as the Python script (dotnet restore, dotnet build --no-restore, dotnet run --no-build --no-restore) and does NOT lock the files

I have tested that for .NET Core 3.0 (regular toolchain), CoreRun and CoreRT

I believe that this PR is ready to be merged ;)

jorive:
:shipit:

PR #228, opened 2019-01-16
Bug fixes and updating comments.

jorive:
- https://github.com/dotnet/performance/pull/218 introduced this bug:
```log
target_framework_moniker = FrameworkAction.get_target_framework_moniker(framework)
NameError: name 'FrameworkAction' is not defined
```
- Pull released `dotnet-install` script instead of latest from `dotnet/cli:master`.
- Minor fixing spacing and follow naming conventions: two lines between global functions, and camel case naming.
- Updated TODO/FIXME comments on the scripts.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -314,8 +314,9 @@ def __run_benchview_scripts(args: list, verbose: bool) -> None:
     bin_directory = micro_benchmarks.BENCHMARKS_CSPROJ.bin_path
 
     # BenchView submission-metadata.py
+    # TODO: Simplify logic. This should be removed and unify repo data.
     submission_name = args.benchview_submission_name
-    is_pr = args.benchview_run_type == 'private' and\
+    is_pr = args.benchview_run_type == 'private' and \
         'BenchviewCommitName' in os.environ
     rolling_data = args.benchview_run_type == 'rolling' and \
         'GIT_BRANCH_WITHOUT_ORIGIN' in os.environ and \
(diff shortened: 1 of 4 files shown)


scripts/benchmarks_ci.py
```diff
@@ -377,13 +380,14 @@ def __run_benchview_scripts(args: list, verbose: bool) -> None:
     if 'Configuration' not in benchview_config:
         benchview_config['Configuration'] = args.configuration
 
-    # Generate existing configs. This may be a good time to unify them?
+    # Generate existing configs.
+    # TODO: Unify configs across all repos?
     submission_architecture = args.architecture
     if args.category.casefold() == 'CoreClr'.casefold():
-        benchview_config['JitName'] = 'ryujit'  # This is currently fixed.
+        benchview_config['JitName'] = 'ryujit'  # FIXME: Remove this.
```
billwert: why aren't we tracking things like this with issues (which would be more discoverable)?

billwert:
LGTM. I'm curious why you've chosen some places to fix line breaks on when the file doesn't seem consistent. For example [here](https://github.com/dotnet/performance/blob/d2500c3aca9cbf8be0a290dd143c2dfcc2edf372/scripts/benchmarks_ci.py#L70) and [here](https://github.com/dotnet/performance/blob/d2500c3aca9cbf8be0a290dd143c2dfcc2edf372/scripts/benchmarks_ci.py#L278). If we're going to enforce style things is there a linter we should be running as a check?

scripts/benchmarks_ci.py
jorive: We should, but I would like to talk about it first.

jorive:
@billwert I always try to run a passive checker for Python programs, a code static checker, and a style guide checker on the scripts before checking them in.

Moving forward I would like to add a CI to perform this checks too, and I could also checkin a sort of "template" for coding style tools to consume.

https://github.com/dotnet/performance/blob/d2500c3aca9cbf8be0a290dd143c2dfcc2edf372/scripts/benchmarks_ci.py#L70 did not show up in my initial lint/codestyle tests, but I could fix it (Generally I only use the default autoformatter from the python extension on vscode, and try to follow some of the rules such as spacing between functions).

https://github.com/dotnet/performance/blob/d2500c3aca9cbf8be0a290dd143c2dfcc2edf372/scripts/benchmarks_ci.py#L278 this is correct.

PR #173, opened 2018-12-06
Port netcoreapp3.0 Windows performance jobs to AzureDevOps

adiaaida:

PR #234, opened 2019-01-22
Default value for the new argument which broke the script

adamsitnik:
#173 introduced new argument called `--bin-directory`. The problem is that when we don't provide it, the script fails. The solution was to add the default value.

```log
  File ".\scripts\benchmarks_ci.py", line 524, in <module>
    __main(sys.argv[1:])
  File ".\scripts\benchmarks_ci.py", line 501, in __main
    csproj_file='MicroBenchmarks.csproj'
  File "C:\Projects\demo\scripts\dotnet.py", line 54, in __init__
    if not bin_directory and not path.isdir(bin_directory):
TypeError: _isdir: path should be string, bytes or os.PathLike, not NoneType
```

### DIFF
--- a/scripts/micro_benchmarks.py
+++ b/scripts/micro_benchmarks.py
@@ -245,7 +245,7 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         '--bin-directory',
         dest='bin_directory',
         required=False,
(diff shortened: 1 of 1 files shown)


jorive:
:shipit:

PR #237, opened 2019-01-23
Added enum performance benchmarks

TylerBrinkley:
Added enum performance benchmarks for [coreclr#22161](https://github.com/dotnet/coreclr/pull/22161).

Benchmarks requiring undefined enum values box the value due to [BenchmarkDotNet#1020](https://github.com/dotnet/BenchmarkDotNet/issues/1020).


src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
```diff
@@ -20,19 +22,534 @@ public enum Colors
...
+        [Benchmark]
+        [Arguments(typeof(BigContiguousEnum), "ABC")]
+        [Arguments(typeof(SmallContiguousEnum), "ABC")]
+        public bool TryParseMissingNonGeneric(Type enumType, string value) => Enum.TryParse(enumType, value, out _);
+        #endregion
+
+        #region HybridSearchCutoff
+        // Last found cutoff was 32
+
+        //[Benchmark]
```
billwert: is it intentional that the benchmark is commented out here?

src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
```diff
@@ -20,19 +22,534 @@ public enum Colors
...
+                }
+                if (values[i] < value)
+                {
+                    return ~(i + 1);
+                }
+            }
+            return -1; // == ~0
+        }
+
+        //[Benchmark]
```
billwert: is it intentional that the benchmark is commented out here?

src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
TylerBrinkley: Yes, it was used to determine at what length of array a linear search is more performant over a binary search. It's not something that would provide much value except for determining the constant which should probably be manually invoked.

src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
```diff
@@ -20,19 +22,534 @@ public enum Colors
...
+        public string ToStringFlags5() => (Colors.Red | Colors.Orange | Colors.Yellow | Colors.Green | Colors.Blue).ToString();
+
+        [Benchmark]
+        //[Arguments(Colors.Orange | Colors.Yellow | Colors.Green)] // BenchmarkDotNet Issue #1020
+        //public string ToStringFlags3(Enum value) => value.ToString();
+        public string ToStringFlags3() => (Colors.Orange | Colors.Yellow | Colors.Green).ToString();
+
+        [Benchmark]
+        [Arguments(Colors.Red)]
+        public string ToStringFlags1(Enum value) => value.ToString();
```
adamsitnik: please update to `0.11.4.1025` and merge `ToStringFlags5`, `ToStringFlags3` and `ToStringFlags1` into one benchmark:

```cs
[Benchmark]
[Arguments(Colors.Red | Colors.Orange | Colors.Yellow | Colors.Green | Colors.Blue)]
[Arguments(Colors.Orange | Colors.Yellow | Colors.Green)]
[Arguments(Colors.Red)]
public string ToStringFlags(Enum value) => value.ToString();
```

The benchmark ID contains all the argument values so there is no need to have a separate method for that

src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
```diff
@@ -20,19 +22,534 @@ public enum Colors
...
+        public string FormatF(Type enumType) => Enum.Format(enumType, Colors.Green | Colors.Yellow | Colors.Orange, "F");
+        #endregion
+
+        [Benchmark]
+        [Arguments(BigContiguousEnum.C, BigContiguousEnum.E)]
+        public int CompareToEnum(Enum value, object target) => value.CompareTo(target);
+
+        [Benchmark]
+        [Arguments(BigContiguousEnum.C)]
+        public int CompareToNull(Enum value) => value.CompareTo(null);
```
adamsitnik: what time does this benchmark report? Is it `0ns`?

src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
```diff
@@ -20,19 +22,534 @@ public enum Colors
...
+        #region Parsing
+        [Benchmark]
+        [Arguments(typeof(Colors))]
+        public object ParseFlagsNonGeneric(Type enumType) => Enum.Parse(enumType, "Red, Orange, Yellow, Green, Blue");
+
+        [Benchmark]
+        public Colors ParseFlagsGeneric() => Enum.Parse<Colors>("Red, Orange, Yellow, Green, Blue");
+
+        [Benchmark]
+        [Arguments(typeof(ByteEnum), "Two")]
```
adamsitnik: could you please use "nameof" here and in the examples below?
```suggestion
        [Arguments(typeof(ByteEnum), nameof(ByteEnum.Two))]
```

src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
```diff
@@ -20,19 +22,534 @@ public enum Colors
...
+        [Arguments(typeof(UInt16Enum))]
+        [Arguments(typeof(Int16Enum))]
+        [Arguments(typeof(UInt32Enum))]
+        [Arguments(typeof(Int32Enum))]
+        [Arguments(typeof(UInt64Enum))]
+        [Arguments(typeof(Int64Enum))]
+        public object ParseValueNonGeneric(Type enumType) => Enum.Parse(enumType, "2");
+
+        [Benchmark]
+        public ByteEnum ParseNameGenericByte() => Enum.Parse<ByteEnum>("Two");
```
TylerBrinkley: It'd be nice if Benchmark.NET supported specifying generic type arguments on methods. With that change we could collapse all of these related methods as we did with the `ParseNameNonGeneric` benchmark. I think it would require adding a `GenericTypeArguments` property on `ArgumentsAttribute`.

src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
TylerBrinkley: `~2ns`

src/benchmarks/micro/corefx/System.Runtime/Perf.Enum.cs
```diff
@@ -20,19 +22,534 @@ public enum Colors
...
+
+        public enum Int64Enum : long
+        {
+            Min = long.MinValue,
+            One = 1,
+            Two = 2,
+            Max = long.MaxValue
+        }
+
+        // 36 members
```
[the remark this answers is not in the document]
TylerBrinkley: In https://github.com/dotnet/coreclr/pull/22161 I've introduced a hybrid search which for enums with more than 32 members uses binary search but for enums with less than or equal to 32 members uses linear search. This number was determined in the now removed benchmarks for determining which method was faster in the worst case for linear search. I've added a comment briefly explaining this.

TylerBrinkley:
Closing due to age.

PR #246, opened 2019-01-29
make netcoreapp3.0 default TFM to simplify command line

adamsitnik:
It's part of making the docs and process simpler after the demo for @vancem

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -61,4 +61,9 @@
     <Compile Remove="corefx\System.Security.Cryptography\Perf.Rfc2898DeriveBytes.cs" />
   </ItemGroup>
 
+  <Target Name="BuildConfigurationCheck" BeforeTargets="Build">
+    <Warning Condition="'$(Configuration)'=='Debug' AND '$(TargetFramework)' == $(DefaultTargetFramework)" 
+        Text="Configuration and Framework were not provided, optimized netcoreapp3.0 build will be used (defaults)." />
+  </Target>
+
 </Project>
--- a/docs/benchmarking-workflow.md
+++ b/docs/benchmarking-workflow.md
@@ -79,7 +79,7 @@ Sample commands:
 - See the list of all available .NET Core 2.1 System.Linq benchmarks:
 
 ```cmd
-dotnet run -f netcoreapp2.1 -- -f System.Linq* --list flat
+dotnet run -f netcoreapp2.1 --filter System.Linq* --list flat
 ```
 
 ```log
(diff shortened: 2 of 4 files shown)


jorive:
Why are we removing the `--` argument to the additional arguments?

```cmd
dotnet.exe run --help
Usage: dotnet run [options] [[--] <additional arguments>...]]

Options:
  -h, --help                            Show command line help.
  -c, --configuration <CONFIGURATION>   The configuration to run for. The default for most projects is 'Debug'.
  -f, --framework <FRAMEWORK>           The target framework to run for. The target framework must also be specified in the project file.
  -r, --runtime <RUNTIME_IDENTIFIER>    The target runtime to run for.
  -p, --project                         The path to the project file to run (defaults to the current directory if there is only one project).
  --launch-profile                      The name of the launch profile (if any) to use when launching the application.
  --no-launch-profile                   Do not attempt to use launchSettings.json to configure the application.
  --no-build                            Do not build the project before running. Implies --no-restore.
  --no-restore                          Do not restore the project before building.
  -v, --verbosity <LEVEL>               Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
  --no-dependencies                     Do not restore project-to-project references and only restore the specified project.
  --force                               Force all dependencies to be resolved even if the last restore was successful.
                                        This is equivalent to deleting project.assets.json.
Additional Arguments:
  Arguments passed to the application that is being run.
```

adamsitnik:
Some time ago `--` became optional for additional arguments. I removed it where it was not mandatory to simplify the instructions

jorive:
I have a proposition, where we change the default behavior of the dotnet cli tools we should let end user know what is happening behind the scenes.
We could add this:
```xml
  <Target Name="BuildConfigurationCheck" BeforeTargets="Build">
    <Warning Condition="'$(Configuration)'=='Debug'" Text="+++ Some warning message +++" />
  </Target>
```
The same for the default framework, but instead of being a `Warning`.
Thoughts?

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -61,4 +61,9 @@
     <Compile Remove="corefx\System.Security.Cryptography\Perf.Rfc2898DeriveBytes.cs" />
   </ItemGroup>
 
+  <Target Name="BuildConfigurationCheck" BeforeTargets="Build">
```
adamsitnik: @jorive I added the warning

obraz

I did not add two separate warnings because it's not possible to determine when the user did not provide the `--framework` and when it did in an explicit way `-f netcoreapp3.0`. I hope that this trade-off is acceptable.

Issue #257, opened 2019-02-07
Add CI correctness jobs for .NET, Core 2.0/2.1/2.2


adamsitnik:
In #246 I have introduced a bug which did not allow to run the benchmarks easily for .NET Core 2.0/2.1/2.2

We need to have a correctness check for not only 3.0 but for the other frameworks as well.

PR #260, opened 2019-02-07
for the correctness check, run the benchmarks exactly once, no warmup, no pilot, no overhead

### LINKED ISSUE #258: The correctness CI job should run every benchmark exactly once (not more)
As of today, the correctness CI job runs with the following arguments:

```cmd
--iterationCount 2
```

While it should be:

```cmd
--iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart
```

The benefit will be much faster correctness check

adamsitnik:
Fixes #258

BDN change and update was required because the possibility to disable overhead from console arguments level was missing  (https://github.com/dotnet/BenchmarkDotNet/commit/20744eaa4f38622f5350b582ec4c44a906341db7)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,7 +34,7 @@ jobs:
           - name: Creator
             value: dotnet-performance
           - name: BDNArguments
-            value: --iterationCount 2
+            value: --iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart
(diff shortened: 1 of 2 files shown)


adamsitnik:
@adiaaida the time dropped from 1h to 12/14 minutes. I am going to merge this PR with no review since it's a very small change, the CI is green and it's blocking me with #257 and #259

PR #261, opened 2019-02-07
add the correctness checks for .NET Core 2.0/2.1/2.2 and full .NET framework

### LINKED ISSUE #257: Add CI correctness jobs for .NET, Core 2.0/2.1/2.2

adamsitnik:
Fixes #257

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -65,51 +65,61 @@ jobs:
               _Category: CoreCLR
               _Framework: netcoreapp3.0
               _BuildConfig: $(_Architecture)_$(_Category)_$(_Framework)
+              _BenchViewArguments: '--benchview-submission-name "$(BenchviewCommitName)" --benchview-machinepool perfsnake --generate-benchview-data --benchview-run-type $(_RunType) $(UploadToBenchview)'
             x86_CoreCLR_netcoreapp30:
               _Architecture: x86
               _Category: CoreCLR
               _Framework: netcoreapp3.0
               _BuildConfig: $(_Architecture)_$(_Category)_$(_Framework)
+              _BenchViewArguments: '--benchview-submission-name "$(BenchviewCommitName)" --benchview-machinepool perfsnake --generate-benchview-data --benchview-run-type $(_RunType) $(UploadToBenchview)'
             x64_CoreCLR_netcoreapp22:
               _Architecture: x64
               _Category: CoreCLR
               _Framework: netcoreapp2.2
               _BuildConfig: $(_Architecture)_$(_Category)_$(_Framework)
+              _BenchViewArguments: ''
             x86_CoreCLR_netcoreapp22:
               _Architecture: x86
               _Category: CoreCLR
               _Framework: netcoreapp2.2
               _BuildConfig: $(_Architecture)_$(_Category)_$(_Framework)
+              _BenchViewArguments: ''
             x64_CoreCLR_netcoreapp21:
               _Architecture: x64
               _Category: CoreCLR
               _Framework: netcoreapp2.1
               _BuildConfig: $(_Architecture)_$(_Category)_$(_Framework)
+              _BenchViewArguments: ''
             x86_CoreCLR_netcoreapp21:
               _Architecture: x86
               _Category: CoreCLR
               _Framework: netcoreapp2.1
               _BuildConfig: $(_Architecture)_$(_Category)_$(_Framework)
+              _BenchViewArguments: ''
             x64_CoreCLR_netcoreapp20:
               _Architecture: x64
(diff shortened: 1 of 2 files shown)


adamsitnik:
@adiaaida @jorive PTAL

obraz

scripts/dotnet.py
```diff
@@ -181,10 +181,22 @@ def get_host_commit_sha(dotnet_path: str = None) -> str:
...
+        # Product Information:
+        #   Version:            2.1.202
+        #   Commit SHA-1 hash:  281caedada
         if 'Host' in decoded_line:
             foundHost = True
+        elif 'Product' in decoded_line:
+            foundHost = True
         elif foundHost and 'Commit' in decoded_line:
-            return decoded_line.strip().split()[1]
+            return decoded_line.strip().split()[-1]
```
adamsitnik: [-1] is the last element. the string gets trimmed first, so this will always be our sha

azure-pipelines.yml
```diff
@@ -65,11 +65,61 @@ jobs:
...
+              _Category: CoreCLR
+              _Framework: net461
+              _BuildConfig: $(_Architecture)_$(_Category)_$(_Framework)
+              _BenchViewArguments: ''
+            x86_CoreCLR_full_framework:
+              _Architecture: x86
+              _Category: CoreCLR
+              _Framework: net461
+              _BuildConfig: $(_Architecture)_$(_Category)_$(_Framework)
+              _BenchViewArguments: ''
```
adamsitnik: @adiaaida I know that adding so many elements to the matrix looks a little bit lame, but I have no better idea. I am open to suggestions

adiaaida: This is expected. I think it looks fine.

Only comment here is I think we should add some comments for like-blocks of definitions. So like above the netcoreapp3.0 ones, a comment saying # netcoreapp3.0 definitions; above the 2.2 # netcoreapp2.2, etc.

adiaaida:
The only other thing I could think of is that there is an each keyword. And example can be found here: https://dnceng.visualstudio.com/internal/_git/dotnet-optimization?path=%2Feng%2Fcollect-optimization-data.yml&version=GBmaster&line=118&lineStyle=plain&lineEnd=119&lineStartColumn=1&lineEndColumn=1

Basically, that goes over the list of artifacts that we've specified in azure-pipelines.yml and pulls them down. The artifact list is specified here: https://dnceng.visualstudio.com/internal/_git/dotnet-optimization?path=%2Fazure-pipelines.yml&version=GBmaster&line=51&lineStyle=plain&lineEnd=52&lineStartColumn=1&lineEndColumn=1

However, I'm not sure if that would work for a matrix, which might not like each. But if we could have some parameter that had a list, say like this:

```
frameworks:
  - name: netcoreapp3.0
  - name: netcoreapp2.2
  - name: netcoreapp2.1
  - name: net461
```

and then above the matrix:

```
${{ each framework in frameworks }}:
  ,,,
  _Framework: ${{ framework.name }}
```

The problem is the naming of each matrix option, which I don't know if can be parameterized. @chcosta would have a better answer for that.

scripts/dotnet.py
jorive: The difference between using element in position 1 vs using the last element is that the current dotnet cli have two tokens and we select the second one as being the SHA. In the future, if this changes, we do not know what is the last token being selected. I would suggest leaving it as-is.

adamsitnik: I think that using the last element is actually more future-proof

For .NET Core 2.0 we had:

```log
Commit SHA-1 hash:  281caedada
```

For .NET Core 2.1+ we have:

```log
Commit:  7a7ca06512
```

I think that if it's ever going to change, somebody is going to add text before the SHA, not after it. Also, now we have a CI that validates that, so if it ever breaks we will know about this with the first broken PR. Maestro sends few PRs a day so we are going to find this out very quickly ;)

adamsitnik:
@adiaaida thank you for the suggestions! I tried to google for some more complex matrix examples but all I could find was one for debug and one for release or one for x64 and one for x86.

I am going to merge this as-is and add a comments and try to use the for loops in my next PR where I will add CoreFX to the list.

scripts/dotnet.py
jorive: Maybe moving forward we should use a regex to match this information. Something along the lines of: `^\W+Commit:\W+\b[0-9a-f]{5,40}\b$`.

adamsitnik: that would be most probably the best solution!

PR #263, opened 2019-02-07
[WIP] Add CoreFX

adamsitnik:
I need to check what fails and get it up and working


jorive:
I feel we should have different functions to retrieve different information, or have a more appropriate parser of the `dotnet --info` output.

PR #277, opened 2019-02-12
Fix environment variable usage in non-windows jobs

adiaaida:


adamsitnik:
@adiaaida thank you for fixing this!

PR #280, opened 2019-02-13
copy python script logs to a folder that gets uploaded by the CI

adamsitnik:


jorive:
Do you need to copy? Cannot archive the logs folder itself?

adamsitnik:
This is what gets published: https://github.com/dotnet/performance/blob/master/eng/common/templates/job/job.yml#L174

btw thanks to @safern for explaining me how it works!

jorive:
I'm not an expert on yml files, but shouldn't you be using `$PathtoPublish` as the destination path for both batch and bash?

adamsitnik:
I think that here `PathtoPublish` is a name of the parameter for the task (I am not an expert myself)

adamsitnik:
It's not going to work because it's two different machines: one is scheduling the job (and can upload), another one is running the job and producing the results (can't upload in an easy way)

PR #288, opened 2019-02-13
Set upload variable as helix precommand

adiaaida:


eng/performance/benchmark_jobs.yml
```diff
@@ -47,6 +49,12 @@ jobs:
             value: ''
           - name: Creator
             value: ''
+          ${{ if eq(parameters.osName, 'windows') }}:
+            - name: HelixPreCommand
+              value: 'set BV_UPLOAD_SAS_TOKEN=$(BV_UPLOAD_SAS_TOKEN)'
+          ${{ if ne(parameters.osName, 'windows') }}:
+            - name: HelixPreCommand
+              value: 'BV_UPLOAD_SAS_TOKEN=$(BV_UPLOAD_SAS_TOKEN)'
```
jorive: Would you need to export it like in bash (I'm not sure how this is handled in yml/helix)?

adiaaida: I am not sure. Possibly. I tried on my mac and just setting it like that was enough, but I expect this might take a small amount of trial and error.

adiaaida: Maybe I'll just export it to be safe

adiaaida:
I have tested this by using a dummy variable + hello world job in a different PR to make sure that setting the variable like this will work for at least windows. The precommand will set the variable, and when echo'd the variable will print the expected output.

PR #300, opened 2019-02-15
Change linux token

adiaaida:
For msbuild/batch/bash reasons, we need to have separate tokens for Linux and Windows. Use the linux token for non-windows jobs.


adiaaida:
Really really close, I think.

PR #301, opened 2019-02-16
Give the MathTests benchmarks a more meaningful name.

jorive:
## What is in this PR at the moment:

- Updated namespace from: `Functions`, to `MathTests.FloatingPointTests`
- Moved Double precision tests from the `MathTests` class to the `DoublePrecisionTests` class (maybe it should be `DoublePrecisionTranscendentalsTests`)
- Moved Single precision tests from the `MathTests` class to the `SinglePrecisionTests` class (maybe it should be `SinglePrecisionTranscendentalsTests`)
- Removed unused constants.

## What is missing:

- There is still a lot of code that could be simplified/cleaned up (a lot of duplication and using inner iterations).

### DIFF
--- a/src/benchmarks/micro/coreclr/Math/Functions/DivRem.cs
+++ b/src/benchmarks/micro/coreclr/Math/Functions/DivRem.cs
@@ -5,7 +5,7 @@
 using System;
 using BenchmarkDotNet.Attributes;
 
-namespace MathTests.FloatingPointTests
+namespace Transcendental
 {
     public partial class MathTests
     {
--- a/src/benchmarks/micro/coreclr/Math/Functions/DivRem.cs
+++ b/src/benchmarks/micro/coreclr/Math/Functions/DivRem.cs
@@ -5,7 +5,7 @@
 using System;
 using BenchmarkDotNet.Attributes;
 
-namespace Functions
+namespace MathTests.FloatingPointTests
 {
     public partial class MathTests
     {
--- a/src/benchmarks/micro/coreclr/Math/Functions/DivRem.cs
+++ b/src/benchmarks/micro/coreclr/Math/Functions/DivRem.cs
@@ -5,7 +5,7 @@
 using System;
 using BenchmarkDotNet.Attributes;
 
-namespace Transcendental
+namespace System.MathBenchmarks
 {
     public partial class MathTests
     {
--- a/src/benchmarks/micro/coreclr/Math/Functions/Double/AbsDouble.cs
+++ b/src/benchmarks/micro/coreclr/Math/Functions/Double/AbsDouble.cs
@@ -4,34 +4,36 @@
 
 using System;
 using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
 
-namespace Functions
+namespace MathTests.FloatingPointTests
 {
-    public partial class MathTests
+    [BenchmarkCategory(Categories.CoreCLR)]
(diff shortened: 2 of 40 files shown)


billwert:
Why wouldn't we follow our guidance and make these mimic the namespace under test? So ```System.Math.<stuff>```. I can get behind keeping ```DoublePrecisionTests``` and ```SinglePrecisionTests```.

billwert:
Hm. I wonder if we should also take this opportunity to rework the tests. The correctness check isn't necessary. This would let us simplify these benchmarks down to just the call to the actual functions being tested, keeping whichever constants are necessary for input to those.

jorive:
@billwert w.r.t. the naming convention `System.Math` it's only used on CoreFx tests, but I'm open to change it here too. The `CoreClr` tests follow a different "convention". They are group under namespaces that describe the type of benchmarks, not necessarily under the library they are testing.

billwert:
Ahh. I missed the path. Sorry!

Up to you. I think it’s nice for things that are directly exposed in the BCL to follow the pattern. Clearly harder for things that target runtime features in more abstract ways.

Thanks,
-Bill

adamsitnik:
I really like this PR but I think that we still have some place for improvement.

The first one is naming. Perhaps the namespace could be `System.Math` the type names: `Double` and `Single` and the method names just the names of the operations?

An example: `System.Math.Double.Abs`

@jorive @billwert what do you think?

adamsitnik:
The other problem is the "pattern" that each benchmark follows here:

```cs
var result = 0.0; var value = -1.0;

for (var iteration = 0; iteration < MathTests.Iterations; iteration++)
{
    value += delta;
    result += Math.SomeOperation(value);
}
```

I think that the reported time for some of the short operations here like Abs can be dominated by the cost of the loop and additional add operation. Moreover we can get into loop body aligment issues and the results are not scaled by BDN as they could be. 

But getting rid of loops might not be easy, because JIT can optimize some of the Math method calls using constant folding and other optimizations.

I need to think about it for a while.

Maybe @AndreyAkinshin has some ideas?

jorive:
@adamsitnik w.r.t. "the namespace: System.Math", `System.Math` is a static class, and you are suggesting to use it as a namespace. Is that appropriate? Does it make sense?
On the same note: [CA1724: Type names should not match namespaces](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1724-type-names-should-not-match-namespaces?view=vs-2017).

adamsitnik:
> Type names should not match namespaces.

You are right, I am sorry, I forgot about it. This is why all the CoreFX tests belong to `System.XYZ.Tests` namespaces.

Speaking of this PR it changes the IDs of the benchmarks and you recently told me that CoreCLR Team wants to have historical data continuation in BenchView. Does it mean that we should not change the names?

jorive:
> Type names should not match namespaces.
> 
> You are right, I am sorry, I forgot about it. This is why all the CoreFX tests belong to System.XYZ.Tests namespaces.
> Speaking of this PR it changes the IDs of the benchmarks and you recently told me that CoreCLR Team wants to have historical data continuation in BenchView. Does it mean that we should not change the names?
> Edit: I believe that this change is a good idea, I am just not sure when is the best time to introduce it

I feel this is the best time to make the change, because the benchmarks measurements themselves will also change.

jorive:
ping

billwert:
LGTM.

PR #303, opened 2019-02-18
[WIP] We want to fail when any of our work items fail

adamsitnik:
I read the following Arcade docs https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Helix/Sdk/Readme.md#all-possible-options and it looks like setting FailOnMissionControlTestFailure to true should give us what we want


adamsitnik:
We need https://github.com/dotnet/arcade/pull/2045 to be published to the NuGet feed.

The latest microsoft.dotnet.helix.sdk, 2.0.0-beta.19116.2 does not contain it (yet)

obraz

PR #319, opened 2019-02-21
Set copyright and package license

tmat:
Prepares repo for change https://github.com/dotnet/arcade/pull/2003 by setting `Copyright` and `PackageLicenseExpression` properties. These values will be required to be set by each repository once https://github.com/dotnet/arcade/pull/2003 is merged.

In order to not break the current builds this change sets the properties conditionally. This condition can be removed once all repos switch to Arcade that has https://github.com/dotnet/arcade/pull/2003.

@markwilkie

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,16 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
-  <PropertyGroup>
-    <ImportNetSdkFromRepoToolset>false</ImportNetSdkFromRepoToolset>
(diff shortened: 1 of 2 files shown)


adamsitnik:
@tmat should we just merge it now or wait until [dotnet/arcade#2003](https://github.com/dotnet/arcade/pull/2003)?

tmat:
@adamsitnik Needs to be merge _before_ https://github.com/dotnet/arcade/pull/2003. Thanks!

PR #320, opened 2019-02-22
Run all benchmarks for both tiered and full_opt optimization levels

adamsitnik:
This PR changes the way we run the benchmarks:

1. No separation for CoreFX and CoreCLR benchmarks, we run all benchmarks and export them to single BenchView container
2. Run the benchmarks not only for optimization-level `tiered` but also `full_opt`

I have also unified the reported BenchView config to: OS name + Optimization Level + Profile

### DIFF
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -69,10 +69,11 @@ jobs:
                 _Framework: ${{ framework }}
                 _OptimizationLevel: full_opt
                 _BuildConfig: ${{ parameters.architecture }}_$(_OptimizationLevel)_$(_Framework)
-              tiered_${{ framework }}:
-                _Framework: ${{ framework }}
-                _OptimizationLevel: tiered
-                _BuildConfig: ${{ parameters.architecture }}_$(_OptimizationLevel)_$(_Framework)
+              ${{ if and(startsWith(framework, 'netcoreapp'), ne(framework, 'netcoreapp2.0')) }}: # Tiered JIT was introduced in .NET Core 2.1
+                tiered_${{ framework }}:
+                  _Framework: ${{ framework }}
+                  _OptimizationLevel: tiered
+                  _BuildConfig: ${{ parameters.architecture }}_$(_OptimizationLevel)_$(_Framework)
         steps:
         - checkout: self
           clean: true
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -85,7 +85,7 @@ jobs:
             HelixTargetQueues: ${{ parameters.queue }}
             HelixPreCommands: $(HelixPreCommand)
             Creator: $(Creator)
-            WorkItemTimeout: 4:00 # 4 hours
+            WorkItemTimeout: 5:00 # 5 hours
             ${{ if eq(parameters.osName, 'windows') }}:
               WorkItemDirectory: '$(Build.SourcesDirectory)\src\benchmarks\micro'
               WorkItemCommand: 'py -3 %HELIX_CORRELATION_PAYLOAD%\benchmarks_ci.py --bin-directory ".\bin" --working-directory . --optimization-level $(_OptimizationLevel) --incremental no --filter * --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -43,7 +43,7 @@ jobs:
             value: '.NET Performance - $(BenchViewRunType) - $(Build.SourceBranchName) $(Build.SourceVersion)'
           # for private runs we want to generate data (--generate-benchview-data ) and upload it (--upload-to-benchview-container specifies where)
           - name: BenchViewArguments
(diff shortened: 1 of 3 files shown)


eng/performance/benchmark_jobs.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             value: '.NET Performance - $(BenchViewRunType) - $(Build.SourceBranchName) $(Build.SourceVersion)'
           # for private runs we want to generate data (--generate-benchview-data ) and upload it (--upload-to-benchview-container specifies where)
           - name: BenchViewArguments
-            value: '--benchview-submission-name "$(BenchviewCommitName)" --benchview-machinepool perfsnake --generate-benchview-data --benchview-run-type $(BenchViewRunType) --upload-to-benchview-container $(_Category)'
+            value: '--benchview-submission-name "$(BenchviewCommitName)" --benchview-machinepool perfsnake --generate-benchview-data --benchview-run-type $(BenchViewRunType) --upload-to-benchview-container all'
```
adamsitnik: @jorive : here I have changed the name of the `--upload-to-benchview-container` from `coreclr` or `corefx` to `all`. Is this OK?

adiaaida: I'd worry about that as we do not currently have an "all" container in the storage account

eng/performance/benchmark_jobs.yml
```diff
@@ -65,14 +65,15 @@ jobs:
         strategy:
           matrix:
             ${{ each framework in parameters.frameworks }}:
-              coreclr_${{ framework }}:
-                _Category: coreclr
+              full_opt_${{ framework }}:
```
adiaaida: I worry about merging coreclr and corefx tests all into one leg. The perf legs that run the full suite with coreclr and corefx separate already run for up to 2 hours (coreclr) and up to 3 hrs (corefx). Merging them together is going to make us go up to 5 hours. At the very least, you'll want to up the workitemtimeout below. But really, I think we should keep them as distributed as possible.

billwert:
@adiaadia has a good point - being done in 3 hours is better than 5. My biggest concern is a) that all the tests run in an identical config and b) that they all eventually show up as the same “run” in the reporting tools. Whether they run on one or two or ten legs is less interesting to me.

Thanks,
-Bill

eng/performance/benchmark_jobs.yml
jorive: @adamsitnik coreclr and corefx are different containers/blob storages with their own SAS tokens. You cannot add any random name.

jorive: Maybe it makes sense having a single blob for all repos? Personally, I always felt it was more manageable having different blobs per repo/feature teams.

eng/performance/benchmark_jobs.yml
jorive: I agree with @adiaaida here. Do not run all tests in the same leg.

eng/performance/benchmark_jobs.yml
billwert: What's the value prop for having the blobs be separate? 

I think from an execution and uploading standpoint, they should be the same. I think it's useful to have all the results for a single "run" be the same, and then have filters in the reporting if folks want to exclude stuff. I think it's *hugely* useful if by default results for CoreFX tests show up for a CoreCLR dev and vice versa - it's possible that those tests get affected by some change and knowing that ahead of time is great.

jorive: Sorry, maybe I did not explain myself properly. How many blobs we use to upload data has nothing to do with grouping and displaying all benchmarks together. It is about managing how much data is generated by different repos with the different permutations on how the benchmarks run, and know how much we spend on Azure storage. It's for monitoring, but there are other ways you could do so.

adiaaida: For the record, it is possible to have the same SAS token for different blobs, as that is what we are doing out of the perf repo (we push to both the coreclr blob/container and the corefx blob/container with the same SAS token).

adamsitnik:
> I'm doing the CoreClr/CoreFx specific configuration names cleanup, and updating the "optimization levels" in a different PR. Please hold on these changes.

Ok!

eng/performance/benchmark_jobs.yml
adiaaida: Can we do a timing comparison between the merged set and just corefx? Are these the 400 longest running benchmarks or the 400 shortest running benchmarks?

adamsitnik: @adiaaida I am going to run all of them tonight and report the results tomorrow

adamsitnik:
Based on #328, with Tiered Jit enabled:

It takes 2h to run CoreCLR benchmarks only

```log
Global total time: 02:02:13 (7333.99 sec), executed benchmarks: 748
```

It takes 3h 23 minutes to run CoreFX benchmarks only

```log
Global total time: 03:23:51 (12231.62 sec), executed benchmarks: 1498
```

To run all microbenchmarks we need 4h 32 minutes:

```log
Global total time: 04:32:25 (16345.61 sec), executed benchmarks: 1942
```

adamsitnik:
With Tiered JIT disabled:

1:48 for CoreCLR

```log
Global total time: 01:48:46 (6526.58 sec), executed benchmarks: 748
```

3:16 for CoreFX

```log
Global total time: 03:16:52 (11812.08 sec), executed benchmarks: 1498
```

3:56 for all benchmarks

```log
Global total time: 03:56:30 (14190.23 sec), executed benchmarks: 1942
```

adamsitnik:
I am closing this one, new PR #340

PR #323, opened 2019-02-24
[Comparer]Take last result 

MarcoRossignoli:
When you specify result folders and not -full.json file and you've more than one results comparer throw.
My strategy is to take last result so it's not a problem if you specify full path or directory.

/cc @adamsitnik @jorive

### DIFF
--- a/src/tools/ResultsComparer/Program.cs
+++ b/src/tools/ResultsComparer/Program.cs
@@ -102,8 +102,8 @@ namespace ResultsComparer
 
         private static IEnumerable<(string id, Benchmark baseResult, Benchmark diffResult)> ReadResults(CommandLineOptions args)
         {
-            var baseFiles = GetFilesToParse(args.BasePath);
-            var diffFiles = GetFilesToParse(args.DiffPath);
+            var baseFiles = GetFilesToParse(args.BasePath).OrderBy(f => new FileInfo(f).CreationTimeUtc).TakeLast(1);
+            var diffFiles = GetFilesToParse(args.DiffPath).OrderBy(f => new FileInfo(f).CreationTimeUtc).TakeLast(1);
 
             if (!baseFiles.Any() || !diffFiles.Any())
                 throw new ArgumentException($"Provided paths contained no {FullBdnJsonFileExtension} files.");
--- a/src/tools/ResultsComparer/Program.cs
+++ b/src/tools/ResultsComparer/Program.cs
@@ -102,11 +102,14 @@ namespace ResultsComparer
 
         private static IEnumerable<(string id, Benchmark baseResult, Benchmark diffResult)> ReadResults(CommandLineOptions args)
         {
-            var baseFiles = GetFilesToParse(args.BasePath).OrderBy(f => new FileInfo(f).CreationTimeUtc).TakeLast(1);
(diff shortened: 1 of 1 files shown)


adamsitnik:
@MarcoRossignoli I am not sure if I understand.

If I point ResultsComparer to folders with multiple results then it throws? I am sure I have been using it for such test case and it did not throw

MarcoRossignoli:
If you do more attempt the tests result has got same tests insider, so throws key already added, steps:
* Run benchmark base in folder a 1 time
* Run benchmark diff more time in folder b
* Run comparer base/difficile

MarcoRossignoli:
@adamsitnik @jorive do you think I can close this?Didn't you hit my issue?

adamsitnik:
@MarcoRossignoli please excuse me for a very delayed response.

Now I understand what problem you are facing, but I am afraid that this is "by design". I expect that each benchmark run will be stored in a dedicated folder. Like `before`, `forceInline`, `useStruct` etc.

I am not 100% sure, but your proposal would change the behaviour to read **only one** file? If so, I can't accept this change.

I think that I should move ResultsComparer to BDN Global tool and add a new mode called "progress" which would do what you are asking for (base vs latests). I will try to increase the priority of https://github.com/dotnet/BenchmarkDotNet/issues/973

MarcoRossignoli:
> @MarcoRossignoli please excuse me for a very delayed response.

No problem, don't worry!

> I am not 100% sure, but your proposal would change the behaviour to read only one file? If so, I can't accept this change. 

After change comparer picks up "last" benchmark run. For instance if you run two time on dedicated folder `after`(my case) you end up to have 2 result in that folder...and if you run comparer it will break, so you need to go to `after` folder and "delete" one of two results.

BTW feel free to close, it's not blocking I'll remove by hand with no problem, thank's for your time!

adamsitnik:
> After change comparer picks up "last" benchmark run

But what if I have few files for different types? like `System.ABC.json` and `System.XYZ.json`

```cs
baseFiles.OrderBy(f => new FileInfo(f).CreationTimeUtc).TakeLast(1).ToArray()
```

I would expect that in such case your change would load only 1 of them?

MarcoRossignoli:
> I would expect that in such case your change would load only 1 of them?

Yes...because I don't know how to generate more than one json "per type" 😃 never did that, but if it's possible you're right, maybe if so I could try to "merge" better(group by/take last)

MarcoRossignoli:
> But what if I have few files for different types? like System.ABC.json and System.XYZ.json

Sorry I didn't know about this feature.

adamsitnik:
> Yes...because I don't know how to generate more than one json

adamsitnik:
> Yes...because I don't know how to generate more than one json 

You can provide multiple filters:

```cmd
--filter *BinaryTrees_2* *BinaryTrees_5* 
```

should result in creating two files (one per type):

```log
BenchmarksGame.BinaryTrees_2-report-full.json
BenchmarksGame.BinaryTrees_5-report-full.json
```

If you want to join them into a single file, you need to use `--join`

MarcoRossignoli:
> If you want to join them into a single file, you need to use --join

Ahhh understood...I always used `--join` sorry to waste your precious time.

adamsitnik:
You haven't wasted any of my time! I am really happy that you are using so many of BDN features and even more happy to show you some more!

PR #330, opened 2019-02-27
send the entire repo to Helix, keep it simple

adamsitnik:


src/benchmarks/micro/Program.cs
```diff
@@ -29,8 +29,7 @@ private static IConfig GetConfig()
                     .WithMinIterationCount(15)
                     .WithMaxIterationCount(20) // we don't want to run more that 20 iterations
                     .AsDefault()) // tell BDN that this are our default settings
-                .WithArtifactsPath(Path.Combine(
-                    Path.GetDirectoryName(typeof(Program).Assembly.Location), "BenchmarkDotNet.Artifacts"))
+                .With(ConfigOptions.DontOverwriteResults) // tell BDN to not overwrite the results
```
jorive: Data generated by the harness should be with the binaries, not drop side-by-side with the source code. I do not agree this should be changed.

scripts/dotnet.py
```diff
@@ -308,7 +308,7 @@ def install(
 
     if platform != 'win32':
         chmod(dotnetInstallScriptPath, S_IRWXU)
-
+
```
jorive: This file only have white spaces changes?

jorive:
Can't Helix send the command to `git clone` the repo, and execute the commands in the repo? Seems like a trivial scenario a machine harness should support?

scripts/benchmarks_ci.py
```diff
@@ -539,5 +533,38 @@ def __main(args: list) -> int:
...
+    if sys.platform == 'win32':
+        RunCommand(
+            'reg export HKLM\SYSTEM\CurrentControlSet\Control\FileSystem backup.reg',
+            verbose=verbose).run()
+        RunCommand(
+            'reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem /v LongPathsEnabled /t REG_DWORD /d 1 /f',
+            success_exit_codes = [0, 1], # we ignore failures on purpose (Access is denied)
+            verbose=verbose).run()
+
+    try:
```
jorive: Could you please explain these changes? They do not seem to belong here, but in the yml file.

jorive: You are solving a lab problem, and I do not feel comfortable with changing user registry.

src/benchmarks/micro/Program.cs
adamsitnik: I believe the opposite. With more than 1.2 million downloads of BenchmarkDotNet where this was always the default setting nobody ever complained about that.
 
The bin is for binaries. Something that humans don't need to interact with like .dlls, .exes and .runtime.json files.
The users should not be going to the bin to find the results, especially if we are using Crapcade and the binaries are not stored in `.\bin\Release` but in `root\artifacts\bin\$projectName\Release\tfm`. Do you expect the users to know this?

scripts/dotnet.py
adamsitnik: Yes. This PR is not ready for the review yet.

adamsitnik:
> Can't Helix send the command to `git clone` the repo, and execute the commands in the repo? Seems like a trivial scenario a machine harness should support?

No, this complex scenario is beyond helix capabilities.

scripts/benchmarks_ci.py
adamsitnik: > You are solving a lab problem, and I do not feel comfortable with changing user registry.

Me neither, but I am trying to solve a problem that is blocking us. How long would it take to change the settings in the lab? I am afraid that a few days at least...

jorive: Is Arcade/Helix able to run two commands? If so, we could move the registry stuff to a `.cmd` file that runs on Windows before everything else.

docs/benchmarkdotnet.md
```diff
@@ -43,7 +43,7 @@ BenchmarkDotNet will protect you from the common pitfalls (even for experienced
 - it consumes the benchmark result to avoid dead code elimination
 - it prevents from inlining of the benchmark by wrapping it with a delegate
 - it prints the results to the console in GitHub markdown, so you can just copy-paste the printed table to GitHub
-- it exports the results to `BenchmarkDotNet.Artifacts\results` so you can store them for later use.
+- it exports the results to `.\BenchmarkDotNet.Artifacts\results` so you can store them for later use.
```
jorive: You need a folder name, not a directory path. You search for `BenchmarkDotNet.Artifacts\results`, not for `.\BenchmarkDotNet.Artifacts\results`.

scripts/benchmarks_ci.py
adamsitnik: That was a very good suggestion, I have moved this logic to `HelixPreCommands` in `.yml`

src/benchmarks/micro/Program.cs
jorive: I still do not agree, but it is a matter of preference. There is always `DIR /AD /S /B BenchmarkDotNet.Artifacts` and `find . -name BenchmarkDotNet.Artifacts -type d`.
If you still decide to push these change, then you must add the folder name to `.gitignore`.

adamsitnik: I have reverted  those changes, it's most probably better to have a separate PR for that.

billwert: FWIW I agree with you @adamsitnik. Having the results buried deeply in the tree never made sense to me.

adamsitnik:
We want to add new benchmarks (ML.NET) to the CI and we can't just send `\src\benchmarks\micro` to Helix, we need to send entire repo and tell the script which benchmarks to run.

But there is a BUG in Helix Client SDK https://github.com/dotnet/arcade/issues/2125 which locks one of the files in the repo and does not allow to send the entire repo. So I had to come up with a workaround that creates a copy of the folders that we need and stores them in "notLocked" directory. Then, we send "notLocked" and it works.

Since we send all we need in one folder, there is no need to use "CorrelationPayloadDirectory" and provide `--bin-directory` or `--working-directory`.

The ARM build is red due to https://github.com/dotnet/performance/issues/333
The net461 build is red due to https://github.com/dotnet/performance/issues/334 (I have added an extra command to enable long paths and it helped to fix the powershell issue for dotnet cli installer, but for some reason the Full Framework benchmarks still fail with PathTooLong exception)

@adiaaida @jorive could you please take a look?

This PR is one big workaround, but as I wrote it's due to the fact that we can't simply send entire repo to Helix and run a script..

I need to merge this PR to be able to merge #325 (it enables ML.NET runs and simply needs different path) to be able to update and merge #320 (it relies on BenchView config changes from #325) to be able to tell if the differences that @billwert has found are due to Tiered JIT or not.

eng/performance/benchmark_jobs.yml
```diff
@@ -76,20 +76,23 @@ jobs:
         steps:
         - checkout: self
           clean: true
+        - ${{ if eq(parameters.osName, 'windows') }}:
+          - script: (robocopy $(Build.SourcesDirectory) $(Build.SourcesDirectory)\notLocked /E /XD $(Build.SourcesDirectory)\notLocked $(Build.SourcesDirectory)\artifacts $(Build.SourcesDirectory)\.git) ^& IF %ERRORLEVEL% LEQ 1 exit 0
```
adamsitnik: the if for the exit code is requried because robocopy returns 1 if  "One of more files were copied successfully"

https://blogs.technet.microsoft.com/deploymentguys/2008/06/16/robocopy-exit-codes/

eng/performance/benchmark_jobs.yml
```diff
@@ -76,20 +76,23 @@ jobs:
...
+              WorkItemDirectory: '$(Build.SourcesDirectory)\notLocked'
+              WorkItemCommand: 'py -3 scripts\benchmarks_ci.py --incremental no --category $(_Category) --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
+              HelixPostCommands: 'reg import backup.reg'
             ${{ if ne(parameters.osName, 'windows') }}:
-              WorkItemDirectory: '$(Build.SourcesDirectory)/src/benchmarks/micro'
-              WorkItemCommand: 'python3.5 $HELIX_CORRELATION_PAYLOAD/benchmarks_ci.py --bin-directory "./bin" --working-directory . --incremental no --category $(_Category) --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
-              CorrelationPayloadDirectory: '$(Build.SourcesDirectory)/scripts'
+              HelixPreCommands: $(HelixPreCommand)
+              WorkItemDirectory: '$(Build.SourcesDirectory)'
+              WorkItemCommand: 'python3.5 scripts/benchmarks_ci.py --incremental no --category $(_Category) --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
```
jorive: I think this should be `python3` instead of `python3.5`

eng/performance/benchmark_jobs.yml
```diff
@@ -76,20 +76,23 @@ jobs:
...
             HelixAccessToken: $(HelixApiAccessToken)
             HelixTargetQueues: ${{ parameters.queue }}
-            HelixPreCommands: $(HelixPreCommand)
             Creator: $(Creator)
             WorkItemTimeout: 4:00 # 4 hours
             ${{ if eq(parameters.osName, 'windows') }}:
-              WorkItemDirectory: '$(Build.SourcesDirectory)\src\benchmarks\micro'
-              WorkItemCommand: 'py -3 %HELIX_CORRELATION_PAYLOAD%\benchmarks_ci.py --bin-directory ".\bin" --working-directory . --incremental no --category $(_Category) --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
-              CorrelationPayloadDirectory: '$(Build.SourcesDirectory)\scripts'
+              HelixPreCommands: '$(HelixPreCommand);reg export HKLM\SYSTEM\CurrentControlSet\Control\FileSystem backup.reg;reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem /v LongPathsEnabled /t REG_DWORD /d 1 /f'
```
adiaaida: Please don't modify machine state. This will change it for everything that comes after it.

adamsitnik: But I restore the settings in the Post command. Is it not enough?

billwert: Will your tool absolutely never crash and are you ensuring the power does not go out before your post runs?

adiaaida: Let me talk to some people. Generally, this isn't something we really want to do simply, at least to me, in the instance that something goes wrong and, say, the post command doesn't get run. I'm not sure if that's a situation we could run into, I just don't want us to run the risk.

adiaaida: Where are we running into the max-path issue, and can we be smarter about our filenames such that we don't?

adamsitnik: The first time is when the python script calls the powershell script to install dotnet cli:

```log
DotNet Install Path: 'C:\dotnetbuild\work\caadb38f-4fcf-4180-968e-fdfd3776b696\Work\2757c072-fe57-4477-8760-154f8f4d9423\Exec\tools\dotnet\x64'
Downloading https://dot.net/v1/dotnet-install.ps1
$ pushd "C:\dotnetbuild\work\caadb38f-4fcf-4180-968e-fdfd3776b696\Work\2757c072-fe57-4477-8760-154f8f4d9423\Exec"
$ powershell.exe -NoProfile -ExecutionPolicy Bypass C:\dotnetbuild\work\caadb38f-4fcf-4180-968e-fdfd3776b696\Work\2757c072-fe57-4477-8760-154f8f4d9423\Exec\tools\dotnet\x64\dotnet-install.ps1 -InstallDir C:\dotnetbuild\work\caadb38f-4fcf-4180-968e-fdfd3776b696\Work\2757c072-fe57-4477-8760-154f8f4d9423\Exec\tools\dotnet\x64 -Architecture x64 -Channel master
dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Sdk/3.0.100-preview4-010525/dotnet-sdk-3.0.100-preview4-010525-win-x64.zip
dotnet-install: Extracting zip from https://dotnetcli.azureedge.net/dotnet/Sdk/3.0.100-preview4-010525/dotnet-sdk-3.0.100-preview4-010525-win-x64.zip
C:\dotnetbuild\work\caadb38f-4fcf-4180-968e-fdfd3776b696\Work\2757c072-fe57-4477-8760-154f8f4d9423\Exec\tools\dotnet\x64\dotnet-install.ps1 : Exception calling "ExtractToFile" with "3" argument(s): "Could not find a part of the path 'C:\dotnetbuild\work\caadb38f-4fcf-4180-968e-fdfd3776b696\Work\2757c072-fe57-4477-8760-154f8f4d9423\Exec\tools\dotnet\x64\packs\Microsoft.AspNetCore.App.Ref\3.0.0-preview4-19123-01\ref\netcoreapp3.0\Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.dll'."
```

adamsitnik: Also here I am doing this only for Windows and we have dedicated Perf Windows queues. Do machines from the perf queues belong to other queues?

Windows.Amd64.ClientRS4.DevEx.15.8.Perf
Windows.10.Amd64.ClientRS1.Perf

adamsitnik: Ok, I was able to work around it by using `CorrelationPayloadDirectory` instead of `WorkItemDirectory`

the difference is the length of the path of the folder where they get checked out in Helix...

```log
C:\dotnetbuild\work\82ac7b90-beba-445c-96a0-c3ed279fccb2\Work\e11567ec-0440-4e29-b6c0-e1f673b64baf\Exec - WorkItemDirectory
C:\dotnetbuild\work\82ac7b90-beba-445c-96a0-c3ed279fccb2\Payload\ - CorrelationPayloadDirectory
```

PR #332, opened 2019-02-27
can I just install python like this?

adamsitnik:
it looks like our ARM builds are failing because somebody decided to remove python3 so I am trying to check if I could install it like this

### DIFF
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -82,14 +82,15 @@ jobs:
             HelixType: 'test/performance_$(_BuildConfig)/'
             HelixAccessToken: $(HelixApiAccessToken)
             HelixTargetQueues: ${{ parameters.queue }}
-            HelixPreCommands: $(HelixPreCommand)
             Creator: $(Creator)
             WorkItemTimeout: 4:00 # 4 hours
             ${{ if eq(parameters.osName, 'windows') }}:
+              HelixPreCommands: $(HelixPreCommand)
               WorkItemDirectory: '$(Build.SourcesDirectory)\src\benchmarks\micro'
               WorkItemCommand: 'py -3 %HELIX_CORRELATION_PAYLOAD%\benchmarks_ci.py --bin-directory ".\bin" --working-directory . --incremental no --category $(_Category) --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
               CorrelationPayloadDirectory: '$(Build.SourcesDirectory)\scripts'
             ${{ if ne(parameters.osName, 'windows') }}:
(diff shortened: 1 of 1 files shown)


adamsitnik:

Our ARM legs started failing due to missing python3 and I hope that this can fix it.

adiaaida:
We don't want to change the machine state of Helix machines. If python needs to be added to the arm queues, we want to do it by updating the creation scripts found in https://dnceng.visualstudio.com/internal/_git/dotnet-helix-machines. I think @ulisesh is working on adding python3 to the arm machines, though possibly not to the main queues.

jorive:
Doesn't this line: https://github.com/dotnet/performance/blob/8d086bd6ae34424af63d9491faff158f5ccb43be/global.json#L10, specifies that Python 3 is required? Are we missing some kind of invocation?

jorive:
Adding some reference:
https://github.com/dotnet/core-eng/issues/5338
https://github.com/dotnet/core-eng/issues/5220

adamsitnik:
@adiaaida thanks for the explanation! I thought that Helix is also spawning a new VM every time we schedule something so I assumed that installing stuff is ok.

@jorive thanks for adding the references, I am going to subscribe for those issues

eng/performance/benchmark_jobs.yml
```diff
@@ -82,14 +82,15 @@ jobs:
...
               WorkItemCommand: 'py -3 %HELIX_CORRELATION_PAYLOAD%\benchmarks_ci.py --bin-directory ".\bin" --working-directory . --incremental no --category $(_Category) --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
               CorrelationPayloadDirectory: '$(Build.SourcesDirectory)\scripts'
             ${{ if ne(parameters.osName, 'windows') }}:
+              HelixPreCommands: '$(HelixPreCommand);sudo apt-get update;sudo apt-get install python3'
               WorkItemDirectory: '$(Build.SourcesDirectory)/src/benchmarks/micro'
-              WorkItemCommand: 'python3.5 $HELIX_CORRELATION_PAYLOAD/benchmarks_ci.py --bin-directory "./bin" --working-directory . --incremental no --category $(_Category) --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
+              WorkItemCommand: 'python3 $HELIX_CORRELATION_PAYLOAD/benchmarks_ci.py --bin-directory "./bin" --working-directory . --incremental no --category $(_Category) --architecture ${{ parameters.architecture }} -f $(_Framework) $(BenchmarkDotNetArguments) $(BenchViewArguments)'
```
jorive: This should still be `python3` instead of `python3.5`

Issue #344, opened 2019-02-28
Port newly added benchmarks from CoreFX/CoreCLR to perf repo


billwert:
The current snapshot of benchmarks has lagged behind the repos. Need to catch up.

jorive:
@billwert Could you please provide a list?

adamsitnik:
I did go through all CoreFX benchmarks today and created the issues to keep track of it.

#357 #358 #359 #360 #362

jorive:
What about CoreClr?

adamsitnik:
> What about CoreClr?

My manager did not tell me to port the CoreCLR benchmarks.

Just kidding. They are on my todo list once I am done with CoreFX I will scan CoreCLR

adamsitnik:
The only missing CoreCLR benchmark was #113 which I have ported in #382

adamsitnik:
All benchmarks have been ported and merged!

PR #350, opened 2019-03-01
Port System.Drawing benchmarks from CoreFX

adamsitnik:


adamsitnik:
Part of #344

adamsitnik:
@JeremyKuhne @ViktorHofer could you please take a look? this is a port from xunit-performance to BenchmarkDotNet, git blame tells me that you are the most familiar people with these benchmarks :D

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -61,4 +62,8 @@
     <Compile Remove="corefx\System.Security.Cryptography\Perf.Rfc2898DeriveBytes.cs" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
```
adamsitnik: the API is not available for .NET Core 2.0

src/benchmarks/micro/corefx/System.Drawing/Perf_Color.cs
```diff
@@ -0,0 +1,95 @@
...
+    [BenchmarkCategory(Categories.CoreFX)]
+    public class Perf_Color 
+    {
+        private static readonly Color[] AllKnownColors;
+
+        private readonly Color _field = DarkSalmon;
+
+        static Perf_Color()
+        {
+            AllKnownColors = new[]
```
adamsitnik: the original benchmark was using reflection to read that https://github.com/dotnet/corefx/blob/9940166bff52155600f25f15da13a83c982526d2/src/System.Drawing.Primitives/tests/Performance/Perf_Color.cs#L18-L22

I want to add CoreRT runs soon, so I did remove the reflection. There is also a chance (very small ofc) that in the future we might add new colors to `System.Drawing.Color` which would be simply seen as a regression (the array would grow and the time as well)

src/benchmarks/micro/corefx/System.Drawing/Perf_Graphics_DrawBeziers.cs
```diff
@@ -0,0 +1,49 @@
...
+            _image = new Bitmap(100, 100);
+            _pen = new Pen(Color.White);
+            _graphics = Graphics.FromImage(_image);
+
+            _point1 = new Point(10, 10);
+            _point2 = new Point(20, 1);
+            _point3 = new Point(35, 5);
+            _point4 = new Point(50, 10);
+
+            _points = new[] {_point1, _point2, _point3, _point4};
```
adamsitnik: the old benchmarks were using 4 points for `DrawBezier_Point` and array of **7 different** points for `DrawBezier_Points`

https://github.com/dotnet/corefx/blob/9940166bff52155600f25f15da13a83c982526d2/src/System.Drawing.Common/tests/Performance/Perf_Graphics_DrawBeziers.cs#L42

I unified that to make sure that we compare apples to apples (same points)

src/benchmarks/micro/corefx/System.Drawing/Perf_Image_Load.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    [BenchmarkCategory(Categories.CoreFX)]
+    public class Perf_Image_Load
+    {
+        private static readonly ImageTestData[] TestCases = {
+            new ImageTestData(ImageFormat.Bmp),
+            new ImageTestData(ImageFormat.Jpeg),
+            new ImageTestData(ImageFormat.Png),
+            new ImageTestData(ImageFormat.Gif)
+        };
```
adamsitnik: the old benchmarks were most probably by accident using only `Bmp`, I have fixed that

https://github.com/dotnet/corefx/blob/9940166bff52155600f25f15da13a83c982526d2/src/System.Drawing.Common/tests/Performance/Perf_Image_Load.cs#L122

src/benchmarks/micro/corefx/System.Drawing/Perf_Image_Load.cs
```diff
@@ -0,0 +1,92 @@
...
+                using (Bitmap bitmap = new Bitmap(Size, Size))
+                using (Pen pen = new Pen(Color.Blue))
+                using (Graphics graphics = Graphics.FromImage(bitmap))
+                {
+                    for (int i = 0; i < 100; i++)
+                    {
+                        graphics.DrawBezier(pen, RandomPoint(), RandomPoint(), RandomPoint(), RandomPoint());
+                    }
+
+                    bitmap.Save(result, format);
```
adamsitnik: the old benchmarks were writing to file

https://github.com/dotnet/corefx/blob/9940166bff52155600f25f15da13a83c982526d2/src/System.Drawing.Common/tests/Performance/Perf_Image_Load.cs#L144

to avoid multimodality of the benchmarks (possibly introduced by disk IO) I decided to use `MemoryStream` instead.

src/benchmarks/micro/corefx/System.Drawing/Perf_Graphics_Transforms.cs
```diff
@@ -0,0 +1,48 @@
...
+                new Point(10, 10), new Point(20, 1), new Point(35, 5), new Point(50, 10),
+                new Point(60, 15), new Point(65, 25), new Point(50, 30)
+            };
+
+            _image = new Bitmap(100, 100);
+            _graphics = Graphics.FromImage(_image);
+        }
+
+        [Benchmark]
+        [AllowedOperatingSystems("Graphics.TransformPoints is not implemented in libgdiplus yet. See dotnet/corefx 20884", OS.Windows)]
```
adamsitnik: this is the only benchmark that can't run on the OSes that we test (Windows RS4, Ubuntu 16.04)

there were some other conditional benchmarks that would not run for Ubuntu 14.04 and other Unixes but we don't test those OSes in our lab so I did not port those conditions

PR #352, opened 2019-03-02
Add ML.NET benchmarks CI legs

adamsitnik:


adamsitnik:

azure-pipelines.yml
```diff
@@ -23,6 +23,8 @@ jobs:
       architecture: x64
       pool: Hosted VS2017
       queue: Windows.Amd64.ClientRS4.DevEx.15.8.Perf
+      csproj: src\benchmarks\micro\MicroBenchmarks.csproj
```
adiaaida: You don't seem to have added the kind parameter here.

eng/performance/benchmark_jobs.yml
```diff
@@ -65,28 +67,19 @@ jobs:
...
+                  _Framework: ${{ framework }}
+                  _CompilationMode: NoTiering
+                  _BuildConfig: ${{ parameters.architecture }}_${{ category }}_$(_Framework)
+                ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}: # to reduce the number of CI legs for every PR
+                  ${{ if and(startsWith(framework, 'netcoreapp'), ne(framework, 'netcoreapp2.0')) }}: # Tiered JIT was introduced in .NET Core 2.1
+                    ${{ category }}_${{ framework }}_tiered:
+                      _Category: ${{ category }}
+                      _Framework: ${{ framework }}
+                      _CompilationMode: Tiered
+                      _BuildConfig: ${{ parameters.architecture }}_${{ category }}_$(_Framework)_$(_CompilationMode)
```
adiaaida: we may need to specify kind in the build config.

azure-pipelines.yml
adamsitnik: For the micro benchmarks I have added it some time ago, but GH does not show it here because it was not changed in this PR.

When you click "View file" you should see it. https://github.com/dotnet/performance/blob/e1cce9b2336c38e2dd3bdd396707452a4d57eb56/azure-pipelines.yml#L22

eng/performance/benchmark_jobs.yml
```diff
@@ -1,12 +1,14 @@
 parameters:
   osName: ''           # required -- windows | linux | macos
   osVersion: ''        # required -- OS version
-  kind: ''             # required -- benchmark kind. As of today, only "micro" benchmarks are supported, we plan to add "scenarios" soon
+  kind: ''             # required -- benchmark kind. As of today, only "micro" and "mlnet" benchmarks are supported, we plan to add "scenarios" soon
```
adiaaida: You could always just default this to micro if you don't want to have to specify it for all of the legs, since most legs are micro.

azure-pipelines.yml
adiaaida: Ah. I see.

adiaaida: I didn't notice that the only thing changed in the line in the template was to add the ml kind as a supported option.

eng/performance/benchmark_jobs.yml
adamsitnik: AzureDevOps has complained about not unique job ids, so I added it to the job id. 

https://github.com/dotnet/performance/blob/e1cce9b2336c38e2dd3bdd396707452a4d57eb56/eng/performance/benchmark_jobs.yml#L20

Do you think that I should add it to both places?

adiaaida: The BuildConfig is the thing that specifies things for mc.dot.net. When we don't have completely unique BuildConfigs between runs that happened on the same queue, mc overwrites the logs (so you could have a failure in the micro leg but not in the ml leg, but if the ml leg finished 2nd, you won't have the logs for the micro leg). I noticed this when first setting up the jobs.

adamsitnik: aa, now I see it. I am going to change it and add a comment

adamsitnik: Done

eng/performance/benchmark_jobs.yml
adamsitnik: As for now yes, but we plan to add scenarios and other stuff so I would prefer to keep it as required with no default value.

PR #355, opened 2019-03-04
don't trigger the CI if only a doc file was changed

adamsitnik:


adamsitnik:
@adiaaida @billwert as we talked last week, we should not trigger the CI if only a doc file was changed

based on https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#paths

billwert:
@adamsitnik @adiaaida does this work? I was surprised to see CI running on my other PR, and even on this PR it seems to have done the whole thing (since it ran 22 checks.) Shouldn't this PR have not had those checks run if this had the desired effect?

billwert:
sorry I guess this one should but #392 seems like it should not?

adiaaida:
I wonder if the **/*.md syntax doesn't work quite how we expect it to.

billwert:
seems we're not the only ones. MicrosoftDocs/vsts-docs#2363. looking at the doc it says the wildcard has to be at the end. so docs/* is fine. @adiaaida do you know where AzDo feature requests get tracked? It'd be nice if this worked.. In the mean time I'll change the paths to be docs/* and all the relative paths to the .md files in the repo.

adiaaida:
@chcosta Who should we contact for AzDO feature requests?

adamsitnik:
> looking at the doc it says the wildcard has to be at the end

it explains why it worked for https://github.com/dotnet/performance/pull/364 but did not work for this PR. It would be great if AzDO could fix it

chcosta:
Please file an issue in https://dnceng.visualstudio.com/internal/_workitems/recentlyupdated/ (set "internal/VSTS" as the Area) and I'll bring this up with Azure DevOps.

billwert:
done thanks @chcosta

PR #356, opened 2019-03-04
add internal build status to the README.md

adamsitnik:
we have two builds: public for PRs and internal that does the actual benchmarks run on real hardware and uploads the results

@jorive @adiaaida @billwert I think that it would be nice to add the Internal Build Status icon to the README.md to make sure that we monitor the status of this build too. What do you think?


adiaaida:
I think the idea is good, but I would probably do it as a table. I think the parentheses makes it confusing.

adamsitnik:
> I think the idea is good, but I would probably do it as a table. I think the parentheses makes it confusing

good suggestion! now looks much better

adiaaida:
Looks awesome!

PR #363, opened 2019-03-05
[PROPOSAL] Remove build status matrix

adamsitnik:
I think that the Build Status Page brings less value than expected:

1. we need to update it anytime we add a new CI leg
2. we need to update it anytime we change CI leg ID
3. it can be confusing:
   - what does it mean that it's green? does it mean we have no regressions?
   - what does it mean that framework X is not present - we don't care?
   - what does it mean that OS Y is not present - we don't care?

I think that the simplest solution is to just remove it and keep two icons for the public and internal build status pages. Once we click it, we got  to Azure DevOps summary with all the legs inside and we can just cick on the red one and see the details.

@jorive @adiaaida @billwert what do you think?


adiaaida:
I'm fine with this change, but I won't be checking the status of the jobs, so I'd wait for Bill to get back from vacation to check it in, as he'll be most likely the one monitoring.

billwert:
I'd kind of like to keep it. I'm not as concerned about how often we change things, as I don't anticipate we will be updating/migrating CI legs around very often (or will we?)

As far as what the matrix means, I think we should change the text here a bit to be more representative of what this really is: functional CI status. Right now it is not differentiated at all from perf results.

I don't feel strongly about this, though. If the cost to keep it is higher than I think it is we can get rid of it.

adamsitnik:
I am going to close it for now and when the matrix becomes a problem in the future we might reconsider removing it again.

PR #366, opened 2019-03-05
Port two benchmarks added recently to Perf.StringBuilder from CoreFX

### LINKED ISSUE #362: Port two benchmarks added recently to Perf.StringBuilder from CoreFX
https://github.com/dotnet/corefx/blob/786cc3edbd25f7323b88aaf672e36518ae30d116/src/System.Runtime/tests/Performance/Perf.StringBuilder.cs#L152-L187

adamsitnik:
fixes #362

there were 2 benchmarks added in https://github.com/dotnet/corefx/pull/33225

https://github.com/dotnet/corefx/blob/786cc3edbd25f7323b88aaf672e36518ae30d116/src/System.Runtime/tests/Performance/Perf.StringBuilder.cs#L152-L187

### DIFF
--- a/src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
+++ b/src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
@@ -137,5 +137,38 @@ namespace System.Tests
 
         [Benchmark]
         public string StringBuilderToString() => _bigStringBuilder.ToString();
+
+#if NETCOREAPP3_0 // Append(ReadOnlyMemory<char> value) is available only for .NET Core 3.0+
+        [Benchmark]
+        [Arguments(100)]
+        [Arguments(1000)]
+        public StringBuilder AppendMemoryAsMemory(int length)
+        {
+            string builtString = length == 100 ? _string100 : _string1000;
+            ReadOnlyMemory<char> memory = builtString.AsMemory();
+            StringBuilder builder = new StringBuilder();
+
+            for (int j = 0; j < NUM_ITERS_APPEND; j++)
+                builder.Append(memory); // Appends a string of length "length" to an increasingly large StringBuilder
+
+            return builder;
+        }
+#endif
+
+        [Benchmark]
+        [Arguments(100)]
+        [Arguments(1000)]
+        public StringBuilder AppendMemoryAsObject(int length)
+        {
+            string builtString = length == 100 ? _string100 : _string1000;
+            object memoryObject = builtString.AsMemory(); // deliberately uses object to force use of memory.ToString() for comparison to AppendAsReadOnlyMemory
+
+            StringBuilder builder = new StringBuilder();
+
+            for (int j = 0; j < NUM_ITERS_APPEND; j++)
+                builder.Append(memoryObject);
+
+            return builder;
+        }
     }
 }
--- a/src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
+++ b/src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
@@ -138,35 +138,14 @@ namespace System.Tests
         [Benchmark]
         public string StringBuilderToString() => _bigStringBuilder.ToString();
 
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
```diff
@@ -137,5 +137,38 @@ public void SetupStringBuilderToString()
...
+            for (int j = 0; j < NUM_ITERS_APPEND; j++)
+                builder.Append(memory); // Appends a string of length "length" to an increasingly large StringBuilder
+
+            return builder;
+        }
+#endif
+
+        [Benchmark]
+        [Arguments(100)]
+        [Arguments(1000)]
```
stephentoub: Do we need both of these?  What are we validating with one that we're not with the other?

src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
```diff
@@ -137,5 +137,38 @@ public void SetupStringBuilderToString()
 
         [Benchmark]
         public string StringBuilderToString() => _bigStringBuilder.ToString();
+
+#if NETCOREAPP3_0 // Append(ReadOnlyMemory<char> value) is available only for .NET Core 3.0+
+        [Benchmark]
+        [Arguments(100)]
+        [Arguments(1000)]
```
stephentoub: Same question as below.

src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
```diff
@@ -137,5 +137,38 @@ public void SetupStringBuilderToString()
 
         [Benchmark]
         public string StringBuilderToString() => _bigStringBuilder.ToString();
+
+#if NETCOREAPP3_0 // Append(ReadOnlyMemory<char> value) is available only for .NET Core 3.0+
```
stephentoub: Is this actually needed?  Seems like the perf would just be better when built against 3.0, but it should still compile against older versions and just bind to the object-based overload, right?

adamsitnik:
@stephentoub you are right, we don't need it.

I have simplified this PR to a single benchmark, which shows that adding the new overload has improved the perf:

```cmd
dotnet run -f netcoreapp2.1 --filter *AppendMemory --runtimes netcoreapp2.1 netcoreapp2.2 netcoreapp3.0
```

|       Method |     Toolchain |     Mean | Allocated Memory/Op |
|------------- |-------------- |---------:|--------------------:|
| AppendMemory | netcoreapp2.1 | 43.76 us |           229.53 KB |
| AppendMemory | netcoreapp2.2 | 43.11 us |           229.53 KB |
| AppendMemory | netcoreapp3.0 | 84.94 us |           198.28 KB |

adamsitnik:
I am blind, I looked only at the memory. Actualy it's slower for .NET Core 3.0. It looks like a Tiered JIT issue.

```
py .\scripts\benchmarks_ci.py -f netcoreapp3.0 --filter *AppendMemory --dotnet-compilation-mode NoTiering
```

|       Method |     Mean | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------- |---------:|------------:|------------:|------------:|--------------------:|
| AppendMemory | 39.53 us |     32.1475 |     15.9922 |           - |           198.28 KB |

```cmd
py .\scripts\benchmarks_ci.py -f netcoreapp3.0 --filter *AppendMemory --dotnet-compilation-mode Tiered
```

|       Method |     Mean | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------- |---------:|------------:|------------:|------------:|--------------------:|
| AppendMemory | 71.41 us |     32.1691 |     15.9314 |           - |           198.28 KB |

I am going to create an issue in CoreCLR

adamsitnik:
Issue reported: https://github.com/dotnet/coreclr/issues/23060

PR #369, opened 2019-03-05
Port System.IO.Pipelines benchmarks from CoreFX

adamsitnik:
@pakrym could you please take a look? I have ported them from https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipelines/tests/Performance/Perf.Pipe.cs

btw was there any reason why the methods were async **void** instead of async **Task**?


pakrym:
> btw was there any reason why the methods were async void instead of async Task?

I don't remember any reason for that.

adamsitnik:
@pakrym thank you!

PR #374, opened 2019-03-06
don't try to upload Full .NET Framework results to BenchView

adamsitnik:


adamsitnik:
@adiaaida this PR disables generating bench view data for Full Framework. We don't want to upload the results for Full Framework from this repo and since it's currently failing the CI because we can't read the commit hash for Full Framework we have decided to simply disable it.

PR #381, opened 2019-03-08
port System.Buffers benchmarks

### LINKED ISSUE #378: Port missing System.Buffers benchmarks from CoreFX
https://github.com/dotnet/corefx/tree/master/src/System.Memory/tests/Performance

/cc @ahsonkhan

adamsitnik:
fixes #378 

these are the last missing CoreFX benchmarks that have not yet been ported

When I was porting the Utf8Parser and Utf8Formatter benchmarks I have realized that we were missing some Parse/TryParse/ToString benchmarks for Byte/SByte/Decimal etc so I just added them


adamsitnik:
@stephentoub @ahsonkhan could you please take a look? these are the last missing CoreFX benchmarks ;)

src/benchmarks/micro/corefx/System.Buffers/Base64EncodeDecodeTests.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        [Benchmark]
+        public OperationStatus Base64Encode() => Base64.EncodeToUtf8(_decodedBytes, _encodedBytes, out _, out _);
+
+        [GlobalSetup(Target = nameof(Base64EncodeDestinationTooSmall))]
+        public void SetupBase64EncodeDestinationTooSmall()
+        {
+            _decodedBytes = ValuesGenerator.Array<byte>(NumberOfBytes);
+            _encodedBytes = new byte[Base64.GetMaxEncodedToUtf8Length(NumberOfBytes) - 1]; // -1
```
stephentoub: Why - 1?  Doesn't that mean some of the encoding operations could fail?

stephentoub: Oh, nevermind, this is the setup for ones that should fail.

src/benchmarks/micro/corefx/System.Buffers/Base64EncodeDecodeTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        [GlobalSetup(Target = nameof(Base64EncodeBaseline))]
+        public void SetupBase64EncodeBaseline()
+        {
+            _decodedBytes = ValuesGenerator.Array<byte>(NumberOfBytes);
+            _encodedChars = new char[Base64.GetMaxEncodedToUtf8Length(NumberOfBytes)];
+        }
+
+        [Benchmark]
+        public int Base64EncodeBaseline() => Convert.ToBase64CharArray(_decodedBytes, 0, _decodedBytes.Length, _encodedChars, 0);
```
stephentoub: Why is this called "Baseline" rather than being named after what the API it's actually measuring?

src/benchmarks/micro/corefx/System.Buffers/Base64EncodeDecodeTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        [GlobalSetup(Target = nameof(Base64Decode))]
+        public void SetupBase64Decode()
+        {
+            _encodedBytes = ValuesGenerator.Array<byte>(NumberOfBytes);
+            _decodedBytes = new byte[Base64.GetMaxEncodedToUtf8Length(NumberOfBytes)];
+        }
+
+        [Benchmark]
+        public OperationStatus Base64Decode() => Base64.DecodeFromUtf8(_encodedBytes, _decodedBytes, out _, out _);
```
stephentoub: Same question... why is this called Base64Decode rather than Base64DecodeFromUtf8, or something like that?

src/benchmarks/micro/corefx/System.Buffers/Base64EncodeDecodeTests.cs
```diff
@@ -0,0 +1,117 @@
...
+#if !NETFRAMEWORK && !NETCOREAPP2_0 // API added in .NET Core 2.1
+        [GlobalSetup(Target = nameof(Base64DecodeBaseline))]
+        public void SetupBase64DecodeBaselinee()
+        {
+            _decodedBytes = ValuesGenerator.Array<byte>(NumberOfBytes);
+            _encodedChars = Convert.ToBase64String(_decodedBytes).ToCharArray();
+        }
+
+        [Benchmark]
+        public bool Base64DecodeBaseline() => Convert.TryFromBase64Chars(_encodedChars, _decodedBytes, out _);
```
stephentoub: Same question.  I'll stop asking about these, but I'm finding the naming confusing.

src/benchmarks/micro/corefx/System.Buffers/Base64EncodeDecodeTests.cs
```diff
@@ -0,0 +1,117 @@
...
+#endif
+    }
+
+    // we want to test InPlace methods so we need to setup every benchmark invocation so we are using [IterationSetup]
+    // to make the Iteration last longer and make the results stable, we are using bigger NumberOfBytes
+    // but due to limitation of BDN where Params have no Target and are applied to entire class
+    // the benchmarks live in a separate class
```
stephentoub: This text is hard to parse without punctuation separating sentences.

src/benchmarks/micro/corefx/System.Buffers/BinaryReadAndWriteTests.cs
```diff
@@ -0,0 +1,300 @@
...
+using static System.Buffers.Binary.BinaryPrimitives;
+
+namespace System.Buffers.Binary.Tests
+{
+    [BenchmarkCategory(Categories.CoreFX, Categories.CoreCLR)]
+    public class BinaryReadAndWriteTests
+    {
+        private readonly static byte[] _arrayLE = GetSpanLE().ToArray();
+        private readonly static byte[] _arrayBE = GetSpanBE().ToArray();
+        private readonly static int[] _oneThousandIntegers = new int[1000];
```
stephentoub: How are you deciding whether to have numeric inputs like this in a Params and the various fields/properties initialized in a global setup routine, versus having static members initialized directly?  I see both approaches in this PR, and I'm not clear on why some are done one way and some are done another.

src/benchmarks/micro/corefx/System.Buffers/BinaryReadAndWriteTests.cs
```diff
@@ -0,0 +1,300 @@
...
+                readStruct.UL1 = ReverseEndianness(readStruct.UL1);
+            }
+
+            return readStruct;
+        }
+
+        [Benchmark]
+        public int[] MeasureReverseEndianness()
+        {
+            var local = _oneThousandIntegers;
```
stephentoub: Nit: var => whatever type it is... same goes for any other occurrences where the type isn't obvious from the RHS

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private int First(ReadOnlySequence<T> sequence)
+        {
+            int consume = 0;
+
+            consume += sequence.First.Length; consume += sequence.First.Length; consume += sequence.First.Length; consume += sequence.First.Length;
+            consume += sequence.First.Length; consume += sequence.First.Length; consume += sequence.First.Length; consume += sequence.First.Length;
+            consume += sequence.First.Length; consume += sequence.First.Length; consume += sequence.First.Length; consume += sequence.First.Length;
+            consume += sequence.First.Length; consume += sequence.First.Length; consume += sequence.First.Length; consume += sequence.First.Length;
```
stephentoub: Nit: it's difficult to see this is being done 16 times.  It'd also be helpful if the count were in the method name.

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            while (!position.Equals(end))
+            {
+                position = sequence.GetPosition(offset, position);
+                consume += position.GetInteger();
+            }
+
+            return consume;
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
stephentoub: Why?

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        private long Slice(ReadOnlySequence<T> sequence)
+        {
+            long consume = 0;
+
+            consume += sequence.Slice(Size / 10 * 0, Size / 10).Length; consume += sequence.Slice(Size / 10 * 1, Size / 10).Length;
+            consume += sequence.Slice(Size / 10 * 2, Size / 10).Length; consume += sequence.Slice(Size / 10 * 3, Size / 10).Length;
+            consume += sequence.Slice(Size / 10 * 4, Size / 10).Length; consume += sequence.Slice(Size / 10 * 5, Size / 10).Length;
+            consume += sequence.Slice(Size / 10 * 6, Size / 10).Length; consume += sequence.Slice(Size / 10 * 7, Size / 10).Length;
+            consume += sequence.Slice(Size / 10 * 8, Size / 10).Length; consume += sequence.Slice(Size / 10 * 9, Size / 10).Length;
```
stephentoub: Same comments.

src/benchmarks/micro/corefx/System.Buffers/Base64EncodeDecodeTests.cs
adamsitnik: good catch, I did not realize that when I was porting it. I think that the initial idea was to use `BaseXYZ` names for `Base64` methods and `XYZBaseline` for `Convert` methods. I am going to change the names and remove `Baseline` keyword from them

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
adamsitnik: ok, I will split the 4x4 matrix into 8x2 one

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
adamsitnik: the first method with this attribute above contains the explanation:

```cs
// make sure that the method does not get inlined for any of the benchmarks and we compare apples to apples
```

should I copy this comment for every use? I personally always add the explanation to the first appearance of it in the file

src/benchmarks/micro/corefx/System.Buffers/BinaryReadAndWriteTests.cs
adamsitnik: If all the benchmarks are using the same input data of the same size and it should not change in the future, I initialize everything in ctor. 

Otherwise, I use Params + Setups.

adamsitnik:
@stephentoub I have pushed a commit with the fixes, PTAL

src/benchmarks/micro/corefx/System.Buffers/BinaryReadAndWriteTests.cs
stephentoub: Thanks.  What's the benefit of that approach versus putting tests into their own class?

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
stephentoub: I see.  Ok.

src/benchmarks/micro/corefx/System.Buffers/Base64Tests.cs
```diff
@@ -39,15 +39,15 @@ public void SetupBase64EncodeDestinationTooSmall()
         [Benchmark]
         public OperationStatus Base64EncodeDestinationTooSmall() => Base64.EncodeToUtf8(_decodedBytes, _encodedBytes, out _, out _);
 
-        [GlobalSetup(Target = nameof(Base64EncodeBaseline))]
+        [GlobalSetup(Target = nameof(ConvertToBase64CharArray))]
         public void SetupBase64EncodeBaseline()
```
stephentoub: Shouldn't the "Setup" methods be renamed as well?

src/benchmarks/micro/corefx/System.Buffers/BinaryReadAndWriteTests.cs
adamsitnik: Setup vs ctor:

- Setup can use values from `[Params]`, ctor can not (they are set after we run the ctor)
- Setup can be async, ctor can not
- ctor can initialize readonly fields, Setup can not
- if  we need some Cleanup then having a Setup+Cleanup seems more intuitive than ctor+finalizer (BDN won't call Dispose if the class implements IDisposable)

when it comes to "putting tests into their own class" I think that it would simply create too many types and it would be harder to find the benchmarks in our reporting tool (in BenchView we have a namespace -> type name -> benchmark name tree)

src/benchmarks/micro/corefx/System.Buffers/Base64Tests.cs
adamsitnik: good catch, #done

src/benchmarks/micro/corefx/System.Buffers/BinaryReadAndWriteTests.cs
stephentoub: Then for consistency why not always use Setup?

adamsitnik: Honestly, I never thought about it from the consistency point of view. I think that it's better to keep it simple (don't use Setup if you don't need to).

adamsitnik: I think that we should discuss and agree on benchmarking best practices and guidelines and write them down into docs.

adamsitnik:
I have created a separate issue for the best practices docs (#396) and got an offline approval from @stephentoub to merge this PR ;)

PR #382, opened 2019-03-08
Port PacketTracer benchmarks

### LINKED ISSUE #113: Port new SoA benchmarks from CoreCLR
https://github.com/dotnet/coreclr/pull/18839

adamsitnik:
Fixes #113 

Please see the commits history for what I have copied and what I have modified.

This is the only and the last missing CoreCLR benchmark


src/benchmarks/micro/coreclr/PacketTracer/Camera.cs
```diff
@@ -0,0 +1,38 @@
...
+        public VectorPacket256 Pos;
+        public VectorPacket256 Forward;
+        public VectorPacket256 Up;
+        public VectorPacket256 Right;
+
+        public static Camera Create(VectorPacket256 pos, VectorPacket256 lookAt)
+        {
+            VectorPacket256 forward = (lookAt - pos).Normalize();
+            VectorPacket256 down =
+                new VectorPacket256(Vector256<float>.Zero, Vector256.Create(-1.0f), Vector256<float>.Zero);
```
jorive: >new VectorPacket256(Vector256<float>.Zero, Vector256.Create(-1.0f), Vector256<float>.Zero) [](start = 16, length = 90)

Should this be a readonly/const variable created/allocated once?

jorive:
@adamsitnik Would it make more sense to place this benchmark under a `SoA.RayTracer` namespace (what is attempting to measure and what it is doing)?

jorive:
This benchmark only runs on `AVX2` and not in `netcoreapp3.0`?

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -68,5 +68,9 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
     <Compile Remove="corefx\System.Drawing\Perf_Graphics*.cs" />
   </ItemGroup>
+  
+    <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp3.0'">
```
adamsitnik: > This benchmark only runs on AVX2 and not in netcoreapp3.0? Is there a way to not bother building on non-x86* architecture?

@jorive it's compiled only for .NET Core 3.0. AFAIK there is no way to exclude it for given architecture (we target AnyCPU here, the target architecture depends on the dotnet cli we use)

src/benchmarks/micro/coreclr/PacketTracer/Camera.cs
adamsitnik: I don't know, I just ported it and made sure it does what expected. I did not have time to analyze all the code

src/benchmarks/micro/MicroBenchmarks.csproj
jorive: Does it build on arm*?

adamsitnik: yes

adamsitnik: The Avx2.IsSupported will just return false

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -8,6 +8,8 @@
     <DebugType>pdbonly</DebugType>
     <DebugSymbols>true</DebugSymbols>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+    <IsARM Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().StartsWith(`arm`, System.StringComparison.InvariantCultureIgnoreCase))'">true</IsARM>
```
adamsitnik: @eerhardt do you know any less hacky way to tell if we are running on ARM?

I based my idea on your GH comment from 2017 ;p

https://github.com/Microsoft/msbuild/issues/539#issuecomment-289930591

eerhardt: You are checking for if the build itself is happening on an ARM device, and not targetting an ARM device? If so, I'd say this is probably the best approach - using `RuntimeInformation`.  We do the same thing in corefx:

https://github.com/dotnet/corefx/blob/2a2d34df7db5d2617d12f55fe5ad8fd208117049/Directory.Build.props#L61-L63

One question - why `InvariantCultureIgnoreCase` instead of `OrdinalIgnoreCase`?

adamsitnik: @eerhardt thanks a lot!

> You are checking for if the build itself is happening on an ARM device, and not targetting an ARM device?

yes, exactly

> One question - why InvariantCultureIgnoreCase instead of OrdinalIgnoreCase?

for some reason, I have a habit to always use InvariantCultureIgnoreCase everywhere based on some bad experience, but I honestly don't remember what experience was it.

I just read more about `OrdinalIgnoreCase` and it looks like a better choice here. Thanks for the suggestion!

adamsitnik:
```suggestion
    <IsARM Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().StartsWith(`arm`, System.StringComparison.OrdinalIgnoreCase))'">true</IsARM>
```

adamsitnik:
@jorive it looks like we have what we want - this benchmark won't be built/run on ARM. Do you want me to rename the namespace/type/method name as suggested?

I have also sent a PR to BDN that will make it much easier for us in the future: https://github.com/dotnet/BenchmarkDotNet/pull/1097

jorive:
@adamsitnik Thanks Adam. I would prefer to rename the namespace/class/method (from *PacketTracer/Render/Run* to something like *StructureOfArray/RayTracer/Render*), but maybe @billwert have a different opinion.

adamsitnik:
I think that this explains why the author of the benchmark did not add a namespace to it in the first place, the naming is just hard ;)

My proposal: HardwareIntrinsics/RayTracer/Render

@jorive what do you say?

adamsitnik:
@jorive @billwert which name should we use? ;)

billwert:
I think if we believe these benchmarks are solely for the purpose of testing hardware intrinsicics, we should go with _HardwareIntrinsics/RayTracer/Render_. If we think there is value in keeping _PacketTracer_ for name recognition perhaps _PacketTracer/HardwareIntrinsics/Render_?

jorive:
The benchmark is more about data storage/layout than "HardwareIntrinsics".

adamsitnik:
> The benchmark is more about data storage/layout than "HardwareIntrinsics".

fiigii:
> what name do you recommend for it?

Packet tracing is a convention name in the graphics area that indicates the SoA vectorized ray tracing algorithm (but this benchmark is a simplified version without BVH). I am okay to rename it in this dotnet repo. Maybe `HardwareIntrinsics/RayTracer/SoA`? That gives us room to add other algorithms under `/RayTracer/` in the future. 

> The benchmark is more about data storage/layout than "HardwareIntrinsics".

Originally, I created this benchmark for hardware intrinsic to leverage the 256-bit SIMD vectors comparing the AoS ray-tracer we had for benchmarking `S.N.Vector<T>`. In .NET Core, it is impossible to implement this program efficiently in `S.N.Vector<T>`, please see  `Transpose`, `public unsafe VectorPacket256(float* vectors)`, and methods in `VectorMath.cs`. So, I think it is definitely about hardware intrinsic.

src/benchmarks/micro/coreclr/PacketTracer/ProducerConsumerCollectionBase.cs
```diff
@@ -0,0 +1,107 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+//
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace System.Collections.Concurrent
```
fiigii: I suggest removing this file. That was directly copied from the original ray tracing benchmark.

src/benchmarks/micro/coreclr/PacketTracer/ObjectPool.cs
```diff
@@ -0,0 +1,67 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace System.Collections.Concurrent
```
fiigii: I suggest removing this file as well and use array in the top-level function instead.

jorive:
Namespace: `HardwareIntrinsics/RayTracer/`
class/method: `SoA/Render`
?

src/benchmarks/micro/coreclr/PacketTracer/ObjectPool.cs
adamsitnik: good point, #done

adamsitnik:
@fiigii thanks for help!

@jorive thanks for the suggestion! `HardwareIntrinsics.RayTracer.SoA.Render` is the new full benchmark name

PR #383, opened 2019-03-11
fix breaking change introduced by latest Helix SDK

adamsitnik:
@dotnet-maestro-bot  has merged a breaking change in https://github.com/dotnet/performance/pull/372 and now our build is red.

this PR tries to fix it..

obraz


adamsitnik:
I was unable to fix it, so I have reverted it in #384

Issue #390, opened 2019-03-18
Output directory is confusing


billwert:
Due to dotnet/cli#10990, if you run the microbenchmarks project without passing -c Release the results end up under a Debug directory, which is confusing.

Worse, if you run a lot of benchmarks the only indication this happened has scrolled far off the terminal window by the time the run is done.

Until that is fixed (if it is ever) I think we should remove the <Optimize> tag from the csproj, so we force a reasonable state.

adamsitnik:
@billwert do I understand correctly that https://github.com/dotnet/performance/pull/391 has solved this issue and we can close it?

billwert:
yup! (I guess I didn't put 'fixes #390' in the right place in the PR, I thought it could be in a comment.)

PR #391, opened 2019-03-18
Remove Optimize define

billwert:
Defining Optimize in the project file is not precisely the same as specifying -c Release. (For example, DEBUG is defined instead of RELEASE, so any code which has tracing will include that tracing.) Instead of continuing to chase down all the ways it is different I think just pulling this out and requiring users to pass -c Release is a good path.

### DIFF
--- a/src/common.props
+++ b/src/common.props
@@ -4,9 +4,6 @@
          and automation failing to clean up the folders once the runs are over -->
     <UseSharedCompilation>false</UseSharedCompilation>
     
-    <!-- to make sure that the users don't always need to specify -c Release when using dotnet run -->
-    <Optimize>true</Optimize>
-    
     <!-- Used by Python script to narrow down the the specified target frameworks to test, and avoid downloading all supported SDKs -->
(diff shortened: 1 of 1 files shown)


billwert:
this fixes #390

billwert:
@adamsitnik good catch, will do.

danmosemsft:
Should there be an error or warning if I _don't_ pass Release? Although IIRC BDN will warn.

jorive:
It should be possible to have a warning/error if `Condition=" '$(Configuration)' != 'Release' "`.

billwert:
BDN will warm and stop. It’s sufficient.

Thanks,
-Bill

billwert:
@danmosemsft: 
// Validating benchmarks:
Assembly MicroBenchmarks which defines benchmarks references non-optimized BenchmarkDotNet.Extensions
        If you own this dependency, please, build it in RELEASE.
        If you don't, you can create custom config with JitOptimizationsValidator.DontFailOnError to disable our custom policy and allow this benchmark to run.
Assembly MicroBenchmarks which defines benchmarks is non-optimized
Benchmark was built without optimization enabled (most probably a DEBUG configuration). Please, build it in RELEASE.

PR #392, opened 2019-03-19
fixup docs

billwert:
In #391 I changed how `dotnet run` needs to be invoked but didn't clean
up the docs.

### DIFF
--- a/docs/benchmarkdotnet.md
+++ b/docs/benchmarkdotnet.md
@@ -86,10 +86,10 @@ py .\scripts\benchmarks_ci.py --frameworks netcoreapp3.0
 
 ### Interactive Mode
 
-To run the benchmarks in interactive mode you have to execute `dotnet run -f $targetFrameworkMoniker` in the folder with benchmarks project.
+To run the benchmarks in interactive mode you have to execute `dotnet run -c Release -f $targetFrameworkMoniker` in the folder with benchmarks project.
 
 ```cmd
-C:\Projects\performance\src\benchmarks\micro> dotnet run -f netcoreapp3.0
+C:\Projects\performance\src\benchmarks\micro> dotnet run -c Release -f netcoreapp3.0
 Available Benchmarks:
   #0   Burgers
   #1   ByteMark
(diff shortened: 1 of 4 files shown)


docs/benchmarking-workflow-corefx.md
```diff
@@ -161,8 +163,8 @@ The next step is to send a PR to this repository with the aforementioned benchma
 
 The real performance investigation starts with profiling. To profile the benchmarked code and produce an ETW Trace file ([read more](./benchmarkdotnet.md#Profiling)):
 
-```
-dotnet run -f netcoreapp3.0 --profiler ETW --filter $YourFilter 
+```cmd
```
adamsitnik: good catch with the missing `cmd` 👍

docs/benchmarking-workflow-corefx.md
```diff
@@ -240,6 +242,7 @@ namespace System
     public static class Console
     {
         public static void WriteHelloWorld() => WriteLine("Hello World!");
+        <!-- markdownlint-disable MD009 -->
```
adamsitnik: nit: I am not sure about this line, GH shows it in the sample code and it looks a little weird:

obraz

jorive: @billwert You need to change ```cs with ```csharp in the block comment.

docs/benchmarkdotnet.md
```diff
@@ -114,45 +114,46 @@ And select one of the benchmarks from the list by either entering its number or
 #### Filtering the Benchmarks
 
 You can filter the benchmarks using `--filter $globPattern` console line argument. The filter is **case insensitive**.
-
+<!-- markdownlint-disable MD029 -->
```
jorive: Is the warning about non-indented bullet points? Can they be fixed instead?

docs/benchmarkdotnet.md
```diff
@@ -114,45 +114,46 @@ And select one of the benchmarks from the list by either entering its number or
...
 ```
 
 6. To print a **joined summary** for all of the benchmarks (by default printed per type), use `--join`:
 
 ```cmd
-dotnet run -f netcoreapp2.1 --filter BenchmarksGame* --join
+dotnet run -c Release -f netcoreapp2.1 --filter BenchmarksGame* --join
 ```
 
+<!-- markdownlint-enable MD029 -->
```
jorive: Same as above.

jorive:
Please do not add markdown warnings tags. Those warnings can be fixed.

docs/benchmarkdotnet.md
billwert: no, it's about having a bunch of stuff between the numbers like we do. It's not really fixable to get the effect we want, which is a series of steps with lots of description in them. That's why I chose to disable this warning; it's not really a problem in this context.

docs/benchmarkdotnet.md
billwert: see above.

docs/benchmarking-workflow-corefx.md
billwert: nah, the actual problem is that the XML comment was not rendering properly in the code block (which makes sense.) cs and csharp seem the same.

I'm just going to delete the blank line instead. :)

billwert:
@jorive the warnings didn't seem trivially reasonable to fix in the context (specifically the list one. The space one could have been fixed, which I have done.)

jorive:
If the markdown renders properly, could you please ignore and remove the tags?

billwert:
@jorive swapped to bullets.

PR #394, opened 2019-03-19
Methods for benchmarking TakeLast

Romasz:
I've added methods for benchmarking System.Linq.TakeLast, however this needed constants definition, hence TakeLast is not available in net472. 

I've also thought about adding _WrapperType.IList_ to _IterationSizeWrapperData()_, but this produces the same results as _WrapperType.NoWrap_.

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -10,6 +10,10 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
+    <DefineConstants>NETFRAMEWORK</DefineConstants>
+  </PropertyGroup>
+
   <ItemGroup>
     <PackageReference Include="Jil" Version="2.15.4" />
     <PackageReference Include="MessagePack" Version="1.7.3.4" />
@@ -54,10 +58,6 @@
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
     <Compile Remove="coreclr\Math\Functions\Single\**" />
   </ItemGroup>
-
-  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0' Or '$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'netcoreapp2.2' Or '$(TargetFramework)' == 'netcoreapp3.0'">
-    <DefineConstants>NETCORE</DefineConstants>
-  </PropertyGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or '$(TargetFramework)' == 'netcoreapp2.0'">
     <Compile Remove="corefx\System.IO.Compression\Brotli.cs" />
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -10,10 +10,6 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
-    <DefineConstants>NETFRAMEWORK</DefineConstants>
-  </PropertyGroup>
-
   <ItemGroup>
     <PackageReference Include="Jil" Version="2.15.4" />
     <PackageReference Include="MessagePack" Version="1.7.3.4" />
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -55,6 +55,10 @@
     <Compile Remove="coreclr\Math\Functions\Single\**" />
   </ItemGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0' Or '$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'netcoreapp2.2' Or '$(TargetFramework)' == 'netcoreapp3.0'">
```
jorive: Wouldn't it be easier to just exclude .NET Framework?
```xml
<ItemGroup Condition=" '$(TargetFrameworkIdentifier)' != '.NETFramework'>
```

src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
```diff
@@ -145,6 +145,31 @@ public void Skip(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType
         public void Take(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType) 
             => Perf_LinqTestBase.Measure(_sizeToPreallocatedArray[size], wrapType, col => col.Take(size - 1), _consumer);
 
+#if NETCORE
```
jorive: Can you replace this with `#if !NETFRAMEWORK`?

src/benchmarks/micro/MicroBenchmarks.csproj
Romasz: Right.

src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
Romasz: Done

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -10,6 +10,10 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
```
adamsitnik: this is not needed, MSBuild does it for us. We are already using `#if !NETFRAMEWORK` in many places: https://github.com/dotnet/performance/search?q=NETFRAMEWORK&unscoped_q=NETFRAMEWORK

Please remove it

src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
```diff
@@ -145,6 +145,31 @@ public void Skip(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType
...
+        public void TakeLast(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType)
+            => Perf_LinqTestBase.Measure(_sizeToPreallocatedArray[size], wrapType, col => col.TakeLast(size - 1), _consumer);
+
+        [Benchmark]
+        [ArgumentsSource(nameof(IterationSizeWrapperData))]
+        public int[] TakeLastToArray(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType)
+        {
+            IEnumerable<int> source = Perf_LinqTestBase.Wrap(_sizeToPreallocatedArray[size], wrapType);
+
+            return source.TakeLast(size - 1).ToArray();
```
adamsitnik: `ToArray()` is a very expensive operation which would dominate the reported time for this benchmark, please remove the `TakeLastToArray` benchmark, what you wrote in `TakeLast` is enough

src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
```diff
@@ -145,6 +145,31 @@ public void Skip(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType
...
+        }
+
+        [Benchmark]
+        [ArgumentsSource(nameof(IterationSizeWrapperData))]
+        public List<int> TakeLastToList(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType)
+        {
+            IEnumerable<int> source = Perf_LinqTestBase.Wrap(_sizeToPreallocatedArray[size], wrapType);
+
+            return source.TakeLast(size - 1).ToList();
```
adamsitnik: same as above, `ToList()` is a very expensive operation which would dominate the reported time for this benchmark, please remove the `TakeLastToList` benchmark

src/benchmarks/micro/MicroBenchmarks.csproj
Romasz: Done

src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
Romasz: Generally the source in the test is already IList so there indeed there is no need to make separate for ToArray/ToList, the NoWrap test shows enough difference.

Done

src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
Romasz: Like above. Done

adamsitnik:
@Romasz once again thanks!

Issue #404, opened 2019-03-26
add CI test to ensure no duplicate test IDs


billwert:
#403 shows that it's possible to check in tests with the same logical ID, which means same set of parameters. We should try hard to prevent this, since it represents wasted time in the lab and also causes problems with tooling (like ResultsComparer.)

PR #407, opened 2019-03-27
Unique values

### LINKED ISSUE #403: UInt tests duplicate 0 as an input
See: https://github.com/dotnet/performance/blob/e720dbf5cf7f288e55ba84196fad6af8011f132a/src/benchmarks/micro/corefx/System.Runtime/Perf.UInt16.cs#L17-L21 

UInt32 and UInt64 use `12345` as their middle value and are fine.

adamsitnik:
Fixes #403 and #404 

@billwert I created a new validator and added it to the recommended config, so if anyone tries to add a duplicated argument the CI is not going to allow for that


jorive:
We need to do something about harnessing and running the unit tests as part of CI.

billwert:
LGTM.

PR #410, opened 2019-03-28
Adds benchmark tests for ReadOnlySequence<T>.FirstSpan 

maryamariyan:
Adds benchmark tests for ReadOnlySequence<T>.FirstSpan covering all cases

Related to https://github.com/dotnet/corefx/issues/33029

cc: @adamsitnik @ahsonkhan

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -38,13 +38,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Reference Include="System.Memory">
-      <HintPath>..\..\..\..\corefx\artifacts\bin\ref\netcoreapp\System.Memory.dll</HintPath>
-      <Private>false</Private>
-    </Reference>
-  </ItemGroup>
-
-  <ItemGroup>
     <None Update="coreclr\BenchmarksGame\Inputs\*.txt">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -20,6 +20,7 @@
     <PackageReference Include="protobuf-net" Version="2.3.7" />
     <PackageReference Include="System.Drawing.Common" Version="4.5.1" />
     <PackageReference Include="System.IO.Pipelines" Version="4.5.3" />
+    <PackageReference Include="System.Memory" Version="4.5.1" />
     <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
     <PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -20,7 +20,6 @@
     <PackageReference Include="protobuf-net" Version="2.3.7" />
     <PackageReference Include="System.Drawing.Common" Version="4.5.1" />
     <PackageReference Include="System.IO.Pipelines" Version="4.5.3" />
-    <PackageReference Include="System.Memory" Version="4.5.1" />
     <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
     <PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
@@ -36,6 +35,13 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\harness\BenchmarkDotNet.Extensions\BenchmarkDotNet.Extensions.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="System.Memory">
+      <HintPath>..\..\..\..\corefx\artifacts\bin\ref\netcoreapp\System.Memory.dll</HintPath>
+      <Private>false</Private>
+    </Reference>
   </ItemGroup>
 
   <ItemGroup>
--- a/src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
+++ b/src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
@@ -76,24 +76,24 @@ namespace System.Buffers.Tests
         public long SliceArray() => Slice(new ReadOnlySequence<T>(_array));
 
         [Benchmark]
-        public int IterateTryGetMemory() => IterateTryGet(new ReadOnlySequence<T>(new ReadOnlyMemory<T>(_array)));
+        public int IterateTryGetMemory() => IterateTryGet(new ReadOnlySequence<T>(_memory));
 
         [Benchmark]
-        public int IterateForEachMemory() => IterateForEach(new ReadOnlySequence<T>(new ReadOnlyMemory<T>(_array)));
+        public int IterateForEachMemory() => IterateForEach(new ReadOnlySequence<T>(_memory));
 
         [Benchmark]
-        public int IterateGetPositionMemory() => IterateGetPosition(new ReadOnlySequence<T>(new ReadOnlyMemory<T>(_array)));
+        public int IterateGetPositionMemory() => IterateGetPosition(new ReadOnlySequence<T>(_memory));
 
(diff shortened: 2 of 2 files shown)


src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -20,7 +20,6 @@
     <PackageReference Include="protobuf-net" Version="2.3.7" />
     <PackageReference Include="System.Drawing.Common" Version="4.5.1" />
     <PackageReference Include="System.IO.Pipelines" Version="4.5.3" />
-    <PackageReference Include="System.Memory" Version="4.5.1" />
```
maryamariyan: Didn't seem to be required. I was able to build and run tests without this. 

for example before even adding new tests (which includes the new FirstSpan API), I wouldn't have needed the local reference below on line 40 and also without this packagereference to System.Memory I could still get the benchmarks running

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
```diff
@@ -26,7 +27,10 @@ public class ReadOnlySequenceTests<T>
         public int IterateGetPositionArray() => IterateGetPosition(new ReadOnlySequence<T>(_array));
 
         [Benchmark(OperationsPerInvoke = 16)]
-        public int FirstArray() => First(new ReadOnlySequence<T>(_array));
+        public int First_Array() => First(new ReadOnlySequence<T>(_array));
```
billwert: Please don't change `[Benchmark]` method names. This will break our reporting.

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
```diff
@@ -26,7 +27,10 @@ public class ReadOnlySequenceTests<T>
...
+
+        [Benchmark(OperationsPerInvoke = 16)]
+        public int FirstSpan_Array() => FirstSpan(new ReadOnlySequence<T>(_array));
```
billwert: Remove _ to be consistent with other names in this file.

billwert:
I didn't leave a comment on each instance of the new underscores. They should all be fixed. Thanks!

src/benchmarks/micro/MicroBenchmarks.csproj
jorive: After removing this package reference, does it still build/run in all target frameworks?

maryamariyan: Good question. Let me check...

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -38,6 +37,13 @@
     <ProjectReference Include="..\..\harness\BenchmarkDotNet.Extensions\BenchmarkDotNet.Extensions.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
jorive: Why do we need this here?

maryamariyan: I needed this so that locally MicroBenchmarks project could pick up my new API (FirstSpan)

According to [benchmarking-workflow-corefx.md#PR](https://github.com/dotnet/performance/blob/master/docs/benchmarking-workflow-corefx.md#PR),
and from my understanding, after my corefx PR (with the new API) gets merged and a new NuGet package is published to the CoreFX NuGet feed, I can remove the reference below to System.Memory.dll and install/update the package consumed by MicroBenchmarks

jorive: After this is merged, could you please open an issue to track this?

jorive: Does it need to be guarded under kind of "netcoreapp3.0"?

maryamariyan: > After this is merged, could you please open an issue to track this?

I don't think I would be merging this PR which is on performance repo, until the corefx repo containing the new API gets merged.
Once that happens I think I should add another commit here that install the correct corefx nuget package for System.Memory.

> Does it need to be guarded under kind of "netcoreapp3.0"?
cc: @adamsitnik what is your recommendation?

src/benchmarks/micro/MicroBenchmarks.csproj
maryamariyan: Yes, without PackageReference to System.Memory it still builds in the following target frameworks:

- net461
- netcoreapp2.0
- netcoreapp2.1
- netcoreapp2.2
- netcoreapp3.0

src/benchmarks/micro/MicroBenchmarks.csproj
ahsonkhan: > I don't think I would be merging this PR which is on performance repo

Right. I don't think the intention was to merge this change. For instance the hint path is specific to @maryamariyan's local build/directories.

> Does it need to be guarded under kind of "netcoreapp3.0"?

I believe so. See my previous comment: https://github.com/dotnet/performance/pull/410/files#r270154876

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
```diff
@@ -28,27 +29,73 @@ public class ReadOnlySequenceTests<T>
...
+        public int FirstMemory() => First(new ReadOnlySequence<T>(_memory));
 
         [Benchmark(OperationsPerInvoke = 10)]
-        public long SliceMemory() => Slice(new ReadOnlySequence<T>(new ReadOnlyMemory<T>(_array)));
+        public long SliceMemory() => Slice(new ReadOnlySequence<T>(_memory));
 
         [GlobalSetup(Targets = new [] { nameof(IterateTryGetSingleSegment), nameof(IterateForEachSingleSegment), nameof(IterateGetPositionSingleSegment), nameof(FirstSingleSegment), nameof(SliceSingleSegment) })]
         public void SetupSingleSegment() => _startSegment = _endSegment = new BufferSegment<T>(new ReadOnlyMemory<T>(_array));
 
+        [GlobalSetup(Targets = new [] { nameof(FirstMemory), nameof(SliceMemory), nameof(IterateTryGetMemory), nameof(IterateForEachMemory), nameof(IterateGetPositionMemory) })]
```
ahsonkhan: Don't we need this setup for `FirstSpanMemory` as well which is if/def'd for .NET Core 3.0 only?

Otherwise, what's the value of `_memory` there?
https://github.com/dotnet/performance/pull/410/files#diff-19d9aaa505dcc18077d456ddc9a86808R37

We might need to if/def this setup too, if that's the suggested approach for adding benchmarks specific to .NET Core 3.0 APIs.

maryamariyan: There is GlobalSetup for FirstSpan* benchmarks inside the ifdef which will call functions outside of ifdef instantiating the _memory

ahsonkhan: Oh right, missed that one. That works :)

src/benchmarks/micro/corefx/System.Buffers/ReadOnlySequenceTests.cs
```diff
@@ -28,27 +29,73 @@ public class ReadOnlySequenceTests<T>
         [Benchmark(OperationsPerInvoke = 16)]
         public int FirstArray() => First(new ReadOnlySequence<T>(_array));
 
+#if NETCOREAPP3_0
```
ahsonkhan: @billwert, @jorive, I am just curious, is this the expected way to single out API benchmarks that only exist on certain TFMs? What if I want to add benchmarks for types that only exist on netstandard2.0, or only exist on .NET Core 2.0+, etc.?

billwert: good question. I don't think we have a solid plan yet. (And this approach is fragile, of course.) @adamsitnik may have considered this - let's see what he says when he's back from vacation.

adamsitnik: @ahsonkhan it depends. if it's a single method or just a few like here, you can use `#if TFM`. If it's an entire type, you should exclude it per certain TFM in the `.csproj` file

https://github.com/dotnet/performance/blob/eaec460c81e07410bb293b68d77315fe90ecd2f9/src/benchmarks/micro/MicroBenchmarks.csproj#L60-L61

ahsonkhan: Sure, I figured that would be the expected approach. That's what I ended up doing for the recent JsonWriter perf tests. Thanks for clarifying.

PR #413, opened 2019-04-02
Basic summary for comparison

### LINKED ISSUE #408: Provide some kind of summary result in ResultsComparer
I have two large test run result sets (787 results each) and would like some overall summary of the improvements and regressions. [Jit-diffs](https://github.com/dotnet/jitutils) has this for code size diffs and we've found it useful for getting an overall impression of a change.

```]
Total bytes of diff: 162 (0.00% of base)
    diff is a regression.
111 total files with size differences (43 improved, 68 regressed), 18 unchanged.
8440 total methods with size differences (4486 improved, 3954 regressed), 191683 unchanged.
```

It would be nice for `ResultsComparer` to give some idea of the distribution of differences, eg a histogram or similar.
* Even crude “improved” “regressed” “same” buckets might be enough…
* Aggregation is tricky but geomean of the base/diff ratios (of medians) is plausible as an overall figure of merit

For example: ```102 regressed, 157 improved, 500 unchanged; geomean (improvement) 1.02```

billwert:
Prints out the number of deltas and the geomean of their ratios.

Fixes #408

### DIFF
--- a/src/tools/ResultsComparer/Program.cs
+++ b/src/tools/ResultsComparer/Program.cs
@@ -84,8 +84,10 @@ namespace ResultsComparer
             better = better.Where(x => GetRatio(x) != double.PositiveInfinity);
             worse = worse.Where(x => GetRatio(x) != double.PositiveInfinity);
 
-            var betterGeoMean = Math.Pow(better.Skip(1).Aggregate(GetRatio(better.First()), (x, y) => x * GetRatio(y)), (1.0 / better.Count()));
-            var worseGeoMean = Math.Pow(worse.Skip(1).Aggregate(GetRatio(worse.First()), (x, y) => x * GetRatio(y)), (1.0 / worse.Count()));
+            var betterGeoMean = Math.Pow(10, better.Skip(1).Aggregate(Math.Log10(GetRatio(better.First())), (x, y) => x + Math.Log10(GetRatio(y))) / better.Count());
+            var worseGeoMean = Math.Pow(10, worse.Skip(1).Aggregate(Math.Log10(GetRatio(worse.First())), (x, y) => x + Math.Log10(GetRatio(y))) / worse.Count());
+
+
             Console.WriteLine("summary:");
             Console.WriteLine($"better: {betterCount}, geomean: {betterGeoMean:F3}");
             Console.WriteLine($"worse: {worseCount}, geomean: {worseGeoMean:F3}");
--- a/src/tools/ResultsComparer/Program.cs
+++ b/src/tools/ResultsComparer/Program.cs
@@ -44,6 +44,8 @@ namespace ResultsComparer
 
             var notSame = GetNotSameResults(args, testThreshold, noiseThreshold).ToArray();
 
+            PrintSummary(notSame);
+
             PrintTable(notSame, EquivalenceTestConclusion.Slower, args);
             PrintTable(notSame, EquivalenceTestConclusion.Faster, args);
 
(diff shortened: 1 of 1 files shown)


billwert:
@AndyAyersMS turns out I also needed this recently. this will add a summary like this:
```
summary:
better: 1013, geomean: 1.514
worse: 617, geomean: 1.337
total diff: 1630
```
We can think about better things as we roll through the other projects in flight but this seems like a reasonable stopgap.

src/tools/ResultsComparer/Program.cs
```diff
@@ -70,6 +72,28 @@ private static IEnumerable<(string id, Benchmark baseResult, Benchmark diffResul
...
+            var worse = notSame.Where(result => result.conclusion == EquivalenceTestConclusion.Slower);
+            var betterCount = better.Count();
+            var worseCount = worse.Count();
+
+            // If the baseline doesn't have the same set of tests, you wind up with Infinity in the list of diffs.
+            // Exclude them for purposes of geomean.
+            better = better.Where(x => GetRatio(x) != double.PositiveInfinity);
+            worse = worse.Where(x => GetRatio(x) != double.PositiveInfinity);
+
+            var betterGeoMean = Math.Pow(better.Skip(1).Aggregate(GetRatio(better.First()), (x, y) => x * GetRatio(y)), (1.0 / better.Count()));
```
jorive: When computing the geometric mean you should use the relationship with logarithms to minimize the chances of overflow. More here: https://en.wikipedia.org/wiki/Geometric_mean#Relationship_with_logarithms

jorive: To generalize the computation, you could also take into account zero values in different metrics (for example allocated memory in current thread metric for Span benchmarks).

jorive: I'm guessing resolved as "Won't fix"?

billwert: Correct.

billwert: Apparently typing a comment and clicking resolve is the wrong flow. I typed up a reasonable reason. 🙂 I’ll see if it’s still on a text box at work or re type it once I get to a real keyboard.

billwert: @jorive 
re: overflow, sure. good suggestion.

re: generalizing: the tool only cares about time/op right now, making it more generic would be a lot more work. I think that's a good note for the work being done on new tools. This is more of a stopgap to scratch the itch Andy and I both had.

AndyAyersMS:
Thanks -- this looks useful.

src/tools/ResultsComparer/Program.cs
```diff
@@ -70,6 +72,30 @@ private static IEnumerable<(string id, Benchmark baseResult, Benchmark diffResul
...
         }
 
+        private static void PrintSummary((string id, Benchmark baseResult, Benchmark diffResult, EquivalenceTestConclusion conclusion)[] notSame)
+        {
+            var better = notSame.Where(result => result.conclusion == EquivalenceTestConclusion.Faster);
+            var worse = notSame.Where(result => result.conclusion == EquivalenceTestConclusion.Slower);
+            var betterCount = better.Count();
+            var worseCount = worse.Count();
+
+            // If the baseline doesn't have the same set of tests, you wind up with Infinity in the list of diffs.
```
adamsitnik: 👍 great, thanks for taking care of this edge case!

PR #437, opened 2019-04-19
React to Utf8JsonWriter API changes and use built-in ABW

ahsonkhan:
Fixes https://github.com/dotnet/performance/issues/436

Follow-up to:
https://github.com/dotnet/corefx/pull/36961
https://github.com/dotnet/performance/pull/427

cc @billwert

Issue #348, opened 2019-02-28
Change "official build" trigger from daily to triggered on core-sdk builds


billwert:
This enables us to be more precise about when we run and get finer grained data. If it gets to being more than a few runs a day we may need to reconsider this.

PR #438, opened 2019-04-19
add explicit dependency to latest CoreFX using maestro dependency flow

adamsitnik:


billwert:
Can you share the plan please?

adamsitnik:
> Can you share the plan please?

Sure! I was going to do that after the build gets green. 

As of today, we don't have an explicit dependency to latest libraries. We reference stable versions of `System.*` and use the python script to download the latest dotnet cli. Script runs the benchmarks using the latest version (because the cli has such `.dlls` in the SDK folder, not because we reference them).

The problem are API changes like the two most recent ones that have changed the existing APIs: we can build the code locally but we get the runtime error when we try to use the changed methods in the CI:
https://github.com/dotnet/performance/pull/437
https://github.com/dotnet/performance/pull/429

And using new methods: we reference old versions, so we can't use new methods locally, but we can in the CI. Example:
https://github.com/dotnet/performance/pull/410

My idea here is to add a maestro dependency to CoreFX, so anytime CoreFX publishes new packages, the maestro updates referenced CoreFX dlls. Like it does for Arcade today.

If it's going to work as expected it should also solve https://github.com/dotnet/performance/issues/348
We would basically get a maestro update for every new System.* package which would allow us to run the benchmarks not once a day, but after every update. It should be up to 4-6 times a day.

ASP.NET, CoreCLR and EF are configured in a very similar way. ASP.NET example:
https://github.com/aspnet/AspNetCore/blob/master/eng/Version.Details.xml#L328
https://github.com/aspnet/AspNetCore/blob/master/eng/Versions.props#L35

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -18,17 +20,30 @@
...
-    <PackageReference Include="System.Memory" Version="4.5.1" />
-    <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
-    <PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
-    <PackageReference Include="System.Runtime.Serialization.Json" Version="4.3.0" />
-    <PackageReference Include="System.Security.Cryptography.Primitives" Version="4.3.0" />
-    <PackageReference Include="System.Threading.Channels" Version="4.5.0" />
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.2" />
-    <PackageReference Include="System.Xml.XmlSerializer" Version="4.3.0" />
+    <PackageReference Include="System.Drawing.Common">
```
jorive: Why the change to `Version` as a child node element?
This should have worked:
```xml
<PackageReference Include="System.Drawing.Common" Version="$(SystemDrawingCommonPackageVersion)" />
```

billwert:
Thanks for the details, @adamsitnik!

This is a great discussion which touches on several issues. It has also prompted me to deepen my understanding of how package versions/references work, which is very helpful.

I think we will end up breaking this apart into a few issues, but I'd like to discuss it here first.

# Current Issues In This Area


1. Microbenchmarks.csproj references old packages, not live packages.
    a. It also references some packages which have moved into the shared runtime at some point.
2. We need more than one run of the repo daily.
3. CI breaks when API surface changes for existing tests.
4. Adding tests using new surface area is hard.

## `PackageReference` version issues

I didn't realize that we had references to dotnet/corefx packages which were outside of the shared runtime. We need to get to use the correct versions of these packages to effectively catch regressions. I propose how in the discussion of running core-sdk below. Fortunately, for our references to things inside the shared runtime package, the shared package wins. This makes our references extraneous.  I think  `Microbenchmarks.csproj` should change to conditionally reference packages based on TFM. A `PackageReference` to non-shared libraries can be used where it is necessary. This clarifies intention.

## Run schedule

As you note, we'd like to run the repo more than once per day. Our current plan, which is in flight with @adiaaida, is to trigger runs of core-sdk when it publishes new builds into the dependency graph. This is similar to what you have here. We consider core-sdk the official build, as it directly tests the configuration we ship. This is an important principle.

The challenge with using core-sdk as the trigger is finding matching versions of the packages not included in the shared runtime. I believe it is possible through BAR to identify the corefx build (hash) which is included in the current core-sdk. We can then use BAR to find the package versions we need and include them in the current run.

I propose we solve the run schedule problem by continuing with our plan to use core-sdk builds. There are a couple details to figure out, such as how to include the non-shared runtime package versions we determine with the PR as it goes through CI. That seems solvable. One benefit of this plan is that maestro will trigger PRs, and we can change the actual run to install the specified version in version.details.xml instead of latest. This will prevent actual runs from happening if CI is broken.

### Contrast with using corefx as the trigger

> Using corefx to trigger as we have here, but continuing to install (and label builds as) the latest core-sdk, is confusing. We wouldn't be using the runtime that came in that core-sdk build. We would be using newer corefx packages than came with that corefx build.
We also continue to have an intention to run corefx/coreclr builds directly. This takes a different shape as those repos do not produce a runnable product, requiring further automation. This moves adding those runs further down the backlog. We should save the dependency pointer to corefx for that purpose.

## CI breaks

I do not believe this is an avoidable problem. Moving the test code out of the repos where the breaks happen necessarily means that those breaks must flow through the system before they can be fixed. The mitigation is that this repo's CI will prevent actual runs from taking place when these breaks exist. It will be clearer what caused the break. It will also mean that unrelated PRs should not break. 

## Adding tests using new surface area is hard

There are two cases for this that I can see:
1. A developer is trying to add new tests for surface area they are actively writing.
2. A developer wants to add tests for surface area that's already in the product.

For 1, I do not think there is an option other than following our [guidance](https://github.com/dotnet/performance/blob/master/docs/benchmarking-workflow-corefx.md#Benchmarking-new-API). For 2, they can validate it using the script (as outlined) or by installing a latest SDK on their machine. Even if we added package references like this, it will still fail if the user doesn't have the proper shared runtime installed:
```cmd
PS E:\source\performance\src\benchmarks\micro [master ≡ +1 ~3 -0 !]> dotnet run -c Release -f netcoreapp3.0
It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '3.0.0-preview5-27618-16' was not found.
```

# Summary

This PR is heading in the direction we want. I do not think it is precisely the place we want to land in the short term, however. My understanding is that the work to enable core-sdk dependency flow was blocked on work to get preview4 out the door. Now that we have shipped preview4, I expect this to happen quickly. Getting this done will alleviate part of this. We'll also prioritize the solution to ensuring we have matching non-shared runtime packages.

I would like to close this PR. I appreciate the discussion it has triggered. I'll open issues to track the separate issues.

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -9,6 +9,8 @@
     <DebugSymbols>true</DebugSymbols>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 
+    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp3.0' ">$(MicrosoftNETCoreAppPackageVersion)</RuntimeFrameworkVersion>
```
billwert: I think this still causes problems, FWIW. Whatever is latest runtime version is likely to be ahead of what is in the shared runtime in core-sdk, so this will break.

PR #441, opened 2019-04-23
Add Roslyn macro benchmarks

agocke:
These benchmarks measure the performance of individual compiler
stages on a frozen version of Roslyn.

### DIFF
--- a/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
+++ b/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
@@ -1,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <IsShipping>false</IsShipping>
+    <StrongNameKeyId>MicrosoftShared</StrongNameKeyId>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="README.md" />
+    <None Include="run-perf.ps1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.1.0-beta1-final" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\harness\BenchmarkDotNet.Extensions\BenchmarkDotNet.Extensions.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
+++ b/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
@@ -10,11 +10,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <None Include="README.md" />
-    <None Include="run-perf.ps1" />
-  </ItemGroup>
-
-  <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.1.0-beta1-final" />
   </ItemGroup>
 
--- a/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
+++ b/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
     <IsShipping>false</IsShipping>
     <StrongNameKeyId>MicrosoftShared</StrongNameKeyId>
   </PropertyGroup>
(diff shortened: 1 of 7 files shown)


agocke:
It looks like this repo uses CoreRun to run against locally built copies, but that's extremely inconvenient for Roslyn as it would mean we would have to build all of CoreFX just to benchmark Roslyn performance. Is there any other way we could override the Roslyn binaries?

billwert:
Hey @agocke!

The CoreRun usage is for local runs of CoreFX/CoreCLR. Actual runs download the latest core-sdk and run against that. (We will be transitioning this to running all core-sdk builds once some work is done to get them properly flowed into dependency flow.)

You can use the [benchmarks_ci script](https://github.com/dotnet/performance/blob/master/scripts/benchmarks_ci.py) to emulate what the lab does. Would that help you test this to ensure you're getting the behavior you want?

An example of running the ML.net benchmarks, which is what we'd want to emulate here:
```
PS E:\source\agocke\src\benchmarks\real-world\Microsoft.ML.Benchmarks [add-roslyn-macro-benchmarks ≡]> py ..\..\..\..\scripts\benchmarks_ci.py --csproj .\Microsoft.ML.Benchmarks.csproj --incremental no --architecture x64 -f netcoreapp3.0 --bdn-arguments "--allCategories mldotnet"
```

agocke:
@billwert Thanks, I'll do that for the lab.

But I also want to make it possible for local runs so that I can compare the results of two local builds of Roslyn on my machine.

src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
```diff
@@ -0,0 +1,25 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <IsShipping>false</IsShipping>
+    <StrongNameKeyId>MicrosoftShared</StrongNameKeyId>
+  </PropertyGroup>
+
+  <ItemGroup>
```
jorive: I do not see these two files in the PR. Is this intended?

billwert:
> But I also want to make it possible for local runs so that I can compare the results of two local builds of Roslyn on my machine.

I *think* this will be fine, and you'd just need to change your package reference to Roslyn when you run locally. The CoreRun instructions are really for folks who are changing things in CoreFX/CoreCLR.

src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
agocke: Ah, I deleted them because they were obsolete and forgot to remove them from the project file. Thanks.

src/benchmarks/real-world/Roslyn/StageBenchmarks.cs
```diff
@@ -0,0 +1,114 @@
...
+using static Microsoft.CodeAnalysis.Compilation;
+
+namespace CompilerBenchmarks
+{
+    public class StageBenchmarks
+    {
+        private Compilation _comp;
+        private CommonPEModuleBuilder _moduleBeingBuilt;
+        private EmitOptions _options;
+        private MemoryStream _peStream;
```
jorive: `_peStream` seems to be setup only in `LoadCompilation` for the `CompileMethodsAndEmit` method/benchmark. On the other hand, `_peStream` is used in the `SerializeMetadata` method/benchmark, but it's not initialized/reset in the global setup for it (the `CompileMethods` method). Is this intentional?

agocke: The Setup method for the other one calls LoadCompilation. Should be fine.

src/benchmarks/real-world/Roslyn/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+            if (!Directory.Exists(sourceDir))
+            {
+                await FileTasks.DownloadAndUnzip(cscSourceDownloadLink, sourceDownloadDir);
+            }
+
+            // Benchmark.NET creates a new process to run the benchmark, so the easiest way
+            // to communicate information is pass by environment variable
+            Environment.SetEnvironmentVariable(Helpers.TestProjectEnvVarName, sourceDir);
+
+            _ = BenchmarkRunner.Run<StageBenchmarks>(config);
```
billwert: Optional: if you use `BenchmarkSwitcher` instead of `BenchmarkRunner`, it's easier to run individual benchmarks locally. This doesn't impact the actual benchmarks so you can change this in the future if you like.

src/benchmarks/real-world/Roslyn/StageBenchmarks.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        [Benchmark]
+        public object SerializeMetadata()
+        {
+            var diagnostics = DiagnosticBag.GetInstance();
+
+            _comp.SerializeToPeStream(
+                _moduleBeingBuilt,
+                new SimpleEmitStreamProvider(_peStream),
```
billwert: does _peStream need to be reset to zero each iteration?

agocke: That's probably a good idea.

src/benchmarks/real-world/Roslyn/Program.cs
agocke: I'll take care of this later. I'd like to start it running first

adamsitnik:
The CI failure seems to be unrelated.

```log
/__w/1/s/.dotnet/sdk/2.1.503/NuGet.targets(114,5): error : Unable to load the service index for source https://dotnet.myget.org/F/symreader/api/v3/index.json. [/__w/1/s/eng/common/helixpublish.proj]
/__w/1/s/.dotnet/sdk/2.1.503/NuGet.targets(114,5): error :   The HTTP request to 'GET https://dotnet.myget.org/F/symreader/api/v3/index.json' has timed out after 100000ms. [/__w/1/s/eng/common/helixpublish.proj]
```

I will try to re-trigger it.

/azp run

adamsitnik:
@agocke @billwert is there any reason why we run those benchmarks only for Windows? What about Linux? It's a matter of adding few lines to the yaml file and I would be happy to do it.

PR #447, opened 2019-04-30
Use our own helix proj file

adiaaida:
Using our own proj file will enable us to split some of the work into multiple pieces, especially when we have scenarios like startup time, etc, that we want to run in parallel.

### DIFF
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -91,7 +91,7 @@ jobs:
         - ${{ if eq(parameters.osName, 'windows') }}:
           - script: (robocopy $(Build.SourcesDirectory) $(Build.SourcesDirectory)\notLocked /E /XD $(Build.SourcesDirectory)\notLocked $(Build.SourcesDirectory)\artifacts $(Build.SourcesDirectory)\.git) ^& IF %ERRORLEVEL% LEQ 1 exit 0
             displayName: Workaround to avoid locked file # https://github.com/dotnet/arcade/issues/2125
-        - template: /eng/common/templates/steps/send-to-helix.yml
+        - template: /eng/performance/send-to-helix.yml
           parameters:
             HelixSource: 'pr/dotnet/performance/$(Build.SourceBranch)' # sources must start with pr/, official/, prodcon/, or agent/
             HelixType: 'test/performance_$(_BuildConfig)/'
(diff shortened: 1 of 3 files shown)


adiaaida:
@billwert Let me know what you think of this so far. I'm going to use this PR to also test splitting our runs into multiple sections, so we know how that works when scenarios are ready.

adiaaida:
Link to MC for one of the legs: https://mc.dot.net/#/user/dotnet-performance/pr~2Fdotnet~2Fperformance~2Frefs~2Fpull~2F447~2Fmerge/test~2Fperformance_x64_coreclr_netcoreapp2.1_micro~2F/20190430.7

How this works:

If you uncheck "Show Failures only," you will see that we have 3 tasks: 1 for the HelixController, 1 called "Echo" and one called "WorkItem." Echo was the test for adding a second HelixWorkItem, which successfully echoed Hello World. WorkItem is the normal run, which also ran as expected.

billwert:
this LGTM. I think if you want to put this in now and then work from there that's a great idea.

adamsitnik:
How are we going to handle Arcade updates in `/eng/common/templates/steps/send-to-helix.yml`? Will it require manual updates of `eng/performance/send-to-helix.yml`?

Is it somehow possible to make `/eng/common/templates/steps/send-to-helix.yml` extend `eng/performance/send-to-helix.yml` ?

adiaaida:
@adamsitnik This change actually moves us to how all the other repos work. The template is just that: a template. It's extremely weak and doesn't give us nearly the amount of freedom to do what we want to do. If you check out coreclr and corefx, you will see that they don't even reference the template. They have their own special logic for the yaml of sending to helix and the project file for sending to helix, and it's MUCH more powerful that using the template. The template was never meant to be the end-all be-all of sending jobs to helix. We were using it because it was there, and I wasn't aware that we could stray from it. When we want to do more than the template allows us to do, this is the recommended path.

adamsitnik:
I understand that we want to have some custom logic (and why we need it) but what I wanted to know is if there is a way to create a new template that derives from an existing template instead of creating a copy of it and adding new stuff to it. To avoid sync issues with the parent template over time and make it more clear what is custom and what is not.

I remember when I've tried to add something to the .yml template in CoreCLR which did not sync with `send-to-helix.yml` for 2 months and the lack of sync (parameter renames, new params etc) made it hard for me to understand how it all works and what I need to change.

adiaaida:
Ah. I see. No, I don’t think there is a way to derive from this template, especially in this case where the proj file is specified in the template. We do, sort of, derive from the job template, but the helix step template isn’t set up for that.

PR #450, opened 2019-05-01
Add option to specify Dotnet Version

adiaaida:
This will enable us to specify which version of the dotnet sdk we want to use based on what we get from darc updates.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -27,6 +27,7 @@ from logging import getLogger
 import os
 import platform
 import sys
+import re
 
 from performance.common import validate_supported_runtime
 from performance.logger import setup_loggers
@@ -54,6 +55,7 @@ if sys.platform == 'linux' and "linux_distribution" not in dir(platform):
 
 def init_tools(
         architecture: str,
+        version: str,
         target_framework_monikers: list,
         verbose: bool) -> None:
     '''
@@ -70,6 +72,7 @@ def init_tools(
     dotnet.install(
         architecture=architecture,
         channels=channels,
+        version=version,
         verbose=verbose,
     )
     benchview.install()
@@ -140,6 +143,24 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
             "%%Y-%%m-%%dT%%H:%%M:%%SZ").'''
     )
 
+    def __is_valid_sdk_version(version:str) -> str:
+        try:
+            if version is None or re.search('\d\.\d+\.\d+', version):
+                return version
+            else:
+                raise ValueError
+        except ValueError:
+            raise ArgumentTypeError(
+                'Version "{}" is in the wrong format'.format(version))
+
+    parser.add_argument(
+        '--version',
+        required=False,
+        default=None,
+        type=__is_valid_sdk_version,
+        help='Version of the dotnet cli to install in the A.B.C format'
+    )
+
     # Generic arguments.
     parser.add_argument(
         '-q', '--quiet',
@@ -199,6 +220,7 @@ def __main(args: list) -> int:
     # Acquire necessary tools (dotnet, and BenchView)
     init_tools(
         architecture=args.architecture,
+        version=args.version,
         target_framework_monikers=target_framework_monikers,
         verbose=verbose
     )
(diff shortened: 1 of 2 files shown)


scripts/dotnet.py
```diff
@@ -568,6 +580,14 @@ def __process_arguments(args: list):
         help='Download DotNet Cli from the Channel specified.'
     )
 
+    install_parser.add_argument(
```
jorive: You should add this to `def add_arguments(parser: ArgumentParser) -> ArgumentParser:`. Then, you would not need to duplicate it in the two scripts. It will be added automatically after the call to `dotnet.argument`.

jorive: I will suggest renaming `--version` to `--dotnet-version` as there might be different One .NET tools added to the mix in the future.

scripts/benchmarks_ci.py
```diff
@@ -140,6 +143,24 @@ def __is_valid_datetime(dt: str) -> str:
...
+        try:
+            if version is None or re.search('\d\.\d+\.\d+', version):
+                return version
+            else:
+                raise ValueError
+        except ValueError:
+            raise ArgumentTypeError(
+                'Version "{}" is in the wrong format'.format(version))
+
+    parser.add_argument(
```
jorive: This would not be needed after suggestions below, except for moving the validation code to `dotnet.py`

scripts/dotnet.py
adiaaida: Done.

scripts/dotnet.py
```diff
@@ -477,16 +478,28 @@ def install(
...
+            '-Architecture', architecture
+    ]
+    if version is not None:
+        cmdline_args = cmdline_args + [
+            '-Version', version,
         ]
         RunCommand(cmdline_args, verbose=verbose).run(
             get_repo_root_path()
         )
+    else:
```
billwert: can we use the same block and just wrap the conditional around putting `-Version` or `-Channel` in the args?

scripts/dotnet.py
```diff
@@ -532,6 +545,25 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         help='{}'.format(CompilationAction.help_text())
     )
 
+    def __is_valid_sdk_version(version:str) -> str:
+        try:
+            if version is None or re.search('\d\.\d+\.\d+', version):
```
billwert: What is the SDK version for preview builds?

scripts/dotnet.py
```diff
@@ -568,6 +601,14 @@ def __process_arguments(args: list):
         help='Download DotNet Cli from the Channel specified.'
     )
 
+    install_parser.add_argument(
+        '--version',
```
billwert: should this also become `--dotnet-version`?

adiaaida: meant to remove that

scripts/dotnet.py
adiaaida: preview builds are \d.\d.\d+-preview#-\d+. This will match both preview and non-preview versions. It matches any string that contains \d.\d+\d+. We might want to harden it down a bit more so we don't match something like blah2.2.2blah.

scripts/benchmarks_ci.py
```diff
@@ -27,6 +27,7 @@
 import os
 import platform
 import sys
+import re
```
jorive: This can be removed.

scripts/dotnet.py
jorive: It might be possible for users to specify `--channels master --version 2.2.1`. To support this scenario we cannot place the usage of both arguments under different path of the same if block.

billwert: What does that mean? Doesn't a version number supersede a channel?

adiaaida: The channels side runs on a list of channels, so I'm not sure how that would work.

adiaaida: Version does supersede channel. But if multiple channels have been chosen (which our current scripting allows for?) we would run for each channel grab this version, which seems like something we aren't interested in doing (even though the later grabs will just say "such and such version is already installed")

jorive: > What does that mean? Doesn't a version number supersede a channel?
This is the scenario:
```cmd
rem The user wants to run benchmarks against the latest master and version 2.0.0 versions of the CLI.
benchmarks_ci.py --channels master -Version 2.0.0
```

jorive: Optionally, it could be `--dotnet-versions` (an array of versions), that you can use to compare preview4 vs preview5.

adiaaida:
Coming back to this. @billwert do you need anything else? I'm working on something that needs this so I'd like to get it checked in soon

PR #452, opened 2019-05-02
Add proposal doc for shared yaml

adiaaida:


adiaaida:
@billwert @DrewScoggins Let me know if you think I've missed anything/haven't considered some angle.

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+- Product repositories do not try to skirt the yaml template to run performance testing only on the tests they care about
+- Setup time for adding performance testing is as simple as adding a new job that references the yaml template that we provide
+
+## Procedural
+
+- Combine the yaml we already use and, likely, the yaml that has been worked on in coreclr (why reinvent the wheel), with any additional logic necessary for pulling down and running the performance repository in product repos
+  - yaml will need to include any additional parameters/knobs that the product repos may want to turn (ie tiered/non-tiered/minopts, pgo/no-pgo, etc)
+  - yaml will need to include all uploading to benchview logic
+  - yaml should be as barebones as possbile with most of the work done in the proj file
+  - yaml should allow sending of just the performance repo as well as sending the performance repo and the core_root directory (for repos that need to specify that)
```
billwert: How does this square with our conversation around reworking things so that we can send a sane correlation package for all runs/scenarios/etc?

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Success Metric(s)
+
+- Product repositories use the provided yaml template to run all performance testing
+- Product repositories do not try to skirt the yaml template to run performance testing only on the tests they care about
+- Setup time for adding performance testing is as simple as adding a new job that references the yaml template that we provide
+
+## Procedural
+
+- Combine the yaml we already use and, likely, the yaml that has been worked on in coreclr (why reinvent the wheel), with any additional logic necessary for pulling down and running the performance repository in product repos
```
billwert: while we're going through this let's make sure to investigate what the yaml in the repos we want to target looks like, including wpf/winforms/roslyn probably.

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+- We do not have to work around darc/dependency flow
+- We do not need to know why the repo triggered a job (ie because coreclr/corefx/etc changed) because the triggers will be in coreclr/corefx
+- While we are still on benchview, this obfuscates a lot of the upload logic so that product repos don't have to worry about getting it wrong
+- When we move off Benchview, this will allow us to have a seemless experience for product repos
+
+## Cons
+
+- This still could enable the product repos to do their own thing for perf testing (ie by not using our template)
+- Scripting might not be quite general enough to work for anyone
+- The product repos will have to pull down the performance repo to do work
```
billwert: is this a point in time problem? Once we've moved to a place where we don't have to build benchmark.net on the end machine, could we build the repo on the build machine and just push down artifacts (like functional tests?)

billwert: We'd still need to sync the perf repo into the build machine of course.

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+- We will need some mechanism of knowing at what version of the performance repo the product repos are running
```
billwert: I think we can know this because we'd just sync the perf repo at some point so we'd know what HEAD is

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+
+- Product repositories use the provided yaml template to run all performance testing
+- Product repositories do not try to skirt the yaml template to run performance testing only on the tests they care about
+- Setup time for adding performance testing is as simple as adding a new job that references the yaml template that we provide
+
+## Procedural
+
+- Combine the yaml we already use and, likely, the yaml that has been worked on in coreclr (why reinvent the wheel), with any additional logic necessary for pulling down and running the performance repository in product repos
+  - yaml will need to include any additional parameters/knobs that the product repos may want to turn (ie tiered/non-tiered/minopts, pgo/no-pgo, etc)
+  - yaml will need to include all uploading to benchview logic
```
brianrob: Why does the yaml need to contain logic for uploading results?  Would upload of results not take place from the helix machines that run the actual tests?

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+## Success Metric(s)
+
+- Product repositories use the provided yaml template to run all performance testing
+- Product repositories do not try to skirt the yaml template to run performance testing only on the tests they care about
+- Setup time for adding performance testing is as simple as adding a new job that references the yaml template that we provide
+
+## Procedural
+
+- Combine the yaml we already use and, likely, the yaml that has been worked on in coreclr (why reinvent the wheel), with any additional logic necessary for pulling down and running the performance repository in product repos
+  - yaml will need to include any additional parameters/knobs that the product repos may want to turn (ie tiered/non-tiered/minopts, pgo/no-pgo, etc)
```
brianrob: Yes - I like this.  It brings up a question for me though.  How are we going to generalize taking parameters like this and (1) applying them, and (2) making sure that they impact the name/config associated with the results so that it's clear what the config was when looking at results?

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+- We don't have to know anything about the builds of the product repos
+- When new repos want to start running performance testing, we don't have to do anything except point them to the template
+- We do not have to work around darc/dependency flow
+- We do not need to know why the repo triggered a job (ie because coreclr/corefx/etc changed) because the triggers will be in coreclr/corefx
+- While we are still on benchview, this obfuscates a lot of the upload logic so that product repos don't have to worry about getting it wrong
+- When we move off Benchview, this will allow us to have a seemless experience for product repos
+
+## Cons
+
+- This still could enable the product repos to do their own thing for perf testing (ie by not using our template)
```
brianrob: I think this is OK.  We can poke and prod.  They could always do their own thing and just not tell us.

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+- Scripting might not be quite general enough to work for anyone
```
brianrob: If it is not sufficient, what is recommendation?  Would we modify that we have or would we expect folks for clone what we have and modify it as necessary?

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+- The product repos will have to pull down the performance repo to do work
+- We will need some mechanism of knowing at what version of the performance repo the product repos are running
+- Will require us to be very precise in how we build it so that it is truly shareable across repos
+- Will require some input from product repo owners
```
brianrob: Why do we have to be concerned about how the perf repo is built to be shareable?  I'd like to believe that the repo build is self-contained and that the interface between the consuming repo and the perf repo is the set of binaries being passed and the shared yaml.

docs/product-repo-plan-proposal.md
```diff
@@ -0,0 +1,56 @@
...
+- The performance repo does not have to know about the product repos
+- We can use scripting that we already have, modify it to allow a supplied corerun, and almost immediately empower the product repos to start performance testing
+- We don't have to know anything about the builds of the product repos
+- When new repos want to start running performance testing, we don't have to do anything except point them to the template
+- We do not have to work around darc/dependency flow
+- We do not need to know why the repo triggered a job (ie because coreclr/corefx/etc changed) because the triggers will be in coreclr/corefx
+- While we are still on benchview, this obfuscates a lot of the upload logic so that product repos don't have to worry about getting it wrong
+- When we move off Benchview, this will allow us to have a seemless experience for product repos
+
+## Cons
```
brianrob: Are there any downsides with respect to using arcade for the shared yaml such as needing to have a story for making breaking changes to yaml where repos may or may not adopt the new yaml immediately?

docs/product-repo-plan-proposal.md
adiaaida: Right now, upload is not integrated with the Helix API. The Benchview uploader is not compatible with the Helix API (I tried, it's not worth forcing them to be together), so until we replace the system, we still need to pass things like the Upload SAS token on private jobs, logic that we don't want to have to have the product repos know about. The fewer places we have to store that sort of logic, the fewer places we have to update when we do move to the new system, which we do plan on integrating with the helix API for uploading of data.

docs/product-repo-plan-proposal.md
adiaaida: My thought is that we would want to be able to share the yaml between the feature repos AND the performance repo, though they will have to send things slightly differently. In the performance repo, we would send just the benchmarks/sane correlation package for the benchmarks that are going to be run as the work item directory. In the feature repos, they will send the sane correlation package as the work item directory, and core_root as the correlation directory. They will always have to send more than we will.

docs/product-repo-plan-proposal.md
adiaaida: I agree. I feel like if they are shamed enough in tactics for not following the guidance of the perf champs, then they will change their behavior, sort of like how people get shamed for not onboarding to arcade and not moving out of jenkins.

docs/product-repo-plan-proposal.md
adiaaida: In an ideal world, any changes to the yaml would maintain the current default behavior so if repos do not adopt the new yaml immediately, they will not break. The goal with the shared yaml is that there are no truly breaking changes.

This is sort of true with any breaking changes in arcade, where we try to maintain original behavior until everyone has a chance to move to the new behavior (see: moving from phases to jobs, etc).

docs/product-repo-plan-proposal.md
adiaaida: Yes, I meant your second comment with this point. And sort of like - we'll need a way to know where in the perf repo we are. I mentioned to Drew adding an additional piece of metadata to the run info that is the perf repo hash, which we could use to easily know if the perf benchmarks moved underneath us between two runs (but not actually break graphs or anything). He thinks that should be no problem

docs/product-repo-plan-proposal.md
adiaaida: Yeah I don't know what I meant by that.

docs/product-repo-plan-proposal.md
adiaaida: Makes sense. I know that coreclr's yaml is some catastrophe of nonsense. But I'm hoping that if we provide them with a template, they will know their repos well enough to be able to insert a new job.

docs/product-repo-plan-proposal.md
brianrob: Ok, so the way I read that is we need to follow that pattern where we overlap changes that would otherwise be breaking to make sure that they aren't until all consumers take the changes.  That's just the cost of using Arcade, or for that matter, anything that is distributed.

docs/product-repo-plan-proposal.md
brianrob: Fair enough.

docs/product-repo-plan-proposal.md
brianrob: Ok, so should I read this as, "Once we switch off of the Benchview uploader, we would move upload to be within the work that takes place on the helix machines"?

billwert:
@adiaaida does our current plan meet the questions asked in this one? can we close it?

PR #453, opened 2019-05-02
Add Partition filter

adiaaida:
The partition filter will allow us to do things like splitting our helix jobs into multiple work items based on a total number of partitions and a partition index.

### DIFF
--- a/src/benchmarks/micro/CommandLineOptions.cs
+++ b/src/benchmarks/micro/CommandLineOptions.cs
@@ -1,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+
+namespace MicroBenchmarks
+{
+    class CommandLineOptions
+    {
+        // Find and parse given parameter with expected int value, then remove it and its value from the list of arguments to then pass to BenchmarkDotNet
+        // Throws ArgumentException if the parameter does not have a value or that value is not parsable as an int
+        public static List<string> ParseAndRemoveIntParameter(List<string> argsList, string parameter, out int? parameterValue)
+        {
+            int parameterIndex = argsList.IndexOf(parameter);
+            parameterValue = null;
+
+            if (parameterIndex != -1)
+            {
+                if (parameterIndex + 1 < argsList.Count && Int32.TryParse(argsList[parameterIndex+1], out int parsedParameterValue))
+                {
+                    // remove --partition-count args
+                    parameterValue = parsedParameterValue;
+                    argsList.RemoveAt(parameterIndex+1);
+                    argsList.RemoveAt(parameterIndex);
+                }
+                else
+                {
+                    throw new ArgumentException(String.Format("{0} must be followed by an integer", parameter));
+                }
+            }
+
+            return argsList;
+        }
+    }
+}
--- a/src/benchmarks/micro/Program.cs
+++ b/src/benchmarks/micro/Program.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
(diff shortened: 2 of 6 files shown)


adiaaida:
Thanks to @adamsitnik for working up a proof of concept for us.

src/harness/BenchmarkDotNet.Extensions/PartitionFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+
+public class PartitionFilter : IFilter
+{
+    private readonly int? _partitionsCount;
+    private readonly int? _partitionIndex; // indexed from 0
+    private int _counter = 0;
+ 
+    public PartitionFilter()
+    {
+        _partitionsCount = ReadFromEnvironmentVariable("PARTITIONS_COUNT");
```
billwert: I'd like it to take command line parameters instead of relying on the environment. This means a more complex change but I think that's goodness.

src/harness/BenchmarkDotNet.Extensions/PartitionFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+        _partitionIndex = ReadFromEnvironmentVariable("PARTITION_INDEX");
+    }
+ 
+    public bool Predicate(BenchmarkCase benchmarkCase)
+    {
+        if (!_partitionsCount.HasValue || !_partitionIndex.HasValue)
```
billwert: nit: this can just check `partitionsCount` since if you didn't specify that it doesn't matter.
```suggestion
        if (!_partitionsCount.HasValue)
```

adiaaida: But if you didn't specify which partition you want, how do you know which one to run? Do we default to 0?

billwert: Sorry. Half a thought. I'd do that as part of argument validation instead of silently fail here. Specifying one without the other isn't a valid state.

src/harness/BenchmarkDotNet.Extensions/PartitionFilter.cs
adiaaida: Done.

src/harness/BenchmarkDotNet.Extensions/PartitionFilter.cs
adiaaida: done.

src/benchmarks/micro/CommandLineOptions.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+
+namespace MicroBenchmarks
+{
+    class CommandLineOptions
```
adamsitnik: could you please move all the parsing logic to `BenchmarkDotNet.Extensions` project? the goal is to keep all the utilities in this helper library while having only benchmarks in `benchmarks/*`

Also if possible it would be great to add a simple test to `src/tests/harness/BenchmarkDotNet.Extensions.Tests`

src/benchmarks/micro/Program.cs
```diff
@@ -12,11 +14,35 @@ namespace MicroBenchmarks
     class Program
     {
         static int Main(string[] args)
-            => BenchmarkSwitcher
+        {
+            var argsList = new List<string>(args);
```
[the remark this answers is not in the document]
adiaaida: I was having a hard time figuring out how to remove the args from the list of arguments we pass through to BDN, and this was what I came up with. That's why it's all outside of the partition filter itself.

src/benchmarks/micro/CommandLineOptions.cs
adiaaida: Done

PR #461, opened 2019-05-07
Use the partition filter to split the Microbenchmarks legs into multiple Helix Work Items

adiaaida:
This change uses the PartitionFilter to split the microbenchmark legs into multiple HelixWorkItems. For right now, I am just testing this to see how it works. We will need to make sure it still works with Benchview upload before checking it in.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -48,14 +48,28 @@ jobs:
       kind: micro
       architecture: x64
       pool: Hosted VS2017
-      queue: Windows.10.Amd64.ClientRS4.DevEx.15.8.Open
+      queue: Windows.10.Amd64.ClientRS4.Open
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
-      categories: ['coreclr', 'corefx']
+      runCategories: 'coreclr corefx' 
       frameworks: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
         - netcoreapp3.0
         - netcoreapp2.2
         - netcoreapp2.1
-        - net461
+
+ # Windows x64 micro benchmarks net461, public correctness job
+- ${{ if eq(variables['System.TeamProject'], 'public') }}:
+  - template: /eng/performance/benchmark_jobs.yml
+    parameters:
+      osName: windows
+      osVersion: RS4
+      kind: micro_net461
+      architecture: x64
+      pool: Hosted VS2017
+      queue: Windows.10.Amd64.ClientRS4.DevEx.15.8.Open
+      csproj: src\benchmarks\micro\MicroBenchmarks.csproj
+      runCategories: 'coreclr corefx'
+      frameworks: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
+        - net461 
 
 # Windows x86 micro benchmarks, public correctness job
 - ${{ if eq(variables['System.TeamProject'], 'public') }}:
@@ -66,9 +80,9 @@ jobs:
       kind: micro
       architecture: x86
       pool: Hosted VS2017
-      queue: Windows.10.Amd64.ClientRS4.DevEx.15.8.Open
+      queue: Windows.10.Amd64.ClientRS4.Open
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
-      categories: ['coreclr', 'corefx']
+      runCategories: 'coreclr corefx'
       frameworks: # for public jobs we want to make sure that the PRs don't break x86
         - netcoreapp3.0
 
@@ -83,7 +97,8 @@ jobs:
       pool: Hosted VS2017
       queue: Windows.10.Amd64.ClientRS1.Perf # using a dedicated private Helix queue (perfsnakes)
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
-      categories: ['coreclr', 'corefx']
+      runCategories: 'coreclr corefx'
+      benchviewCategory: 'coreclr'
       frameworks: # for private jobs we want to benchmark .NET Core 3.0 only
         - netcoreapp3.0
       
@@ -98,7 +113,8 @@ jobs:
       pool: Hosted VS2017
       queue: Windows.10.Amd64.ClientRS1.Perf # using a dedicated private Helix queue (perfsnakes)
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
-      categories: ['coreclr', 'corefx']
+      runCategories: 'coreclr corefx'
+      benchviewCategory: 'coreclr'
       frameworks: # for private jobs we want to benchmark .NET Core 3.0 only
         - netcoreapp3.0
         
@@ -114,7 +130,7 @@ jobs:
       queue: Ubuntu.1604.Amd64.Open
       container: ubuntu_x64_build_container
       csproj: src/benchmarks/micro/MicroBenchmarks.csproj
-      categories: ['coreclr', 'corefx']
+      runCategories: 'coreclr corefx'
       frameworks: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
         - netcoreapp3.0
         - netcoreapp2.2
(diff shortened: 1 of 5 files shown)


adiaaida:
@jorive What happens in benchview right now if we get uploads from the same commit hash/benchview name/etc, but with different benchmarks. For example, leg A uploads results for benchmark foo at hash 1, and leg B uploads results for benchmark bar at hash 1. Will their results merge together in the database? I imagine this could muck up the places where we merge results (geomean levels).

adiaaida:
@billwert I think I got this working. We should talk about the following things:

1. How many partitions do we want? I did 5 because that was easy to get working. We have 10 windows and 10 linux machines at the moment, but obviously, we will have more in the future.
2. Do we want to partition only official builds, or both official and PR runs (ie, functional runs)? I was thinking just official runs because we have SO MANY legs, which means SO MANY VMs in functional, but I can be persuaded either way.
3. Are there any other legs we should consider splitting (this only splits the coreclr/corefx runs, not the ML or other csprojs that we run)?

adiaaida:
Some data about the most recent run, compared to a non-partitioned PR run:

**Partitioned**
* The slowest run takes about 8 minutes per partition (when looking at the time actually spent running the partition)
* The legs are still taking up to 54 minutes due to waiting for machines (the scale set controls how many VMs from the queue we can spin up at a time). We are waiting for up to 42 minutes for a machine
* Due to the time waiting and the number of legs we run, we experience diminishing returns in partitioning PRs

**Non-Partitioned**
* Slowest run takes 28 minutes
* Slowest leg take 51 minutes, so we are waiting for up around 23 minutes for a machine.

I'm going to run this PR a few more times to see if maybe I just hit a bad time for the queue or something, but if the average is that partitioned prs take as long as non-partitioned prs, then partitioning doesn't make sense for PRs. For official builds, where we run for 4+ hours and only run 2 or so legs, then we should definitely be partitioning. Figuring out the partition count will require some telemetry.

adamsitnik:
One thing that we can reconsider is filtering by categories which were introduced to split the benchmarks into partitions. 
Since we have the partitioner we could stop filtering by categories and let the partitioner do the job. But still keep them in the code to allow for CoreCLR/FX jobs from the repos.

It would reduce the number of jobs by half and would ensure that we don't run any benchmarks twice (5-10% of benchmarks belong to both CoreFX and CoreCLR and we run them twice).

```diff
  - name: BenchmarkDotNetArguments
-    value: '--allCategories $(_Category)'
+    value: '--filter * --partitionIndex $(_PartitionIndex) --partitionsCount $(PartitionsCount)'
```

billwert:
> Since we have the partitioner we could stop filtering by categories and let the partitioner do the job. But still keep them in the code to allow for CoreCLR/FX jobs from the repos.

This is a good idea. (Though don't assume we will like to see CoreCLR not running CoreFX tests and vice versa. One of the useful applications of centralized tests is that we can run them all everywhere.)

adiaaida:
@adamsitnik That would mean completely removing the category parameter, correct?

adamsitnik:
> That would mean completely removing the category parameter, correct?

Yes. The partitioner combined would become the only filter.

adiaaida:
Update on timing for PR runs with 5 partitions:

|         | With Partitioning | Without Partitioning |
| ---- | ----------------- | -------------------- |
| windows average | 32 min | 35 min |
| windows max | 43 min | 44 min |
| ubuntu average | 9 min | 30 min |
| ubuntu max | 9 min | 37 min |

The Ubuntu legs are obviously taking better advantage of the partitioning than the windows legs because they use a queue with a larger scale set (up to 300 vms, rather than 60). I've tried moving the windows legs to the larger windows queue, but I was getting weird python urllib errors, so I moved them back off. I think in another PR, I will try to switch to that other queue and work through why we are seeing the urllib failures there (maybe some firewall nonsense?), but for now, I think this is about as good as we are going to get.

azure-pipelines.yml
```diff
@@ -48,14 +48,28 @@ jobs:
...
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
-      categories: ['coreclr', 'corefx']
+      runCategories: 'coreclr corefx' 
       frameworks: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
         - netcoreapp3.0
         - netcoreapp2.2
         - netcoreapp2.1
-        - net461
+
+ # Windows x64 micro benchmarks net461, public correctness job
```
billwert: what's the reasoning for splitting this out?

eng/performance/benchmark_jobs.yml
```diff
@@ -33,7 +34,7 @@ jobs:
             value: dotnet-performance
           # for public runs we want to run the benchmarks exactly once, no warmup, no pilot, no overhead
           - name: BenchmarkDotNetArguments
-            value: '--allCategories $(_Category) --iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart --stopOnFirstError true'
+            value: '--anyCategories ${{ parameters.runCategories }} --iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart --stopOnFirstError true'
```
adamsitnik: I think that it can be simplified and we can filter by providing `--filter *` instead of `--anyCategories`

So far we have 3 projects with BDN benchmarks:

- Roslyn with only Roslyn benchmarks
- ML.NET with only ML.NET benchmarks
- MicroBenchmarks with CoreFX, CoreCLR and ThirdParty benchmarks. The benchmarks from `ThirdParty` category are 40-60 benchmarks for the most popular 3rd party serializers that we have excluded before to save some time when running the benchmarks as part of the CI. But now when we have the possibility to parition them and send to multiple machines and imho we don't need to do this anymore and we can run all of them.

```suggestion
            value: "--filter '*' --iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart --stopOnFirstError true"
```

The asterisk needs to be wrapped with single quote characters on Unix because otherwise the OS replaces it with the list of files in current folder.

eng/performance/helix.proj
```diff
@@ -18,10 +18,32 @@
...
+    <Partition Include="Partition0" Index="0" />
+    <Partition Include="Partition1" Index="1" />
+    <Partition Include="Partition2" Index="2" />
+    <Partition Include="Partition3" Index="3" />
+    <Partition Include="Partition4" Index="4" />
+  </ItemGroup>
+
+  <!-- 
+    Partition the Microbenchamrks project, but nothing else
+    TODO: only for oficial builds
```
adamsitnik: Maybe this TODO could be solved by adding the partitions Ids and count to the `benchmark_jobs.yml`?

By doing so we could define:
- defaults used my ML.NET and Roslyn that don't require partitioning
- different values for official and PR buils


Sth like:

```yml
// part of eng/performance/benchmark_jobs.yml 
  partitionCount: 1     # optional -- the number of partitions
  frameworks: [ '0' ]    # optional -- partition ids, indexed from 0

// part of  azure-pipelines.yml for PR runs
  partitionCount: 2
  frameworks: [ '0', '1' ]

// part of  azure-pipelines.yml for official runs
  partitionCount: 5
  frameworks: [ '0', '1', '2', '3', '4' ]
```

azure-pipelines.yml
adiaaida: net461 needs to be on a VM with VS installed on it, which means it needs to be on a DevEx queue. I'm hoping to be able to move the netcoreapp legs to the Windows.10.Amd64.Open queue, so I split net461 out in preparation for that.

eng/performance/benchmark_jobs.yml
adiaaida: We tried that. There are many benchmarks in the microbenchmark project that are not triggered when you specify coreclr or corefx that are triggered by filter *. The serializer benchmarks are broken, which causes everything to break. They need to be fixed before we can include them. Is there another way to disable benchmarks while using filter *?

eng/performance/helix.proj
adiaaida: I actually decided that I'd just partition PRs as well, since it does have the possibility of saving us time. I will remove the TODO.

adiaaida:
@billwert The benchviewCategory parameter is used to tell benchview which azure container to upload the results to. We can only upload to containers that exist, so I've passed it as a separate parameter. Would it make more sense if I renamed the parameter to benchviewContainer? This parameter will go away when we start uploading to the Kusto database using the Helix API.

billwert:
> This parameter will go away when we start uploading to the Kusto database using the Helix API.

Sounds great. I just wanted to understand the semantic. Thanks!

azure-pipelines.yml
billwert: cool. We may be able to take advantage of the new framework reference package but that can be another time.

adiaaida:
Opened #473 for the serializer benchmarks.

PR #464, opened 2019-05-08
Stop supporting netcoreapp2.0

adiaaida:
2.0 has gone out of support for .Net Core, so we should stop supporting
it in our performance testing framework. In particular, when you pull
down dotnet with -Channel 2.0, you are actually pulling down a dotnet at
version 2.1.200. There is an issue with this version where, if you run
on a machine where another run has executed nuget in the last 30 minutes,
you will get various errors of the "error NU3004: The package is not
signed" variety, which is causing a large number of our builds to fail.
Additionally, since netcoreapp2.0 is officially no longer supported, we
are not even running functional testing on the release/2.0 branches out
of coreclr or corefx. Therefore, we should not be running functional
testing of our performance tests for the netcoreapp2.0 bits.

This change removes netcoreapp2.0 support from the performance repo and
updates docs to reflect this.


adiaaida:
FYI: @adamsitnik

jorive:
I do not recall if there are build MACROS across the source code and csproj files.

adiaaida:
I just realized I missed two things in the microbenchmarks csproj.

There are few other mentions of netcoreapp2.0 in CscBenchmark and JitBench. I didn't touch those because I didn't want to change the tests. In JitBench's case, I don't even know if that runs, so I left it alone. Also left in the dmlib benchmark and docker-harness, which are benchmarks that I don't even know if we care about running anymore.

jorive:
We should also remove the usage of `NETCOREAPP2_0`

adiaaida:
Thanks! I'll search for that.

PR #468, opened 2019-05-10
fix up status page

billwert:
It looks like the format of the status links changed. I dont't see a
likely cause for this in our repo so perhaps AzDO changed something.


billwert:
@adiaaida I'm not just merging this because I'm not 100% sure this wasn't due to some other change you made. (I looked but nothing jumped out.) If you don't anticipate any of your other changes breaking this soon go ahead and merge it.

adiaaida:
My changes will definitely break this because I'm changing the configuration names (no more corefx vs coreclr). I'll try to work that into my partitioning change.

PR #488, opened 2019-05-21
enforce Server GC for ArrayPool benchmarks

adamsitnik:


src/benchmarks/micro/corefx/System.Buffers/ArrayPoolTests.cs
```diff
@@ -152,6 +153,7 @@ public async Task ProducerConsumer()
     [BenchmarkCategory(Categories.CoreFX)]
     [GenericTypeArguments(typeof(byte))] // value type
     [GenericTypeArguments(typeof(object))] // reference type
+    [GcServer(true)] // Server GC is the most important scenario for ArrayPool
```
stephentoub: Is doing this per test suite the right thing to do?  Or would it be better to do GcServer(false) runs for all benchmarks and GcServer(true) runs for all benchmarks?

I imagine that many of our benchmarks are allocation-heavy and would be influenced by client vs server GC.

billwert: I would hope the vast majority of our benchmarks do not allocate, or allocate very little. I certainly don't want to duplicate the matrix on this pivot without strong evidence it makes a big difference and is useful. My guess is that doing this experiment to highlight any unexpected surprises once and track those as individual work items is the better strategy.

adamsitnik: We have very few allocation-heavy benchmarks. I think that what we do now (don't enforce Server GC) is the best thing.

stephentoub:
> We have very few allocation-heavy benchmarks

I would have guessed that, outside of parsing and span formatting, most of our benchmarks allocate. You're saying that the vast majority of our benchmarks incur 0 allocations?

billwert:
> I would have guessed that, outside of parsing and span formatting, most of our benchmarks allocate.

Good question. I was curious so I did some analysis on a result set I ran a couple days ago.

Right now it breaks down like this

- 2388 total benchmarks
- 1035 allocate anything at all
- 793 have a Gen 0 collection
- 244 have a Gen 1 collection
- 23 have a Gen 2 collection

(Run on a machine with 64GB of RAM.)


Looking at percentiles:

| bucket | bytes / op          |
|-----|-------------|
| 0.1 | 40          |
| 0.2 | 64          |
| 0.3 | 112         |
| 0.4 | 168         |
| 0.5 | 504         |
| 0.6 | 2072        |
| 0.7 | 7678.4      |
| 0.8 | 18313.6     |
| 0.9 | 161641.6    |
| 1   | 13112000048 |

Same, but for Gen 0 collections. Note that this isn't scaled per operation, and is the total over all operations. Scaling per operation puts the 90th percentile at 0.03 GC/op.

| bucket | Gen 0 / iteration |
|--------|-------|
| 0.1    | 0     |
| 0.2    | 0     |
| 0.3    | 3     |
| 0.4    | 7     |
| 0.5    | 11    |
| 0.6    | 22    |
| 0.7    | 45    |
| 0.8    | 79.4  |
| 0.9    | 165   |
| 1      | 4166  |

I'm surprised we get up to allocating 500b/operation by the 50th percentile. However, I don't think there's really evidence supporting doubling the cost of the run (which currently is about 5 hours.) It's possible we'd get meaningfully different results out of the top allocators (which are things like ByteMark, BenchmarksGame, Burgers, etc.) 

Thoughts?

billwert:
Regarding the actual change, it looks like there's nothing that will nicely flow into the reporting system. Ideally we would have a good flag when some benchmark wasn't "normally configured" for the run. The only place this is flagged is in the description of the Job, which doesn't get reported anywhere.

In general my philosophy around tests like this in labs is to guard against future code changes impacting them. One could argue that this change is ensuring that in the "normal" config we expect, server GC, it isn't impacted by server GC. In opposition, I would argue that we should have specific benchmarks targeting the GC, and that we'd also capture such things through server perf runs and the like. It's also true that this is a regression suite, not an absolute number suite - this data isn't intended to be used in any absolute way, so whether or not the GC is server or workstation shouldn't matter as long as we have a consistent trend.

I'm torn on this one. I think this is a change, and precedent, I would rather not set.

adamsitnik:
> You're saying that the vast majority of our benchmarks incur 0 allocations?

By "allocation-heavy benchmarks" I meant benchmarks that allocate a lot of memory and/or the ones that do Gen 1/2 collects. Not the ones that allocate at all.

adamsitnik:
@billwert thank you for the data. It's much better to have this conversation backed with the numbers.

We all agree that both Server and Workstation GC performance is important. We also know that we don't have unlimited CPU power.

Perhaps we should add one more CI leg for the Server GC benchmark runs? But make sure that it is triggered only once a day? (today the CI is triggered once a day + when a PR gets merged to the perf repo). I can extend the BDN json file to include this information.

billwert:
> Perhaps we should add one more CI leg for the Server GC benchmark runs?

I'm really not convinced that's necessary. I have seen very few instances where a given test showed a regression in only one GC configuration, and it wasn't caught by some other mechanism. (GC testing, ASP.net testing, etc.) The data shows that ~10% of tests do a Gen 1, and ~1% do a Gen 2.

Let me put the question differently: What do we base an expectation that adding the run has value on? What do we think the likelihood it catches an actionable bug is?

VSadov:
Assuming that most benchmarks are throughput oriented, have modest allocations and are not multithreaded, workstation GC would be an appropriate default. It could also result in more stable measurements since there are fewer things that may work differently from run to run.
I would probably disable background marking as well, - just to make results more deterministic.

Where ServerGC is more interesting is where we have concurrent allocations. Once in a while an allocation may need an exclusive access to the heap and would take a lock. In server GC these locks are nearly never contested since there is a heap per core, but on a workstation GC there is a single shared heap and locking may lead to random spikes and outliers. Basically server GC may have more stable results when there are concurrent allocations.

In this particular case we do have allocations and do have concurrent scenarios, so ServerGC makes sense just from reducing the noise point of view.

VSadov:
Adding a whole dimension of server/workstation runs seems a bit of an overkill. Many tests would not care about the differences and for those that do care the difference in results would be expected and thus not very actionable.

src/benchmarks/micro/corefx/System.Buffers/ArrayPoolTests.cs
```diff
@@ -13,6 +13,7 @@ namespace System.Buffers.Tests
...
```
VSadov: Did you confirm this made it switch to server GC? 
Somehow a similar switch did not work for me, but it is possible I did something else wrong. Just want to confirm this works when used properly.

stephentoub:
My main concern is that the primary environment we have where the kinds of benchmarks we have in this repo matter is ASP.NET, and ASP.NET defaults to using server GC.  Seems like if we're going to measure one environment configuration, it should be the one the main performance-focused app model uses as a default.

stephentoub:
(We also have other benchmarks that do concurrent allocations.  If we're going to only selectively apply server GC to some tests, let's at least be methodical about which ones we apply it to.)

VSadov:
Using `Task.Run(...)` in a benchmark might be a strong hint for ServerGC. Also `new Thread(...)`

billwert:
@VSadov do you have local runs that show noise reductions in this benchmark thanks to server GC? Looking at the results for the last week this test has been in, I do not see any evidence of significant GC activity. It's true that currently the bytes allocated metric isn't perfectly accurate on tests that allocate on many threads, but the count of GCs should be. Almost none of the tests ever triggers a Gen 0 GC, which suggests that the allocation rates are low enough that any noise from stalls due to locks in the GC isn't happening.

VSadov:
@billwert - I could not convince the test to run with Server GC. Somehow it seems to ignore the attribute and still runs with workstation GC. For the scenarios that do not allocate it would not be a lot of difference, obviously. 

Some more background - my change reduces the cost of array allocations when it is ok for the arrays to not be zeroed out. ArrayPool does not have such contract, so when it does need to allocate an array, it could use such API  (`GC.AllocateUninitializedArray`). 
As a result I am primarily interested in scenarios that do allocate. - to see if there are benefits to have from the API. 
There are definitely benefits in the Consumer/Producer scenario. I assume it allocates or we would not see any difference.

Overall - It is not guaranteed that server GC will result in less noisy results, but there are reasons to expect that. 
Also, since we have scenarios that specifically rent/use/return concurrently, it feels like server GC would be closer approximation of the environment where the feature will be used.

adamsitnik:
As discussed offline with @billwert I am going to close this issue. The plan is to add Server GC runs for the entire portfolio of benchmarks, run them for a few weeks and validate the usefulness after that.

PR #490, opened 2019-05-21
Enforce IDE to display source files excluded from compilation

adamsitnik:
@ahsonkhan reported that when he opens the MicroBenchmarks.sln in VS, all the source files excluded conditionally from the compilation are not visible. Example: the new JSON serializer benchmarks.

I checked that and it's true for both VS 2017 and VS 2019. Editing these files is hard: it requires to go to the `.csproj` file and removing the `<Compile Remove` rules. As soon as we do that, the files appear in IDE, but we get compilation errors for the TFM that was excluding them.

I found a fix/workaround for that: Combine `Compile Remove` with `Content Include` which basically tells the IDE to not compile the files, but include them in the solution explorer as content.

```xml
    <Compile Remove="**\*netcoreapp*.cs" />
    <Content Include="**\*netcoreapp*.cs" /> <!-- makes VS display the exluded files in solution explorer -->
````

I did not want to duplicate this for every file we exclude, so I introduced a new pattern for conditional compilation based on file extension:

- all .NET Core only files ends with `.netcoreapp.cs` instead of `.cs`
- all .NET Core 3.0 only files ends with `.netcoreapp30.cs` instead of `.cs`

It works for both VS  2017 and 2019:

VS 2017 before:

obraz

VS 2017 after:

obraz

VS 2019 before:

obraz

VS 2019 after:

obraz

@eerhardt @safern is there a cleaner way to do that?

### CODE REVIEW

**Comment by eerhardt** [comment]
In ML.NET, we do:

https://github.com/dotnet/machinelearning/blob/ce46faacd6af7c5e744b1157ec9d707e7c9d6d58/src/Microsoft.ML.CpuMath/Microsoft.ML.CpuMath.csproj#L17-L18

```xml
    <!-- Workaround https://github.com/dotnet/project-system/issues/935 -->
    <None Include="**/*.cs" />
```

linking to https://github.com/dotnet/project-system/issues/935 and /cc @davkean.

billwert:
I wonder if this got fixed in VS2019.. It doesn't seem to be empirically.. @davkean?

billwert:
My biggest qualm with this is at some point we will stop running runtimes that didn't support these APIs. We're going to be left with a weird history of `netcore30.cs`, `.netcore31.cs`, and so on. I like @eehardt's suggestion a lot, if we can't figure out when the IDE will just do the right thing.

ahsonkhan:
> My biggest qualm with this is at some point we will stop running runtimes that didn't support these APIs. 

I agree. I don't think we should rename files for new APIs especially since they will continue to be what makes up "netcoreapp". If anything, we may want non-netcoreapp APIs to be renamed, especially if we are only running/tracking perf tests for .NET Core. I understand that has its own down-sides too, so it isn't ideal.

adamsitnik:
I totally agree the solution provided by @eerhardt is way cleaner.

I reverted my changes and used it.

Thanks @eerhardt !

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -68,10 +64,7 @@
     <Compile Remove="coreclr\Math\Functions\Double\Asinh.cs" />
     <Compile Remove="coreclr\Math\Functions\Double\Atanh.cs" />
     <Compile Remove="coreclr\Math\Functions\Double\Cbrt.cs" />
-    <Compile Remove="coreclr\Math\Functions\Single\Acosh.cs" />
-    <Compile Remove="coreclr\Math\Functions\Single\Asinh.cs" />
-    <Compile Remove="coreclr\Math\Functions\Single\Atanh.cs" />
-    <Compile Remove="coreclr\Math\Functions\Single\Cbrt.cs" />
+    <Compile Remove="coreclr\Math\Functions\Single\**" />
```
ahsonkhan: We have this here but then again in the other ItemGroup at Line 76-78 below under:
```csproj
<ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0')">
```
How come?

Also, we use wildcards for Single but not for Double, why?

ahsonkhan: Also, are we use we want to remove all of them for Single as well? I would have thought most of them are fine to keep (and only the 4 listed previously need to be excluded):
https://github.com/dotnet/performance/tree/master/src/benchmarks/micro/coreclr/Math/Functions/Single

adamsitnik: The Single API does not exist for Full .NET Framework. ILogB, Log2, ScaleB use APIs added in .NET Core 3.0.

So we exclude all Single tests for Full Framework and those which use .NET Core 3.0 API for non .NET Core 3.0

eerhardt: Why not `**\*.cs` at the end? If someone adds a .txt, .md, or .resx file to the folder, it will cause compiler errors.

adamsitnik: Good point, fixed

eerhardt: Oh - at first look I missed these were `Remove` elements, so I guess it didn't matter much (since the other file types shouldn't have been in `Compile` to start with). But this is clearer IMO - thanks!

billwert:
This passed all the checks - I think it's just stuck due to the webhook problem.

PR #493, opened 2019-05-21
Work around a perf issue in a StringBuilder benchmark

kouvel:
Related to https://github.com/dotnet/coreclr/issues/23060
- Worked around a perf issue in a `StringBuilder` benchmark when `QuickJitForLoops` is enabled

### DIFF
--- a/src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
+++ b/src/benchmarks/micro/corefx/System.Runtime/Perf.StringBuilder.cs
@@ -139,7 +139,7 @@ namespace System.Tests
         public string StringBuilderToString() => _bigStringBuilder.ToString();
 
         [Benchmark]
-        public StringBuilder AppendMemory()
+        public int AppendMemory()
         {
             ReadOnlyMemory<char> memory = _string100.AsMemory();
             StringBuilder builder = new StringBuilder();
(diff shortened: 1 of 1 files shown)


kouvel:
@adamsitnik

adamsitnik:
Explanation can be found here: https://github.com/dotnet/coreclr/issues/23060#issuecomment-494565895

billwert:
does this impact the measurement at all?

adamsitnik:
> does this impact the measurement at all?

Not anymore. Now, when `QuickJitForLoops` is disabled by default (change introduced in some of the recent .NET Core previews) it does not affect the measurements (I have checked that).

adamsitnik:
BTW diabling QuickJitForLoops by default is reflected on the BenchView charts for this benchmark:

obraz

billwert:
Sorry I actually just meant the return type change. :)

adamsitnik:
> just meant the return type change

This should also not affect the results. BDN always generates a method that has the same arguments and return type to measure the overhead. So the `Result = Actual - Overhead` should return same value for StringBuilder and int.

The only case when it can affect the result is when JIT can optimize the method to return a const. An example:

```cs
[Benchmark]
public int AllocateArray() => new byte[100].Length;
```

If JIT one day becomes smart enough to replace it to "return 100" then the benchmark result is going to change.

PR #499, opened 2019-05-23
New reporting library

billwert:
This library enables writing results to the new database.

This commit also modifies the BenchmarkDotNet.Extensions library to additonally
write out the new format.

### DIFF
--- a/src/harness/BenchmarkDotNet.Extensions/PerfLabExporter.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/PerfLabExporter.cs
@@ -1,0 +1,98 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Exporters;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Parameters;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using Reporting;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+
+namespace BenchmarkDotNet.Extensions
+{
+    internal class PerfLabExporter : ExporterBase
+    {
+        protected override string FileExtension => "json";
+        protected override string FileCaption => "perf-lab-report";
+        public PerfLabExporter()
+        {
+        }
+        public override void ExportToLog(Summary summary, ILogger logger)
+        {
+            var reporter = Reporter.CreateReporter();
+            if (reporter == null) // not running in the perf lab
+                return;
+
+            var nameDelegate = typeof(BenchmarkCase).Assembly.GetType("BenchmarkDotNet.Exporters.FullNameProvider").GetMethod("GetBenchmarkName", BindingFlags.Static | BindingFlags.NonPublic);
+
+
+            foreach (var report in summary.Reports)
+            {
+                var test = new Test();
+                //test.Name = FullNameProvider.GetBenchmarkName(report.BenchmarkCase);
+                test.Name = (string)nameDelegate.Invoke(null, new[] { report.BenchmarkCase });
+                test.Categories = report.BenchmarkCase.Descriptor.Categories;
+                var results = from result in report.AllMeasurements
+                              where result.IterationMode == Engines.IterationMode.Workload && result.IterationStage == Engines.IterationStage.Result
+                              orderby result.LaunchIndex, result.IterationIndex
+                              select new { result.Nanoseconds, result.Operations };
+                test.Counters.Add(new Counter
+                {
+                    Name = "Duration of single invocation",
+                    TopCounter = true,
+                    DefaultCounter = true,
+                    HigherIsBetter = false,
+                    MetricName = "ns",
+                    Results = (from result in results
+                               select result.Nanoseconds / result.Operations).ToList()
+                });
+                test.Counters.Add(new Counter
+                {
+                    Name = "Duration",
+                    TopCounter = true,
+                    DefaultCounter = false,
+                    HigherIsBetter = false,
+                    MetricName = "ms",
+                    Results = (from result in results
+                               select result.Nanoseconds).ToList()
+                });
+
+                test.Counters.Add(new Counter
+                {
+                    Name = "Operations",
+                    TopCounter = false,
+                    DefaultCounter = false,
+                    HigherIsBetter = true,
+                    MetricName = "Count",
+                    Results = (from result in results
+                               select  (double)result.Operations).ToList()
+                });
+
+                foreach (var metric in report.Metrics.Keys)
+                {
+                    var m = report.Metrics[metric];
+                    test.Counters.Add(new Counter
+                    {
+                        Name = m.Descriptor.DisplayName,
+                        TopCounter = false,
+                        DefaultCounter = false,
+                        HigherIsBetter = m.Descriptor.TheGreaterTheBetter,
+                        MetricName = m.Descriptor.Unit,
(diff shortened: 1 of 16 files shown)


src/harness/BenchmarkDotNet.Extensions/PerfLabExporter.cs
```diff
@@ -0,0 +1,98 @@
...
+        public PerfLabExporter()
+        {
+        }
+        public override void ExportToLog(Summary summary, ILogger logger)
+        {
+            var reporter = Reporter.CreateReporter();
+            if (reporter == null) // not running in the perf lab
+                return;
+
+            var nameDelegate = typeof(BenchmarkCase).Assembly.GetType("BenchmarkDotNet.Exporters.FullNameProvider").GetMethod("GetBenchmarkName", BindingFlags.Static | BindingFlags.NonPublic);
```
billwert: Once we update BDN this type will be public and this can be changed.

src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            build = new Build
+            {
+                Repo = environment.GetEnvironmentVariable("PERFLAB_REPO"),
+                Branch = environment.GetEnvironmentVariable("PEFRFLAB_BRANCH"),
+                Architecture = environment.GetEnvironmentVariable("PERFLAB_BUILDARCH"),
+                Locale = environment.GetEnvironmentVariable("PERFLAB_LOCALE"),
+                GitHash = environment.GetEnvironmentVariable("PERFLAB_HASH"),
+                BuildName = environment.GetEnvironmentVariable("PERFLAB_BUILDNUM"),
+                TimeStamp = DateTime.Parse(environment.GetEnvironmentVariable("PERFLAB_BUILDTIMESTAMP")),
```
adiaaida: I feel like this would be easier if we figured out the timestamp in-code, if we are going to use git to tell us the timestamp.

src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            var jsonobj = new
+            {
+                build,
+                os,
+                run,
+                tests
+            };
+            return JsonConvert.SerializeObject(jsonobj,
+                                               Formatting.Indented,
```
adiaaida: nit: The formatting of this looks a little weird to me

src/tools/Reporting/Reporting/Reporter.cs
billwert: We will - I'm going to put it in benchmarks_ci.py for now, and in the future when we have the build information on the helix build machine we will figure it out there. I don't want the reporting library calling out to git itself.

jorive: Take a look at `build.py` from the BenchView scripts. It has the logic there.

adiaaida: It also turns out that we don’t have git info on the helix machines

src/harness/BenchmarkDotNet.Extensions/PerfLabExporter.cs
```diff
@@ -0,0 +1,98 @@
...
+                    TopCounter = true,
+                    DefaultCounter = true,
+                    HigherIsBetter = false,
+                    MetricName = "ns",
+                    Results = (from result in results
+                               select result.Nanoseconds / result.Operations).ToList()
+                });
+                test.Counters.Add(new Counter
+                {
+                    Name = "Duration",
```
adamsitnik: We used to have this metric in BenchView, when we were using the `IterationTime` as the result because xunit-performance did not scale the results (`Scaled=IterationTime/InnerInvocationTime`) But BDN scales the results and reports the mentioned above "Duration of a single invocation" which is the scaled value.

More: https://github.com/dotnet/performance/blob/master/docs/porting-to-benchmarkdotnet.md#scaling-of-the-results

With BDN we can easily change the `IterationTime` and it's going to use the provided value and scale the results:

https://github.com/dotnet/performance/blob/07b782ef1efb2acad301c6ec74f24c2e30ce802b/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L18

For example: in the future, we could change it to 200ms to save 1/5 of the execution time (if we make sure it does not affect the results). If we ever do that, the "Duration" will change but "Duration of single invocation" will not. 

We should consider supporting only "Duration of single invocation" in the new reporting system. Or even choose a better name ;)

src/tools/Reporting/Reporting/Counter.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public bool TopCounter { get; set; }
+
+        public bool DefaultCounter { get; set; }
+
+        public bool HigherIsBetter { get; set; }
+
+        public string MetricName { get; set; }
+
+        public IList<double> Results { get; set; }
```
adamsitnik: nit: maybe `IReadOnlyList` to make sure no new elements are added?

src/tools/Reporting/Reporting/Os.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Reporting
+{
+    public class Os
```
adamsitnik: nit: all other types in this project have full names, so maybe a better name would be `OperatingSystem`?

src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -0,0 +1,111 @@
...
+        private Os os;
+        private Build build;
+        private List<Test> tests = new List<Test>();
+        protected IEnvironment environment;
+
+        private Reporter() { }
+
+        public void AddTest(Test test)
+        {
+            if (tests.Any(t => t.Name.Equals(test.Name)))
```
adamsitnik: nit: consider using `Dictionary<string, Test>` instead

src/tools/Reporting/Reporting/Counter.cs
```diff
@@ -0,0 +1,23 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Reporting
+{
+    public class Counter
+    {
+        public string Name { get; set; }
```
adamsitnik: nit: if `Build` is sealed, should this type be too?

src/harness/BenchmarkDotNet.Extensions/PerfLabExporter.cs
billwert: Thanks for the info! I chose to report everything that was being reported for continuity. "Duration of a single invocation" is the "default" and will be what people see first. Other counters like this are for additional context. We can always turn one off if it turns out to not be worth it.

src/tools/Reporting/Reporting/Counter.cs
billwert: Heh, I didn't actually intend to seal build. I'll make a note to clean that up at some point.

src/tools/Reporting/Reporting/Counter.cs
billwert: BDN happens to know all the results at once but you could imagine another tool which wants to add results as it parses them out (a linear parser for ETLs, maybe.)

src/tools/Reporting/Reporting/Os.cs
billwert: Probably. I worked backward from what we typed in the json (which I just typed out by hand as it's the contract between parts of the system.) I think it isn't worth changing now (since it would change in a bunch of places) but I will keep the feedback in mind for future!

PR #500, opened 2019-05-24
Make ArtifactsUploader more generic

billwert:
Since we're not using the benchview tools to prepare results uploads it makes sense to move away from that suite entirely. This uploader was very nice but needed a few things so it could be used more generally.

Changes:

Renamed several args, made some optional

Added --skipCompression to just upload loose files

Upgrade to 3.0

fix tests

Remove unused argument isPR

### DIFF
--- a/src/tools/ArtifactsUploader/ArtifactsUploader/Properties/launchSettings.json
+++ b/src/tools/ArtifactsUploader/ArtifactsUploader/Properties/launchSettings.json
@@ -1,8 +1,0 @@
-{
-  "profiles": {
-    "ArtifactsUploader": {
-      "commandName": "Project",
-      "commandLineArgs": "--container results --branch csfiles --artifacts . --searchPatterns *cs --storageUrl https://pvscmdupload.blob.core.windows.net --token \"?sv=2018-03-28&ss=bf&srt=sco&sp=rwdlac&se=2019-05-24T07:15:29Z&st=2019-05-23T23:15:29Z&spr=https&sig=IjS1XEZdEm0dfieRK3hccg22%2Bh%2BX%2FrtwMDFMdBqNPYo%3D\" --job testjob --skipCompression true"
-    }
-  }
-}
--- a/src/tools/ArtifactsUploader/ArtifactsUploader/Properties/launchSettings.json
+++ b/src/tools/ArtifactsUploader/ArtifactsUploader/Properties/launchSettings.json
@@ -1,0 +1,8 @@
+{
+  "profiles": {
+    "ArtifactsUploader": {
+      "commandName": "Project",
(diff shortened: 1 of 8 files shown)


billwert:
I don't see anywhere this is used in the repo, so I didn't fix up any usage of it. If it is being used somewhere let me know.

billwert:
I'm also very open to just removing this tool entirely and relying on [azcopy](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10) directly (though that would take a little scripting to do file discovery, so this is probably a reasonable place to be.)

billwert:
Actually, looking at it again, I think azcopy does precisely what we want.. Hm.

adamsitnik:
We were planning to use this tool to upload the trace files from perf runs to Azure Blob Storage. It was a year ago when the traces were uploaded to Jenkins. But then the reorg came and we have never used this tool.

As far as I remember the reason why I did not use azcopy was because we wanted to preserve the folder structure (namespace/typename/benchmark.etl) and zip it. One of the reasons could be the azcopy did not support all the OSes we wanted to run the benchmarks for.

Having said that, if azcopy meets your new requirements as of today, this tool can be removed.
If not, it can be changed to fulfill the new requirements.

src/tools/ArtifactsUploader/ArtifactsUploader/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "ArtifactsUploader": {
+      "commandName": "Project",
+      "commandLineArgs": "--container results --branch csfiles --artifacts . --searchPatterns *cs --storageUrl https://pvscmdupload.blob.core.windows.net --token \"?sv=2018-03-28&ss=bf&srt=sco&sp=rwdlac&se=2019-05-24T07:15:29Z&st=2019-05-23T23:15:29Z&spr=https&sig=IjS1XEZdEm0dfieRK3hccg22%2Bh%2BX%2FrtwMDFMdBqNPYo%3D\" --job testjob --skipCompression true"
```
adiaaida: We should not put the token in this json file.

billwert: great catch. I missed this file getting created. My bad. The token already expired but I am going to roll keys on this storage account anyway.

billwert:
Great. I do think azcopy probably fits our needs - we will try that and I will delete this tool if it does.

src/tools/ArtifactsUploader/ArtifactsUploader/Properties/launchSettings.json
jorive: I would add this file to `.gitignore`, so it does not happen in the future.

billwert: I was going to but found some evidence (1, 2) that perhaps it's more integral than the old .user project files..

1: https://github.com/github/gitignore/pull/2705
2: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-run?tabs=netcore21

jorive: Optionally, there is a CI tool that scan the repo to catch these things (though it would be too late once it's on a PR).

PR #501, opened 2019-05-24
S.C.Sort<T> add extra custom types and overloads to microbenchmark

nietras:
@jkotas here the extra types for sort microbenchmarks as discussed in https://github.com/dotnet/coreclr/pull/24419

### DIFF
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -39,6 +39,7 @@ namespace System.Collections
     [InvocationCount(InvocationsPerIteration)]
     [DisassemblyDiagnoser(recursiveDepth: 3)]
     [InliningDiagnoser]
+    [EtwProfiler]
     public class Sort<T>
     {
         private const int InvocationsPerIteration = 40000;
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -5,7 +5,6 @@
 using System.Collections.Generic;
 using System.Linq;
 using BenchmarkDotNet.Attributes;
-using BenchmarkDotNet.Diagnostics.Windows.Configs;
 using BenchmarkDotNet.Extensions;
 using MicroBenchmarks;
 
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -12,6 +12,7 @@ namespace System.Collections
 {
     [BenchmarkCategory(Categories.CoreCLR, Categories.Collections, Categories.GenericCollections)]
     [GenericTypeArguments(typeof(int))] // value type, Array sort in native code
+    [GenericTypeArguments(typeof(string))] // reference type, Array sort in native code
     [GenericTypeArguments(typeof(IntStruct))] // custom value type, sort in managed code
     [GenericTypeArguments(typeof(IntClass))] // custom reference type, sort in managed code
     [GenericTypeArguments(typeof(BigStruct))] // custom value type, sort in managed code
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -28,7 +28,7 @@ namespace System.Collections
         private int _iterationIndex = 0;
         private T[] _values;
         private T[][] _arrays;
-        private List<T>[] _lists = null;
+        private List<T>[] _lists;
 
         [GlobalSetup]
         public void Setup() => _values = GenerateValues();
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -78,7 +78,7 @@ namespace System.Collections
             return count;
         }
 
-        T[] GenerateValues()
+        private T[] GenerateValues()
         {
             if (typeof(T) == typeof(IntStruct))
             {
@@ -101,12 +101,12 @@ namespace System.Collections
             }
         }
 
-        sealed class ComparableComparerClass : IComparer<T>
+        private sealed class ComparableComparerClass : IComparer<T>
         {
             public int Compare(T x, T y) => x.CompareTo(y);
         }
 
-        readonly struct ComparableComparerStruct : IComparer<T>
+        private readonly struct ComparableComparerStruct : IComparer<T>
         {
             public int Compare(T x, T y) => x.CompareTo(y);
         }
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -28,17 +28,13 @@ namespace System.Collections
         private int _iterationIndex = 0;
         private T[] _values;
         private T[][] _arrays;
-        private List<T>[] _lists;
+        private List<T>[] _lists = null;
 
         [GlobalSetup]
         public void Setup() => _values = GenerateValues();
 
         [IterationCleanup]
-        public void CleanupIteration()
-        {
-            _iterationIndex = 0; // after every iteration end we set the index to 0
-            _lists = null; // after every iteration end we must clear the lists since `Utils.FillCollections` reuses lists
-        }
+        public void CleanupIteration() => _iterationIndex = 0; // after every iteration end we set the index to 0
 
         [IterationSetup(Targets = new []{ nameof(Array), nameof(Array_ComparerClass),
             nameof(Array_ComparerStruct), nameof(Array_Comparison) })]
@@ -57,7 +53,7 @@ namespace System.Collections
         public void Array_Comparison() => System.Array.Sort(_arrays[_iterationIndex++], (x, y) => x.CompareTo(y));
 
         [IterationSetup(Target = nameof(List))]
-        public void SetupListIteration() => Utils.FillCollections(ref _lists, InvocationsPerIteration, _values);
+        public void SetupListIteration() => Utils.ClearAndFillCollections(ref _lists, InvocationsPerIteration, _values);
 
         [Benchmark]
         public void List() => _lists[_iterationIndex++].Sort();
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -11,35 +11,17 @@ using MicroBenchmarks;
 
 namespace System.Collections
 {
-    // TODO: Where should the below types reside?
-
-    public readonly struct IntStruct : IComparable<IntStruct>
-    {
-        readonly int _value;
-
-        public IntStruct(int value) => _value = value;
-
-        public int CompareTo(IntStruct other) => _value.CompareTo(other._value);
-    }
-
-    public class IntClass : IComparable<IntClass>
-    {
-        readonly int _value;
-
-        public IntClass(int value) => _value = value;
-
-        public int CompareTo(IntClass other) => _value.CompareTo(other._value);
-    }
-
     [BenchmarkCategory(Categories.CoreCLR, Categories.Collections, Categories.GenericCollections)]
     [GenericTypeArguments(typeof(int))] // value type, Array sort in native code
     [GenericTypeArguments(typeof(IntStruct))] // custom value type, sort in managed code
     [GenericTypeArguments(typeof(IntClass))] // custom reference type, sort in managed code, compare fast
+    [GenericTypeArguments(typeof(BigStruct))] // custom value type, sort in managed code
     [GenericTypeArguments(typeof(string))] // reference type, compare slow
     [InvocationCount(InvocationsPerIteration)]
-    public class Sort<T>
+    public class Sort<T> where T : IComparable<T>
     {
-        private const int InvocationsPerIteration = 5000;
+        static readonly ComparableComparerClass _comparableComparerClass = new ComparableComparerClass();
+        private const int InvocationsPerIteration = 1000;
 
         [Params(Utils.DefaultCollectionSize)]
         public int Size;
@@ -56,20 +38,30 @@ namespace System.Collections
         [IterationCleanup]
         public void CleanupIteration() => _iterationIndex = 0; // after every iteration end we set the index to 0
 
-        [IterationSetup(Target = nameof(Array))]
+        [IterationSetup(Targets = new []{ nameof(Array), nameof(Array_ComparerClass),
+            nameof(Array_ComparerStruct), nameof(Array_Comparison) })]
         public void SetupArrayIteration() => Utils.FillArrays(ref _arrays, InvocationsPerIteration, _values);
 
         [Benchmark]
         public void Array() => System.Array.Sort(_arrays[_iterationIndex++], 0, Size);
 
-        //[IterationSetup(Target = nameof(List))]
+        [Benchmark]
+        public void Array_ComparerClass() => System.Array.Sort(_arrays[_iterationIndex++], 0, Size, _comparableComparerClass);
+
+        [Benchmark]
+        public void Array_ComparerStruct() => System.Array.Sort(_arrays[_iterationIndex++], 0, Size, new ComparableComparerStruct());
+
+        [Benchmark]
+        public void Array_Comparison() => System.Array.Sort(_arrays[_iterationIndex++], (x, y) => x.CompareTo(y));
+
+        [IterationSetup(Target = nameof(List))]
         public void SetupListIteration() => Utils.FillCollections(ref _lists, InvocationsPerIteration, _values);
 
-        //[Benchmark]
+        [Benchmark]
         public void List() => _lists[_iterationIndex++].Sort();
 
-        //[BenchmarkCategory(Categories.LINQ)]
-        //[Benchmark]
+        [BenchmarkCategory(Categories.LINQ)]
+        [Benchmark]
         public int LinqQuery()
         {
             int count = 0;
@@ -78,8 +70,8 @@ namespace System.Collections
             return count;
         }
 
-        //[BenchmarkCategory(Categories.LINQ)]
-        //[Benchmark]
+        [BenchmarkCategory(Categories.LINQ)]
+        [Benchmark]
         public int LinqOrderByExtension()
         {
             int count = 0;
@@ -104,6 +96,15 @@ namespace System.Collections
             {
                 return ValuesGenerator.ArrayOfUniqueValues<T>(Size);
             }
+        }
+
+        sealed class ComparableComparerClass : IComparer<T>
+        {
+            public int Compare(T x, T y) => x.CompareTo(y);
+        }
+        readonly struct ComparableComparerStruct : IComparer<T>
+        {
+            public int Compare(T x, T y) => x.CompareTo(y);
         }
     }
 }
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -102,6 +102,7 @@ namespace System.Collections
         {
             public int Compare(T x, T y) => x.CompareTo(y);
         }
+
         readonly struct ComparableComparerStruct : IComparer<T>
         {
             public int Compare(T x, T y) => x.CompareTo(y);
--- a/src/benchmarks/micro/corefx/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Sort.cs
@@ -34,7 +34,11 @@ namespace System.Collections
         public void Setup() => _values = GenerateValues();
 
         [IterationCleanup]
-        public void CleanupIteration() => _iterationIndex = 0; // after every iteration end we set the index to 0
+        public void CleanupIteration()
+        {
+            _iterationIndex = 0; // after every iteration end we set the index to 0
+            _lists = null; // after every iteration end we must clear the lists since `Utils.FillCollections` reuses lists
+        }
 
         [IterationSetup(Targets = new []{ nameof(Array), nameof(Array_ComparerClass),
             nameof(Array_ComparerStruct), nameof(Array_Comparison) })]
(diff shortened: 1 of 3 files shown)


src/benchmarks/micro/corefx/System.Collections/Sort.cs
```diff
@@ -10,26 +10,49 @@
...
+
+    public class IntClass : IComparable<IntClass>
+    {
+        readonly int _value;
+
+        public IntClass(int value) => _value = value;
+
+        public int CompareTo(IntClass other) => _value.CompareTo(other._value);
+    }
+
```
jkotas: It may be useful to also cover large struct with multiple fields.

nietras: Yes absolutely, I will add later when we have uncovered the Span regressions :)

nietras:
@adamsitnik thanks for the help in https://github.com/dotnet/coreclr/pull/24419 here I am adding some more tests but also trying to get assembly for the relevant code (through `DisassemblyDiagnoser`, but this does not work with the run:
```
dotnet run -c Release -f netcoreapp3.0 --filter **Collections.Sort* --coreRun 
C:\git\oss\coreclr-base\bin\tests\Windows_NT.x64.Release\Tests\Core_Root\CoreRun.exe 
C:\git\oss\coreclr\bin\tests\Windows_NT.x64.Release\Tests\Core_Root\CoreRun.exe
```
do you have any pointers on how this might be made to work? I'm interested in the assembly all the way down into `System.Private.CoreLib`.

Also see gist: https://gist.github.com/nietras/fc3dd8b145e0ea1f9ac93b76d67c1d42

billwert:
I see this is still WIP, so you will probably go back and uncomment things, but I wanted to state that in general we should not be removing or modifying parameters to existing benchmarks without very strong reasons (basically, they just don't work). We'd rather not disrupt existing test trends.

Thanks for helping us make our tests better!

nietras:
> you will probably go back and uncomment things, but I wanted to state that in general we should not be removing or modifying parameters to existing benchmarks without very strong reasons (basically, they just don't work). We'd rather not disrupt existing test trends

@billwert understood :), and yes this is very much WIP and I am committing temporary changes I make to get a tight dev loop and to keep a log of things I did. I will of course clean everything up before ready to merge. And at this point I will point out changes I think that should be made (apart from additional tests) e.g. insufficient invocation counts for int compared to string.

src/benchmarks/micro/corefx/System.Collections/Sort.cs
```diff
@@ -5,31 +5,56 @@
...
+    public readonly struct IntStruct : IComparable<IntStruct>
+    {
+        readonly int _value;
+
+        public IntStruct(int value) => _value = value;
+
+        public int CompareTo(IntStruct other) => _value.CompareTo(other._value);
+    }
+
+    public class IntClass : IComparable<IntClass>
```
nietras: @billwert where should these extra types reside?

adamsitnik:
@nietras the BDN disassembly diagnoser does not handle indirect calls (virtual methods) and native parts of CoreCLR.

But CoreCLR itself can produce the best JIT dump. Here is how you can do it:
https://github.com/dotnet/coreclr/blob/master/Documentation/building/viewing-jit-dumps.md
https://mijailovic.net/2018/07/05/generated-code/#coreclr

src/benchmarks/micro/corefx/System.Collections/Sort.cs
billwert: I think if they're just used by this benchmark in the file is fine. Putting them in a file next to Sort.cs is also fine.

src/benchmarks/micro/corefx/System.Collections/Sort.cs
nietras: @jkotas added `BigStruct`, does it look ok?

src/benchmarks/micro/corefx/System.Collections/Sort.cs
```diff
@@ -11,34 +11,48 @@
...
+    [GenericTypeArguments(typeof(IntStruct))] // custom value type, sort in managed code
+    [GenericTypeArguments(typeof(IntClass))] // custom reference type, sort in managed code, compare fast
+    [GenericTypeArguments(typeof(BigStruct))] // custom value type, sort in managed code
+    [GenericTypeArguments(typeof(string))] // reference type, compare slow
     [InvocationCount(InvocationsPerIteration)]
-    public class Sort<T>
+    public class Sort<T> where T : IComparable<T>
     {
+        static readonly ComparableComparerClass _comparableComparerClass = new ComparableComparerClass();
         private const int InvocationsPerIteration = 1000;
```
nietras: @billwert I have kept the `1000` InvocationsPerIteration, but do note this is very small for the small types.

nietras:
> BDN disassembly diagnoser does not handle indirect calls (virtual methods) and native parts of CoreCLR

@adamsitnik ah yes, too bad, thanks for the pointers though!

nietras:
Something is not right here. `List<T>.Sort` is a lot faster (2.4us) than `Array.Sort` (8.5us) in these tests, which of course makes no sense since `List<T>.Sort` just calls `Array.Sort`. @jkotas any idea why this could be the case?

``` ini

BenchmarkDotNet=v0.11.3.1003-nightly, OS=Windows 10.0.17763.503 (1809/October2018Update/Redstone5)
Intel Core i7-8700 CPU 3.20GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=3.0.100-preview5-011568
  [Host]     : .NET Core 3.0.0-preview5-27626-15 (CoreCLR 4.6.27622.75, CoreFX 4.700.19.22408), 64bit RyuJIT
  Job-KXBXKF : .NET Core ? (CoreCLR 4.6.27803.0, CoreFX 4.700.19.22901), 64bit RyuJIT
  Job-GPLUIV : .NET Core ? (CoreCLR 4.6.27808.0, CoreFX 4.700.19.22901), 64bit RyuJIT

Runtime=Core  InvocationCount=1000  IterationTime=250.0000 ms  
MaxIterationCount=20  MinIterationCount=15  UnrollFactor=1  
WarmupCount=1  

```


# `IntStruct`
|               Method |     Toolchain | Size |      Mean |     Error |    StdDev |    Median |       Min |       Max | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|--------------------- |-------------- |----- |----------:|----------:|----------:|----------:|----------:|----------:|------:|--------:|------------:|------------:|------------:|--------------------:|
|                Array | \coreclr-base |  512 |  8.527 us | 3.7468 us | 4.3148 us | 11.264 us |  3.699 us | 12.627 us |  1.00 |    0.00 |           - |           - |           - |                   - |
|                Array |      \coreclr |  512 |  7.398 us | 2.9548 us | 3.4028 us |  6.485 us |  3.612 us | 11.755 us |  1.01 |    0.55 |           - |           - |           - |                   - |
|                      |               |      |           |           |           |           |           |           |       |         |             |             |             |                     |
|  Array_ComparerClass | \coreclr-base |  512 | 21.746 us | 0.8421 us | 0.9697 us | 21.445 us | 20.485 us | 23.057 us |  1.00 |    0.00 |           - |           - |           - |                64 B |
|  Array_ComparerClass |      \coreclr |  512 | 21.227 us | 0.7182 us | 0.8271 us | 20.840 us | 20.475 us | 22.683 us |  0.98 |    0.06 |           - |           - |           - |                64 B |
|                      |               |      |           |           |           |           |           |           |       |         |             |             |             |                     |
| Array_ComparerStruct | \coreclr-base |  512 | 21.978 us | 0.1853 us | 0.1643 us | 21.924 us | 21.763 us | 22.345 us |  1.00 |    0.00 |           - |           - |           - |                88 B |
| Array_ComparerStruct |      \coreclr |  512 | 23.244 us | 0.7342 us | 0.8455 us | 23.643 us | 21.525 us | 24.104 us |  1.05 |    0.04 |           - |           - |           - |                88 B |
|                      |               |      |           |           |           |           |           |           |       |         |             |             |             |                     |
|     Array_Comparison | \coreclr-base |  512 | 22.059 us | 0.8001 us | 0.9214 us | 22.666 us | 20.515 us | 22.749 us |  1.00 |    0.00 |           - |           - |           - |                   - |
|     Array_Comparison |      \coreclr |  512 | 21.765 us | 0.8586 us | 0.9888 us | 22.273 us | 20.278 us | 22.809 us |  0.99 |    0.08 |           - |           - |           - |                   - |
|                      |               |      |           |           |           |           |           |           |       |         |             |             |             |                     |
|                 List | \coreclr-base |  512 |  2.445 us | 0.0580 us | 0.0620 us |  2.440 us |  2.311 us |  2.597 us |  1.00 |    0.00 |           - |           - |           - |                   - |
|                 List |      \coreclr |  512 |  2.899 us | 0.0885 us | 0.1020 us |  2.915 us |  2.574 us |  3.070 us |  1.18 |    0.03 |           - |           - |           - |                   - |
|                      |               |      |           |           |           |           |           |           |       |         |             |             |             |                     |
|            LinqQuery | \coreclr-base |  512 | 44.199 us | 1.0197 us | 0.9538 us | 44.263 us | 42.778 us | 46.191 us |  1.00 |    0.00 |      1.0000 |           - |           - |              6544 B |
|            LinqQuery |      \coreclr |  512 | 44.612 us | 0.7270 us | 0.6071 us | 44.691 us | 43.585 us | 45.881 us |  1.01 |    0.03 |      1.0000 |           - |           - |              6544 B |
|                      |               |      |           |           |           |           |           |           |       |         |             |             |             |                     |
| LinqOrderByExtension | \coreclr-base |  512 | 43.689 us | 0.5569 us | 0.4651 us | 43.720 us | 42.796 us | 44.466 us |  1.00 |    0.00 |      1.0000 |           - |           - |              6544 B |
| LinqOrderByExtension |      \coreclr |  512 | 44.544 us | 0.8497 us | 0.8346 us | 44.415 us | 43.528 us | 46.366 us |  1.02 |    0.02 |      1.0000 |           - |           - |              6544 B |

src/benchmarks/micro/corefx/System.Collections/Sort.cs
```diff
@@ -11,34 +11,47 @@
...
         private T[][] _arrays;
         private List<T>[] _lists;
 
         [GlobalSetup]
-        public void Setup() => _values = ValuesGenerator.ArrayOfUniqueValues<T>(Size);
-        
+        public void Setup() => _values = GenerateValues();
+
         [IterationCleanup]
         public void CleanupIteration() => _iterationIndex = 0; // after every iteration end we set the index to 0
```
nietras: @billwert @jkotas  I think there is a bug in the List benchmark, the `CleanupIteration` should reset `_lists = null` since otherwise, the Lists will be used, that is how `Utils.FillCollections` work it seems, unlike `UtilsFillArray`, the way `Utils.FillCollections` is a bit prone to errors for tests that are "destructive".

nietras:
Yes, the problem was a bug in the benchmark. After this Array and List give same results as they should. 

|               Method | Size |      Mean |     Error |    StdDev |   Median |       Min |      Max | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|--------------------- |----- |----------:|----------:|----------:|---------:|----------:|---------:|------------:|------------:|------------:|--------------------:|
|                Array |  512 |  9.799 us | 1.8174 us | 2.0929 us | 11.18 us |  3.805 us | 11.30 us |           - |           - |           - |                   - |
|  Array_ComparerClass |  512 | 21.634 us | 0.8926 us | 1.0280 us | 22.34 us | 20.091 us | 22.65 us |           - |           - |           - |                64 B |
| Array_ComparerStruct |  512 | 23.036 us | 0.3148 us | 0.2945 us | 22.87 us | 22.679 us | 23.56 us |           - |           - |           - |                88 B |
|     Array_Comparison |  512 | 22.322 us | 0.7610 us | 0.8764 us | 22.86 us | 20.630 us | 23.08 us |           - |           - |           - |                   - |
|                 List |  512 | 10.093 us | 2.9595 us | 3.4082 us | 12.21 us |  3.767 us | 12.27 us |           - |           - |           - |                   - |
|            LinqQuery |  512 | 46.353 us | 0.8793 us | 0.9408 us | 46.47 us | 44.819 us | 48.46 us |      1.0000 |           - |           - |              6544 B |
| LinqOrderByExtension |  512 | 45.585 us | 0.6159 us | 0.5143 us | 45.54 us | 45.014 us | 46.85 us |      1.0000 |           - |           - |              6544 B |

src/benchmarks/micro/corefx/System.Collections/Sort.cs
adamsitnik: @nietras Thanks for finding and fixing the bug!

The goal of the method executed as `IterationSetup` is to take less time than the benchmarked method itself. If we allocate a new list and add all the elements to it it's going to take more time than Sort itself. When we run such a benchmark it will be harder to get stable results.

Could you please fix this bug by not setting the `list` to `null` in the IterationCleanup but instead implementing a new method that will just put the elements using indexer? Sth like:

```cs
internal static void SetItems<TCollection, TValues>(ref TCollection[] collections, int collectionsCount, TValues[] keys)
    where TCollection : IList<TValues>, new()
{
    if (collections == null)
        collections = Enumerable.Range(0, collectionsCount).Select(_ => new TCollection()).ToArray();

    foreach (var collection in collections)
        for (int i = 0; i < keys.Length; i++)
            collection[i] = keys[i];

    if (collections.Any(collection => collection.Count != keys.Length)) // we dont use Debug.Assert here because this code will be executed mostly in Release
        throw new InvalidOperationException();
}
```

Thanks!

nietras: @adamsitnik no problem, but won't what you suggest fail if we get more size test e.g.:
```
        [Params(512, 256)]
        public int Size;
```
If `IterationSetup` should be fast, why not allocated these before that? 

The other thing is that the `Utils` helper methods seem prone to errors e.g.:
```csharp
internal static void FillCollections<TCollection, TValues>(ref TCollection[] collections, int collectionsCount, TValues[] keys)
    where TCollection : ICollection<TValues>, new()
{
    if (collections == null)
        collections = Enumerable.Range(0, collectionsCount).Select(_ => new TCollection()).ToArray();

    foreach (var collection in collections.Where(collection => collection.Count < keys.Length))
    foreach (var value in keys)
        collection.Add(value);
    
    if(collections.Any(collection => collection.Count != keys.Length)) // we dont use Debug.Assert here because this code will be executed mostly in Release
        throw new InvalidOperationException();
}
```
If you change `collectionsCount` during benchmark e.g. from 100 to 10, this will not fail and you will have more collections than "asked for". Depending on the benchmark this could be a problem.

nietras: @adamsitnik  the main source of variance here is in fact that `InvocationsPerIteration` should be different depending on `T` although this of course gets a separate run per `T`. In some ways, the code implicitly depends a lot on how `BenchmarkDotNet` executes the benchmark, which may be ok of course.

nietras: @adamsitnik I have added `ClearAndFillCollections` instead, this ensure the length always matches the `TValues[]` given.

nietras: @adamsitnik also added check for `collectionsCount` let me know if that is ok :)

ahsonkhan:
> Yes, the problem was a bug in the benchmark. After this Array and List give same results as they should.

Looking at some of the error/std dev numbers, there seems quite a bit of variation (e.g. ~3 us on a test that takes ~10 us). I noticed previously (here: https://github.com/dotnet/coreclr/pull/24419#issuecomment-495593099) you mentioned that the tests were running quickly. 

>  StdDev is quite high for some of this, and the test run seemed to run for quite a short time here

What I usually do for local runs, is remove the custom settings for benchmarks runs within this repo. That will result in more accurate results (and the perf test will take longer to run). Hopefully that reduces the error as well. Try that before trying to infer whether one implementation approach is better than the other because the perf result just might be too noisy to make inferences on, at this level.

So, I just comment this out in the `RecommendedConfig`:
https://github.com/dotnet/performance/blob/0eec798bb592aaf8ca676d0cb8c79d22b6404e6f/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L17-L20

@adamsitnik, maybe we should call this out in the guide somewhere (unless you have a better inner-dev-loop workflow that might be more optimal). Usually accuracy trumps test run time when doing this type of investigations related to perf work/optimizations.

nietras:
>  result in more accurate results (and the perf test will take longer to run)

@ahsonkhan I did exactly that :), changing `InvocationsPerIteration` to 40.000 for `IntStruct` for example. Note that the `Sort<T>` test runs differently than the general tests, it does not use `RecommededConfig`. 

As I pointed out a few times:

> main source of variance here is in fact that `InvocationsPerIteration` should be different depending on `T`

The `Sort<T>` really should have different `InvocationsPerIteration` for different `T`, unfortunately it is not that easy to change with the attribute usage for this currently. `1000` is fair enough for `string` where compares are expensive, but not for `IntStruct` where compares are orders of magnitudes faster.

If `Sort<T>` had been running with the `RecommendedConfig` it would have been fine, but we can't do that here since `Sort<T>` is a "destructive" test that changes input. So iterations needs to be controlled.

nietras:
This PR is ready for review/merge, and needs to go in before https://github.com/dotnet/coreclr/pull/24419 as according to plan in https://github.com/dotnet/coreclr/pull/24419#issuecomment-496026316 Note that that PR contains more benchmark results.

src/benchmarks/micro/corefx/System.Collections/Sort.cs
```diff
@@ -64,5 +77,38 @@ public int LinqOrderByExtension()
                 count++;
             return count;
         }
+
+        T[] GenerateValues()
```
ahsonkhan: nit: add access modifier, explicitly

src/benchmarks/micro/corefx/System.Collections/Sort.cs
```diff
@@ -64,5 +77,38 @@ public int LinqOrderByExtension()
...
+                var values = ValuesGenerator.ArrayOfUniqueValues<int>(Size);
+                return (T[])(object)values.Select(v => new BigStruct(v)).ToArray();
+            }
+            else
+            {
+                return ValuesGenerator.ArrayOfUniqueValues<T>(Size);
+            }
+        }
+
+        sealed class ComparableComparerClass : IComparer<T>
```
ahsonkhan: Same here/below. Add access modifier keyword.

src/benchmarks/micro/corefx/System.Collections/Sort.cs
nietras: fixed

nietras:
@adamsitnik CC: @jkotas @ahsonkhan I have removed `string` from the test case, as it makes little sense testing it for `Sort` given the circumstances, and with this gone `InvocationsPerIteration` can be bumped to `50000`giving us reliable results for the actually interesting test cases.

If this looks good, can we please merge? :)

adamsitnik:
@nietras I was hoping that I will be just able to commit the suggestions to your fork and merge the PR. But I lack the permissions.

obraz

Could you please apply my suggested changes? As soon as you do that I am going to merge this PR

src/benchmarks/micro/corefx/System.Collections/Sort.cs
```diff
@@ -11,36 +11,48 @@
 namespace System.Collections
 {
     [BenchmarkCategory(Categories.CoreCLR, Categories.Collections, Categories.GenericCollections)]
-    [GenericTypeArguments(typeof(int))] // value type
-    [GenericTypeArguments(typeof(string))] // reference type
+    [GenericTypeArguments(typeof(int))] // value type, Array sort in native code
```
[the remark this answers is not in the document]
nietras: @adamsitnik `string` is the reason the benchmark take and insubordinate amount of time e.g. 15 minutes with 50000 iterations. There is no big value in keeping it in my mind, at 5000 it will still be slow. We are taking magnitudes slower, hence it dominates the total benchmark run, at the cost of precision for the other tests.

However, I have approved your suggestion, then you can decide if you still want to keep it :)

src/benchmarks/micro/corefx/System.Collections/Sort.cs
```diff
@@ -11,36 +11,48 @@
...
+
+        [Benchmark]
+        public void Array_ComparerStruct() => System.Array.Sort(_arrays[_iterationIndex++], 0, Size, new ComparableComparerStruct());
+
+        [Benchmark]
+        public void Array_Comparison() => System.Array.Sort(_arrays[_iterationIndex++], (x, y) => x.CompareTo(y));
+
         [IterationSetup(Target = nameof(List))]
-        public void SetupListIteration() => Utils.FillCollections(ref _lists, InvocationsPerIteration, _values);
+        public void SetupListIteration() => Utils.ClearAndFillCollections(ref _lists, InvocationsPerIteration, _values);
```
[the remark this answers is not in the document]
nietras: thanks 😄

nietras:
> please apply my suggested changes?

@adamsitnik done, but see my notes on `string` and iteration count.

adamsitnik:
@nietras thank you! and once again appologies for the very delayed review

adamsitnik:
> done, but see my notes on string and iteration count.

thanks! I am working on https://github.com/dotnet/BenchmarkDotNet/pull/1157 to get rid of this problem once and for all

nietras:
@adamsitnik thanks for merging, are these benchmarks run as part of an official build that one can use a baseline/reference?

adamsitnik:
>  are these benchmarks run as part of an official build

yes, every benchmark that we have here and is part of the CoreFX or CoreCLR category is run on our dedicated hardware a few times a day and is used to prevent from regressions in .NET Core

nietras:
@adamsitnik where can you see the results of this?

adamsitnik:
> where can you see the results of this?

We have an internal Reporting System for that. But @DrewScoggins is working on a Power-BI based solution and maybe in the future we might make the results public

PR #504, opened 2019-05-24
Add new automation support to runner

DrewScoggins:
We need to set the cli commit hash and commit date to the environment
before we run the benchmarks for the new reporting tool. I have taken
the code from where we do that as part of the sending to benchview and
put it right before we run the benchmarks. I also make sure that this is
reset after we run.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -237,6 +237,13 @@ def __main(args: list) -> int:
     # Run micro-benchmarks
     if not args.build_only:
         for framework in args.frameworks:
+            #Before we run the benchmarks we need to set the CommitDate in the environment
+            #for the nex reporting tool
+            target_framework_moniker = micro_benchmarks.FrameworkAction.get_target_framework_moniker(framework)
+            commit_sha = dotnet.get_dotnet_sdk(target_framework_moniker, args.cli)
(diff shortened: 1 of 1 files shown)


DrewScoggins:
@billwert @adiaaida PTAL

scripts/benchmarks_ci.py
```diff
@@ -237,13 +237,22 @@ def __main(args: list) -> int:
...
+            os.environ['PERFLAB_HASH'] = commit_sha
+            os.environ['PERFLAB_BUILDTIMESTAMP'] = source_timestamp
             micro_benchmarks.run(
                 BENCHMARKS_CSPROJ,
                 args.configuration,
                 framework,
                 verbose,
                 args
             )
+            os.environ['PERFLAB_HASH'] = ""
```
jorive: I would just remove them from the environment instead of setting them to empty string.

jorive: maybe use something like `os.unsetenv` or `os.environ.pop`?

adiaaida:
And net461 (ie full framework runs) don't have this hash apparently?

jorive:
Those parameters should be added as optional script parameters, and passed on automation (yml file).

DrewScoggins:
Yeah, this currently will only work for our sdk runs, but since that is all we are doing right now I feel OK with that. I went ahead and made a change to make sure we don't break net461, but once the CI passed I am going to check this in.

PR #506, opened 2019-05-25
Enable uploading perf-lab-report.json files

billwert:
Instead of trying to extract functionality from the benchview scripts for uploading, add a new script that does it. Uses AzCopy to copy files up.

### DIFF
--- a/scripts/azcopy.py
+++ b/scripts/azcopy.py
@@ -3,7 +3,6 @@
 import os
 from os import sys, path, makedirs, environ
 from glob import glob
-from performance.logger import setup_loggers
 from urllib.request import urlopen
 from tarfile import TarFile
 from zipfile import ZipFile
@@ -72,7 +71,6 @@ class AzCopy:
 
     @staticmethod
     def upload_results(containerPath: str, verbose: bool) -> None:
-        setup_loggers(verbose)
         if(os.environ.get('PERFLAB_UPLOAD_TOKEN') != None):
             files = glob(path.join(get_artifacts_directory(), '**','*perf-lab-report.json'), recursive=True)
             if files:
--- a/scripts/azcopy.py
+++ b/scripts/azcopy.py
@@ -1,34 +1,36 @@
-#!/usr/bin/env python3
-    
-import os
-from os import sys, path, makedirs, environ
-from glob import glob
-from urllib.request import urlopen
-from tarfile import TarFile
-from zipfile import ZipFile
-from logging import getLogger
+"""AzCopy"""
 
-from performance.common import get_tools_directory, get_artifacts_directory
-from performance.common import RunCommand
+import os
+from os import makedirs, path, sys
+from glob import glob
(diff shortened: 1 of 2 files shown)


jorive:
@billwert What did you mean with "Use a different formatting technqiue due to old python on linux machines"? The new script explicitly targets python3 (`#!/usr/bin/env python3`). Could you please ellaborate?

scripts/azcopy.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python3
```
jorive: If the script is meant to be used as a module and standalone, could you please make it executable (`chmod +x`)?

billwert:
> @billwert What did you mean with "Use a different formatting technqiue due to old python on linux machines"? The new script explicitly targets python3 (`#!/usr/bin/env python3`). Could you please ellaborate?

Honestly, just a hunch. The code worked fine on the Windows legs and on my WSL with 3.6.x but failed in the Ubuntu legs in the lab. I suspect those are earlier than 3.6 which is when that style of string formatting came into the language. (ref [PEP 498](https://www.python.org/dev/peps/pep-0498/))

jorive:
I thought the scripts work for 3.5+. I'm curious about the python version installed in the lab and the error you were getting.

billwert:
> I thought the scripts work for 3.5+. I'm curious about the python version installed in the lab and the error you were getting.

```
+ python3 --version
Python 3.5.2
```

The Windows machines have 3.7.

adiaaida:
For the Ubuntu machines, we do a apt-get install of python3-dev, which gets us 3.5.1-3 for the 16.04 machines. For 18.04, it's 3.6.5-3. If we want 3.7 on the linux machines, we will need to explicitly ask the core eng team to add that to the queues (by having the vms install python3.7-dev and any additional associated packages).

billwert:
I don't have a strong desire - knowing we'll get closer to "modern" when we go to 18.04 is great.

jorive:
Thanks for "making the code look more like python" and keeping the script as a module :)

scripts/benchmarks_ci.py
```diff
@@ -142,6 +143,12 @@ def __is_valid_datetime(dt: str) -> str:
             "%%Y-%%m-%%dT%%H:%%M:%%SZ").'''
     )
 
+    parser.add_argument('--upload-to-perflab-container',
```
jorive: One last thing you could do when using `argparse` is to add validation to the user input. Currently, the argument does not explain that the input is a "path" (I would rename the argument), and maybe validate that the path exist?

billwert: It turns out that is not a path on disk but is a path in the results blob. It's also sort of optional - if you just specify `--upload-to-perflab-contanier` with no argument it will go to the root of the results container. (That may never be what we want but it might be?) Gonna leave it.

PR #512, opened 2019-05-28
Add python script to generate the HelixPreCommands script that will set up the environment

adiaaida:


scripts/benchmarks_ci.py
```diff
@@ -250,8 +251,6 @@ def __main(args: list) -> int:
                 target_framework_moniker = micro_benchmarks.FrameworkAction.get_target_framework_moniker(framework)
                 commit_sha = dotnet.get_dotnet_sdk(target_framework_moniker, args.cli)
                 source_timestamp = dotnet.get_commit_date(framework, commit_sha, args.cli_repository)
-                os.environ['PERFLAB_HASH'] = commit_sha
-                os.environ['PERFLAB_BUILDTIMESTAMP'] = source_timestamp
```
billwert: does it make sense to just roll back the entire change that brought this in at this point? (maybe it doesn't matter.)

scripts/ci_setup.py
```diff
@@ -0,0 +1,202 @@
...
+        default=[],
+        help='Configurations used in the build in key=value format'
+    )
+
+    return parser
+
+
+def __process_arguments(args: list):
+    parser = ArgumentParser(
+        description='Tool to run .NET micro benchmarks',
```
billwert: is this copied from somewhere? Doesn't seem like the right description.

scripts/dotnet.py
```diff
@@ -303,23 +308,25 @@ def run(self,
...
     if not groups:
         raise ValueError("Unknown target framework: {}".format(framework))
 
     FrameworkVersion = namedtuple('FrameworkVersion', ['major', 'minor'])
     version = FrameworkVersion(int(groups.group(1)), int(groups.group(2)))
 
+    return version
+
+def get_sdk_path(version: str, dotnet_path: str = None) -> str:
+    """Gets the dotnet Host version from the `dotnet --info` command."""
```
billwert: this seems like the wrong docstring?

scripts/ci_setup.py
adiaaida: Yup.

scripts/dotnet.py
adiaaida: I'll fix. Copy paste

scripts/benchmarks_ci.py
adiaaida: Oh, right, extra delete!

scripts/ci_setup.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python3
+
+from argparse import ArgumentParser, ArgumentTypeError
```
jorive: `ArgumentTypeError`, `datetime.datetime`, `platform` imported but not used.

scripts/ci_setup.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python3
```
jorive: If this is a stand anole script, could you please check it in as executable?

PR #513, opened 2019-05-28
Add missing output type to project

billwert:


billwert:
I got tripped up by the project type GUID mismatching the target framework choice, which was breaking it loading in VS.

PR #519, opened 2019-05-29
Azcopy fix

billwert:


scripts/azcopy.py
```diff
@@ -73,18 +73,37 @@ def upload_files(self, search_path: str):
...
+                                               filename)
+                    os.rename(file, newname)
+
+                renamed_files = glob(path.join(
+                                        get_artifacts_directory(),
+                                        '**',
+                                        '*perf-lab-report.json'), recursive=True)
+
+
+                dirname = path.dirname(renamed_files[0])
```
adiaaida: Does this ever get used?

adiaaida: Oh, nevermind. I found it.

scripts/benchmarks_ci.py
```diff
@@ -259,8 +260,8 @@ def __main(args: list) -> int:
 
         benchview.run_scripts(args, verbose, BENCHMARKS_CSPROJ)
 
-        if(args.upload_to_perflab_container): 
-            AzCopy.upload_results(args.upload_to_perflab_container, verbose)
+        if args.upload_to_perflab_container: 
+            AzCopy.upload_results('', verbose=verbose)
```
jorive: Is empty string intended?

scripts/benchmarks_ci.py
```diff
@@ -147,7 +147,8 @@ def __is_valid_datetime(dt: str) -> str:
     parser.add_argument('--upload-to-perflab-container',
         dest="upload_to_perflab_container",
         required=False,
-        help="Container to upload perf lab results to."
+        help="Container to upload perf lab results to.",
```
jorive: Do we need default value here?

scripts/benchmarks_ci.py
billwert: Yes.

scripts/benchmarks_ci.py
billwert: good catch - need to update help. (though no one is ever actually going to use this but us.)

PR #520, opened 2019-05-29
Use Windows.10.Amd64.Open queue

adiaaida:


adiaaida:
This was... moderately faster.

adiaaida:
Since our test machines are client, we want to stay client.

Issue #523, opened 2019-05-30
JSON microbenchmarks for JIL use the old MS Dateformat


Tornhoof:
The default DateTime Format in JIL looks like this:
"\/Date(1559314917689+0000)\/"

As you want to compare apples with apples, you need to pass Options.ISO8601 to each JIL invocation, then it will output the ISO8061 format the other serializers use.

adamsitnik:
Fixed by #524

PR #524, opened 2019-05-30
Change Datetime format of JIL to ISO8601

Tornhoof:
Solves #523


Tornhoof:
Sorry, I don't really now why it is the default, but JSON.NET also used that format by default in early versions (~2012) and maybe it's for compatibility reasons.

PR #527, opened 2019-05-30
Specify artifacts directory for ci

adiaaida:


billwert:
ugh, wish we'd gotten to talk 30 seconds more before you did this. BenchmarkDotNet already takes `--artifacts` on the command line. It appears that command line arguments win over properties on the Config in code, so I think we can just simplify the change to be the yml only.

adiaaida:
Cool cool. I'll do that. I didn't actually realize that (and why weren't we doing this in the lab already?)

adamsitnik:
As far as I remember, the python script that uploads the data to BenchView expects the artifacts to be in a certain directory (something like `artifacts\bin\MicroBenchmarks\Release\netcoreapp3.0\BenchmarkDotNet.Artifacts`). Recently generating BenchView data for PR runs was disabled in #508 so it's not tested anymore as part of the PRs. 

What I am afraid of is that this change is going to break the BenchView data generator and uploader, but we are going to find out about this after it gets merged with the first "official" CI run.

Please make sure that it's not going to break it.

billwert:
Good point. I thought it was more flexible but it is a string in there.

https://github.com/dotnet/performance/blob/2578ce569f9d46a0ecc3ef03e03835cc33a3d136/scripts/benchview.py#L357

Still worth fixing but needs a little more care.

Also worth turning back on generating data for this PR to ensure it’s right, and revert that part when we are happy.

Thanks,
-Bill

adiaaida:
I will do some testing with --generate-benchview-data turned on.

adiaaida:
@adamsitnik the Roslyn benchmarks are set up kind of strange where they don't pass args to the Run method. It also uses BenchmarkRunner.Run instead of BenchmarkSwitcher.Run. What's the best way to fix this so that the --artifacts parameter gets sent to the Roslyn benchmarks?

billwert:
Just change what it's hard coded to to point to where the new location is. Once we make this change we're not going to change it again (while we still use Benchview). We're only jumping through this hoop to keep current compat with the `benchmarks_ci` script so it isn't broken by overwriting artifacts if someone runs it with two frameworks.

billwert:
If that means just adding parsing of this specific argument to the Roslyn benchmarks that's also a fine way to go.

billwert:
LGTM

adamsitnik:
> the Roslyn benchmarks are set up kind of strange where they don't pass args to the Run method. It also uses BenchmarkRunner.Run instead of BenchmarkSwitcher.Run. What's the best way to fix this so that the --artifacts parameter gets sent to the Roslyn benchmarks?

I think that the best way to solve it is to replace the usage of `BenchmarkRunner` with `BenchmarkSwitcher`.

adamsitnik:
@adiaaida I have created https://github.com/dotnet/performance/pull/533 to address the Roslyn issue.

adiaaida:
@adamsitnik Thank you! When that goes in, I'll revert my changes to Roslyn and we can just use that.

PR #532, opened 2019-05-31
Make 'Duration' not a TopCounter

billwert:


billwert:
@DrewScoggins merge this when you're ready.

PR #534, opened 2019-06-02
Improve parsing test

agocke:
The current parsing test includes all the file IO for reading
the files off disk instead of just the parsing. The new version
of the test only parses. The test config is also adapted to
adjust to the shortened total time of the test.

### DIFF
--- a/src/benchmarks/real-world/Roslyn/Program.cs
+++ b/src/benchmarks/real-world/Roslyn/Program.cs
@@ -6,18 +6,37 @@ using System.IO;
 using System.Threading.Tasks;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Extensions;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Diagnosers;
+using BenchmarkDotNet.Exporters.Json;
+using BenchmarkDotNet.Columns;
 
 namespace CompilerBenchmarks
 {
     public class Program
     {
+        private static IConfig CustomConfig(DirectoryInfo artifactsPath, ImmutableHashSet<string> mandatoryCategories, int? partitionCount = null, int? partitionIndex = null)
+            => DefaultConfig.Instance
+                .With(Job.Default) // tell BDN that this are our default settings
+                .WithArtifactsPath(artifactsPath.FullName)
+                .With(MemoryDiagnoser.Default) // MemoryDiagnoser is enabled by default
+                .With(new OperatingSystemFilter())
+                .With(new PartitionFilter(partitionCount, partitionIndex))
+                .With(JsonExporter.Full) // make sure we export to Json (for BenchView integration purpose)
+                .With(new PerfLabExporter())
+                .With(StatisticColumn.Median, StatisticColumn.Min, StatisticColumn.Max)
+                .With(TooManyTestCasesValidator.FailOnError)
+                .With(new UniqueArgumentsValidator()) // don't allow for duplicated arguments #404
+                .With(new MandatoryCategoryValidator(mandatoryCategories));
+
         public static async Task<int> Main(string[] args)
         {
             await Setup();
 
             return BenchmarkSwitcher
                 .FromAssembly(typeof(Program).Assembly)
-                .Run(args, RecommendedConfig.Create(
+                .Run(args, CustomConfig(
                     artifactsPath: new DirectoryInfo(Path.Combine(Path.GetDirectoryName(typeof(Program).Assembly.Location), "BenchmarkDotNet.Artifacts")),
                     mandatoryCategories: ImmutableHashSet.Create("Roslyn")))
                 .ToExitCode();
--- a/src/benchmarks/real-world/Roslyn/Program.cs
+++ b/src/benchmarks/real-world/Roslyn/Program.cs
@@ -18,7 +18,7 @@ namespace CompilerBenchmarks
     {
         private static IConfig CustomConfig(DirectoryInfo artifactsPath, ImmutableHashSet<string> mandatoryCategories, int? partitionCount = null, int? partitionIndex = null)
             => DefaultConfig.Instance
-                .With(Job.Default) // tell BDN that this are our default settings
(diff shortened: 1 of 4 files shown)


billwert:
@agocke I think you got tripped up by how benchmarkdotnet builds out jobs.

@adamsitnik I must admit that I do not have a firm grasp on the job concept in BenchmarkDotNet - specifically what the default job actually is, or what all the variants mean. I observed that RecommendedConfig calls `.AsDefault()` so copied it, and it fixed the problem (we were getting duplicate jobs), but I do not know if this is actually the correct fix. Thoughts?

src/benchmarks/real-world/Roslyn/StageBenchmarks.cs
```diff
@@ -18,11 +20,36 @@ public class StageBenchmarks
...
+        public SyntaxTree[] Parsing()
         {
-            return Helpers.CreateReproCompilation();
+            var trees = new SyntaxTree[_files.Length];
+            Parallel.For(0, _files.Length, index =>
+            {
+                trees[index] = SyntaxFactory.ParseSyntaxTree(_files[index], _parseOptions);
+            });
+            return trees;
         }
```
adamsitnik: `Parallel.For` uses a heuristic that decides how many tasks to use to parallelize given code. This makes the benchmarks very often unstable (if you run it `n` times the results will most probably vary).

What is the point of making this benchmark parallel?

Maybe it would be good to have two implementations? Sequential and parallel?

src/benchmarks/real-world/Roslyn/StageBenchmarks.cs
```diff
@@ -18,11 +20,36 @@ public class StageBenchmarks
...
+                    _files[index] = SourceText.From(fstream);
+                }
+            });
+        }
 
         [Benchmark]
-        public object Parsing()
+        public SyntaxTree[] Parsing()
         {
-            return Helpers.CreateReproCompilation();
```
adamsitnik: Should the `CreateReproCompilation` method be removed now because it's not used anymore?

adamsitnik:
> I must admit that I do not have a firm grasp on the job concept in BenchmarkDotNet - specifically what the default job actually is, or what all the variants mean. I observed that RecommendedConfig calls .AsDefault() so copied it, and it fixed the problem (we were getting duplicate jobs), but I do not know if this is actually the correct fix. Thoughts?

BenchmarkDotNet runs the benchmark for every `Job` defined in a `Config`. 
For example: if we run `--runtimes netcoreapp2.1 netcoreapp2.2` this is represented as a config with two jobs. 

`Job.Default` is a set of recommended, default settings (Recommended would be a better name here): warmup, pilot, some certain level of accuracy etc.

The problems with the API started when I have introduced command line arguments support. How should the arguments passed by a user from command line level affect the config: replace the values provided in the code? Or extend the Config with a new Job?

```cs
vod Main(string[] args)
    => BenchmarkSwitcher
        .FromAssembly(typeof(Program).Assembly)
        .Run(args, DefaultConfig.Instance.With(Job.Default.WithWarmupCount(1)));
```

```cmd
dotnet run --warmupCount 2
```

And now back to the unfortunate `job.AsDefault()` - it means - this is my default job, if I provide anything from the console line arguments, the settings of `job.AsDefault()` should be overwritten.

Also, when we pass some settings that simply require new job to be created, the `.AsDefault()` job instance is cloned and extended with provided settings.

Example:

```cs
vod Main(string[] args)
    => BenchmarkSwitcher
        .FromAssembly(typeof(Program).Assembly)
        .Run(args, DefaultConfig.Instance.With(Job.Default.WithWarmupCount(1)));
```

```cmd
dotnet run --runtimes netcoreapp2.1 netcoreapp2.2
```

Is going to give us two jobs: .NET Core 2.1 and 2.2 and both of them are going to have `WarmupCount=1`.

I am open to suggestions, especially to introducing a new method with better name that would do the same as `.AsDefault()` but have more intuitive name.

src/benchmarks/real-world/Roslyn/Program.cs
```diff
@@ -6,18 +6,37 @@
...
+                .WithArtifactsPath(artifactsPath.FullName)
+                .With(MemoryDiagnoser.Default) // MemoryDiagnoser is enabled by default
+                .With(new OperatingSystemFilter())
+                .With(new PartitionFilter(partitionCount, partitionIndex))
+                .With(JsonExporter.Full) // make sure we export to Json (for BenchView integration purpose)
+                .With(new PerfLabExporter())
+                .With(StatisticColumn.Median, StatisticColumn.Min, StatisticColumn.Max)
+                .With(TooManyTestCasesValidator.FailOnError)
+                .With(new UniqueArgumentsValidator()) // don't allow for duplicated arguments #404
+                .With(new MandatoryCategoryValidator(mandatoryCategories));
```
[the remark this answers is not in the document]
billwert: This is probably a better answer.

src/benchmarks/real-world/Roslyn/Program.cs
```diff
@@ -6,18 +6,37 @@
...
+using BenchmarkDotNet.Exporters.Json;
+using BenchmarkDotNet.Columns;
 
 namespace CompilerBenchmarks
 {
     public class Program
     {
+        private static IConfig CustomConfig(DirectoryInfo artifactsPath, ImmutableHashSet<string> mandatoryCategories, int? partitionCount = null, int? partitionIndex = null)
+            => DefaultConfig.Instance
+                .With(Job.Default.AsDefault()) // tell BDN that this are our default settings
```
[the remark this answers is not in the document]
agocke: The main problem is that I've reduced the parsing test in time, which means that the default settings doesn't seem to be getting very reliable results.

src/benchmarks/real-world/Roslyn/StageBenchmarks.cs
agocke: That probably makes sense. The simple answer here is: that's what we do in the compiler, so that's what I'm interested in measuring.

agocke:
I'll close this PR until I'm ready to make the suggested changes

PR #535, opened 2019-06-02
String Comparers benchmarks

adamsitnik:
In https://github.com/dotnet/corefx/issues/37691 it turned out that we had a huge difference for Windows vs Linux for `StringComparer.InvariantCultureIgnoreCase`.

I added the `GetHashCode` benchmark for https://github.com/dotnet/corefx/issues/37691 and `Compare` benchmarks for https://github.com/dotnet/corefx/issues/29462

/cc @danmosemsft @mjsabby

### DIFF
--- a/src/benchmarks/micro/corefx/System.Runtime/Perf.StringComparer.cs
+++ b/src/benchmarks/micro/corefx/System.Runtime/Perf.StringComparer.cs
@@ -18,7 +18,7 @@ namespace System.Tests
             StringComparison.InvariantCulture, StringComparison.InvariantCultureIgnoreCase)]
         public StringComparison Comparison { get; set; }
 
-        private string _input, _same, _lastCharacterDifferent;
+        private string _input, _same;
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/corefx/System.Runtime/Perf.StringComparer.cs
```diff
@@ -0,0 +1,60 @@
...
+using BenchmarkDotNet.Extensions;
+using MicroBenchmarks;
+
+namespace System.Tests
+{
+    [BenchmarkCategory(Categories.CoreFX, Categories.CoreCLR)]
+    public class Perf_StringComparer
+    {
+        [Params(
+            128, // stackalloc path
```
jorive: I like the covered cases :)

jorive: **Thanks for the comments too

PR #538, opened 2019-06-04
run Roslyn benchmarks for Ubuntu as well

adamsitnik:
The Roslyn benchmarks should work OOTB on Linux, it would be nice to have them to spot and fix any Windows/Linux perf differences.


adamsitnik:
The change was trivial and I want to get some results before I merge https://github.com/dotnet/coreclr/pull/24889 so I am merging it right now

PR #543, opened 2019-06-04
Json micro benchmarks related to comment parsing

krwq:
cc: @ahsonkhan 

<s>Do not merge until outerloop tests are available:
https://github.com/dotnet/performance/issues/545 (remove the limit change before merging)</s>
I have reduced number of test cases but still I'd  consider moving those to outerloop

Before (`SegmentSize = 0` => single segment):

|                       Method | CommentHandling | SegmentSize |        TestCase |        Mean |       Error |      StdDev |      Median |         Min |         Max | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|----------------------------- |---------------- |------------ |---------------- |------------:|------------:|------------:|------------:|------------:|------------:|------------:|------------:|------------:|--------------------:|
| Utf8JsonReaderCommentParsing |           Allow |           0 |  ShortMultiLine |    119.2 ns |   2.3512 ns |   2.3092 ns |    119.2 ns |    116.0 ns |    123.5 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |           0 |   LongMultiLine |  7,262.3 ns |  68.0316 ns |  60.3082 ns |  7,245.3 ns |  7,180.4 ns |  7,369.4 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |           0 | ShortSingleLine |    112.6 ns |   1.3544 ns |   1.2669 ns |    112.4 ns |    110.3 ns |    115.1 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |           0 |  LongSingleLine |    153.2 ns |   1.9873 ns |   1.7617 ns |    153.6 ns |    150.0 ns |    155.6 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |         100 |  ShortMultiLine |    144.3 ns |   2.5630 ns |   2.3974 ns |    144.3 ns |    141.0 ns |    148.4 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |         100 |   LongMultiLine | 26,204.4 ns | 406.7528 ns | 360.5758 ns | 26,132.8 ns | 25,629.4 ns | 27,043.3 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |         100 | ShortSingleLine |    136.6 ns |   2.3334 ns |   2.1827 ns |    136.7 ns |    133.5 ns |    141.3 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |         100 |  LongSingleLine |    949.6 ns |   8.7820 ns |   8.2147 ns |    948.8 ns |    936.8 ns |    966.4 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |           0 |  ShortMultiLine |    117.5 ns |   2.3642 ns |   2.4279 ns |    117.4 ns |    113.7 ns |    123.1 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |           0 |   LongMultiLine | 12,353.9 ns | 145.6122 ns | 129.0814 ns | 12,360.8 ns | 12,076.5 ns | 12,527.6 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |           0 | ShortSingleLine |    107.3 ns |   0.7509 ns |   0.7024 ns |    107.3 ns |    106.3 ns |    108.6 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |           0 |  LongSingleLine |    162.1 ns |   2.7758 ns |   2.4607 ns |    162.7 ns |    157.6 ns |    165.7 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |         100 |  ShortMultiLine |    141.5 ns |   2.0972 ns |   1.8591 ns |    142.0 ns |    138.4 ns |    143.8 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |         100 |   LongMultiLine | 18,561.6 ns | 297.7161 ns | 263.9175 ns | 18,545.6 ns | 18,219.1 ns | 19,066.3 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |         100 | ShortSingleLine |    131.8 ns |   1.4226 ns |   1.2611 ns |    132.0 ns |    129.5 ns |    133.8 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |         100 |  LongSingleLine |    908.4 ns |   7.5598 ns |   7.0714 ns |    908.5 ns |    892.0 ns |    918.7 ns |           - |           - |           - |                   - |

After:

|                       Method | CommentHandling | SegmentSize |        TestCase |        Mean |      Error |     StdDev |      Median |         Min |         Max | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|----------------------------- |---------------- |------------ |---------------- |------------:|-----------:|-----------:|------------:|------------:|------------:|------------:|------------:|------------:|--------------------:|
| Utf8JsonReaderCommentParsing |           Allow |           0 |  ShortMultiLine |    118.0 ns |   2.269 ns |   2.122 ns |    118.1 ns |    114.4 ns |    121.6 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |           0 |   LongMultiLine | 12,277.3 ns | 161.879 ns | 151.422 ns | 12,236.1 ns | 12,045.7 ns | 12,546.7 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |           0 | ShortSingleLine |    115.2 ns |   1.169 ns |   1.094 ns |    115.6 ns |    113.0 ns |    117.0 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |           0 |  LongSingleLine |    171.4 ns |   1.973 ns |   1.846 ns |    171.4 ns |    168.4 ns |    174.2 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |         100 |  ShortMultiLine |    143.7 ns |   1.690 ns |   1.581 ns |    143.6 ns |    141.7 ns |    146.8 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |         100 |   LongMultiLine | 13,962.4 ns | 133.750 ns | 125.110 ns | 13,964.7 ns | 13,736.6 ns | 14,189.3 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |         100 | ShortSingleLine |    138.8 ns |   2.723 ns |   2.675 ns |    138.1 ns |    135.3 ns |    144.4 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |           Allow |         100 |  LongSingleLine |  1,089.2 ns |  19.731 ns |  18.456 ns |  1,091.4 ns |  1,061.9 ns |  1,123.2 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |           0 |  ShortMultiLine |    122.5 ns |   2.196 ns |   1.714 ns |    123.1 ns |    119.7 ns |    125.2 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |           0 |   LongMultiLine |  7,276.8 ns |  94.828 ns |  84.062 ns |  7,268.8 ns |  7,178.5 ns |  7,481.4 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |           0 | ShortSingleLine |    119.1 ns |   1.275 ns |   1.130 ns |    118.6 ns |    117.9 ns |    121.6 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |           0 |  LongSingleLine |    171.7 ns |   1.728 ns |   1.616 ns |    172.2 ns |    168.0 ns |    173.5 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |         100 |  ShortMultiLine |    136.6 ns |   2.113 ns |   1.977 ns |    137.3 ns |    133.1 ns |    139.2 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |         100 |   LongMultiLine | 14,067.7 ns | 152.104 ns | 134.836 ns | 14,031.2 ns | 13,890.4 ns | 14,399.2 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |         100 | ShortSingleLine |    135.2 ns |   2.637 ns |   2.590 ns |    136.5 ns |    131.3 ns |    139.9 ns |           - |           - |           - |                   - |
| Utf8JsonReaderCommentParsing |            Skip |         100 |  LongSingleLine |  1,050.9 ns |  20.643 ns |  22.088 ns |  1,046.3 ns |  1,020.9 ns |  1,102.3 ns |           - |           - |           - |                   - |


src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
```diff
@@ -0,0 +1,283 @@
...
+                var state = new JsonReaderState(options: new JsonReaderOptions { CommentHandling = commentHandling });
+                var json = new Utf8JsonReader(data, isFinalBlock: true, state);
+
+                while (json.Read())
+                {
+                }
+            }
+        }
+
+        private static ReadOnlySequence<byte> GetSequence(byte[] dataUtf8, int segmentSize)
```
krwq: GetSequence and dependants (everything to the bottom of this file) are copied over from test helpers in corefx

ahsonkhan:
@krwq, can you share the performance results for these tests (just to establish a baseline)?
Consider commenting out the following to get more accurate results (even though they will take longer to run): https://github.com/dotnet/performance/blob/2e05afce5332d5e0a308e6208112b4205d3faf75/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L17-L20

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
```diff
@@ -0,0 +1,283 @@
...
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+using MicroBenchmarks.Serializers;
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace System.Text.Json.Serialization.Tests
```
ahsonkhan: The namespace needs to be fixed. Remove Serialization.

```suggestion
namespace System.Text.Json.Tests
```

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
```diff
@@ -0,0 +1,283 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
ahsonkhan: Can the file have comments in the name since all these tests are specific to payloads with comments?

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
krwq: oops, copy & paste fail

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+        private byte[] s_singleShortSingleLineComment;
+        private byte[] s_multipleShortSingleLineComment;
+        private byte[] s_singleLongSingleLineComment;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            // every test case ends with {} so that we can re-use it for both Skip and Allow options
+            s_singleEmptyMultiLineComment = Encoding.UTF8.GetBytes("/**/{}");
+            s_multipleEmptyMultiLineComment = Encoding.UTF8.GetBytes("/**//**//**//**//**//**//**//**//**//**//**//**//**//**//**//**//**//**/{}");
```
ahsonkhan: Is it necessary to test both single-line and multi-line comments? Shouldn't the perf characteristics of both generally be the same?

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+            s_multipleEmptySingleLineComment = Encoding.UTF8.GetBytes("//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n{}");
+            s_singleShortSingleLineComment = Encoding.UTF8.GetBytes("//asdasd\n{}");
+            s_multipleShortSingleLineComment = Encoding.UTF8.GetBytes("//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n{}");
+            s_singleLongSingleLineComment = Encoding.UTF8.GetBytes("//" + new string('c', 2000) + "\n{}");
+
+            MultiSegmentSetup();
+        }
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void SingleEmptyMultiLineComment_CommentHandling_Allow() => GenericReadTest(s_singleEmptyMultiLineComment, JsonCommentHandling.Allow);
```
ahsonkhan: The comment handling enum (allow and skip) can be Params instead, to avoid code duplication.

Maybe something like:
```C#
[Params(JsonCommentHandling.Skip, JsonCommentHandling.Allow)]
public JsonCommentHandling CommentHandling;
```

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void MultipleShortSingleLineComment_CommentHandling_Skip() => GenericReadTest(s_multipleShortSingleLineComment, JsonCommentHandling.Skip);
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void SingleLongSingleLineComment_CommentHandling_Skip() => GenericReadTest(s_singleLongSingleLineComment, JsonCommentHandling.Skip);
+
+        private static void GenericReadTest(byte[] dataUtf8, JsonCommentHandling commentHandling)
+        {
+            for (int i = 0; i < Iterations; i++)
```
ahsonkhan: I would remove `Iterations` here and also remove `OperationsPerInvoke`. They are not needed for a test like this.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+            s_singleShortMultiLineComment = Encoding.UTF8.GetBytes("/*asdasd*/{}");
+            s_multipleShortMultiLineComment = Encoding.UTF8.GetBytes("/*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*/{}");
+            s_singleLongMultiLineComment1 = Encoding.UTF8.GetBytes("/*" + new string('c', 2000) + "*/{}");
+            s_singleLongMultiLineComment2 = Encoding.UTF8.GetBytes("/*" + new string('\n', 2000) + "*/{}");
+
+            s_singleEmptySingleLineComment = Encoding.UTF8.GetBytes("//\n{}");
+            s_multipleEmptySingleLineComment = Encoding.UTF8.GetBytes("//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n{}");
+            s_singleShortSingleLineComment = Encoding.UTF8.GetBytes("//asdasd\n{}");
+            s_multipleShortSingleLineComment = Encoding.UTF8.GetBytes("//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n//asdasd\n{}");
+            s_singleLongSingleLineComment = Encoding.UTF8.GetBytes("//" + new string('c', 2000) + "\n{}");
```
ahsonkhan: nit: Let's move the comment to be after the payload (rather than being the first token). The same applies to other test cases.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+            // every test case ends with {} so that we can re-use it for both Skip and Allow options
+            s_singleEmptyMultiLineComment = Encoding.UTF8.GetBytes("/**/{}");
+            s_multipleEmptyMultiLineComment = Encoding.UTF8.GetBytes("/**//**//**//**//**//**//**//**//**//**//**//**//**//**//**//**//**//**/{}");
+            s_singleShortMultiLineComment = Encoding.UTF8.GetBytes("/*asdasd*/{}");
+            s_multipleShortMultiLineComment = Encoding.UTF8.GetBytes("/*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*//*asdasd*/{}");
+            s_singleLongMultiLineComment1 = Encoding.UTF8.GetBytes("/*" + new string('c', 2000) + "*/{}");
+            s_singleLongMultiLineComment2 = Encoding.UTF8.GetBytes("/*" + new string('\n', 2000) + "*/{}");
+
+            s_singleEmptySingleLineComment = Encoding.UTF8.GetBytes("//\n{}");
+            s_multipleEmptySingleLineComment = Encoding.UTF8.GetBytes("//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n//\n{}");
```
ahsonkhan: I think an array of JSON elements with comments in between is a better test case than lots of comments together (particularly for skip), since that ends up creating a tight loop while reading/skipping comments which is unlikely to happen for any realistic payload.

So, something like:
`[1//\n,2//\n,3//\n,...]`

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Threading.Tasks;
+
+namespace System.Text.Json.Serialization.Tests
+{
+    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
+    public partial class Utf8JsonReaderTests
+    {
+        private const int Iterations = 1_000_000;
+
+        private byte[] s_singleEmptyMultiLineComment;
```
ahsonkhan: nit: It might be better to have a TestCase enum type to capture all the different test cases to make the results and test cases more readable.

Something like the following:
https://github.com/dotnet/corefxlab/blob/4cefa703d08c2ddc8f950d9ca211cfde06c8f484/tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs#L47-L53

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+        private byte[] s_multipleEmptySingleLineComment;
+        private byte[] s_singleShortSingleLineComment;
+        private byte[] s_multipleShortSingleLineComment;
+        private byte[] s_singleLongSingleLineComment;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            // every test case ends with {} so that we can re-use it for both Skip and Allow options
+            s_singleEmptyMultiLineComment = Encoding.UTF8.GetBytes("/**/{}");
```
ahsonkhan: I don't know how useful this test case is, especially compared to the other test case which has a comment which contains some characters. The only reason why it would matter is if we were comparing the overhead between a payload with comments (and mode == skip) and one with no comments at all (and mode == disallow). But, we are not doing that comparison in these tests.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+        public void SingleLongMultiLineComment1_CommentHandling_Allow() => GenericReadTest(s_singleLongMultiLineComment1, JsonCommentHandling.Allow);
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void SingleLongMultiLineComment2_CommentHandling_Allow() => GenericReadTest(s_singleLongMultiLineComment2, JsonCommentHandling.Allow);
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void SingleEmptySingleLineComment_CommentHandling_Allow() => GenericReadTest(s_singleEmptySingleLineComment, JsonCommentHandling.Allow);
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void MultipleEmptySingleLineComment_CommentHandling_Allow() => GenericReadTest(s_multipleEmptySingleLineComment, JsonCommentHandling.Allow);
```
ahsonkhan: The test case name is capturing too many concepts that it is hard to parse. I would have trouble reasoning about the results as well.

Let's break it down `MultipleEmptySingleLineComment_CommentHandling_Allow`:
- Multiple - This is the count of how many comments show up (Single vs Multiple) - Maybe count params?
- Empty - So, empty is the talking about the contents of the comment itself, vs short/long.
- SingleLine - OK, so this is multi-line comment vs single line (/**/ vs //\n), which I think we can live without OR add a - - bool params to say singleline true/false)
- Allow - this can be fixed with a CommentHandling params

We should figure out a cleaner way to represent such test cases.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace System.Text.Json.Serialization.Tests
+{
+    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
+    public partial class Utf8JsonReaderTests
+    {
+        private const int Iterations = 1_000_000;
```
ahsonkhan: Get rid of `Iterations`.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
```diff
@@ -0,0 +1,132 @@
...
+using MicroBenchmarks.Serializers;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace System.Text.Json.Serialization.Tests
+{
+    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
+    public partial class Utf8JsonReaderTests
```
ahsonkhan: I would update the class name to indicate this is specifically about comments (similar to the file name change).
Maybe: Utf8JsonReader_Comments

```suggestion
    public partial class Utf8JsonReader_Comments
```

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
```diff
@@ -0,0 +1,283 @@
...
+
+            return CreateReadOnlySequence(buffers);
+        }
+
+        private static ReadOnlySequence<byte> CreateReadOnlySequence(params byte[][] buffers)
+        {
+            if (buffers.Length == 1)
+                return new ReadOnlySequence<byte>(buffers[0]);
+            var list = new List<Memory<byte>>();
+            foreach (var buffer in buffers)
```
ahsonkhan: nit: don't use var here

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
```diff
@@ -0,0 +1,283 @@
...
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void MultiSegment_MultipleShortSingleLineCommentSequence100_CommentHandling_Skip() => MultiSegment_GenericReadTest(s_multipleShortSingleLineCommentSequence100, JsonCommentHandling.Skip);
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void MultiSegment_SingleLongSingleLineCommentSequence100_CommentHandling_Skip() => MultiSegment_GenericReadTest(s_singleLongSingleLineCommentSequence100, JsonCommentHandling.Skip);
+
+        private static void MultiSegment_GenericReadTest(ReadOnlySequence<byte> data, JsonCommentHandling commentHandling)
+        {
+            for (int i = 0; i < Iterations; i++)
```
ahsonkhan: A lot of the single segment feedback applies here  (for multi-segment) as well.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
```diff
@@ -0,0 +1,283 @@
...
+        public void MultiSegment_SingleLongMultiLineComment1Sequence100_CommentHandling_Skip() => MultiSegment_GenericReadTest(s_singleLongMultiLineComment1Sequence100, JsonCommentHandling.Skip);
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void MultiSegment_SingleLongMultiLineComment2Sequence100_CommentHandling_Skip() => MultiSegment_GenericReadTest(s_singleLongMultiLineComment2Sequence100, JsonCommentHandling.Skip);
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void MultiSegment_SingleEmptySingleLineCommentSequence100_CommentHandling_Skip() => MultiSegment_GenericReadTest(s_singleEmptySingleLineCommentSequence100, JsonCommentHandling.Skip);
+
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void MultiSegment_MultipleEmptySingleLineCommentSequence100_CommentHandling_Skip() => MultiSegment_GenericReadTest(s_multipleEmptySingleLineCommentSequence100, JsonCommentHandling.Skip);
```
ahsonkhan: Ah... one more dimension/concept (`MultiSegment`).

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
```diff
@@ -0,0 +1,283 @@
...
+            s_singleLongSingleLineCommentSequence1 = GetSequence(s_singleLongSingleLineComment, 1);
+
+            s_singleEmptyMultiLineCommentSequence100 = GetSequence(s_singleEmptyMultiLineComment, 100);
+            s_multipleEmptyMultiLineCommentSequence100 = GetSequence(s_multipleEmptyMultiLineComment, 100);
+            s_singleShortMultiLineCommentSequence100 = GetSequence(s_singleShortMultiLineComment, 100);
+            s_multipleShortMultiLineCommentSequence100 = GetSequence(s_multipleShortMultiLineComment, 100);
+            s_singleLongMultiLineComment1Sequence100 = GetSequence(s_singleLongMultiLineComment1, 100);
+            s_singleLongMultiLineComment2Sequence100 = GetSequence(s_singleLongMultiLineComment2, 100);
+
+            s_singleEmptySingleLineCommentSequence100 = GetSequence(s_singleEmptySingleLineComment, 100);
```
ahsonkhan: What are we getting from the 1 vs 100 segment split? Should we instead be more intentional and just pick the location where the split occurs (i.e. within a comment rather than somewhere outside of it)?

ahsonkhan:
cc @adamsitnik, @billwer

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
krwq: no, multi-segment has a bit more checks - it's essentially different code

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
krwq: does it make any difference except for visual?

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
krwq: this is copy paste of corefx code

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
krwq: for me interesting cases are:
- many segments, segment iteration perf
- 1 or few segments

I think 1 and 100 achieve this reasonably well (this could use Params though which I didn't know it existed)

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
ahsonkhan: > no, multi-segment has a bit more checks and logic - it's essentially different code

I was asking about single-line and multi-line **comments**, not multi-segment. Does your statement still hold true?

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
ahsonkhan: The reader special cases the "first" token, and although generally falls down the same code path, I want to avoid that special code path.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
ahsonkhan: Yep, I know. It was an oversight there too and we should fix it (also was in test code). It was likely missed in review.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.MultiSegment.Tests.cs
ahsonkhan: > many segments, segment iteration perf

I don't see measuring the perf of many small-sized segments being relevant at all. If we were to regress this case in order to optimize something else (like say reasonably sized chunks of ~4K get faster), I would make that trade off. Hence, us tracking and worrying about the perf of small segments seems unnecessary to me and send the wrong signal. We shouldn't be optimizing for this case.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
krwq: @ahsonkhan still different code - it's possible to regress one but not another

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.Comments.Tests.cs
```diff
@@ -0,0 +1,156 @@
...
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+using MicroBenchmarks.Serializers;
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
```
ahsonkhan: Do we need this using directive? Why?

krwq:
@jorive yo can remove the `*NO MERGE*` now :)

krwq:
@ahsonkhan perf results are mostly the same and the only regression which I see is one piece of code I haven't changed (single segment multi-line comment) which I'm not sure what to make of (I have build corefx from same commit hash but on different days, presumably there is no difference related to that?)

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.SingleSegment.Tests.cs
krwq: removed empty comments tests

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.Comments.Tests.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        // 0 => single segment
+        [Params(0, 100)]
+        public int SegmentSize;
+
+        private bool MultiSegment => SegmentSize != 0;
+        private byte[] _jsonPayload;
+        private ReadOnlySequence<byte> _jsonPayloadSequence;
+
+        [ParamsSource(nameof(TestCaseValues))]
```
adamsitnik: just a hint for future: BenchmarkDotNet has a `[ParamsAllValues]` attribute that expands given `[Params]` property with all enum values

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Utf8JsonReader.Comments.Tests.cs
```diff
@@ -0,0 +1,138 @@
...
+    public partial class Utf8JsonReaderCommentsTests
+    {
+        [Params(JsonCommentHandling.Skip, JsonCommentHandling.Allow)]
+        public JsonCommentHandling CommentHandling;
+
+        // 0 => single segment
+        [Params(0, 100)]
+        public int SegmentSize;
+
+        private bool MultiSegment => SegmentSize != 0;
```
ahsonkhan: I think this bool is what we should have as the Params (true for single segment, false for multisegment), rather than SegmentSize.

1) SegmentSize is confusing since 0 is supposed to mean single segment (but the person reviewing the perf results wouldn't know that).
2) SegmentSize of 100 is inaccurate since based on the payload, we won't have 100 segments.
3) It's only used in GobalSetup, whereas the bool is used in the actual test.

```C#
[Params(true, false)]
public bool SingleSegment;

[GlobalSetup]
public void Setup()
{
   _jsonPayload = Encoding.UTF8.GetBytes(s_testCases[TestCase]);
   _jsonPayloadSequence = GetSequence(_jsonPayload, 100);
}
```

That way, the perf results are more meaningful, imo (and doesn't require additional mapping to various concepts which require looking at perf test implementation).

krwq: For me interesting optimization point is also SegementSize=1, with current state I can quickly add that case

ahsonkhan: If we stick to the test we have atm, having `SegmentSize` as a column in the test result doesn't make sense. Having SingleSegment/MultiSegment as a column makes more sense.

Alternatively, we add two separate tests - one for single segment, and one for multi-segment.

> For me interesting optimization point is also SegementSize=1

In reality, we should only optimize for segment/buffer sizes of 4k/8k, and not 1-100 byte segments. I don't see much value in measuring the perf, and optimizing for, any unreasonable segment sizes.

PR #548, opened 2019-06-06
update JsonSerializer.WriteAsync API usage (the order of arguments has changed)

adamsitnik:


adamsitnik:
Since the build is red due to this change I am going to merge it now (all the checks passed)

obraz

PR #551, opened 2019-06-08
Modify Utf8JsonReader Comments perf test (reduce test name size and make output clearer).

ahsonkhan:
From https://github.com/dotnet/performance/pull/543#discussion_r291728581

Primarily to make the output more reasonable to read and to remove the redundant `Utf8JsonReader` within the fully qualified test name (especially for the result comparer view - https://github.com/dotnet/corefx/pull/38036#issuecomment-500092168):
`System.Text.Json.Tests.Utf8JsonReaderCommentsTests.Utf8JsonReaderCommentParsing`

Also, trying to be consistent with the other perf test format.

cc @krwq, @adamsitnik

### DIFF
--- a/src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
+++ b/src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
@@ -11,16 +11,14 @@ using System.Linq;
 namespace System.Text.Json.Tests
 {
     [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
-    public partial class Utf8JsonReaderCommentsTests
+    public partial class Utf8JsonReader_Comments
     {
         [Params(JsonCommentHandling.Skip, JsonCommentHandling.Allow)]
         public JsonCommentHandling CommentHandling;
 
-        // 0 => single segment
-        [Params(0, 100)]
-        public int SegmentSize;
+        [Params(true, false)]
+        public bool SingleSegment;
 
-        private bool MultiSegment => SegmentSize != 0;
         private byte[] _jsonPayload;
         private ReadOnlySequence<byte> _jsonPayloadSequence;
 
@@ -41,18 +39,14 @@ namespace System.Text.Json.Tests
         public void Setup()
         {
             _jsonPayload = Encoding.UTF8.GetBytes(s_testCases[TestCase]);
-
-            if (MultiSegment)
-            {
-                _jsonPayloadSequence = GetSequence(_jsonPayload, SegmentSize);
-            }
+            _jsonPayloadSequence = GetSequence(_jsonPayload, 100);
         }
 
         [Benchmark]
-        public void Utf8JsonReaderCommentParsing()
+        public void EmptyLoop()
         {
             var state = new JsonReaderState(options: new JsonReaderOptions { CommentHandling = CommentHandling });
-            Utf8JsonReader reader = MultiSegment ?
+            Utf8JsonReader reader = !SingleSegment ?
                 new Utf8JsonReader(_jsonPayloadSequence, isFinalBlock: true, state) :
                 new Utf8JsonReader(_jsonPayload, isFinalBlock: true, state);
 


src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
```diff
@@ -41,18 +39,14 @@ public partial class Utf8JsonReaderCommentsTests
         public void Setup()
         {
             _jsonPayload = Encoding.UTF8.GetBytes(s_testCases[TestCase]);
-
-            if (MultiSegment)
-            {
-                _jsonPayloadSequence = GetSequence(_jsonPayload, SegmentSize);
-            }
+            _jsonPayloadSequence = GetSequence(_jsonPayload, 100);
```
krwq: you've mentioned 4096 would be a better test

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
```diff
@@ -41,18 +39,14 @@ public partial class Utf8JsonReaderCommentsTests
...
         }
 
         [Benchmark]
-        public void Utf8JsonReaderCommentParsing()
+        public void EmptyLoop()
```
krwq: ConsumeAll perhaps

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
```diff
@@ -41,18 +39,14 @@ public partial class Utf8JsonReaderCommentsTests
...
         {
             var state = new JsonReaderState(options: new JsonReaderOptions { CommentHandling = CommentHandling });
-            Utf8JsonReader reader = MultiSegment ?
+            Utf8JsonReader reader = !SingleSegment ?
```
krwq: You an as well swap expressions and remove `!`

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
```diff
@@ -11,16 +11,14 @@
 namespace System.Text.Json.Tests
 {
     [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
-    public partial class Utf8JsonReaderCommentsTests
+    public partial class Utf8JsonReader_Comments
```
krwq: Maybe would be better to put Utf8JsonReader somewhere in the namespace and making a class name just `CommentsTests`

ahsonkhan: I thought about that, but then the test namespace doesn't follow the API namespace structure. I believe I was told that the pattern is `<API name space>.Tests`.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
ahsonkhan: Maybe. However, we don't really "consume". I would think consume to mean observing each token (either via `reader.ValueSpan` or reader.GetString/GetInt/etc.)

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
ahsonkhan: We should re-consider the test, sure, but I am primarily making a cosmetic change here.

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
ahsonkhan: Will do.

krwq:
I think it was my antivirus before which was causing the weird perf results :man_facepalming:

src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Comments.cs
krwq: for me, reading is consuming (also matches naming convention used in the product)

billwert:
@ahsonkhan @adamsitnik @krwq is this ready?

krwq:
nothing blocking on my side, would be nice if we've addressed the feedback though. @ahsonkhan, have time to address it?

ahsonkhan:
I'll try to do an update today before I leave. Thanks for the reminder.

adamsitnik:
@ahsonkhan ping

PR #561, opened 2019-06-12
Disable tiered runs

billwert:
We discussed this a while ago, finally getting to it.

### DIFF
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -106,14 +106,6 @@ jobs:
                   _PerfLabArguments: $(PerfLabArguments)
                 ${{ if not(startsWith(framework, 'netcoreapp')) }}: # we don't want to upload non-.NET Core results to BenchView
                   _BenchViewArguments: ''
-              ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}: # to reduce the number of CI legs for every PR
-                ${{ if startsWith(framework, 'netcoreapp') }}: # Tiered JIT was introduced in .NET Core 2.1
-                  ${{ framework }}_tiered:
-                    _Framework: ${{ framework }}
-                    _Configs: CompilationMode=$(_CompilationMode)
-                    _CompilationMode: Tiered
-                    _BuildConfig: ${{ parameters.architecture }}_$(_Framework)_$(_CompilationMode)_${{ parameters.kind }}
-                    _BenchViewArguments: $(BenchViewArguments)
         steps:
         - checkout: self
           clean: true
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -98,8 +98,7 @@ jobs:
             ${{ each framework in parameters.frameworks }}:
               ${{ framework }}:
                 _Framework: ${{ framework }}
-                _CompilationMode: NoTiering
-                _Configs: CompilationMode=$(_CompilationMode)
+                _Configs: CompilationMode=Tiered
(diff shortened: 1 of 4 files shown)


billwert:
the goal here is to end up with a "default" mode that doesn't do anything other than how the product comes. This is how we'll operate going forward - we'll never touch this default, and if we add configs we'll just differentiate them from the default.

I left 'Tiered' as the config name for now to keep consistency on benchview until we shut it down and transition folks to looking at the new stuff we have. We can clean this up in the DB after the fact.

scripts/dotnet.py
```diff
@@ -588,7 +593,7 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         required=False,
         action=CompilationAction,
         choices=CompilationAction.modes(),
-        default=CompilationAction.tiered(),
+        default=CompilationAction.noenv(),
```
billwert: used this name so as not to conflict with `default` on the base class `Action`

eng/performance/benchmark_jobs.yml
```diff
@@ -98,22 +98,13 @@ jobs:
             ${{ each framework in parameters.frameworks }}:
               ${{ framework }}:
                 _Framework: ${{ framework }}
-                _CompilationMode: NoTiering
-                _Configs: CompilationMode=$(_CompilationMode)
+                _Configs: CompilationMode=Tiered
```
jorive: Did you mean `Default` here?

billwert: see comment in the PR conversation.

jorive: I guess a better question is: What's the configuration generated by the scripts after changing the command line arguments below? I think that removing `--dotnet-compilation-mode` will generate a new trace because the new compilation mode is `Default`.
Maybe I'm missing something here, but I believe the changes will generate a trace that is: `.NET Compilation Mode=[Default]`

billwert: Ah, good point. I forgot that we didn't update what goes into the Benchview scripts yet to use the new variable. Let me think about this - maybe this is where we just pull the plug on Benchview. (It's happening in a couple weeks anyway.)

jorive: If that's the case, we should remove all BenchView related scripts/commands/logic (it would be a good cleanup).

adamsitnik: > maybe this is where we just pull the plug on Benchview. (It's happening in a couple weeks anyway.)

We are now very close to the release date and I am using BenchView daily to validate that our fixes actually fix perf problems.

Could you please pull the plug after the 3.0 release or after we switch to the new reporting system?

billwert: Of course. We'll transition smoothly, don't worry.

billwert: For now I'm just going to hard code the value for benchview as well - I want to get to turning off the extra run we no longer need, and want to get to where we're collecting a baseline set of data without any `COMPlus` variables set. Removing upload to benchview can be another step.

scripts/benchview.py
```diff
@@ -400,7 +400,7 @@ def __run_scripts(
     #   This value should be optional, and set when applicable.
     # benchview_config['Jit'] = 'RyuJIT'
     benchview_config['Framework'] = framework
-    benchview_config['.NET Compilation Mode'] = args.dotnet_compilation_mode
+    benchview_config['.NET Compilation Mode'] = 'Tiered'
```
jorive: Instead of hardcoding, wouldn't it make more sense to delay this particular change? OR hardcode it on the yml file?

jorive: I think the latter makes more sense.

billwert: I do not see a meaningful difference between the two. We already hardcode other values here. This code is going to be gone in a few weeks.

PR #566, opened 2019-06-16
make sure all StringComparison are tested, remove one Count permutation to not reach the test cases limit

adamsitnik:
I realized that when I added the `StringComparison` benchmarks I did not add the test cases for `StringComparison.CurrentCulture` and `StringComparison.CurrentCultureIgnoreCase`.

To make sure that we test all of them I have used `[ParamsAllValues]` which is a BDN way of saying generate Params for all values of a given enum.

Our limit for test cases is 16, with only this change we would have 18.

In order to reduce the number of test cases, I have removed the `Count=2 * 1024 * 1024` test case. So we have 12 instead of 18 test cases.


adamsitnik:
It should be already green...

obraz

PR #574, opened 2019-06-18
Generalizeyml

adiaaida:


scripts/performance/common.py
```diff
@@ -150,9 +150,10 @@ def verbose(self) -> bool:
         '''Enables/Disables verbosity.'''
         return self.__verbose
 
-    def run(self, working_directory: str = None) -> None:
+    def run(self, working_directory: str = None) -> str:
         '''Executes specified shell command.'''
         should_pipe = self.verbose
+        output = ''
```
jorive: nit: can make `output` a list (`output = []`), and join at the end of the function.

adiaaida: This is a work in progress and not ready for review. I am just testing things at the moment.

PR #576, opened 2019-06-19
Generalize scripting to prepare for cross-repo use

adiaaida:
This change does the following:

1. Adds a --get-perf-hash, which will get the hash of the performance
repo where the script running resides. This will be used when running in
the product repos. This will cause us to run git rev-parse HEAD to get
the hash.
2. Changes the config string to not add quotes for windows (quotes are
needed in bash, but not cmd. Currently, windows configs are being added
as key: "CompilationMode; value: Tiered". We'll need to fix it in the
database.

### DIFF
--- a/scripts/ci_setup.py
+++ b/scripts/ci_setup.py
@@ -6,9 +6,11 @@ from logging import getLogger
 import os
 import sys
 
+from subprocess import check_output
+
 from performance.common import get_repo_root_path
 from performance.common import get_tools_directory
-from performance.common import RunCommand
+from performance.common import push_dir
 from performance.common import validate_supported_runtime
 from performance.logger import setup_loggers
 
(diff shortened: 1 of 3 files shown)


scripts/ci_setup.py
```diff
@@ -6,7 +6,9 @@
 import os
 import sys
 
+from performance.common import get_repo_root_path
 from performance.common import get_tools_directory
+from performance.common import RunCommand
```
jorive: instead of using `RunCommand`, I will recommend using `from subprocess import check_output`. Dotnet.py uses if to parse output. Modifying RunCommand will affect other instances where buffering for parsing is not needed.

adiaaida: will do

PR #587, opened 2019-06-25
Revert "Change the artifacts directory to be HELIX_WORKITEM_UPLOAD_ROOT"

adiaaida:
Reverts dotnet/performance#585

Unfortunately, the Os Onboarding rollout does not automatically rollout to on-prem machines, so we are now failing our internal jobs. https://github.com/dotnet/core-eng/issues/6801 tracks updating the perf machines. When that goes in, we will check that back in.


adiaaida:
fyi @billwert

PR #588, opened 2019-06-25
Fix duplication problems in the uploader

billwert:
HELIX_CORRELATION_ID can cause collisions when tests live on more than one partition.

Swapping to HELIX_WORKITEM_ID should fix this, as each machine gets its own
work item ID.


adiaaida:
So the reason we were using correlation id was because that's what's used to form the results container. We may want to think a little harder on this before making this kind of change. Probably actually need to update the database schema.

PR #590, opened 2019-06-26
Use BuildConfig in the name of the work item

adiaaida:
Currently, we set the HelixType to be _BuildConfig. This is messy, and causes us to have a bunch of MC.net pages to go look at. Instead, we can name the test runs according to _BuildConfig, and have everything in mc.net point to test/performance/. It is much neater. See: https://mc.dot.net/#/user/dotnet-performance/pr~2Fdotnet~2Fperformance~2Frefs~2Fpull~2F590~2Fmerge/test~2Fperformance~2F/20190626.2

### DIFF
--- a/eng/performance/helix.proj
+++ b/eng/performance/helix.proj
@@ -26,11 +26,11 @@
     <PartitionCount>5</PartitionCount>
   </PropertyGroup>
   <ItemGroup>
-    <Partition Include="Partition0" Index="0" />
-    <Partition Include="Partition1" Index="1" />
-    <Partition Include="Partition2" Index="2" />
-    <Partition Include="Partition3" Index="3" />
-    <Partition Include="Partition4" Index="4" />
+    <Partition Include="$(_BuildConfig).Partition0" Index="0" />
+    <Partition Include="$(_BuildConfig).Partition1" Index="1" />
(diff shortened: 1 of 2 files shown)


eng/performance/helix.proj
```diff
@@ -44,7 +44,7 @@
     </HelixWorkItem>
   </ItemGroup>
   <ItemGroup Condition="!$(TargetCsproj.Contains('MicroBenchmarks.csproj')) Or '$(Architecture)' == 'arm64'">
-    <HelixWorkItem Include="WorkItem">
+    <HelixWorkItem Include="$(_BuildConfig).WorkItem">
```
billwert: now that I look at this, can we just omit the Include on this tag? We don't reference it like we do Partition in the other branch.

adiaaida: The Include is how the HelixWorkItem is named. For some reason, the ml and roslyn runs aren't showing up, though, so I need to look into this further.

billwert: Aha. It isn't clear just from the code that this is how Helix chooses the name entirely. Would have expected `Name` to be a property on `HelixWorkItem`. Thanks for clarifying.

adiaaida:
So it turns out doing this still overwrites stuff. Closing, alas.

PR #597, opened 2019-06-28
Update JsonSerializer benchmarks

adiaaida:
JsonSerializer API changed.

### DIFF
--- a/src/benchmarks/micro/corefx/System.Text.Json/Serializer/WriteJson.cs
+++ b/src/benchmarks/micro/corefx/System.Text.Json/Serializer/WriteJson.cs
@@ -26,23 +26,23 @@ namespace System.Text.Json.Serialization.Tests
             _value = DataGenerator.Generate<T>();
 
             _memoryStream = new MemoryStream(capacity: short.MaxValue);
-            await JsonSerializer.WriteAsync(_memoryStream, _value);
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/corefx/System.Text.Json/Serializer/WriteJson.cs
```diff
@@ -26,23 +26,23 @@ public async Task Setup()
...
 
             _memoryStream = new MemoryStream(capacity: short.MaxValue);
-            await JsonSerializer.WriteAsync(_memoryStream, _value);
+            await JsonSerializer.SerializeAsync(_memoryStream, _value);
         }
 
         [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
         [Benchmark]
-        public string SerializeToString() => JsonSerializer.ToString(_value);
+        public string Serialize() => JsonSerializer.Serialize(_value);
```
billwert: Please don't change the benchmark name (it will break the trend.) I know it's sort of confusing, but the name `SerializeToString` is still the logical operation so I think it's okay.

adiaaida: Oh sorry!

PR #610, opened 2019-07-03
Add perf tests for BitArray

BruceForstall:
Add tests for And, Or, Xor, LeftShift, RightShift

### DIFF
--- a/src/benchmarks/micro/corefx/System.Collections/Perf.BitArray.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Perf.BitArray.cs
@@ -53,7 +53,7 @@ namespace System.Collections.Tests
         [Benchmark]
         public BitArray BitArrayIntArrayCtor() => new BitArray(_ints);
 
-        [GlobalSetup(Targets = new [] { nameof(BitArraySetAll), nameof(BitArrayNot), nameof(BitArrayGet), nameof(BitArrayRightShift), nameof(BitArrayLeftShift) })]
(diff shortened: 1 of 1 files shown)


BruceForstall:
@adamsitnik Looks like net461 doesn't have LeftShift/RightShift
```
[2019/07/03 18:53:04][INFO] corefx\System.Collections\Perf.BitArray.cs(78,55): error CS1061: 'BitArray' does not contain a definition for 'RightShift' and no accessible extension method 'RightShift' accepting a first argument of type 'BitArray' could be found (are you missing a using directive or an assembly reference?) [C:\dotnetbuild\work\42f98a9c-fd7b-4660-bff6-2169a4c04152\Payload\src\benchmarks\micro\MicroBenchmarks.csproj]
[2019/07/03 18:53:04][INFO] corefx\System.Collections\Perf.BitArray.cs(81,54): error CS1061: 'BitArray' does not contain a definition for 'LeftShift' and no accessible extension method 'LeftShift' accepting a first argument of type 'BitArray' could be found (are you missing a using directive or an assembly reference?) [C:\dotnetbuild\work\42f98a9c-fd7b-4660-bff6-2169a4c04152\Payload\src\benchmarks\micro\MicroBenchmarks.csproj]
```
It's available on Core from 2.0 onwards (https://docs.microsoft.com/en-us/dotnet/api/system.collections.bitarray.leftshift)

Is there a way to conditionalize based on that?

I actually didn't care much about those, since I wasn't working on them, but just noticed they were missing a test. So I could just remove them.

BruceForstall:
Ok I found `#if !NETFRAMEWORK`

PR #611, opened 2019-07-03
Add benchmarks for predicate Linq methods

jvandertil:
Added benchmarks for FirstOrDefault, Single, SingleOrDefault, Last,
LastOrDefault, and Any methods that accept a predicate, compares
them with chaining calls the same methods without a predicate and a call
to Where.

### DIFF
--- a/src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
+++ b/src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
@@ -146,6 +146,7 @@ namespace System.Linq.Tests
         public int LastOrDefaultWithPredicate(int size, int iterationCount, Perf_LinqTestBase.WrapperType wrapType)
             => Perf_LinqTestBase.Wrap(_sizeToPreallocatedArray[size], wrapType).LastOrDefault(x => x % 2 == 0);
 
+        [Benchmark]
         [ArgumentsSource(nameof(IterationSizeWrapperData))] // for some reason the size and iteration arguments are ignored for this benchmark
         public void Cast_ToBaseClass(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType)
         {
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
```diff
@@ -230,7 +281,7 @@ public bool Contains_ElementNotFound(int size, int iterationCount, Perf_LinqTest
 
         [Benchmark]
         [ArgumentsSource(nameof(IterationSizeWrapperData))]
-        public bool Contains_FirstElementMatches(int size, int iterationCount, Perf_LinqTestBase.WrapperType wrapType)
+        public bool Contains_FirstElementMatches(int size, int iteration, Perf_LinqTestBase.WrapperType wrapType)
```
billwert: Changing parameter names breaks test identity in our reporting system - please revert all name changes.

jvandertil: That is unfortunate, I have reverted the commit. Is the name `iteration` or `iterationCount` preferred for the new tests?

jvandertil:
These benchmarks will not hit code that has been special cased for `List<T>`, unfortunately the wrapping is part of the benchmark timing so wrapping things into `List<T>` does not seem possible at this time (since the entire array will first be copied into the `List`). It could be done by rewriting the benchmarks to accept a `IEnumerable<T>` and do the wrapping in the `IterationSizeWrapperData` method.

src/benchmarks/micro/corefx/System.Linq/Perf.Linq.cs
billwert: consistency is the strong preference, so in this case I would use `iterationCount` for new tests on this type / in this file.

jvandertil: Alright, that was already the case. Thanks!

adamsitnik:
I can see that this commit https://github.com/dotnet/performance/pull/611/commits/b8d33aa48b54cb9340c829e64ea264eb496ed987 was the change requested by @billwert so I am going to merge this PR now

PR #620, opened 2019-07-08
Microbenchmark Design Guidelines

### LINKED ISSUE #396: Create docs for benchmarking best practices and guidelines
As discussed with @stephentoub in #381 we simply need to write the knowledge down and publish it in a form of docs which should help us to ensure high quality and consistency.

adamsitnik:
I have finally found some time to write the docs.

Direct link to user-friendly markdown file: https://github.com/adamsitnik/performance/blob/benchmarkingBestPractices/docs/microbenchmark-design-guidelines.md

@jorive @billwert please take a look

/cc @danmosemsft 

Fixes #396


docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+## Mindset
+
+Writing Benchmarks is much different than writing Unit Tests. So before you start coding, you need to change your mindset first.
+
+### Benchmarks are not Unit Tests
+
+When writing Unit Tests, we ideally want to test all methods and properties of the given type. We also test both the happy and unhappy paths. The result of every Unit Test run is a single value: passed or failed.
+
+Benchmarks are different. First of all, the result of a benchmark run is never a single value. It's a whole distribution, described with values like mean, standard deviation, min, max and so on. To get a meaningful distribution, the benchmark has to be executed many, many times. **This takes a lot of time**. With the current [recommended settings](https://github.com/dotnet/performance/blob/51d8f8483b139bb1edde97f917fa436671693f6f/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L17-L20) used in this repository, it takes on average six seconds to run a single benchmark. 
+The public surface of .NET Standard 2.0 API has `36 519` methods. If we had 1 benchmark for every public method, it would take two and a half days to run the benchmarks.
```
billwert: While the precision is nice we aren't going to keep this up to date and I don't want PRs "fixing" it :)

```suggestion
The public surface of .NET Standard 2.0 API has tens of thousands of methods. If we had 1 benchmark for every public method, it would take two and a half days to run the benchmarks.
```

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+This is only one of the reasons why writing Benchmarks is different than writing Unit Tests.
```
danmosemsft: They also take time to analyze, have false positives,. ...

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+```
+
+**Note:** If you need to clean up resources after the benchmark run, you should use the corresponding `[GlobalCleanup]` attribute. It's going to be executed only once, after all benchmark iterations.
+
+### IterationSetup
+
+If your benchmark requires a clean state for every invocation, you need to use the `[IterationSetup]` attribute. Unfortunately, just using the `[IterationSetup]` attribute is not enough to get stable results. You also need to make sure that the benchmark itself performs enough of computations for a single invocation to run longer than 100ms. The closer to 250ms the more stable results you are going to get.
+
+```cs
+[Params(1000 * 1000 * 200)] // allows for stable iteraiton around 200ms
```
danmosemsft: typo

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+    public override string ToString() => FormatName;
+
+    private static Stream CreateTestImage(ImageFormat format)
+    // omitted for brevity
+}
+```
+
+#### Generic benchmarks
+
+BenchmarkDotNet supports generic classes. It's allows for having a dedicated test case for Value and Reference Types.
```
danmosemsft: it's -> it

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+[GlobalSetup(Target = nameof(Reverse))]
+public void SetupReverse()
+
+[GlobalSetup(Targets = new [] { nameof(Array), nameof(Span), nameof(ReadOnlySpan)})]
+public void SetupArray()
+```
+
+**Note:** If you need to clean up resources after the benchmark run, you should use the corresponding `[GlobalCleanup]` attribute. It's going to be executed only once, after all benchmark iterations.
+
+### IterationSetup
```
billwert: I think given that everyone who uses this feature hits the failure case, it is worth it to call out that unless you take special care (as you describe) your benchmark will be entirely invalid. In the general case we should specifically recommend people avoid this feature, imo.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+* BenchmarkDotNet **does not require the user to provide the number of iterations and invocations per iteration**, it implements a smart heuristic based on standard error and runs the benchmark until the results are not stable.
+* BenchmarkDotNet runs every benchmark in a separate process, process isolation allows avoiding side-effects. The more memory allocated by given benchmark, the bigger the difference for in-proc vs out-proc execution.
+* BenchmarkDotNet was designed to make accurate nano-benchmarks with repeatable results possible, to achieve that it does many things, including overhead calculation and subtraction (it benchmarks an empty method with the same signature and subtract the average value from results).
+* BenchmarkDotNet removes outliers by default (this repo is configured to remove only the upper outliers)
+* BenchmarkDotNet creates a type which derives from type with benchmarks. So the type with benchmarks must **not** be **sealed** and it can **not** be **static** and it has to be **public**. It also has to be a `class` (no structs support).
+
+## Setup
+
+Let's write a benchmark that measures the performance of reverting an array of 1 000 integers:
```
billwert:
```suggestion
Let's write a benchmark that measures the performance of reversing an array of 1 000 integers:
```

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+```cmd
+dotnet run -c Release -f netcoreapp3.0 --filter *.Reverse --profiler ETW
+```
+
+And open the produced trace file with [PerfView](https://github.com/Microsoft/perfview):
+
+![Regression](img/setup_array_reverse_profiler.png)
+
+As you can see, reversing the array took only `26.6%` of the benchmark execution time!! The rest was spent on executing the array creation logic. What does it mean? It means that the given benchmark is measuring the performance of creating and reversing the array. Not just reversing the array.
```
billwert:
```suggestion
As you can see, reversing the array took only `26.6%` of the benchmark execution time! The rest was spent on executing the array creation logic. What does it mean? It means that the given benchmark is measuring the performance of creating and reversing the array. Not just reversing the array.
```

docs/microbenchmark-design-guidelines.md
billwert: yeah, was going to suggest a short list and keep this expansion as a double click into why fewer tests is better.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+This is why **the initialization logic should be always separated from the benchmark**. If possible of course.
```
billwert: I think it is almost always possible. There are very few cases where the initialization isn't something you can remove - I'd advocate for writing docs that drive the behavior we want (which in this case is "think hard about it and strive to do the right thing.")
```suggestion
This is why **the initialization logic should be always separated from the benchmark**.
```

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+```cs
+[GlobalSetup(Target = nameof(Reverse))]
+public void SetupReverse()
+
+[GlobalSetup(Targets = new [] { nameof(Array), nameof(Span), nameof(ReadOnlySpan)})]
+public void SetupArray()
+```
+
+**Note:** If you need to clean up resources after the benchmark run, you should use the corresponding `[GlobalCleanup]` attribute. It's going to be executed only once, after all benchmark iterations.
```
billwert: Worth it to call out here that an example of required cleanup would be any files created on the disk by the benchmark process.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+### IterationSetup
+
+If your benchmark requires a clean state for every invocation, you need to use the `[IterationSetup]` attribute. Unfortunately, just using the `[IterationSetup]` attribute is not enough to get stable results. You also need to make sure that the benchmark itself performs enough of computations for a single invocation to run longer than 100ms. The closer to 250ms the more stable results you are going to get.
```
billwert: I feel like this needs a glossary at the top: what is an iteration versus an invocation? etc. Terms to define are at least:

- iteration
- invocation
- duration (total time reported)
- duration of single invocation (actual metric we track)
...

docs/microbenchmark-design-guidelines.md
billwert: this is another pitfall: 200ms on your Macbook may be 80ms on our lab machines, or vice versa. Another reason why I really want there to be a big red flag on this section.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+```
+
+The benchmark is not going to measure the performance of `Slice` operation, but similar to previous example it's going to measure the performance of creating a `Span` from `Array` and the `Slice` operation.
+
+To solve this problem we can use `OperationsPerInvoke` property of the `[Benchmark]` attribute.
+
+```cs
+private byte[] _nonEmptyArray = new byte[18];
+
+[Benchmark(OperationsPerInvoke = 16)]
```
billwert: Is there any guidance or principle around how I should choose this number? Will BDN scale the number of invocations with this, since we'll hit the 250ms ceiling still?

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+![Regression](img/setup_array_reverse_profiler.png)
+
+As you can see, reversing the array took only `26.6%` of the benchmark execution time!! The rest was spent on executing the array creation logic. What does it mean? It means that the given benchmark is measuring the performance of creating and reversing the array. Not just reversing the array.
+
+This is why **the initialization logic should be always separated from the benchmark**. If possible of course.
+
+### GlobalSetup
+
+Public method marked with `[GlobalSetup]` attribute is going to be executed **exactly once**, before running the benchmark for the first time.
```
billwert:
```suggestion
A public method marked with `[GlobalSetup]` attribute is going to be executed **exactly once**, before running the benchmark for the first time.
```

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+private int[] _array;
+
+[GlobalSetup]
+public void SetupReverse() => _array = Enumerable.Range(0, 1000).ToArray();
+
+[Benchmark]
+public void Reverse() => Array.Reverse(_array);
+```
+
+If given `[GlobalSetup]` does not specify the `Target(s)` benchmark name(s), it's executed for every benchmark declared in a given class.
```
billwert:
```suggestion
Any method marked `[GlobalSetup]` which does not specify the `Target(s)` benchmark name(s) is executed for every benchmark declared in a given class.
```

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+    100, // comment explaining why it's needed to have this test case
+    1_000, // comment explaining why it's needed to have this test case
+    10_000)] // comment explaining why it's needed to have this test case
+public int Size { get; set; }
+
+[GlobalSetup]
+public void PrepareArray() => _array = new int[Size];
+```
+
+However, **`[Params]` are applied to all the benchmarks in given class**. So if few benchmarks require different `[Params]` values they have to be split into separate types (current BenchmarkDotNet implementation limitation).
```
billwert: Is there a BDN issue tracking this limitation? Will it go away in the future? If not, I'd elide the parenthetical here and leave the statement as is.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+[Benchmark]
+[ArgumentsSource(nameof(Cultures))]
+public DateTime Parse(CultureInfo cultureInfo)
+    => DateTime.Parse("10/10/2010 12:00:00 AM", cultureInfo);
+```
+
+**Note:** If you need to use the argument in the setup method, then instead of using `ArgumentsSource` you should use `ParamsSource`.
+
+**Note:** the time spent for initializing an argument is not included in the time reported by the harness. It means that you can have a custom complex type with initialization logic in it's ctor and setup the benchmark in that way:
+
```
billwert: It might be worth noting here that `ArgumentsSource` methods are only called once per execution of the benchmark, and the cached results are used per invocation. This is something that confused me for a while.

docs/microbenchmark-design-guidelines.md
billwert: or it's -> this

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+In this particular benchmark, the list is growing with every benchmark invocation. `List<T>` is internally using an `Array` to store all the elements. When the array is not big enough to store one more element, a two times bigger array is allocated and all elements are copied from the old array to the new one. It means that every next `Add` operation takes more time. We might also get `OutOfMemoryException` at some point in time.
+
+**Benchmarks should not have any side effects**. 
+
+### Dead Code Elimination
+
+To prevent from [dead code elimination](https://en.wikipedia.org/wiki/Dead_code_elimination) BenchmarkDotNet consumes the result returned from a benchmark and writes it to a `volatile` field ([code](https://github.com/dotnet/BenchmarkDotNet/blob/94863ab4d024eca04d061423e5aad498feff386b/src/BenchmarkDotNet/Engines/Consumer.cs)).
+
+The only thing that you need to remember is to **return the result from the benchmark**. Even if you think that as of today the JIT is not going to eliminate the code.
```
billwert:
```suggestion
The only thing that you need to remember is to **return the result from the benchmark**. This is required practice even if you know that the JIT will not eliminate code in a particular case.
```

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+By relying on the BDN mechanism you are going to avoid loop alignment issues. BenchmarkDotNet handles that on its own side when your benchmark does not have a loop.
+
+### Method Inlining
+
+BenchmarkDotNet prevents from inlining the benchmarked method by wrapping it into a delegate (delegates can not be inlined as of today). The cost of delegate invocation is excluded by running separate run for Overhead calculation.
+
+The benchmark methods don't need to have `[MethodImpl(MethodImplOptions.NoInlining)]` attribute applied.
+
+## Summary
```
billwert: I think the summary section could actually go to the top, with links to the sub-sections. It's a good general overview, and might spark people to think "oh, I hadn't thought about arguments. I should think about that" without reading the whole doc.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+* Choose a good name because you won't be able to change it after the first results are uploaded to the Reporting System.
+* Move initialization logic to a Setup method. Don't forget to clean up the resources in a corresponding Cleanup method.
+* If your benchmark needs input data, it should be always exactly the same data. Use `ValuesGenerator` to generate the data.
+* You should have one test case per one code path. Not more.
```
billwert: this feels a little ambiguous between "benchmarks should do one thing" and "one benchmark for each code path I am trying to test". You are saying the second here - maybe just be clearer?

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+* Focus on the most common use cases, not edge cases.
```
billwert: Call out again that performance of error cases is almost never interesting.

danmosemsft:
Looks great to me.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+Benchmarks are different. First of all, the result of a benchmark run is never a single value. It's a whole distribution, described with values like mean, standard deviation, min, max and so on. To get a meaningful distribution, the benchmark has to be executed many, many times. **This takes a lot of time**. With the current [recommended settings](https://github.com/dotnet/performance/blob/51d8f8483b139bb1edde97f917fa436671693f6f/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L17-L20) used in this repository, it takes on average six seconds to run a single benchmark. 
+The public surface of .NET Standard 2.0 API has tens of thousands of methods. If we had 1 benchmark for every public method, it would take two and a half days to run the benchmarks.
+
+This is only one of the reasons why writing Benchmarks is different than writing Unit Tests.
+
+The goal of benchmarking is to test the performance of all the methods that are frequently used (hot paths) and should be performant. The focus should be on the most common use cases, not edge cases.
```
gfoidl:
```suggestion
The goal of benchmarking is to test the performance of all the methods that are frequently used (hot paths) and should be performant. **The focus should be on the most common use cases, not edge cases**.
```
But some emphasis on this? I think this is a essential statement.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+The Reporting System is used to monitor the performance changes and detect regressions. If the body of the benchmark is changed, the old name is kept and the reported time increases, it's going to be marked as regression and require a dedicated performance investigation. We want to avoid such false alarms and hence we want to reduce the number of changes in existing benchmarks to a minimum.
+
+![Regression](img/benchmark_id_regression.png)
+
+If you have some good reasons for changing the implementation of the benchmark you should remove the old one and introduce a new one with a new name.
+
+## BenchmarkDotNet
+
+BenchmarkDotNet is the benchmarking harness used in this repository. If you are new to BenchmarkDotNet, you should read [this doc](./benchmarkdotnet.md).
```
gfoidl:
```suggestion
BenchmarkDotNet is the benchmarking harness used in this repository. If you are new to BenchmarkDotNet, you should read [this introduction to BenchmarkDotNet](./benchmarkdotnet.md).
```

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+
+Key things that you need to remember: 
+
+* BenchmarkDotNet **does not require the user to provide the number of iterations and invocations per iteration**, it implements a smart heuristic based on standard error and runs the benchmark until the results are not stable.
```
gfoidl:
```suggestion
* BenchmarkDotNet **does not require the user to provide the number of iterations and invocations per iteration**, it implements a smart heuristic based on standard error and runs the benchmark until the results are stable.
```

docs/microbenchmark-design-guidelines.md
gfoidl: Maybe also worth to add this hint to the xml-docs of `IterationSetup`?

gfoidl: Maybe also worth to add this hint to the xml-docs of `IterationSetup`?

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+[Benchmark]
+public Span<byte> Slice()
+{
+    Span<byte> span = new Span<byte>(_nonEmptyArray);
+
+    return span.Slice(span.Length / 2);
+}
+```
+
+The benchmark is not going to measure the performance of `Slice` operation, but similar to previous example it's going to measure the performance of creating a `Span` from `Array` and the `Slice` operation.
```
gfoidl:
```suggestion
The benchmark is not going to measure the performance of `Slice` operation, but similar to previous example of reversing an array it's going to measure the performance of creating a `Span` from `Array` and the `Slice` operation.
```

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+[Benchmark(OperationsPerInvoke = 16)]
+public Span<byte> Slice()
+{
+    Span<byte> span = new Span<byte>(_nonEmptyArray); // create it once
+
+    // perform OperationsPerInvoke-many operations
+    span = span.Slice(1); span = span.Slice(1); span = span.Slice(1); span = span.Slice(1);
+    span = span.Slice(1); span = span.Slice(1); span = span.Slice(1); span = span.Slice(1);
+    span = span.Slice(1); span = span.Slice(1); span = span.Slice(1); span = span.Slice(1);
+    span = span.Slice(1); span = span.Slice(1); span = span.Slice(1); span = span.Slice(1);
```
gfoidl: Please add a note why or why not 
```c#
// ...
for (int i = 0; i < 16; ++i)
{
    span = span.Slice(1);
}
// ...
```
should be used, as it would be tempting.

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+        first = ref Unsafe.Add(ref first, 1);
+        last = ref Unsafe.Add(ref last, -1);
+    } while (Unsafe.IsAddressLessThan(ref first, ref last));
+}
+```
+
+Does it make sense to test the code paths that throw?
+
+ * No, because we would be measuring the performance of throwing and catching the exceptions. That was not the goal of this benchmark.
+ * No, because throwing exceptions should be exceptional and [exceptions should not be used to control flow](https://docs.microsoft.com/en-US/visualstudio/profiling/da0007-avoid-using-exceptions-for-control-flow?view=vs-2019). It's an edge case, we should focus on [common use cases, not edge cases](#Benchmarks-are-not-Unit-Tests).
```
gfoidl: 👍  for the links

docs/microbenchmark-design-guidelines.md
```diff
@@ -0,0 +1,557 @@
...
+    }
+    return result;
+}
+```
+
+By relying on the BDN mechanism you are going to avoid loop alignment issues. BenchmarkDotNet handles that on its own side when your benchmark does not have a loop.
+
+### Method Inlining
+
+BenchmarkDotNet prevents from inlining the benchmarked method by wrapping it into a delegate (delegates can not be inlined as of today). The cost of delegate invocation is excluded by running separate run for Overhead calculation.
```
gfoidl:
```suggestion
BenchmarkDotNet prevents from inlining the benchmarked method by wrapping it into a delegate (delegates can not be inlined as of today). The cost of delegate invocation is excluded by a separate run for Overhead calculation.
```

docs/microbenchmark-design-guidelines.md
adamsitnik: @billwert good point, I added following sentence (in bold):

**If you don't, your benchmark will be entirely invalid.**

docs/microbenchmark-design-guidelines.md
adamsitnik: Good point, I extended the BDN summary with a link to [this BDN doc](https://benchmarkdotnet.org/articles/guides/how-it-works.html) that explains what these things are using words and pseudocode.

docs/microbenchmark-design-guidelines.md
adamsitnik: Yes, it's going to scale the number of invocations.

I added following sentence to the description:

**Note:** `OperationsPerInvoke` should be big enough to amortize the "setup" cost.

docs/microbenchmark-design-guidelines.md
adamsitnik: I made the following sentence bold and underline. + "NOT" as uppercase (Vance style ;) )

<ins>**If using `[GlobalSetup]` is enough, you should NOT be using `[IterationSetup]`**</ins>

docs/microbenchmark-design-guidelines.md
adamsitnik: Another good point! It makes me wonder what are the other things that are obvious to me but surprising to the end user ;)

adamsitnik: Ok, let's see what the users are going to say https://twitter.com/SitnikAdam/status/1148555464394911744

docs/microbenchmark-design-guidelines.md
adamsitnik: good point, I added it as a comment in the sample code

adamsitnik:
@danmosemsft @billwert @gfoidl big thanks for the review! I have applied all the suggestions and fixed the issues

billwert:
Thanks @adamsitnik!

PR #626, opened 2019-07-11
Add core sdk dependency

adiaaida:
Will set us up for taking per-build dependency updates from core-sdk


adiaaida:
cc/ @billwert 

I'm going to check this in quickly so that we can see if we get the core-sdk dependency update that is currently running.

billwert:
:shipit:

PR #627, opened 2019-07-11
Add reference assemblies

billwert:


billwert:
fails due to dotnet/benchmarkdotnet#1197. I have another idea to unblock it though..

adamsitnik:
I saw https://github.com/dotnet/BenchmarkDotNet/issues/1197 and I need to wrap my head around it. Tomorrow I am OOF so I won't have a solution this week.

billwert:
This is going to have to wait until we get a fix from BDN. The `Directory.Build.props` approach is blocked specifically.

PR #630, opened 2019-07-11
add benchmark for Guid.Equals

adamsitnik:
https://github.com/dotnet/coreclr/issues/25644 made me realize that we don't have any `Guid.Equals` benchmark

### DIFF
--- a/src/benchmarks/micro/corefx/System.Runtime/Perf.Guid.cs
+++ b/src/benchmarks/micro/corefx/System.Runtime/Perf.Guid.cs
@@ -12,10 +12,16 @@ namespace System.Tests
     {
         const string guidStr = "a8a110d5-fc49-43c5-bf46-802db8f843ff";
 
+        private readonly Guid _guid = new Guid(guidStr);
+        private readonly Guid _same = new Guid(guidStr);
+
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/corefx/System.Runtime/Perf.Guid.cs
```diff
@@ -12,10 +12,16 @@ public class Perf_Guid
...
+        private readonly Guid _same = new Guid(guidStr);
+
         [Benchmark]
         public Guid NewGuid() => Guid.NewGuid();
 
         [Benchmark]
         public Guid ctor_str() => new Guid(guidStr);
+
+        [Benchmark]
+        public bool EqualsSame() => _guid.Equals(_same);
```
gfoidl: Add also a benchmark for the unequal case?

```c#
private readonly Guid _other = Guid.NewGuid();
// ...
[Benchmark]
public bool EqualsFalse() => _guid.Equals(_other);
```

And I would rename this benchmark:

```suggestion
        public bool EqualsTrue() => _guid.Equals(_same);
```
as `EqualsSame` reminds me of `_guid.Equals(_guid)`.

PR #631, opened 2019-07-11
Use the performance template to run testing

adiaaida:
We can't check this in until preview7 ships since it removed all publishing to benchview. I wanted to get it tested to make sure everything is working, so creating this PR now.


adiaaida:
Looks like maestro didn't propagate file permissions. This is being tracked here: https://github.com/dotnet/arcade/issues/2862

PR #634, opened 2019-07-11
Do not install dotnet with -Channel if versions are specified

adiaaida:
When we supply a version, but still pull down with -Channel, we will use
whichever sdk is newer. So if we are trying to check an older version,
or if there is a new version between when we start a run and when we actually
run, we will be testing the "wrong" version, ie, not the version we specified.

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -545,7 +545,12 @@ def install(
                 get_repo_root_path()
             )
 
-    if channels:
+    # Only check channels if versions are not supplied.
+    # When we supply a version, but still pull down with -Channel, we will use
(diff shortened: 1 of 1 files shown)


jorive:
Can this be fixed at the input level instead? Make the options mutually exclusive?

billwert:
we need this now because we just turned up the frequency of runs a ton, and this is causing us immediate problems. we can open an issue to clean it up in the future.

adiaaida:
As far as I can tell, that would be a significant amount of engineering effort. I have looked at this in many different ways, but there are too many ways that we are currently calling the install function to make that fix reasonable.

PR #637, opened 2019-07-11
Move to the new perf queues

adiaaida:
Move to the new perf queues. The linux queue has 25 machines and the windows queue has 30 machines, with more being added daily.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -89,11 +89,11 @@ jobs:
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: windows
-      osVersion: RS5
+      osVersion: 19H1
       kind: micro
       architecture: x64
       pool: Hosted VS2017
-      queue: Windows.10.Amd64.ClientRS5.Perf # using a dedicated private Helix queue (perfsnakes)
+      queue: Windows.10.Amd64.19H1.Tiger.Perf # using a dedicated private Helix queue (perfsnakes)
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
       runCategories: 'coreclr corefx'
       benchviewCategory: 'coreclr'
@@ -119,11 +119,11 @@ jobs:
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: windows
-      osVersion: RS5
+      osVersion: 19H1
       kind: micro
       architecture: x86
       pool: Hosted VS2017
-      queue: Windows.10.Amd64.ClientRS5.Perf # using a dedicated private Helix queue (perfsnakes)
+      queue: Windows.10.Amd64.19H1.Tiger.Perf # using a dedicated private Helix queue (perfsnakes)
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
       runCategories: 'coreclr corefx'
       benchviewCategory: 'coreclr'
@@ -171,7 +171,7 @@ jobs:
       kind: micro
       architecture: x64
       pool: Hosted Ubuntu 1604
-      queue: Ubuntu.1804.Amd64.Perf # using a dedicated private Helix queue (perfsnakes)
+      queue: Ubuntu.1804.Amd64.Tiger.Perf # using a dedicated private Helix queue (perfsnakes)
       container: ubuntu_x64_build_container
       csproj: src/benchmarks/micro/MicroBenchmarks.csproj
       runCategories: 'coreclr corefx'
@@ -231,11 +231,11 @@ jobs:
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: windows
-      osVersion: RS5
+      osVersion: 19H1
       kind: mlnet
       architecture: x64
       pool: Hosted VS2017
-      queue: Windows.10.Amd64.ClientRS5.Perf # using a dedicated private Helix queue (perfsnakes)
+      queue: Windows.10.Amd64.19H1.Tiger.Perf # using a dedicated private Helix queue (perfsnakes)
       csproj: src\benchmarks\real-world\Microsoft.ML.Benchmarks\Microsoft.ML.Benchmarks.csproj
       runCategories: 'mldotnet'
       benchviewCategory: 'mldotnet'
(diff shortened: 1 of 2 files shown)


billwert:
is there an open queue we can move Windows CI to as well?

adiaaida:
A ClientRS5 queue? Yes. I'll change that.

azure-pipelines.yml
```diff
@@ -89,7 +89,21 @@ jobs:
...
+      kind: micro
+      architecture: x64
+      pool: Hosted VS2017
+      queue: Windows.10.Amd64.ClientRS5.Perf # using a dedicated private Helix queue (perfsnakes)
+      csproj: src\benchmarks\micro\MicroBenchmarks.csproj
+      runCategories: 'coreclr corefx'
+      benchviewCategory: 'coreclr'
+      frameworks: # for private jobs we want to benchmark .NET Core 3.0 only
+        - netcoreapp3.0
+- ${{ if and(ne(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'Schedule')) }}:
```
adamsitnik: Why the Windows version depends on the trigger (Windows RS**5** for PullRequest and RS**1** for the Scheduled triggers)? How are we going to differentiate the uploaded results in the Reporting System (there can be some differences I guess)?

```yml
# Windows x64 micro benchmarks, private job
- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
    parameters:
      osVersion: RS5
      queue: Windows.10.Amd64.ClientRS5.Perf # using a dedicated private Helix queue (perfsnakes)
- ${{ if and(ne(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'Schedule')) }}:
    parameters:
      osVersion: RS1
      queue: Windows.10.Amd64.ClientRS1.Perf # using a dedicated private Helix queue (perfsnakes)
```

azure-pipelines.yml
```diff
@@ -116,16 +144,16 @@ jobs:
...
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: ubuntu
-      osVersion: 1604
+      osVersion: 1804
       kind: micro
       architecture: x64
       pool: Hosted Ubuntu 1604
-      queue: Ubuntu.1604.Amd64.Open
+      queue: Ubuntu.1804.Amd64.Open
```
adamsitnik: awesome! I am super happy to see the Ubuntu 18.04 being added to the list of tracked OSes!

/cc @tannergooding @danmosemsft ft

azure-pipelines.yml
```diff
@@ -116,16 +144,16 @@ jobs:
...
+# Ubuntu 1804 x64 micro benchmarks, public correctness job
 - ${{ if eq(variables['System.TeamProject'], 'public') }}:
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: ubuntu
-      osVersion: 1604
+      osVersion: 1804
       kind: micro
       architecture: x64
       pool: Hosted Ubuntu 1604
```
adamsitnik: nit: should the name of the pool be updated as well?

```suggestion
      pool: Hosted Ubuntu 1804
```

azure-pipelines.yml
adiaaida: To question 1: The RS5 queue is the new queue for the PerfTigers, the RS1 queue is the PerfSnakes. We want to run on the PerfTigers for all runs, but for a little while, we also want to maintain some duplicate runs on the snakes so we can understand if we're seeing a regression compared to earlier runs, up until the point that we have enough history from the Tigers.

In the new database, we retain information about the queue that a job was run in, so we can separate the results out based on queue. We did this knowing we would have instances where we changed machines, and wanted to keep old results separate from new results.

azure-pipelines.yml
adiaaida: There is no Hosted Ubuntu 1804 queue, I believe. I had changed that, but it broke it. It also does not matter which hosted queue we are on, as that's just where we trigger the helix runs from (we don't build on build machines).

azure-pipelines.yml
adamsitnik: @adiaaida thanks for the explanation! now I get it

azure-pipelines.yml
tannergooding: A few questions:
* What are the plans, if any, for testing interim builds, like Ubuntu 19.04 or 19.10 when it comes out?
* Are there any goals at getting onto new LTS builds (e.g. 20.04) sooner after they come out?
* Given that there are 2-3 LTS builds out at a given time, what are the plans for the lifetime of the perf machines for each version?

adiaaida: For perf, it's a bit of a challenge to move to new OSes. @billwert would probably be the best person to talk about in terms of future planning. I would guess that we will only run perf on LTS builds, but it requires us to split our pools, and we have limited hardware. So when we move to a new LTS, we will lose support for previous versions. We are just now moving from 1604 to 1804, and 1804 has been out for over a year.

billwert: yeah, it would be my goal to run only LTSes, I think. At some point in the future perhaps we'll have an imaging story and will have more flexibility. Right now the OS is installed on bare metal, so shifting is a big deal. It also means that, as Michelle says, if we tried to run 2 LTSes at once we'd have to cut the pool in half. I would rather not do that. In the future we'll be able to respond more quickly. I think we'll pick up new LTSes when a branch of the product starts supporting that LTS. (So for 20.04, .NET Core 5 next year?)

There is a complication to this that older products may not work on newer LTS versions, so our ability to baseline things properly may be impacted. Need to think about that a bit.

billwert:
for some reason I thought we were blocked from using `HELIX_WORKITEM_UPLOAD_ROOT` on our machines because they were physical and some deployment tech didn't work properly on them. Can you clarify (likely just to correct me? :))

adiaaida:
The snakes have not been updated (hence the if not RS1 and if RS1), but the new machines and any VM queue should have the latest helix scripts.

azure-pipelines.yml
```diff
@@ -42,11 +42,11 @@ jobs:
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: windows
-      osVersion: RS4
+      osVersion: RS5
       kind: micro
       architecture: x64
       pool: Hosted VS2017
-      queue: Windows.10.Amd64.ClientRS4.Open
+      queue: Windows.10.Amd64.ClientRS5.Open
```
billwert: I guess we need the new queue names and to update this. (in several spots.)

adiaaida: Yes. I am going to wait until it is rolled out and the machines have been updated to make that change.

adiaaida: Made the change. This can't be checked in until the new rollout to create the queue and the machines have been moved to the new queue name.

adiaaida:
@ilyas1974 Will you let us know when the queue name change is done? We can check this in once it is.

adiaaida:
@ilyas1974 Will you let us know when the queue name change is done? We can check this in once it is.

PR #653, opened 2019-07-12
Move the download of the benchview scripts to ci_setup

adiaaida:
With the influx of new runs, we have started seeing 403 errors when
downloading the benchview scripts because we are hitting the server too
frequently. Since we do not need to download the scripts in every
partition, just once on the build machine, we should move the download
to the ci_setup script. This change does that, and adds a
skip-download-benchview flag to be used in the lab so we don't try to
download them again.

Another option would be to not delete the benchview folder if it exists
already and not attempt to download the scripts if they are already on
the machine, but I did not go that route because I understand that if
there are updates to the scripts, we would want to overwrite them (on
personal machines, I guess). A third option would have been to finally
just check the benchview scripts into the performance repo, but I
decided against that since benchview will only be in use until preview7
ships.


adiaaida:
It looks like the real issue is that we hit our quota and Azure disabled the site that hosts the benchview scripts. We may need a different solution than this.

adiaaida:
I have changed this to adding the benchview scripts to the performance repo.

jorive:
Can we temporarily update the Azure quota instead (to avoid the 403 error)? Or add the nuget package to other feed? If we are getting rid of BenchView, I do not see the point of adding this scripts here.
On the other hand, this will not work just as-is. The packaging step brings down external-fixed-version Python dependencies too.

scripts/tools/submission-metadata.json
```diff
@@ -0,0 +1 @@
+{"created": "2019-01-30T21:57:23Z", "cuid": "cjrjqegqb0000ms7411y3ajxt", "description": null, "email": "dotnet-bot@microsoft.com", "name": "Test", "runs": []}
```
jorive: This file is not needed.

adiaaida:
Upping the quota requires spending money, and I'm not willing to make that decision without talking to someone first. I didn't realize that the dependency folder was strictly necessary (why? don't most of these things come from python itself?). I'm trying to unblock us as quickly as possible, so I don't care how it gets fixed, as long as it does.

jorive:
> Upping the quota requires spending money, and I'm not willing to make that decision without talking to someone first. I didn't realize that the dependency folder was strictly necessary (why? don't most of these things come from python itself?). I'm trying to unblock us as quickly as possible, so I don't care how it gets fixed, as long as it does.

Why? Because the scripts use modules that are OSS and not part of the Python standard library.

jorive:
How about using myget as the temporary feed? That seems to scale better, we have used it in the past, and it is a quick/small fix.

billwert:
I just spent money.

PR #654, opened 2019-07-12
Use different scripting to upload arm64 files

adiaaida:


scripts/performance/common.py
```diff
@@ -87,6 +87,28 @@ def get_artifacts_directory() -> str:
     '''
     return os.path.join(get_repo_root_path(), 'artifacts')
 
+def rename_upload_files(files) -> None:
```
jorive: Is this function going to be used outside `upload.py`?

adiaaida: Yes. It is used in azcopy.py.

adiaaida:
@DrewScoggins PTAL. I'd like to get this in today.

scripts/upload.py
```diff
@@ -0,0 +1,80 @@
...
+def upload_data(url, data):
+    headers = { "x-ms-blob-type": "BlockBlob", "Content-Type": "application/json" }
+    req = urllib.request.Request(url=url, data=data, method='PUT', headers=headers)
+    with urllib.request.urlopen(req) as f:
+        pass
+    if not (f.status == 200 or f.status == 201):
+        raise ConnectionError("Upload to url {} failed with status {} and reason {}".format(url, f.status, f.reason))
+
+
+def load_json_from_binary(data):
```
billwert: unused?

scripts/performance/common.py
```diff
@@ -87,6 +87,28 @@ def get_artifacts_directory() -> str:
...
+    for file in files:
+        directory_name = os.path.dirname(file)
+        filename = os.path.basename(file)
+        newname = "{0}-{1}".format(os.path.join(
+                                    directory_name,
+                                    os.getenv('HELIX_WORKITEM_ID')),
```
billwert: nit: can you make passing in HELIX_WORKITEM_ID a param since this is now a common function?

adiaaida: Sounds good.

scripts/performance/common.py
```diff
@@ -87,6 +87,28 @@ def get_artifacts_directory() -> str:
     '''
     return os.path.join(get_repo_root_path(), 'artifacts')
 
+def rename_upload_files(files, helix_workitem_id) -> None:
```
billwert: double super nit: maybe:
```suggestion
def rename_upload_files(files, differentiator) -> None:
```

PR #667, opened 2019-07-16
make sure that the arrays used by IndexerBench are always unaligned

adamsitnik:
I was looking at the numbers I found for IndexerBench in https://github.com/dotnet/coreclr/issues/25726 and I was not sure if this an alignment issue or not. 

To make sure that we are always comparing apples to apples I think it's worth to ensure that the arrays are **not** aligned. Not aligned (instead always aligned) to make sure we don't track the performance of best, but rare scenario.

@billwert this change should make the `IndexerBench` results more stable. I think that we should start using the new helper methods in more benchmarks, to get rid of the array alignment issues and the noise it always creates during investigations.

### DIFF
--- a/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
 
 namespace BenchmarkDotNet.Extensions
@@ -19,6 +20,22 @@ namespace BenchmarkDotNet.Extensions
                 return Array<T>(byte.MaxValue).First(value => !value.Equals(default));
             else
                 return ArrayOfUniqueValues<T>(2).First(value => !value.Equals(default));
+        }
+
+        /// <summary>
+        /// generates an array of unique random values
+        /// the array is guaranteed to be not 16-byte aligned
+        /// </summary>
+        /// <remarks>make sure to Free the handle in a Cleanup method</remarks>
+        public static T[] UnalignedArrayOfUniqueValues<T>(int count, out GCHandle pinnedArrayHandle)
+            where T : struct // GCHandle supports only pinning arrays of primitive and blittable types
+        {
+            T[] randomUnique = ArrayOfUniqueValues<T>(count);
+            T[] unalignedUnique = AllocateAlignedArray<T>(count, false, out pinnedArrayHandle);
+
+            System.Array.Copy(randomUnique, unalignedUnique, count);
+
+            return unalignedUnique;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


billwert:
> Not aligned (instead always aligned) to make sure we don't track the performance of best, but rare scenario.

The runtime in fact goes to some trouble to make sure things are often aligned. Do you have evidence of this being a problem?

For benchmarking purposes the absolute number isn't the real value - saying 2ns here or 5ns there doesn't matter, it's about consistency. I think having something that forces alignment is okay, but having both flavors feels like overkill.

adamsitnik:
> The runtime in fact goes to some trouble to make sure things are often aligned

Are you sure? I thought that we don't enforce any array alignment policies.

> Do you have evidence of this being a problem?

A good example is `IniArray` benchmark. It's very dependent on the alignment.

> but having both flavors feels like overkill

Do you mean from the repo-wide perspective? To have it everywhere or nowhere?

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -114,5 +147,23 @@ private static string GenerateRandomString(Random random, int minLength, int max
 
             return builder.ToString();
         }
+
+        private static T[] AllocateAlignedArray<T>(int size, bool aligned, out GCHandle pinnedArrayHandle)
```
jorive: nit: The method name (`AllocateAlignedArray`) is misleading when you are calling it to allocate an un-aligned array.

billwert:
> Are you sure? I thought that we don't enforce any array alignment policies.

ECMA 335 asserts that objects must be aligned (12.6.2 Alignment).

The array allocation itself will always be aligned. If I change your function to check for four or eight byte alignment, it never finds an unaligned array. This matches my expectation. As it is written, on my machine it happens to find one after a couple iterations, but that's because it happens to find an address that is not 16 byte aligned. (But is 8.)

```
PS E:\work\scratch\align > dotnet run
1628557722400
1628557739624
```

Perhaps the root issue here is that this is an array of `byte`. Necessarily not every single byte access can be aligned. Even then, I am somewhat curious why this change made the test more stable for you. I believe two things are true:
1)  As the tests are walking the array sequentially, the processor should be doing a good job of pulling data into cache, at which point the cost of unaligned access is diminished.
2) As the test consistently uses the same array, the cost of unaligned access is amortized across the run and should remain stable. I don't believe that single byte access will incur any of the normal penalties associated with alignment issues (principally, when a value crosses a cache line boundary.)

Can you share assembly diffs so we can understand if there is another change not being considered?

> Do you mean from the repo-wide perspective? To have it everywhere or nowhere?

I mean from this perspective: Numbers in this repo aren't intended to be used by anyone to say "this method takes N nanoseconds." They are only internally consistent. Choosing whether to test the "best" or "worst" outcome in this case doesn't matter, as long as we are consistent about it. If we can detect inconsistency, we should fix it in one way.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -21,6 +22,22 @@ public static T GetNonDefaultValue<T>()
                 return ArrayOfUniqueValues<T>(2).First(value => !value.Equals(default));
         }
 
+        /// <summary>
+        /// generates an array of unique random values
+        /// the array is guaranteed to be not 16-byte aligned
+        /// </summary>
+        /// <remarks>make sure to Free the handle in a Cleanup method</remarks>
+        public static T[] UnalignedArrayOfUniqueValues<T>(int count, out GCHandle pinnedArrayHandle)
```
billwert: As indicated in the comment on the main thread, I do not think this works as intended.

billwert:
Actually, thinking about this for another minute, accessing a byte **must** be aligned. I think we need a more in depth analysis of what is causing noise in this benchmark and how this change might be impacting it.

adamsitnik:
@billwert thanks for providing a detailed answer, I was not aware of that.

PR #669, opened 2019-07-17
Startup measurement tools

billwert:
This tool is the mechanical part of enabling startup time testing.

There are a handful of things it still needs. I'm checking this MVP in
to unblock teammates.

I'll open an issue to track these known issues:
1) needs to choose TFM dynamically (similar to microbenchmarks) to enable tests on 2.2
2) WPF parser isn't done
3) Refactoring opportunities between the parsers.

### DIFF
--- a/src/tools/ScenarioMeasurement/ScenarioMeasurement.sln
+++ b/src/tools/ScenarioMeasurement/ScenarioMeasurement.sln
@@ -3,9 +3,11 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
 VisualStudioVersion = 16.0.28801.201
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Startup", "Startup\Startup.csproj", "{756B0602-0D94-4EB7-A987-25BF1ED7B712}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Startup", "Startup\Startup.csproj", "{756B0602-0D94-4EB7-A987-25BF1ED7B712}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Util", "Util\Util.csproj", "{6B673656-6DE3-4971-A859-91AFDBFE5424}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Util", "Util\Util.csproj", "{6B673656-6DE3-4971-A859-91AFDBFE5424}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Reporting", "..\Reporting\Reporting\Reporting.csproj", "{1A871E0D-F5E0-4A61-B336-D7B080BC30A5}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -21,6 +23,10 @@ Global
 		{6B673656-6DE3-4971-A859-91AFDBFE5424}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{6B673656-6DE3-4971-A859-91AFDBFE5424}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{6B673656-6DE3-4971-A859-91AFDBFE5424}.Release|Any CPU.Build.0 = Release|Any CPU
+		{1A871E0D-F5E0-4A61-B336-D7B080BC30A5}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{1A871E0D-F5E0-4A61-B336-D7B080BC30A5}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{1A871E0D-F5E0-4A61-B336-D7B080BC30A5}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{1A871E0D-F5E0-4A61-B336-D7B080BC30A5}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/src/tools/ScenarioMeasurement/ScenarioMeasurement.sln
+++ b/src/tools/ScenarioMeasurement/ScenarioMeasurement.sln
@@ -1,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28801.201
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Startup", "Startup\Startup.csproj", "{756B0602-0D94-4EB7-A987-25BF1ED7B712}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Util", "Util\Util.csproj", "{6B673656-6DE3-4971-A859-91AFDBFE5424}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{756B0602-0D94-4EB7-A987-25BF1ED7B712}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{756B0602-0D94-4EB7-A987-25BF1ED7B712}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{756B0602-0D94-4EB7-A987-25BF1ED7B712}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{756B0602-0D94-4EB7-A987-25BF1ED7B712}.Release|Any CPU.Build.0 = Release|Any CPU
+		{6B673656-6DE3-4971-A859-91AFDBFE5424}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{6B673656-6DE3-4971-A859-91AFDBFE5424}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{6B673656-6DE3-4971-A859-91AFDBFE5424}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{6B673656-6DE3-4971-A859-91AFDBFE5424}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {7C86D749-5052-44A5-BBEC-DC01C3B6CD92}
+	EndGlobalSection
+EndGlobal
--- a/src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
+++ b/src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
@@ -1,0 +1,104 @@
+﻿using Microsoft.Diagnostics.Tracing;
+using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Diagnostics.Tracing.Session;
+using Reporting;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ScenarioMeasurement
+{
+    // [EventSource(Guid = "9bb228bd-1033-5cf0-1a56-c2dbbe0ebc86")]
+    // class PerfLabGenericEventSource : EventSource
+    // {
+    //     public static PerfLabGenericEventSource Log = new PerfLabGenericEventSource();
+    //     public void Startup() => WriteEvent(1);
+    // }
+
+    internal class GenericStartupParser : IParser
+    {
+        public void EnableKernelProvider(TraceEventSession kernel)
+        {
+            kernel.EnableKernelProvider((KernelTraceEventParser.Keywords)(KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.Thread | KernelTraceEventParser.Keywords.ContextSwitch));
+        }
+
+        public void EnableUserProviders(TraceEventSession user)
+        {
+            user.EnableProvider("PerfLabGenericEventSource");
+        }
+
+        public IList<Counter> Parse(string mergeTraceFile, string processName)
+        {
+            var results = new List<double>();
+            var threadTimes = new List<double>();
+            double threadTime = 0;
+            var ins = new Dictionary<int, double?>();
+            double start = -1;
+            int? pid = null;
+            using (var source = new ETWTraceEventSource(mergeTraceFile))
+            {
+
+                source.Kernel.ProcessStart += evt =>
+                {
+                    if (evt.ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (pid.HasValue)
+                        {
+                            throw new Exception("found a start when we already had a start");
+                        }
+                        pid = evt.ProcessID;
+                        start = evt.TimeStampRelativeMSec;
+                    }
+                };
+
+                source.Kernel.ThreadCSwitch += evt =>
+                {
+                    if (!pid.HasValue) // we're currently in a measurement interval
+                        return;
+
+                    if (evt.NewProcessID != pid && evt.OldProcessID != pid) // but this isn't it
+                        return;
+
+                    if (evt.OldProcessID == pid && ins.ContainsKey(evt.OldThreadID))
+                    {
+                        if (!ins[evt.OldThreadID].HasValue)
+                        {
+                            return;
+                        }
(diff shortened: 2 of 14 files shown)


src/tools/ScenarioMeasurement/Util/Util.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace ScenarioMeasurement
+{
+
+    public class Util
+    {
+        [DllImport("User32.Dll")]
+        private static extern long SetCursorPos(int x, int y);
```
brianrob: I don't think you should block on this, but one thing we should consider is how we're going to handle platform specific stuff.  Right now, the system depends on an ETL, but you could imagine it working with an eventpipe file or LTTng file.  There will always be some stuff like this pinvoke that are platform specific and so we should think about how we want to handle these things (statically at compile time vs. dynamically at runtime).

src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
```diff
@@ -0,0 +1,104 @@
...
+            using (var source = new ETWTraceEventSource(mergeTraceFile))
+            {
+
+                source.Kernel.ProcessStart += evt =>
+                {
+                    if (evt.ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (pid.HasValue)
+                        {
+                            throw new Exception("found a start when we already had a start");
```
brianrob: Do we want to handle cases where there are multiple processes spawned but we care about a specific one?  For example, the SDK?

src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
```diff
@@ -0,0 +1,104 @@
...
+                    if (evt.OldProcessID == pid && ins.ContainsKey(evt.OldThreadID))
+                    {
+                        if (!ins[evt.OldThreadID].HasValue)
+                        {
+                            return;
+                        }
+                        threadTime += evt.TimeStampRelativeMSec - ins[evt.OldThreadID].Value;
+                        ins.Remove(evt.OldThreadID);
+                    }
+                    else
```
brianrob: This should be confirming that ```evt.OldProcessID == pid```.  The if above could fail if there is a lost event and we hit a switch out first at which point it would be treated as a switch in.

src/tools/ScenarioMeasurement/Startup/TimeToMainParser.cs
```diff
@@ -0,0 +1,90 @@
...
+            using (var source = new ETWTraceEventSource(mergeTraceFile))
+            {
+                
+                source.Kernel.ProcessStart += evt =>
+                {
+                    if(evt.ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        if(pid.HasValue)
+                        {
+                            throw new Exception("found a start when we already had a start");
```
brianrob: Same thing here.  Do we need to support selection of a process where there might be more than one of the same name (e.g. dotnet.exe)?

src/tools/ScenarioMeasurement/Startup/TimeToMainParser.cs
```diff
@@ -0,0 +1,90 @@
...
+                    if (evt.OldProcessID == pid && ins.ContainsKey(evt.OldProcessID))
+                    {
+                        if (!ins[evt.OldThreadID].HasValue)
+                        {
+                            return;
+                        }
+                        threadTime += evt.TimeStampRelativeMSec - ins[evt.OldThreadID].Value;
+                        ins.Remove(evt.OldThreadID);
+                    }
+                    else
```
brianrob: Same thing here.

src/tools/ScenarioMeasurement/Startup/TimeToMainParser.cs
```diff
@@ -0,0 +1,90 @@
...
+                        if(pid.HasValue)
+                        {
+                            throw new Exception("found a start when we already had a start");
+                        }
+                        pid = evt.ProcessID;
+                        start = evt.TimeStampRelativeMSec;
+                    }
+                };
+
+                source.Kernel.ThreadCSwitch += evt =>
```
brianrob: It would be nice to only have one copy of a bunch of this logic since it's the same across parsers and is just complicated enough.  I wouldn't block on this though.

src/tools/ScenarioMeasurement/Startup/Startup.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
```
brianrob: As is, this will allow any runtime that can support netcoreapp2.2 to run this.  Do we care about automatically taking new patch versions if the repo gets upgraded by Arcade?  If so, we should specify the runtimeframeworkversion.  I'm thinking  about this from the angle of controlling variables.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -0,0 +1,178 @@
...
+                        int measurementDelay = 15,
+                        string appArgs = "",
+                        string logFileName = "",
+                        string workingDir = "",
+                        bool warmup = true,
+                        bool guiApp = true,
+                        bool skipProfileIteration = false,
+                        string reportJsonPath = "")
+        {
+            Logger logger = new Logger(String.IsNullOrEmpty(logFileName) ? $"{appExe}.startup.log" : logFileName);
```
brianrob: I'm thinking we might also want to have an option to abort all sessions - like ```perfview abort```.  Might also be good to run this at the beginning of any ```startup.exe``` run.

src/tools/ScenarioMeasurement/Startup/Startup.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <LangVersion>Preview</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.41" />
```
jorive: nit: I think you do not need this here because it's already included in the `Utils.csproj`

src/tools/ScenarioMeasurement/Util/Util.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <RootNamespace>ScenarioMeasurement</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.41" />
```
jorive: nit: I think latest version is `2.0.44`

src/tools/ScenarioMeasurement/ScenarioMeasurement.sln
```diff
@@ -0,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
jorive: Do you want to include the `..\..\Reporting\Reporting\Reporting.csproj` to the solution (included in `Startup.csproj`)?

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -0,0 +1,178 @@
...
+                    parser = new TimeToMainParser();
+                    break;
+                case MetricType.GenericStartup:
+                    parser = new GenericStartupParser();
+                    break;
+                    //case MetricType.WPF:
+                    //    parser = new WPFParser();
+                    //    break;
+            }
+            using (var kernel = new TraceEventSession(KernelTraceEventParser.KernelSessionName, kernelTraceFile))
```
jorive: You might be able to create a single trace event session for kernel and user data if you target recent Windows OS (there are some comments on TraceEvent about it), then you only need to deal with a single file and skip the merge of etl.

src/tools/ScenarioMeasurement/ScenarioMeasurement.sln
billwert: Good question. I went back and forth on this, as logically `Reporting` is an independent component used elsewhere in the system. That said, the weirdness of how the project system works differently in VS versus the CLI is annoying. I'll put it in there.

jorive: The way the `csproj` project dependencies are treated and not added automatically in the solution is weird.

src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
billwert: I've thought about this a lot. There's two common buckets of this:

1) A process starts many processes to finish a job (dotnet, roslyn, VS). In those cases the customer experience we're trying to measure is the outermost process. That's what this code currently does. If we wanted to log sub-process splits and have it show up in the reporting tools, we could build that feature. My take is that you'll see that easily enough in the profile, but I could be convinced pretty easily.
2) App models which have other launchers (`dotnet run`, UWP apps?) My intent right now is that we will publish all our tests so we have the minimal host experience, and not use `dotnet run`. (I recognize the importance of `dotnet run` and think we should test it in the context of our SDK testing.) When we start to support UWP apps, depending on the hosting model, we may need to implement the ability to specify "host process" versus "target process" so we report the correct one. We built that into the last version of these tools.

In both cases I think this is complexity we can add as needed.

src/tools/ScenarioMeasurement/Startup/Startup.csproj
billwert: I had been planning to just use the runtime under test, because I'm a huge dogfooding fan. It also simplifies our test deployment - right now we only have the runtime under test available to us in the test pipeline. We could of course pull in a 2.2 stable runtime also. 

My guess is that the runtime this runs against is very unlikely to introduce variables that matter (since this process sits mostly quiescent while the test happens), and if we do see those variables I really want to know about them. Let's chat if you think I'm wrong. :)

src/tools/ScenarioMeasurement/Startup/Startup.csproj
billwert: Good catch. I chose to remove it from `Utils.csproj` instead as that project has no usages and could be used in non-TraceEvent based things. (If I refactor more in the future I will maybe make a `TracingUtils` library.)

src/tools/ScenarioMeasurement/Startup/TimeToMainParser.cs
billwert: Yeah, this refactoring is definitely on my list.

src/tools/ScenarioMeasurement/Util/Logger.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        }
+        public void Log(string message)
+        {
+            Console.WriteLine(message);
+        }
+
+        public void LogVerbose(string message)
+        {
+            Console.WriteLine("message");
```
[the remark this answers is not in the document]
billwert: Good catch. This will get fleshed out into a more robust log solution also, of course. I just need to figure out precisely what.

src/tools/ScenarioMeasurement/Util/Util.cs
billwert: Yup.

src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
brianrob: I'm with you that scenario 2 is reasonable to bring up if/when we need it.  For scenario 1, unless I've missed something I don't think it works today.  If you were to run `dotnet build` it would look for a matching process name before process ID and if there were multiple dotnet processes then it would throw.  This would not be an issue for other scenarios that don't spawn processes of their own.

src/tools/ScenarioMeasurement/Startup/TimeToMainParser.cs
brianrob: Cool.

src/tools/ScenarioMeasurement/Startup/Startup.csproj
brianrob: I'll think about this.  Don't block on me.

src/tools/ScenarioMeasurement/Util/Logger.cs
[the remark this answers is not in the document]
brianrob: Yup, sounds good.

src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
billwert: Ah, good catch. I'll fix it.

src/tools/ScenarioMeasurement/Startup/Startup.cs
billwert: Going on the backlog.

src/tools/ScenarioMeasurement/Startup/Startup.cs
billwert: Will take a look. Would clean things up a bit. (Though I note PerfView doesn't do this either..)

billwert:
Tracking todos in #672

src/tools/ScenarioMeasurement/Startup/Startup.cs
brianrob: I'd be interested to know what version of the OS allows for this.  Unless it goes pretty far back (like Win7) I wouldn't make this change, as inevitably, we'll want to run this on old operating systems some day.

src/tools/ScenarioMeasurement/Startup/WPFParser.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Microsoft.Diagnostics.Tracing.Session;
+using Reporting;
+using System.Collections.Generic;
+
+namespace ScenarioMeasurement
+{
+    internal class WPFParser : IParser
```
jorive: Could remove this parser from this PR.

src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
```diff
@@ -0,0 +1,97 @@
...
+    internal class GenericStartupParser : IParser
+    {
+        public void EnableKernelProvider(TraceEventSession kernel)
+        {
+            kernel.EnableKernelProvider((KernelTraceEventParser.Keywords)(KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.Thread | KernelTraceEventParser.Keywords.ContextSwitch));
+        }
+
+        public void EnableUserProviders(TraceEventSession user)
+        {
+            user.EnableProvider("PerfLabGenericEventSource");
```
ooooolivia: post-merge question: can we skip creating a user provider if it's not used anywhere?

brianrob: > Can we skip creating a user provider and just having an empty method for the interface?

Yes, you can choose not to create the session if you don't need it.  Though it's probably easiest from a code perspective to not worry about that and always create it.  The cost of having the session around is negligible.

>Is "PerfLabGenericEventSource" pre-defined somewhere?

ETW does not have a hard requirement that providers are registered anywhere.  The key is that if you have something that emits an event, then you probably need something that knows how to read it.  There are a few options around this - one of them is to register providers with the operating system.  For `EventSource` implementations, we generally don't do this - instead the `EventSource` writes its metadata into the trace and then the consuming application just knows how to pull the metadata out of the trace and use it.  In terms of this app, you should expect to see a class called PerfLabGenericEventSource somewhere in the apps that are being traced, as they'll need to emit an event that represents the end of start-up.

ooooolivia: Looks like  PerfLabGenericEventSource class is commented out

brianrob: Yup.  Not sure if that's intentional or not.  @billwert?

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            if (!failed)
+            {
+                logger.Log("Parsing..");
+                TraceEventSession.Merge(new[] { kernelTraceFile, userTraceFile }, traceFileName);
+                var counters = parser.Parse(traceFileName, Path.GetFileNameWithoutExtension(appExe));
+                foreach (var counter in counters)
+                {
+                    logger.Log($"{counter.Name,-15}: {counter.Results.Average():F3} {counter.MetricName}");
```
ooooolivia: Are we going to add more statistics here, instead of just average?  It seems for non-micro-benchmarks more values like min, max, std ... could help prove the stability of the results.

src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
billwert: It's intentional. The expectation is that apps/tests that want to use it will just copy it in. There's very little chance we will meaningfully change it in the future in a way that might break some measurement (we're more likely to just make a new one). I don't want to build a separate DLL because I don't want to introduce a module load into the path of logging events.

src/tools/ScenarioMeasurement/Startup/Startup.cs
billwert: We can add anything we want.

PR #673, opened 2019-07-17
Use the azure storage api to upload files

adiaaida:
We recently released a python api for uploading blobs to azure blob
storage. Switch all of our uploading to the perflab storage container to
use the api, rather than azcopy. This also fixes uploading on arm64.


adiaaida:
Just so everyone knows: I am waiting for internal testing to show that everything is working correctly, and then I will check this in.

PR #676, opened 2019-07-18
Add System.IO.Pipelines benchmarks from AspNetCore repo, fixes #317

### LINKED ISSUE #317: Add System.IO.Pipelines benchmarks from AspNetCore repo
https://github.com/aspnet/AspNetCore/pull/4535

adamsitnik:
I am doing 2.2  vs 3.0 and I need to make sure that we did not regress the Pipelines

The code is from https://github.com/aspnet/AspNetCore/pull/4535

### DIFF
--- a/src/benchmarks/micro/corefx/System.IO.Pipelines/PipeThroughputBenchmark.cs
+++ b/src/benchmarks/micro/corefx/System.IO.Pipelines/PipeThroughputBenchmark.cs
@@ -1,0 +1,114 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Buffers;
+using System.IO.Pipelines;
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/corefx/System.IO.Pipelines/PipeThroughputBenchmark.cs
```diff
@@ -0,0 +1,114 @@
...
+        private const int InnerLoopCount = 512;
+
+        private PipeReader _reader;
+        private PipeWriter _writer;
+        private MemoryPool<byte> _memoryPool;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            _memoryPool = KestrelMemoryPool.Create();
```
pakrym: I think we can use the default pool and drop dependency on kestrel

adamsitnik: @pakrym good point, I have changed it to use the Shared one instead.

PR #682, opened 2019-07-19
Move manually triggered PR to newer OSes.

adiaaida:


adiaaida:
@billwert this PR demonstrates how we can trigger different pipelines. Basically what I have done is this:

1. Create the new pr.yml file and checked it in.
2. Go to the dnceng public piplines list and click "New."
3. Set up the pipeline as from github, dotnet/performance as the repo, path to pr.yml as the yml file.
4. Once the new pipeline is created, edit it to change the name to performance-pr, edited the pr triggers to enable secrets (for sending to helix), checked the check box for comments to require a comment to trigger on all pull requests (so that the run is only triggerable via comment).
5. Save.

Then when I create a new PR, I can trigger it by running azp run performance-pr.

When we want to run on physical hardware, we'll change the name of the queue to be the physical hardware queue, and then we'll be good. We will probably want to do that in the template and move this yml file to use the template instead of basically a copy of azure-pipelines, and I will also want to make changes there to support runs on physical hardware (so we have actual perf vs functional testing via the template). Should be a quick fix.

PR #693, opened 2019-07-23
Add ProcessTimeParser

ooooolivia:
- ProcessTimeParser measures the elapsed time of the process.
- Tested it with dotnet sdk build commands
- Todo:  1) add iteration setup and cleanup 2) add more statistics to the results

### DIFF
--- a/src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
+++ b/src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
@@ -17,7 +17,6 @@ namespace ScenarioMeasurement
 
         public void EnableUserProviders(TraceEventSession user)
         {
-            return;
         }
 
         public IList<Counter> Parse(string mergeTraceFile, string processName)
--- a/src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
+++ b/src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
@@ -1,0 +1,94 @@
+﻿using Microsoft.Diagnostics.Tracing;
+using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Diagnostics.Tracing.Session;
+using Reporting;
+using System;
+using System.Collections.Generic;
+
+
+namespace ScenarioMeasurement
+{
+    internal class ProcessTimeParser : IParser
+    {
+        public void EnableKernelProvider(TraceEventSession kernel)
+        {
+            kernel.EnableKernelProvider((KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.Thread | KernelTraceEventParser.Keywords.ContextSwitch));
+        }
+
+        public void EnableUserProviders(TraceEventSession user)
+        {
+            user.EnableProvider("PerfLabProcessTimeEventSource");
+        }
+
+        public IList<Counter> Parse(string mergeTraceFile, string processName)
+        {
+            var results = new List<double>();
+            double threadTime = 0;
+            var threadTimes = new List<double>();
+            var ins = new Dictionary<int, double>();
+            double start = -1;
+            int? pid = null;
+            var imageFileName = "";
+            using (var source = new ETWTraceEventSource(mergeTraceFile))
+            {
+
+                source.Kernel.ProcessStart += evt =>
+                {
+                    if (evt.ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (pid.HasValue)
+                        {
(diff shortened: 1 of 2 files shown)


src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
```diff
@@ -0,0 +1,94 @@
...
+    internal class ProcessTimeParser : IParser
+    {
+        public void EnableKernelProvider(TraceEventSession kernel)
+        {
+            kernel.EnableKernelProvider((KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.Thread | KernelTraceEventParser.Keywords.ContextSwitch));
+        }
+
+        public void EnableUserProviders(TraceEventSession user)
+        {
+            user.EnableProvider("PerfLabProcessTimeEventSource");
```
brianrob: You should not need to enable a user provider here because the events that you use below are only kernel events.

ooooolivia: This is what I thought. However it would break merging files in Startup.cs because no user trace file will be created. If necessary we could change the conditions of merging files?

billwert: As discussed put a check in Startup.cs for the existence of the user trace, so we don't have to turn on a fake user trace here.

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
```diff
@@ -17,7 +17,7 @@ public void EnableKernelProvider(TraceEventSession kernel)
 
         public void EnableUserProviders(TraceEventSession user)
         {
-            user.EnableProvider("PerfLabProcessTimeEventSource");
+            return;
```
billwert: because this method returns void, you don't need a return statement here. It can be an empty method body.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -120,7 +120,14 @@ static void checkArg(string arg, string name)
...
                 logger.Log("Parsing..");
-                TraceEventSession.Merge(new[] { kernelTraceFile, userTraceFile }, traceFileName);
+                if (File.Exists(userTraceFile))
+                {
+                    TraceEventSession.Merge(new[] { kernelTraceFile, userTraceFile }, traceFileName);
+                }
+                else
+                {
+                    TraceEventSession.Merge(new[] { kernelTraceFile }, traceFileName);
+                }
```
billwert: What about something like:
```cs
var files = new List<string>();
files.Add(kernelTraceFile);
if(File.Exists(userTraceFile))
{
    files.Add(userTraceFile);
}
TraceEventSession.Merge(files.ToArray(), traceFileName);
```

ooooolivia: Yah this looks nicer

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -116,7 +120,15 @@ static void checkArg(string arg, string name)
             if (!failed)
             {
                 logger.Log("Parsing..");
-                TraceEventSession.Merge(new[] { kernelTraceFile, userTraceFile }, traceFileName);
+                var files = new System.Collections.Generic.List<string>
```
billwert: nit: add the using at the top of the file instead of using the full typename here. This follows our general conventions.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -116,7 +120,15 @@ static void checkArg(string arg, string name)
...
+                {
+                    kernelTraceFile
+                };
```
billwert: nit: you can put this all on one line:
```suggestion
var files = new System.Collections.Generic.List<string> { kernelTraceFile };
```

src/tools/ScenarioMeasurement/Startup/Startup.cs
ooooolivia: second thought: this could be a learning to me and I wonder more specifically why importing and creating a list object  is better than using the if-else in terms of efficiency. Reasons I can think of: 1)  it's easier to add more than two files in the future.  2) avoid writing repetitive code.

billwert: Those are exactly the reasons I suggested this change. This approach is less efficient in terms of performance, but this isn't an area where that matters and the difference is trivial in this case.

PR #704, opened 2019-07-25
Add iteration set up to Startup Tool

ooooolivia:
- Startup.cs: Added iteration set up option before each iteration 
- ProcessTimeParser.cs: 
1) Changed it to matching event's commandLine instead, which is more accurate when events associated with setup/cleanup share the same process name with those of the actual test (ex: dotnet.exe)
2) Todo: This means we need to change the same thing for other metric types as well
- Util\ProcessHelper.cs: handle the case when the exit code is not 0 and give a ExitedWithError, instead of Success
- Todo: 
1) add iteration clean up
2) fix test log issue

### DIFF
--- a/src/tools/ScenarioMeasurement/Startup/IParser.cs
+++ b/src/tools/ScenarioMeasurement/Startup/IParser.cs
@@ -8,6 +8,6 @@ namespace ScenarioMeasurement
     {
         void EnableUserProviders(TraceEventSession user);
         void EnableKernelProvider(TraceEventSession kernel);
-        IList<Counter> Parse(string mergeTraceFile, string processName);
+        IEnumerable<Counter> Parse(string mergeTraceFile, string processName, IList<int> pids);
     }
 }
--- a/src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
+++ b/src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
@@ -19,7 +19,7 @@ namespace ScenarioMeasurement
         {
         }
 
-        public IList<Counter> Parse(string mergeTraceFile, string processName)
+        public IEnumerable<Counter> Parse(string mergeTraceFile, string processName, IList<int> pids)
         {
             var results = new List<double>();
             double threadTime = 0;
@@ -27,13 +27,12 @@ namespace ScenarioMeasurement
             var ins = new Dictionary<int, double>();
             double start = -1;
             int? pid = null;
-            var imageFileName = "";
             using (var source = new ETWTraceEventSource(mergeTraceFile))
             {
 
                 source.Kernel.ProcessStart += evt =>
                 {
-                    if (evt.ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase))
+                    if (processName.Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase) && pids.Contains(evt.ProcessID))
                     {
                         if (pid.HasValue)
                         {
@@ -42,7 +41,6 @@ namespace ScenarioMeasurement
                         }
                         pid = evt.ProcessID;
                         start = evt.TimeStampRelativeMSec;
-                        imageFileName = evt.ImageFileName;
                     }
                 };
 
@@ -70,9 +68,7 @@ namespace ScenarioMeasurement
 
                 source.Kernel.ProcessEndGroup += evt =>
                 {
-                    // not using processname here because it's not tracked by the state machine when the process just ended
-                    // but ImageFileName exists
-                    if (pid.HasValue && evt.ProcessID == pid && evt.ImageFileName == imageFileName)
+                    if (pid.HasValue && pid == evt.ProcessID)
                     {
                         results.Add(evt.TimeStampRelativeMSec - start);
                         pid = null;
--- a/src/tools/ScenarioMeasurement/Startup/Startup.cs
+++ b/src/tools/ScenarioMeasurement/Startup/Startup.cs
@@ -1,10 +1,10 @@
-﻿using Microsoft.Diagnostics.Tracing.Parsers;
-using Microsoft.Diagnostics.Tracing.Session;
-using Reporting;
-using System;
+﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Collections.Generic;
+using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Diagnostics.Tracing.Session;
+using Reporting;
 
 namespace ScenarioMeasurement
 {
@@ -35,6 +35,8 @@ namespace ScenarioMeasurement
         /// <param name="guiApp">true: app under test is a GUI app, false: console</param>
         /// <param name="skipProfileIteration">true: skip full results iteration</param>
         /// <param name="reportJsonPath">path to save report json</param>
+        /// <param name="iterationSetup">command to set up before each iteration</param>
+        /// <param name="setupArgs">arguments of iterationSetup</param>
         /// <returns></returns>
         static int Main(string appExe,
                         MetricType metricType,
@@ -42,6 +44,8 @@ namespace ScenarioMeasurement
                         string traceFileName,
                         bool processWillExit = false,
                         int iterations = 5,
+                        string iterationSetup = "",
+                        string setupArgs = "",
                         int timeout = 60,
                         int measurementDelay = 15,
                         string appArgs = "",
@@ -74,6 +78,15 @@ namespace ScenarioMeasurement
                 WorkingDirectory = workingDir,
                 GuiApp = guiApp
             };
+
+            // create iteration set up process helper
+            logger.Log($"Iteration set up: {iterationSetup} (args: {setupArgs})");
+            ProcessHelper setUpProcHelper = null;
+            if (!String.IsNullOrEmpty(iterationSetup))
+            {
+                setUpProcHelper = SetupIteration(iterationSetup, setupArgs, workingDir);
+            }
+
             Util.Init();
 
             if (warmup)
@@ -99,6 +112,8 @@ namespace ScenarioMeasurement
                     //    parser = new WPFParser();
                     //    break;
             }
+
+            var pids = new List<int>();
             using (var kernel = new TraceEventSession(KernelTraceEventParser.KernelSessionName, kernelTraceFile))
             {
                 parser.EnableKernelProvider(kernel);
@@ -107,13 +122,31 @@ namespace ScenarioMeasurement
                     parser.EnableUserProviders(user);
                     for (int i = 0; i < iterations; i++)
                     {
-                        var result = procHelper.Run();
-                        if (result != ProcessHelper.Result.Success)
+                        // set up iteration
+                        if (setUpProcHelper != null)
                         {
-                            logger.Log($"failed. result: {result}");
+                            var setUpResult = setUpProcHelper.Run().result;
+                            if (setUpResult != ProcessHelper.Result.Success)
+                            {
+                                logger.Log($"Failed to set up. Result: {setUpResult}");
+                                failed = true;
+                                break;
+                            }
+                        }
+
+                        // run iteration
+                        var runResult = procHelper.Run();
+                        if (runResult.result == ProcessHelper.Result.Success)
+                        {
+                            pids.Add(runResult.pid);
+                        }
+                        else
+                        {
+                            logger.Log($"Failed to run. Result: {runResult.result}");
                             failed = true;
                             break;
                         }
+
                     }
                 }
             }
@@ -127,7 +160,7 @@ namespace ScenarioMeasurement
                     files.Add(userTraceFile);
                 }
                 TraceEventSession.Merge(files.ToArray(), traceFileName);
-                var counters = parser.Parse(traceFileName, Path.GetFileNameWithoutExtension(appExe));
+                var counters = parser.Parse(traceFileName, Path.GetFileNameWithoutExtension(appExe), pids);
                 foreach (var counter in counters)
                 {
                     logger.Log($"{counter.Name,-15}: {counter.Results.Average():F3} {counter.MetricName}");
@@ -146,7 +179,6 @@ namespace ScenarioMeasurement
                         File.WriteAllText(reportJsonPath, reporter.GetJson());
                     }
                 }
-             
             }
 
             File.Delete(kernelTraceFile);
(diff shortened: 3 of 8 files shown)


src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
```diff
@@ -19,21 +19,21 @@ public void EnableUserProviders(TraceEventSession user)
...
             int? pid = null;
-            var imageFileName = "";
             using (var source = new ETWTraceEventSource(mergeTraceFile))
             {
 
                 source.Kernel.ProcessStart += evt =>
                 {
-                    if (evt.ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase))
+                    // comparing command line here to separate the test from set up and clean up commands with the same process name
+                    if (commandLine.Equals(evt.CommandLine.Replace("\"",""), StringComparison.OrdinalIgnoreCase))
```
ooooolivia: In commandLine value, ETW wraps the exe file with quotes so I have to remove it.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -184,5 +213,17 @@ static void checkArg(string arg, string name)
             return (failed ? -1 : 0);
 
         }
+        
+        private static ProcessHelper SetUpIteration(string command, string args, string workingDir)
+        {
+            var procHelper = new ProcessHelper()
```
ooooolivia: Need more feedback on the properties of the iteration set up process; not sure whether to make Timeout and MeasurementDelay settable.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -35,13 +35,17 @@ class Startup
...
+        /// <param name="setUpArgs">arguments of iterationSetup</param>
         /// <returns></returns>
         static int Main(string appExe,
                         MetricType metricType,
                         string scenarioName,
                         string traceFileName,
                         bool processWillExit = false,
                         int iterations = 5,
+                        string iterationSetUp = "",
+                        string setUpArgs = "",
```
billwert: nit: casing.
```suggestion
                        string setupArgs = "",
```

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -35,13 +35,17 @@ class Startup
         /// <param name="guiApp">true: app under test is a GUI app, false: console</param>
         /// <param name="skipProfileIteration">true: skip full results iteration</param>
         /// <param name="reportJsonPath">path to save report json</param>
+        /// <param name="iterationSetUp">command to set up before each iteration</param>
+        /// <param name="setUpArgs">arguments of iterationSetup</param>
```
billwert: nit: casing
```suggestion
        /// <param name="setupArgs">arguments of iterationSetup</param>
```

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -35,13 +35,17 @@ class Startup
         /// <param name="guiApp">true: app under test is a GUI app, false: console</param>
         /// <param name="skipProfileIteration">true: skip full results iteration</param>
         /// <param name="reportJsonPath">path to save report json</param>
+        /// <param name="iterationSetUp">command to set up before each iteration</param>
```
billwert: nit: casing
```suggestion
        /// <param name="iterationSetup">command to set up before each iteration</param>
```

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -35,13 +35,17 @@ class Startup
...
+        /// <param name="setUpArgs">arguments of iterationSetup</param>
         /// <returns></returns>
         static int Main(string appExe,
                         MetricType metricType,
                         string scenarioName,
                         string traceFileName,
                         bool processWillExit = false,
                         int iterations = 5,
+                        string iterationSetUp = "",
```
billwert: nit: casing
```suggestion
                        string iterationSetup = "",
```

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
```diff
@@ -108,6 +108,10 @@ public Result Run()
                     sw.WriteLine("Standard error:");
                     sw.WriteLine(error.ToString());
                 }
+                if (process.ExitCode != 0)
```
billwert: If we're going to assert exit codes, we need a configuration to know what valid exit codes are. For example, robocopy returns a 1 if everything worked and a 0 if there were no errors, but no copying was done. [Source](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/robocopy#exit-return-codes)

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
```diff
@@ -70,9 +69,7 @@ public IList<Counter> Parse(string mergeTraceFile, string processName)
 
                 source.Kernel.ProcessEndGroup += evt =>
                 {
-                    // not using processname here because it's not tracked by the state machine when the process just ended
-                    // but ImageFileName exists
-                    if (pid.HasValue && evt.ProcessID == pid && evt.ImageFileName == imageFileName)
+                    if (pid.HasValue && evt.ProcessID == pid && commandLine.Equals(evt.CommandLine.Replace("\"", ""), StringComparison.OrdinalIgnoreCase))
```
billwert: I know we had trouble with the processName field on end events. Is CommandLine valid in this case?

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
```diff
@@ -19,21 +19,21 @@ public void EnableUserProviders(TraceEventSession user)
         {
         }
 
-        public IList<Counter> Parse(string mergeTraceFile, string processName)
+        public IList<Counter> Parse(string mergeTraceFile, string commandLine)
```
billwert: You'll need to change the logic in the other parsers to account for this. Perhaps we should instead record the PID from the process executions and pass in a list here?

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
brianrob: You can probably simplify your logic here by getting the PID of the process under measurement when you start it and then save it for use during parsing.  That way, you don't have to worry about getting all of the escaping correct.

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
ooooolivia: Yes evt.CommandLine does not have the same issue with process name

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
ooooolivia: We could if the process run method can return a tuple of (result, pid) ?

src/tools/ScenarioMeasurement/Startup/Startup.cs
brianrob: If I understand correctly, MeasurementDelay shouldn't be needed here because there is no measurement happening during setup.  Timeout could be interesting because you want to make sure that setup doesn't run forever.  Is there a default timeout?

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
ooooolivia: That makes sense! Is there any other way to detect the failure instead of giving a false positive result?

src/tools/ScenarioMeasurement/Startup/Startup.cs
ooooolivia: Yes the default timeout is 60 seconds. I assume the setup won't be an GUIApp and the workingDir is the same as the main iteration.

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
billwert: Not really. Historically when we've valued exit codes we've defaulted to 0 being passing and optionally taking in a list of other acceptable exit codes. We could do something like that here if we wanted to. Did you encounter a situation where we returned non-0 and would have returned Success?

ooooolivia: Not yet. Currently we are running only dotnet.exe for the main iteration and cleanup and most likely it only returns 0 for success. If Startup Tool is only used and maintained by us, it seems checking 0 is enough.

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
```diff
@@ -19,21 +19,22 @@ public void EnableUserProviders(TraceEventSession user)
...
             int? pid = null;
-            var imageFileName = "";
+            string imageFile = "";
             using (var source = new ETWTraceEventSource(mergeTraceFile))
             {
 
                 source.Kernel.ProcessStart += evt =>
                 {
-                    if (evt.ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase))
+                    // comparing command line here to separate the test from set up and clean up commands with the same process name
```
billwert: comment is old now.

src/tools/ScenarioMeasurement/Startup/IParser.cs
```diff
@@ -8,6 +8,6 @@ internal interface IParser
     {
         void EnableUserProviders(TraceEventSession user);
         void EnableKernelProvider(TraceEventSession kernel);
-        IList<Counter> Parse(string mergeTraceFile, string processName);
+        IList<Counter> Parse(string mergeTraceFile, string processName, List<int> pids);
```
billwert: Does this have to be `List<T>` or can it be an `IList<T>`? Or maybe even `IEnumerable<T>`? (The return type here probably doesn't need to be `IList<T>` either. If you wanted to change it to `IEnumerable<T>` that'd be great.)

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
```diff
@@ -70,9 +71,7 @@ public IList<Counter> Parse(string mergeTraceFile, string processName)
 
                 source.Kernel.ProcessEndGroup += evt =>
                 {
-                    // not using processname here because it's not tracked by the state machine when the process just ended
-                    // but ImageFileName exists
-                    if (pid.HasValue && evt.ProcessID == pid && evt.ImageFileName == imageFileName)
+                    if (pid.HasValue && imageFile.Equals(evt.ImageFileName, StringComparison.OrdinalIgnoreCase) && pid == evt.ProcessID)
```
billwert: Because we're just using PID now we don't need to check the image file name here. It's not going to happen that the PID is reused while the process is still running.

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
```diff
@@ -19,21 +19,22 @@ public void EnableUserProviders(TraceEventSession user)
...
+        public IList<Counter> Parse(string mergeTraceFile, string processName, List<int> pids)
         {
             var results = new List<double>();
             double threadTime = 0;
             var threadTimes = new List<double>();
             var ins = new Dictionary<int, double>();
             double start = -1;
             int? pid = null;
-            var imageFileName = "";
+            string imageFile = "";
```
billwert: can be removed.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -74,6 +78,17 @@ static void checkArg(string arg, string name)
...
+
+            // create iteration set up process helper
+            logger.Log($"Iteration set up: {iterationSetup} (args: {setupArgs})");
+            ProcessHelper setUpProcHelper = null;
+            if (!String.IsNullOrEmpty(iterationSetup))
+            {
+                setUpProcHelper = SetupIteration(iterationSetup, setupArgs, workingDir);
+            }
+
+            // TODO: create clean up proc helper
```
billwert: Are you going to do this now? If not let's remove the TODO and we'll just know we need to come back to this if we ever have a scenario that warrants it.

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
```diff
@@ -38,7 +38,7 @@ public class ProcessHelper
         /// <param name="appArgs">Optional arguments</param>
         /// <param name="workingDirectory">Optional working directory (defaults to current directory)</param>
         /// <returns></returns>
-        public Result Run()
+        public (Result, int) Run()
```
billwert: Further thought from our conversation. This is fine as it is. One minor advantage of moving the names here (instead of tuple deconstruction) is that it gives you an idea of what the intent of these fields is when you are looking at the method. `(Result result, int pid)` tells me immediately what is going on. Still up to you but I think it's a minor point in favor of a change.

src/tools/ScenarioMeasurement/Startup/ProcessTimeParser.cs
ooooolivia: I guess it is still the same concern that Brian mentioned - there's a small chance the pid could be recycled in the same tracing session. I assume it means after one process ended, the next process would (for a very little chance) reuses the pid in the same tracing session. I'm not sure what you mean by "the process is still running".

billwert: This is the end event for the process. It's incredibly unlikely to get a process end event for the PID we have stored for another instance of the process. For this to happen we would need:

1) We capture a PID and store it into `pid`
2) We miss the end event for whatever reason, and do not execute this code which clears `pid`
3) We miss the process start event for another instance of our test process **using the same PID**
4) We capture the end event for this second process.

In light of this I think we can simplify and not capture `imageFile`.

DrewScoggins: I think that number three is not quite correct. It does not have to be our test process that starts, as any process that got assigned the recycled pid would trigger this ProcessEngGroup code, and we would not even have to miss the start event for this process, because it would not have the same name as our target process and we would never run that code.

billwert: Sorry, I was describing the situation where `imageFile` saves us (but wasn't clear about that), which I think is very unlikely. I agree that some other process could use the PID and we'd get into a weird state if we miss 2. (I also think that's very very unlikely, particularly since we're using such low volume eventing for measurement.)

src/tools/ScenarioMeasurement/Startup/IParser.cs
ooooolivia: Yes. However the most specific type the parameter can be is IList because it uses the "Contains" method below.

PR #721, opened 2019-07-31
Add benchmarks for special cased LINQ methods

jvandertil:
This pull request adds benchmarks for LINQ methods that have special cased implementations for `List<T>` and `T[]` as implemented in https://github.com/dotnet/corefx/pull/39200.

Since `Single` and `SingleOrDefault` use the same code path internally, the benchmarks for `SingleOrDefault` have been removed (in line with `First` and `FirstOrDefault`).

This pull request is not completely finished yet, links to GitHub still have to be added.
Should I update those after the PR is merged, or how can I make sure the link stays valid while the PR is open?.

`Any(predicate)` does not have a really interesting code path, it delegates to `TryGetFirst()` internally, remove the benchmarks for `Any(pred)`? 

/cc @adamsitnik

### DIFF
--- a/src/benchmarks/micro/corefx/System.Linq/Perf.Enumerable.cs
+++ b/src/benchmarks/micro/corefx/System.Linq/Perf.Enumerable.cs
@@ -69,10 +69,13 @@ namespace System.Linq.Tests
 
         public IEnumerable<object> FirstPredicateArguments()
         {
-            // First(predicate) has 2 code paths: OrderedEnumerable and IEnumerable
+            // First(predicate) has 4 code paths: OrderedEnumerable, Array, List, and IEnumerable
+            // TODO: Update link to GitHub.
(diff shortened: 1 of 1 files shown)


adamsitnik:
Hi @jvandertil 

> This pull request is not completely finished yet, links to GitHub still have to be added.
Should I update those after the PR is merged, or how can I make sure the link stays valid while the PR is open?.

I think that you can replace the permalinks with regular links to master so when your PR to CoreFX gets merged the links are still going to point to the right thing. 

Example:

```diff
- https://github.com/dotnet/corefx/blob/dcf1c8f51bcdbd79e08cc672e327d50612690a25/src/System.Linq/src/System/Linq/First.cs
+ https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/First.cs
```

As soon as you do that I am going to merge this PR.


Thanks!

jvandertil:
Hey @adamsitnik, I've updated the links for the benchmarks that I hit with this PR. If you want I could also update the other links, but they should still point to the correct implementation.

PR #722, opened 2019-08-01
Add iteration cleanup to Startup Tool

ooooolivia:
- Add iteration cleanup (it follows the same pattern as setup but happens after each iteration)
- Minor changes to variable naming
- The test logs turned out to be fine! Will investigate in why it disappeared previously


ooooolivia:
Merged with the dotnet-maestro's dependency update and pushed again. Will try to submit a pull request in the morning next time.

PR #756, opened 2019-08-09
 Add netcoreapp5.0, move netcoreapp3.0 to release/3.0.1xx channel

adiaaida:
This change adds netcoreapp5.0, since that is now what is pulled down when we download dotnet from the master channel.

It moves netcoreapp3.0 to pull down a release/3.0.1xx dotnet, since the 3.0 channel doesn't seem to be functional.

It does not move us to using netcoreapp5.0 because we get the following error:

```
[2019/08/09 09:32:29][INFO] The provided runtime "netcoreapp5.0" is invalid. Available options are: clr, core, corert, mono, net461, net462, net47, net471, net472, netcoreapp2.0, netcoreapp2.1, netcoreapp2.2, netcoreapp3.0.
```

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -261,10 +261,7 @@ def __main(args: list) -> int:
                 '**',
                 '*perf-lab-report.json')
 
-            returncode = upload.upload(globpath, 'results', 'PERFLAB_UPLOAD_TOKEN', 'https://pvscmdupload.blob.core.windows.net')
-
-            if returncode != 0:
-                return returncode
+            upload.upload(globpath, 'results', 'PERFLAB_UPLOAD_TOKEN', 'https://pvscmdupload.blob.core.windows.net')
                 
         # TODO: Archive artifacts.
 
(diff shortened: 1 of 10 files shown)


adiaaida:
@adamsitnik is the error I'm seeing for netcoreapp5.0 from BenchmarkDotNet?

adiaaida:
@adamsitnik can you also help me understand why for the projects that originally only had 1 target framework, it now can't find the exes (because they're being built under Release/<framework> instead of just Release)? What needs to be added to make those look in the right place?

adiaaida:
Checking this in to get us green. I'll figure out the netcoreapp5.0 for ML/Roslyn later. And we can turn it on for 5.0 once BDN is ready.

adamsitnik:
@adiaaida yes, we are using an old version of BDN that does not support .NET Core 3.1, .NET Core 5.0 and .NET 4.8

The good thing is that I've already implemented this support in https://github.com/dotnet/BenchmarkDotNet/pull/1188

I am going to send a separate PR to see if the support that I've implemented works

PR #511, opened 2019-05-27
update BenchmarkDotNet to latest version to take advantage of improvements

adamsitnik:
The update contains a lof of minor improvements and bug fixes, but the most important are:

### "Use new .NET Core 3.0 API to get the total number of allocated bytes for all threads" 

https://github.com/dotnet/BenchmarkDotNet/commit/f54055a23bbf8490499166ea90947c966619c26c So far BDN was using `GC.GetAllocatedBytesForCurrentThread` which returns the value for **current thread only**. Recently new API  was added, `GC.GetTotalAllocatedBytes` which returns the value of allocated managed memory for all threads. By using it we get the proper numbers. For multi-threaded benchmarks (few dozens in the enitre portfolio) we are going to get new, bigger numbers. This is a change of reported values, but also a **bug fix** (cc @stephentoub).

### "Use Median instead of Mean for overhead calculations" 

https://github.com/dotnet/BenchmarkDotNet/commit/d9901baacd89a212ee6d1e46fc6ec5dabc0e9d77  

One of the techniques BDN uses to allow for accurate nano-benchmarking is overhead deduction. BDN measures the overhead by benchmarking a method of same return type and arguments as the benchmarked method. The difference is that the "overhead" method does nothing for `void` and returns `default` for methods returning anything. 

So far the actual results was: 

```
Result(n) = Actual(n) - Average(Overhead)
```

But as @AndyAyersMS has spotted, using `Average` for deducing the overhead is bad when we have some outliers that affect the Average:

```log
1.69
1.65
1.70
1.83
2.60 <-- an outlier -->
1.76
1.69
1.69
1.73
1.64
1.66
1.67
1.71
1.68
1.64
```

The implementation was changed, now BDN uses Median.

How it's going to affect the reported results? The nano-benchmarks are going to get more stable results, we are going to have fewer results affected by outliers. 

### Power Plans

By default BDN is now enforcing High Performance Power plan when running benchmarks on Windows. https://github.com/dotnet/BenchmarkDotNet/commit/0dfa37efbba24f1384f2c9726f2dcce30827bc26

To make sure it does not affect the results reported in our lab, I have disabled this behaviour in the RecommendedConfig type (`.DontEnforcePowerPlan()`). So this new feature is not going to affect the results, we might consider enabling it in the futre.

### Other improvements

[...]

PR #758, opened 2019-08-10
[NO MERGE] check if new version of BDN supports .NET Core 5.0

adamsitnik:
The purpose of this PR is to check if the new version of BDN supports .NET Core 5.0 and everything works as expected.

/cc @adiaaida @billwert

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -50,6 +50,7 @@ jobs:
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
       runCategories: 'coreclr corefx' 
       frameworks: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
+        - netcoreapp5.0
         - netcoreapp3.0
         - netcoreapp2.2
         - netcoreapp2.1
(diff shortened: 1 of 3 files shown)


adiaaida:
Looks like everything checked out. So if you want to change this from a draft to a real PR, looks good to me.

billwert:
@adamsitnik did we figure out the remaining diffs?

billwert:
@adamsitnik I see now that this build probably is the CI job for dotnet/BenchmarkDotNet#1188. I am curious if there's a path where BDN can be more accepting to TFMs it doesn't know about, in line with the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle). While I agree that we should be updating BDN with some frequency, I would very much like to never be broken in this manner again.

adiaaida:
I just want to note that getting an update to BDN that supports netcoreapp5.0 is blocking our ability to run per-commit builds in coreclr and corefx (a fact that I just realized while talking to Jarret). So if we can up the priority on figuring out why we're seeing the weird performance impact in your other PR, and get a stable version of BDN that supports netcoreapp5.0, that would be great!

adiaaida:
Thought: could we conditionalize the package reference for netcoreapp5.0 so that we could get that working without affecting any of the earlier releases until we understand the performance differences we saw? We've already switched to the tigers from the snakes, so there's already a break in our graphs, and this seems like a good time to make that switch.

@billwert @adamsitnik

adamsitnik:
I've updated #511 so I am closing this one

PR #764, opened 2019-08-13
[WIP] Test pr a/b testing

adiaaida:


adiaaida:
@billwert @DrewScoggins This is what I've put together for A/B testing for PRs. Right now it uses ResultsComparer, which will print out to the log any regressions/improvements, which I think could be pretty useful, though not when it's so hard to get to our logs. This change would allow repos to do the following:

* Supply a baseline CoreRun and a diff CoreRun and compare the results.
* Only supply a diff CoreRun and then compare against the latest sdk (particularly useful if they don't want to figure out how to get an n-1 build of their repo).
* Run the comparison on physical hardware.

What this doesn't do:

* Write any results to file which could then be uploaded
* Any sort of decision on whether or not the results are good
* Give guidance on how to get a baseline build of the product.
* Use Drew's tool

If we want to write results to a file, we may then be able to save them to a place that helix could find them, and then write a parser that will report back to AzDO regressions (in the form of failed tests). I'm not sure if we really want to report failure on builds due to perf, but the only way we get the data saved is if a run is deemed "failing." Also, since the comparer is running as a post step, that won't fail the build (only the workitem command can fail the build).

billwert:
Good progress! Questions:

>  * Supply a baseline CoreRun and a diff CoreRun and compare the results.
> * Only supply a diff CoreRun and then compare against the latest sdk (particularly useful if they don't want to figure out how to get an n-1 build of their repo).

When a PR happens and produces a build for testing, what is its baseline? Does it get merged up to the tip of the branch before it runs tests? If not, we should probably require the first option. If it does get merged up, we can use the second option. My concern is that if the base of the PR is actually N-10, and we test against N-1, we risk phantom regressions slipping into the PR run and losing trust the service. 

> * Any sort of decision on whether or not the results are good
> * Use Drew's tool

Obviously we'll only need one of these. For now using ResultsComparer is fine and we can swap in the better tool when we unify it with ResultsComparer.



> * Write any results to file which could then be uploaded (and paragraph associated.)

I wonder if the right thing is to request the ability to mark a pipeline/run/etc as "keep all logs/results" as a feature. I can understand why most folks don't want this, but we definitely do.

Ideally I'd like this wired back to GitHub so that the person who submits the PR can easily see the results. What does it take to have a bot comment for us in this instance?

adiaaida:
> When a PR happens and produces a build for testing, what is its baseline? 

When a PR happens, runs only get scheduled the change can be merged into master. So N-1 would be tip of master, N would be tip of master + any commits in the PR. The question is whether or not tip of master = the current most recent sdk in dotnet (ie, has it made it all the way through the pipeline and been published or not). The answer there is likely (and by likely, I mean almost definitely) no. So the goal would be to get the repo owners to figure out how they'd produce an N-1 build (probably, build current PR, save. git checkout HEAD~1 or some such, build again, save again). But we will need to work with the repo owners to decide the best way for them to do this, and it might be different in various repos. Certainly their builds will be.

> I wonder if the right thing is to request the ability to mark a pipeline/run/etc as "keep all logs/results" as a feature.

I'll talk to EngServices people. I also wonder if we save the results to some file in the upload directory, if they will be uploaded. I could do some experimenting. We could also do something like (pseudo-code):

```
dotnet run resultscomparer >> %HelixUploadFolder%\resultscomparer.log
type %HelixUploadFolder%\resultscomparer.log
```
And get both the results in a file and in the console log.

> Ideally I'd like this wired back to GitHub so that the person who submits the PR can easily see the results. What does it take to have a bot comment for us in this instance?

To be honest, I have absolutely no idea. I know that the maestro bot will write a comment "I'm waiting on XYZ jobs to complete," but I don't know what it would take for a bot to be able to know what job its looking at, find the correct helix correlation id, read the results file, and then write markdown as a comment (or even just a link to that file). I can talk to EngServices here too, to see if they have any thoughts. I have never worked with any of our bots.

adiaaida:
After discussing with @alexperovich and @billwert, we are going to get ResultsComparer to spit out xml results that we already have parsers for so that we can report regressions/improvements through AzDO's test reporting.

scripts/dotnet.py
```diff
@@ -511,6 +511,15 @@ def remove_dotnet(architecture: str) -> str:
     '''
     rmtree(__get_directory(architecture))
 
+def shutdown_server(verbose:bool) -> None:
+    '''
+    Shutsdown the dotnet server
+    '''
+    cmdline = [
+        'dotnet', 'build-server', 'shutdown'
```
adamsitnik: nice, I had no idea that such command exists! 👍

src/tools/ResultsComparer/Program.cs
```diff
@@ -185,6 +196,75 @@ private static void ExportToCsv((string id, Benchmark baseResult, Benchmark diff
             Console.WriteLine($"CSV results exported to {csvPath.FullName}");
         }
 
+        private static void ExportToXml(List<string> slower, List<string> faster, FileInfo xmlPath)
```
adamsitnik: I think that it would be better to implement a method similar to `ExportToCsv` that uses strong types instead of strings. Strings need to be split and work until we change the way the results are printed.

adiaaida: I wanted to be able to separate out the slower benchmarks from the faster benchmarks without having to implement the math twice. The XML file is only meant to be read by the xunit parser in Helix. Maybe I should change the parameter name and function name to make that clearer? This is also a stopgap until Drew implements this in the new visualization tool.

adamsitnik: > I wanted to be able to separate out the slower benchmarks from the faster benchmarks without having to implement the math twice

this information is available in the `conclusion` field. 


```cs
ExportToXml(notSame, args.XmlPath);

private static void ExportToXml((string id, Benchmark baseResult, Benchmark diffResult, EquivalenceTestConclusion conclusion)[] notSame, FileInfo xmlPath)
{
    using (XmlWriter writer = XmlWriter.Create(xmlPath.Open(FileMode.OpenOrCreate, FileAccess.Write, FileShare.Write)))
    {
        foreach (var slower in notSame.Where(x => x.conclusion == EquivalenceTestConclusion.Slower))
        {
            writer.WriteStartElement("test");
            writer.WriteAttributeString("name", slower.id);
            writer.WriteAttributeString("type", slower.baseResult.Type);
            writer.WriteAttributeString("method", slower.baseResult.Method);
            writer.WriteAttributeString("time", "0");
            writer.WriteAttributeString("result", "Fail");
            writer.WriteStartElement("failure");
            writer.WriteAttributeString("exception-type", "Regression");
            writer.WriteElementString("message", $"{slower.id} has regressed, was {slower.baseResult.Statistics.Median} is {slower.diffResult.Statistics.Median}");
            writer.WriteEndElement();
        }
    }
}
```

adiaaida: Ok, sure.

adiaaida: That worked really well. Thanks for the insight Adam. I'll be putting this part of the change as a separate PR.

scripts/dotnet.py
adiaaida: Yeah, I knew about it but had to re-look it up when I started working on this change.

adiaaida:
@billwert I also wanted to mention that you can see the failing results (aka regressions) on the checks page: https://github.com/dotnet/performance/pull/764/checks?check_run_id=200645699, which almost gets you what you wanted when you mentioned having AzDO report back to the github job regressions.

billwert:
That's pretty good!

adiaaida:
This has been checked into arcade!

PR #768, opened 2019-08-14
Fix artifact directory for windows PerfSnakes and arm64

adiaaida:
We added conditionals to only use the upload directory for new hardware, but forgot about arm64, and didn't update the osVersion for the windows perfsnakes correctly.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -297,7 +297,7 @@ jobs:
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: windows
-      osVersion: 10
+      osVersion: RS1
       kind: micro
       architecture: x64
       pool: Hosted VS2017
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -312,7 +312,7 @@ jobs:
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: windows
-      osVersion: 10
+      osVersion: RS1
```
billwert: this is probably an unrelated question but I thought we were going to move these to an RS5 pool of some sort..

adiaaida: I was thinking that if we wanted to keep the graphs consistent, we would do that post-release of 3.0, at which point, the machines will get all of the helix updates that allow them to use the artifacts directory.

PR #772, opened 2019-08-15
Add run kind to config

DrewScoggins:


eng/performance/benchmark_jobs.yml
```diff
@@ -106,7 +106,7 @@ jobs:
             ${{ each framework in parameters.frameworks }}:
               ${{ framework }}:
                 _Framework: ${{ framework }}
-                _Configs: CompilationMode=Tiered
+                _Configs: CompilationMode=Tiered;RunKind=${{ parameters.kind }}
```
billwert: Will this enforce a requirement that these never have spaces in them? I don't think we have any that do today but who knows what someone would type in the future?

DrewScoggins: I don't believe so. Do we have the equivalent of String.Replace in YAML? Are you thinking that we should quote this string?

DrewScoggins:
This needs to happen to make it easier to ensure that we are grabbing the right build in the report generator.

eng/performance/benchmark_jobs.yml
billwert: Just asking - seems like the kind of thing that might bite us later, but I don't have a specific concern.

eng/performance/benchmark_jobs.yml
```diff
@@ -106,7 +106,7 @@ jobs:
             ${{ each framework in parameters.frameworks }}:
               ${{ framework }}:
                 _Framework: ${{ framework }}
-                _Configs: CompilationMode=Tiered
+                _Configs: CompilationMode=Tiered;RunKind="${{ parameters.kind }}"
```
adiaaida: ci-setup.py's --build-configs option takes a space separated list. My suggestion would be to make each of these options space separated and put the quotes around the entire key=value, not around just the value. I.e. change it to

```
_Configs: "CompilationMode=Tiered" "RunKind=${{parameters.kind}}" 
```

But we should also check the logs to make sure that python isn't doing wonky things with the quotation marks.

billwert: we have code in another place that parses this as semi-colon delimited, though.

We should pick one and make everything align.

billwert: (I vote for semicolons..)

adiaaida: Configs is only used by ci_setup.py.

billwert: Ahh. It's super confusing that ci_setup is what makes this the semicolon delimited string that gets parsed later. Why'd we do it that way instead of representing the string the same way everywhere?

DrewScoggins: I honestly don't care which of these that we do, so for now I am just going to go with the spaces and quotes. It does what we want, and is probably the safest.

adiaaida: I was just following the example of other arguments that have multiple inputs. I guess it didn't really matter.

PR #780, opened 2019-08-19
Some minor .NET Core 5.0 fixes

adamsitnik:
I forgot that we have some `#if NETCOREAPP3_0` statements which would make given benchmarks run for .NET Core 3.0 but not 5.0 ;)

I've used `#if !NETFRAMEWORK && !NETCOREAPP2_1 && !NETCOREAPP2_2` as a fix, it does not look perfect but as far as I know there is no way to express something like "#if newer than NETCOREAPP3_0`. 

I've also fixed .csprojs that VS was complaining about.

### DIFF
--- a/src/benchmarks/micro/coreclr/System.Reflection/Activator.cs
+++ b/src/benchmarks/micro/coreclr/System.Reflection/Activator.cs
@@ -21,7 +21,7 @@ namespace System.Reflection
         [Benchmark]
         public object CreateInstanceType() => System.Activator.CreateInstance(typeof(T));
 
-#if NETFRAMEWORK || NETCOREAPP3_0 // API available in Full .NET Framework and .NET Core 3.0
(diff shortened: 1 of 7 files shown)


adiaaida:
LGTM. Don't worry about this in this PR if you think it is out of scope, but did we update Roslyn/ML to work on netcoreapp5.0 yet?

adiaaida:
Also adding Drew since I will be moving back to Eng Services in 2 weeks.

DrewScoggins:
LGTM

gfoidl:
> we have some #if NETCOREAPP3_0 statements which would make given benchmarks run for .NET Core 3.0 but not 5.0

Therefore patterns like
```c#
#if NETCOREAPP2_1
    // ...
#elif NETCOREAPP3_0
    // ...
#else
    #error Update TargetFrameworks
#endif
```
can be used. So an update can't be forgotten as an error will be fired by the compiler.

> "#if newer than NETCOREAPP3_0`.

There's an issue around for that, but I don't find it right now unfortunately.

PR #783, opened 2019-08-20
Add XUnit-style xml output for ResultsComparer

adiaaida:
Running in AzDO, helix has support for parsing xml output from test legs
if the xml is in a particular format. This change adds the --xml option
which enables us to output these xunit-style xml output files that we
can use when running ResultsComparer in the lab to compare two
performance runs.

### DIFF
--- a/src/tools/ResultsComparer/Program.cs
+++ b/src/tools/ResultsComparer/Program.cs
@@ -9,6 +9,7 @@ using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading;
+using System.Xml;
 using BenchmarkDotNet.Mathematics;
 using BenchmarkDotNet.Mathematics.StatisticalTesting;
 using CommandLine;
@@ -57,6 +58,7 @@ namespace ResultsComparer
             PrintTable(notSame, EquivalenceTestConclusion.Faster, args);
 
             ExportToCsv(notSame, args.CsvPath);
+            ExportToXml(notSame, args.XmlPath);
         }
 
         private static IEnumerable<(string id, Benchmark baseResult, Benchmark diffResult, EquivalenceTestConclusion conclusion)> GetNotSameResults(CommandLineOptions args, Threshold testThreshold, Threshold noiseThreshold)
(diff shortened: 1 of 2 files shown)


src/tools/ResultsComparer/Program.cs
```diff
@@ -185,6 +187,54 @@ private static void ExportToCsv((string id, Benchmark baseResult, Benchmark diff
...
+             if (xmlPath.Exists)
+                xmlPath.Delete();
+
+            using (XmlWriter writer = XmlWriter.Create(xmlPath.Open(FileMode.OpenOrCreate, FileAccess.Write, FileShare.Write)))
+            {
+                writer.WriteStartElement("performance-tests");
+                foreach (var slower in notSame.Where(x => x.conclusion == EquivalenceTestConclusion.Slower))
+                {
+                    writer.WriteStartElement("test");
+                    writer.WriteAttributeString("name", slower.id);
```
adamsitnik: does `WriteAttributeString` takes care of escaping the string?

benchmark id contains the benchmark arguments which can be literally anything, including `<` and `>` characters. Example:

```cs
[Benchmark]
[Arguments(">>>>")]
public int Example(string arg) => arg.GetHashCode();
```

src/tools/ResultsComparer/Program.cs
```diff
@@ -185,6 +187,54 @@ private static void ExportToCsv((string id, Benchmark baseResult, Benchmark diff
...
+                    writer.WriteElementString("reason", $"{faster.id} has improved, was {faster.baseResult.Statistics.Median} is {faster.diffResult.Statistics.Median}.");
+                    writer.WriteEndElement();
+                }
+
+                writer.WriteEndElement();
+                writer.Flush();
+            }
+
+            Console.WriteLine($"XML results exported to {xmlPath.FullName}");
+            return;
```
adamsitnik: nit: the return is not needed

```suggestion
```

src/tools/ResultsComparer/Program.cs
```diff
@@ -185,6 +187,54 @@ private static void ExportToCsv((string id, Benchmark baseResult, Benchmark diff
...
+                    writer.WriteAttributeString("type", faster.baseResult.Type);
+                    writer.WriteAttributeString("method", faster.baseResult.Method);
+                    writer.WriteAttributeString("time", "0");
+                    writer.WriteAttributeString("result", "Skip");
+                    writer.WriteElementString("reason", $"{faster.id} has improved, was {faster.baseResult.Statistics.Median} is {faster.diffResult.Statistics.Median}.");
+                    writer.WriteEndElement();
+                }
+
+                writer.WriteEndElement();
+                writer.Flush();
```
adamsitnik: good idea to flush the writer! 👍

src/tools/ResultsComparer/Program.cs
DrewScoggins: It does, I just tested it.

src/tools/ResultsComparer/Program.cs
DrewScoggins: Doesn't flush get called as part of disposing the writer object from the using?

adiaaida: TBH, I copied the flush from the code I found when I googled "C# write to xml file"

adiaaida: Internet suggests that inside the using statement it isn't necessary. Anyone have any strong opinions?

DrewScoggins: Not really.

PR #784, opened 2019-08-20
Gracefully shutdown the dotnet build-server after run

adiaaida:
The dotnet process keeps running, even after a build or run has
completed. Before exiting the pyhton script, call dotnet build-server
shutdown to gracefully shutdown dotnet so we do not have zombie
processes still running.

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -513,7 +513,7 @@ def remove_dotnet(architecture: str) -> str:
 
 def shutdown_server(verbose:bool) -> None:
     '''
-    Shutsdown the dotnet server
+    Shuts down the dotnet server
     '''
     cmdline = [
         'dotnet', 'build-server', 'shutdown'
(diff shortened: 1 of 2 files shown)


billwert:
Am I to infer that Helix isn't able to kill processes when it tries to delete work item / correlation directories?

adiaaida:
Helix can, but on the a/b comparison, our scripting tries to delete the dot net directory, but it’s still running, so we crash and burn. This basically helps bypass that issue. It also fixes the issue when working locally, which is something I do, so it’s nice to have

DrewScoggins:
LGTM

scripts/dotnet.py
```diff
@@ -511,6 +511,16 @@ def remove_dotnet(architecture: str) -> str:
     '''
     rmtree(__get_directory(architecture))
 
+def shutdown_server(verbose:bool) -> None:
+    '''
+    Shutsdown the dotnet server
```
adamsitnik: nit: typo
```suggestion
    Shuts down the dotnet server
```

billwert:
Got it. I was curious because we're going to have testing that doesn't run through benchmarks_ci soon enough. We'll need to keep this in mind.

PR #786, opened 2019-08-21
Allow multiple frameworks for ResultsComparer

adiaaida:


scripts/ci_setup.py
```diff
@@ -233,10 +233,12 @@ def __main(args: list) -> int:
                 out_file.write(variable_format % ('PERFLAB_BUILDTIMESTAMP', source_timestamp))
                 out_file.write(variable_format % ('PERFLAB_CONFIGS', config_string))
                 out_file.write(variable_format % ('DOTNET_VERSION', dotnet_version))
+                out_file.write(variable_format % ('PYTHON_SCRIPT_TARGET_FRAMEWORKS'), framework)
 
         else:
             with open(args.output_file, 'w') as out_file:
                 out_file.write(variable_format % ('PERFLAB_INLAB', '0'))
+                out_file.write(variable_format % ('PYTHON_SCRIPT_TARGET_FRAMEWORKS'), framework)
```
billwert: I like this - we have other tools that will fall into this problem. If we're just going to set this globally, can we do a slightly larger change and just rename the variable to `PERFLAB_TARGET_FRAMEWORKS`? The `benchmarks_ci.py` script can still set/use it.

src/tools/ResultsComparer/ResultsComparer.csproj
```diff
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>$(PYTHON_SCRIPT_TARGET_FRAMEWORKS)</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">netcoreapp2.1;netcoreapp3.0;netcoreapp5.0</TargetFrameworks>
```
billwert: I wonder if in this case we should go the other way - leave it as 2.1 in the case that it isn't set. That will work fine on almost anyone's machine, and in the lab we'll have `netcoreapp5.0` where we need it. (`Microbenchmarks.csproj` is more of a special case because we expect it to only run with latest CLIs. Other tools can be more relaxed.)

src/tools/ResultsComparer/Program.cs
```diff
@@ -136,7 +136,7 @@ private static void PrintTable((string id, Benchmark baseResult, Benchmark diffR
 
             var table = data.ToMarkdownTable().WithHeaders(conclusion.ToString(), conclusion == EquivalenceTestConclusion.Faster ? "base/diff" : "diff/base", "Base Median (ns)", "Diff Median (ns)", "Modality");
 
-            foreach (var line in table.ToMarkdown().Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries))
+            foreach (var line in table.ToMarkdown().Split(Environment.NewLine))
```
billwert: What will be an empty entry now?

adiaaida: That was a mistake. Has been undone.

src/tools/ResultsComparer/ResultsComparer.csproj
adiaaida: Smart.

scripts/ci_setup.py
adiaaida: Would you want it changed to PERFLAB_TARGET_FRAMEWORKS everywhere, or just in this case?

billwert: Everywhere. (Which is really just a few more `csproj` files and the python script that sets it, I think.)

adiaaida: Cool.

src/benchmarks/real-world/Microsoft.ML.Benchmarks/Microsoft.ML.Benchmarks.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
+    <TargetFrameworks>$(PERFLAB_TARGET_FRAMEWORKS)</TargetFrameworks>
```
billwert: I think this needs the same conditional as the other projects.

src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
```diff
@@ -4,6 +4,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
+    <TargetFrameworks>$(PERFLAB_TARGET_FRAMEWORKS)</TargetFrameworks>
```
billwert: I think this needs the same conditional as the other projects.

scripts/ci_setup.py
adiaaida: I updated it in MicroBenchmarks.csproj, and added it to Roslyn and ML. If Roslyn/ML fail, I'll probably back out that change.

billwert: Fair!

PR #817, opened 2019-08-26
Change logging of Startup Tool

ooooolivia:
1)	Log the output of subprocesses in the main window, instead of creating a .log file for each iteration (it’s hard to find which file contains the error when there are many iterations) 
2)	Give more statistics like max and min

### DIFF
--- a/src/tools/ScenarioMeasurement/Startup/Startup.cs
+++ b/src/tools/ScenarioMeasurement/Startup/Startup.cs
@@ -1,10 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Data;
 using System.IO;
 using System.Linq;
 using Microsoft.Diagnostics.Tracing.Parsers;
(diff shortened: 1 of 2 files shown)


src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -252,5 +252,18 @@ private static ProcessHelper CreateProcHelper(string command, string args, strin
             };
             return procHelper;
         }
+
+        private static void WriteResultTable(IEnumerable<Counter> counters, Logger logger)
+        {
+            logger.Log($"{"Metric",-15}|{"Average",-15}|{"Max",-15}|{"Min",-15}");
```
brianrob: Nit: I would propose that Min be written before max - this is commonly what folks expect.

ooooolivia: good point!

brianrob:
LGTM.  Thanks!

PR #836, opened 2019-08-29
Add new perf benchmark for generic dictionary expansion

fadimounir:
@billwert PTAL

### DIFF
--- a/src/benchmarks/micro/coreclr/perflab/DictinoaryExpansion.cs
+++ b/src/benchmarks/micro/coreclr/perflab/DictinoaryExpansion.cs
@@ -1,0 +1,158 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace PerfLabTests
+{
+    public class GenClass<T>
+    {
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public Type FuncOnGenClass(int level)
+        {
+            switch (level)
+            {
+                case 0: return typeof(T);
+                case 1: return typeof(List<T>);
+                case 2: return typeof(List<List<T>>);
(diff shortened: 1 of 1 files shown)


billwert:
I’ll take a look when I am back from vacation. Thanks!

src/benchmarks/micro/coreclr/perflab/DictinoaryExpansion.cs
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) Microsoft. All rights reserved.
```
adamsitnik: the file name contains a typo  `Dictinoary`

src/benchmarks/micro/coreclr/perflab/DictinoaryExpansion.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public static void FuncOnGenClassTest()
+        {
+            var o1 = new GenClass<string>();
+            var o2 = new GenClass<object>();
+            var o3 = new GenClass<DictionaryExpansion>();
+
+            for (int j = 0; j < DictionaryExpansion.s_Iterations; j++)
```
adamsitnik: You should not need to add a loop here, BenchmarkDotNet scales the number of inner iterations using a dynamic heuristic based on a `IterationTime` setting:

https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#Loops

src/benchmarks/micro/coreclr/perflab/DictinoaryExpansion.cs
```diff
@@ -0,0 +1,158 @@
...
+                    GenFunc2<object>(i);
+
+                for (int i = 0; i < 15; i++)
+                    GenFunc<DictionaryExpansion>(i);
+            }
+        }
+
+        public static int s_Iterations = 100000;
+
+        [Benchmark]
```
adamsitnik: would it be possible to add some comment here explaining what the benchmark is measuring? A link to some coreclr docs would be enough

fadimounir:
Thanks for reviewing this @adamsitnik. Should be ready for merging now.

PR #849, opened 2019-09-03
docs: profiling local CoreFX build using VTune

adamsitnik:
This is a short summary of the recent @tannergooding talk.

https://github.com/adamsitnik/performance/blob/081b701626182d249db0964a038f9d7af217e9e1/docs/profiling-workflow-corefx.md#VTune

@tannergooding could you please take a look?


docs/profiling-workflow-corefx.md
```diff
@@ -18,6 +18,12 @@
       - [Analyzing the Results](#Analyzing-the-Results)
       - [Viewing Source Code](#Viewing-Source-Code)
       - [Identifying Regressions](#Identifying-Regressions)
+  - [VTune](#VTune)
```
tannergooding: This should have a separate section or at least briefly call out `AMD uProf` which is the similar program for AMD machines.

adamsitnik: good point, we should definitely have AMD uProf section

@tannergooding do I remember correctly from your presentation that I can run AMD uProf on Intel machine? So I should be able to create similar docs myself without need to access AMD machine?

docs/profiling-workflow-corefx.md
```diff
@@ -389,3 +395,156 @@ PerfView has a built-in support for identifying regressions. To use it you need
...
+
+When we open the Flame Graph we can see that the Call Stack ends at `FindEntry` method. So `Equals` and `GetHashCode` got devirtualized and inlined.
+
+![Inlined](img/vtune_why_not_perfview_flamegraph.png)
+
+**You should start every investigation with VS Profiler or PerfView. When you get to a point where you clearly need information on CPU instruction level and you are using Intel hardware, use VTune.**
+
+#### Identifying Hotspots
+
+Run VTune **as Administrator** and click `New Project`:
```
tannergooding: This should call out that Hyper-V should be disabled in the OS and VTx/other virtualization options should be disabled in the BIOS so that instruction based sampling is supported (VTune will generally warn, but sometimes not very visibly).

tannergooding: Ah, I see the comment below now

docs/profiling-workflow-corefx.md
tannergooding: Yes, they should work for the most part (VTune on AMD and uProf on Intel); but various options won't be supported and you may not get as accurate traces/etc.

tannergooding: If it isn't working well enough, I can cover the same example for AMD uProf and modify the doc after this PR is merged (just create an issue an assign it to me)

adamsitnik: @tannergooding excellent! Since I am now working on the PerfCollect and dotnet trace doc I am going to create a separate issue and assign you. Thanks!

PR #850, opened 2019-09-03
docs: profiling local CoreFX build using PerfCollect

adamsitnik:
The goal of this doc is to show how to profile local CoreFX build using PerfCollect. It's super important because it's the easiest way to get native call stacks.

https://github.com/adamsitnik/performance/blob/35c6ba8bd1f032821dbb81b85602ad99647f3a35/docs/profiling-workflow-corefx.md#PerfCollect

@tarekgh @roji could you please take a look?


adamsitnik:
@tarekgh that was fast! thanks for the review!

tarekgh:
@adamsitnik it LGTM. Thanks for getting this documented. it is useful to have it handy.

roji:
A bit late but the guide looks great! Hope to not need to go into native stack traces but I now know what to do if I need to :)

PR #867, opened 2019-09-09
Fix lab break

### LINKED ISSUE #865: Compilation error with latest Roslyn
The following code no longer compiles:

https://github.com/dotnet/performance/blob/97d49cfc30502b835122d360ca4b93f29215d4cc/src/benchmarks/micro/corefx/System.Buffers/ArrayPoolTests.cs#L39-L47

Compiler reports the following error:

```
corefx/System.Buffers/ArrayPoolTests.cs(44,21): error CS0019: Operator '==' cannot be applied to operands of type 'T' and 'default' [/Users/filipnavara/agent/_work/1/s/work/performance/src/benchmarks/micro/MicroBenchmarks.csproj]
```

This is caused by breaking change in the compiler (https://github.com/dotnet/roslyn/issues/36492) that was previously leading to miscompiled code. Originally reported as https://github.com/dotnet/roslyn/issues/38568, moving here.

billwert:
dotnet/roslyn#37596 includes a breaking change in the 5.0 branch. This function is
intended to ensure reads happen from the array, so the test isn't strictly necessary.

Fixes #865

### DIFF
--- a/src/benchmarks/micro/corefx/System.Buffers/ArrayPoolTests.cs
+++ b/src/benchmarks/micro/corefx/System.Buffers/ArrayPoolTests.cs
@@ -19,7 +19,6 @@ namespace System.Buffers.Tests
         private readonly T[][] _nestedArrays = new T[NestedDepth][];
         private const int Iterations = 100_000;
         private const int NestedDepth = 8;
-
         [Params(4096)]
         public int RentalSize;
 
@@ -36,14 +35,14 @@ namespace System.Buffers.Tests
 
         private static void Clear(T[] arr) => arr.AsSpan().Clear();
 
-        private static int CountDefault(T[] arr)
+        private static T IterateAll(T[] arr)
         {
-            int count = 0;
+            T ret = default;
             foreach (T item in arr)
             {
-                if (item == default) count++;
+                ret = item;
             }
-            return count;
+            return ret;
         }
 
         [Benchmark(OperationsPerInvoke = Iterations)]
@@ -55,7 +54,7 @@ namespace System.Buffers.Tests
                 T[] arr = pool.Rent(RentalSize);
                 if (ManipulateArray) Clear(arr);
                 if (Async) await Task.Yield();
-                if (ManipulateArray) CountDefault(arr);
+                if (ManipulateArray) IterateAll(arr);
                 pool.Return(arr);
             }
         }
@@ -75,7 +74,7 @@ namespace System.Buffers.Tests
                         T[] arr = pool.Rent(RentalSize);
                         if (ManipulateArray) Clear(arr);
                         if (Async) await Task.Yield();
-                        if (ManipulateArray) CountDefault(arr);
+                        if (ManipulateArray) IterateAll(arr);
                         pool.Return(arr);
                     }
                 })));
@@ -97,7 +96,7 @@ namespace System.Buffers.Tests
 
                 for (int j = _nestedArrays.Length - 1; j >= 0; j--)
                 {
-                    if (ManipulateArray) CountDefault(_nestedArrays[j]);
+                    if (ManipulateArray) IterateAll(_nestedArrays[j]);
                     pool.Return(_nestedArrays[j]);
                 }
             }
@@ -122,7 +121,7 @@ namespace System.Buffers.Tests
 
                     while (reader.TryRead(out T[] buffer))
                     {
-                        if (ManipulateArray) CountDefault(buffer);
+                        if (ManipulateArray) IterateAll(buffer);
                         pool.Return(buffer);
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/corefx/System.Buffers/ArrayPoolTests.cs
```diff
@@ -36,14 +35,14 @@ public class RentReturnArrayPoolTests<T>
...
         {
-            int count = 0;
+            T ret = default;
             foreach (T item in arr)
             {
-                if (item == default) count++;
+                ret = item;
             }
-            return count;
+            return ret;
```
filipnavara: I don't know how much of a concern it is but I think Mono+LLVM can eliminate this kind of loops and possibly screw up the results. Wouldn't it be possible to just change the condition to `item == ret` here? (ie. save the default value into variable and compare with that)

billwert: You can't use the equality operator on `T`s. The correct way to compare like that is with `EqualityComparer<T>.Default.Equals()`, which turns this into a test for `EqualityComparer<T>.Default.Equals()`. :)

Chatting with @stephentoub the actual goal of this loop is just to ensure the entire array is read. We're pretty sure the CoreCLR compilers are not smart enough to skip the enumeration. I don't have a Mono+LLVM setup handy - can you test this?

I'm going to check this in for now because we need to unblock the lab. If this is unsuitable for Mono we can figure out another strategy.

filipnavara: The current default optimizations in Mono + LLVM JIT don't eliminate it. However, the [proposed ones](https://github.com/mono/mono/pull/16436) and the ones used for AOT compilation do eliminate it:

```
*** ASM for HelloWorld.Program:IterateAll<byte> ***
/var/folders/z9/3dmjh44j775b2_wsv97s8fd80000gn/T/.9W4MZO:
(__TEXT,__text) section
loWorld_Program_IterateAll_byte_:
0000000000000000        movl    0x18(%rdi), %eax
0000000000000003        testl   %eax, %eax
0000000000000005        jle     0x11
0000000000000007        addl    $-0x1, %eax
000000000000000a        cltq
000000000000000c        movb    0x20(%rdi,%rax), %al
0000000000000010        retq
0000000000000011        xorl    %eax, %eax
0000000000000013        retq
***
```

adamsitnik: In the future we should consider using [Consumer](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Engines/Consumer.cs) that would prevent from the elimination:

```cs
private readonly Consumer _consumer = new Consumer();

foreach (var item in collection)
   _consumer.Consume(item);
```

Or just redesign the benchmark. I am not sure if I understand correctly why the `ArrayPool` benchmark is trying to access every item in the array.

stephentoub: > In the future we should consider using Consumer that would prevent from the elimination

Have you measured the impact of that when running on arm? I suspect you'll find the overhead incurred from the volatile writes impacts the benchmarks.

> I am not sure if I understand correctly why the ArrayPool benchmark is trying to access every item in the array.

To simulate what code actually does with the arrays, impact caches accordingly, take a corresponding amount of time between rent and return, etc.

adamsitnik: > Have you measured the impact of that when running on arm? 

Not yet. BTW @janvorli why volatile writes are so expensive on ARM?

>  I suspect you'll find the overhead incurred from the volatile writes impacts the benchmarks.

👍 

> To simulate what code actually does with the arrays, impact caches accordingly, take a corresponding amount of time between rent and return, etc.

Thanks for the explanation. I need to wrap my head around these benchmarks. We should most probably consider adding some ArrayPool benchmarks to @sebastienros aspnet/benchmarks that would allow us to test the perf under high parallel load.

stephentoub: > why volatile writes are so expensive on ARM?

Jan can correct me if something's changed recently, but last I knew we generate a full barrier for volatile writes on arm.

billwert: > In the future we should consider using [Consumer](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Engines/Consumer.cs) that would prevent from the elimination:

Yeah, I'm aware of the feature. I tried it in this case and the impact to the existing benchmark was quite a lot (even on x64). My goal was to keep the numbers as close to where they were while fixing the problem.

stephentoub: > do you happen to know that?

Looks like https://github.com/dotnet/coreclr/issues/11545.

PR #874, opened 2019-09-12
Remove parsing and make binding/emit benchmarks more reliable

agocke:
I've been looking at the results that have built up from the first set of perf tests I uploaded and found a few things:

- The results are unfortunately pretty noisy.

- The noisiest results are the non-deterministic threading tests

- Even straight-line single threaded code (metadata serialization) has some noise

I've made some changes to try and help. First, I've removed the parsing tests. I think parsing is closer to a micro benchmark and I'm going to move it out of the e2e test suite. Second, I've turned off concurrent compilation. This increases test time, but seems to reduce variance. Lastly, I've stopped using the default RecommendedConfig. I think the Default Job parameters and a tighter standard error requirement is better for us. We may end up taking slightly longer to run the tests, but each test should only be a couple seconds at most.

### DIFF
--- a/src/benchmarks/real-world/Roslyn/Program.cs
+++ b/src/benchmarks/real-world/Roslyn/Program.cs
@@ -22,8 +22,8 @@ namespace CompilerBenchmarks
                 .Run(args, RecommendedConfig.Create(
                                artifactsPath: new DirectoryInfo(Path.Combine(Path.GetDirectoryName(typeof(Program).Assembly.Location),
                                                                              "BenchmarkDotNet.Artifacts")),
-                               mandatoryCategories: ImmutableHashSet.Create("Roslyn"))
-                           .With(Job.Default.WithMaxRelativeError(0.01)))
+                               mandatoryCategories: ImmutableHashSet.Create("Roslyn"),
+                               job: Job.Default.WithMaxRelativeError(0.01)))
                 .ToExitCode();
         }
 
--- a/src/benchmarks/real-world/Roslyn/Program.cs
+++ b/src/benchmarks/real-world/Roslyn/Program.cs
@@ -4,8 +4,10 @@ using System;
 using System.Collections.Immutable;
 using System.IO;
 using System.Threading.Tasks;
+using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Extensions;
+using BenchmarkDotNet.Jobs;
 
 namespace CompilerBenchmarks
 {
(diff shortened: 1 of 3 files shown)


src/benchmarks/real-world/Roslyn/Program.cs
```diff
@@ -15,12 +17,16 @@ public static async Task<int> Main(string[] args)
...
-                .Run(args, RecommendedConfig.Create(
-                    artifactsPath: new DirectoryInfo(Path.Combine(Path.GetDirectoryName(typeof(Program).Assembly.Location), "BenchmarkDotNet.Artifacts")),
-                    mandatoryCategories: ImmutableHashSet.Create("Roslyn")))
-                .ToExitCode();
+            var summary = BenchmarkRunner.Run<StageBenchmarks>(
+                RecommendedConfig.Create(
+                               artifactsPath: new DirectoryInfo(Path.Combine(Path.GetDirectoryName(typeof(Program).Assembly.Location),
+                                                                             "BenchmarkDotNet.Artifacts")),
+                               mandatoryCategories: ImmutableHashSet.Create("Roslyn"))
+                           .With(Job.Default
```
adamsitnik: `.With` does not replace `Job` defined in `RecommendedConfig` but instead adds one more `Job` to the `Config`. So the benchmarks are actually executed twice.

Could you please extend `RecommendedConfig.Create` with an optional `Job` argument and use it instead of the recommended default if it was provided?

src/benchmarks/real-world/Roslyn/Program.cs
```diff
@@ -15,12 +17,16 @@ public static async Task<int> Main(string[] args)
         {
             await Setup();
 
-            return BenchmarkSwitcher
-                .FromAssembly(typeof(Program).Assembly)
-                .Run(args, RecommendedConfig.Create(
-                    artifactsPath: new DirectoryInfo(Path.Combine(Path.GetDirectoryName(typeof(Program).Assembly.Location), "BenchmarkDotNet.Artifacts")),
-                    mandatoryCategories: ImmutableHashSet.Create("Roslyn")))
-                .ToExitCode();
+            var summary = BenchmarkRunner.Run<StageBenchmarks>(
```
adamsitnik: I know that as of today this project defines a single type with benchmarks, but I would rather prefer to keep using `BenchmarkSwitcher`. It supports more than 1 type (but it requires to provide the `--filter`  argument or choosing type from the list.)

Moreover, it supports parsing console line arguments, which are passed by the CI to the benchmark app. A good example is `--stopOnFirstError true` which stops running the benchmarks after first error. I am not 100% sure but most probably the CI failure is due to removing the `.Run(args` part

agocke:
@adamsitnik Before i tried to make the change, I was hitting some kind of duplicate test key. I'm not sure what was going on there.

src/benchmarks/real-world/Roslyn/Program.cs
agocke: Ah, OK, this explains the behavior I was seeing. Lemme follow up with that.

agocke:
@adamsitnik I've fixed some stuff up but I'm not sure what's causing these errors

billwert:
@agocke our CI is busted right now and I am working on it. I'll poke this after I merge the fix.

agocke:
Ah, great, thanks. I'm also working on a subsequent set of changes to support Roslyn analyzers.

PR #882, opened 2019-09-16
Add release/3.1 and turn off 3.0 for now

billwert:
The 3.0 release branch isn't installing properly, which blocks our runs. We can turn it
back on after GA. Also adding tracking release/3.1.1xx.


billwert:
This is now just updating to point to the 3.1.1xx release branch since there's no new TFM

DrewScoggins:
LGTM

PR #887, opened 2019-09-18
Add benchmark and modify ConsoleMandel

CarolEidt:
Add a bilinear interpolation benchmark and modify the ConsoleMandel benchmark to include Double scenarios and change to use the versions that utilize `Vector<int>` and `Vector<long>` for comparisons.

### DIFF
--- a/src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
+++ b/src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
@@ -268,6 +268,10 @@ public class BilinearTest
     }
 #endif // !NETCOREAPP2_1 && !NETCOREAPP2_2 && !NETFRAMEWORK
 
+    // This method is currently unused. It is useful, when generating a new vectorized version
+    // of the benchmark (e.g. with a different set of HW intrinsics), to ensure that it is
+    // correct (i.e. produces the same result as the scalar version).
+    //
     public static bool CheckResult(double[] output, double[] vectorOutput)
     {
         double eps = 1e-16;
@@ -282,17 +286,17 @@ public class BilinearTest
         return true;
     }
 
-    [Benchmark(Description = "BilinearInterpol_Scalar")]
+    [Benchmark]
     [BenchmarkCategory(Categories.CoreCLR)]
-    public double[] Test0()
+    public double[] Interpol_Scalar()
     {
         output = BilinearInterpol(input, A, minXA, maxXA, B, minXB, maxXB, weightB);
         return output;
     }
 
-    [Benchmark(Description = "BilinearInterpol_Vector")]
+    [Benchmark]
     [BenchmarkCategory(Categories.CoreCLR)]
-    public double[] Test1()
+    public double[] Interpol_Vector()
     {
         double[] vectorOutput = BilinearInterpol_Vector(input, A, minXA, maxXA, B, minXB, maxXB, weightB);
         return vectorOutput;
@@ -300,8 +304,8 @@ public class BilinearTest
 
 #if !NETCOREAPP2_1 && !NETCOREAPP2_2 && !NETFRAMEWORK
     [BenchmarkCategory(Categories.CoreCLR)]
-    [Benchmark(Description = "BilinearInterpol_AVX")]
-    public double[] Test2()
+    [Benchmark]
+    public double[] Interpol_AVX()
     {
         if (Avx2.IsSupported)
         {
--- a/src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
+++ b/src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
@@ -1,8 +1,12 @@
 ﻿using System;
 using System.Linq;
 using System.Numerics;
+
+// HW Intrinsic APIs are available only in .NET Core 3.0+
+#if !NETCOREAPP2_1 && !NETCOREAPP2_2 && !NETFRAMEWORK
 using System.Runtime.Intrinsics;
 using System.Runtime.Intrinsics.X86;
+#endif
 
 using BenchmarkDotNet.Attributes;
 using MicroBenchmarks;
@@ -182,6 +186,7 @@ public class BilinearTest
         return z;
     }
 
+#if !NETCOREAPP2_1 && !NETCOREAPP2_2 && !NETFRAMEWORK
     private static unsafe double[] BilinearInterpol_AVX(
                                             double[] x,
                                             double[] A,
@@ -261,6 +266,7 @@ public class BilinearTest
         }
         return z;
     }
+#endif // !NETCOREAPP2_1 && !NETCOREAPP2_2 && !NETFRAMEWORK
 
     public static bool CheckResult(double[] output, double[] vectorOutput)
     {
@@ -292,6 +298,7 @@ public class BilinearTest
         return vectorOutput;
     }
 
+#if !NETCOREAPP2_1 && !NETCOREAPP2_2 && !NETFRAMEWORK
     [BenchmarkCategory(Categories.CoreCLR)]
     [Benchmark(Description = "BilinearInterpol_AVX")]
     public double[] Test2()
@@ -304,4 +311,5 @@ public class BilinearTest
         }
         return null;
     }
+#endif
 }
--- a/src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
+++ b/src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
@@ -1,0 +1,307 @@
+﻿using System;
+using System.Linq;
+using System.Numerics;
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+public class BilinearTest
+{
+    // These must be a multiple of the largest available vector of doubles.
+    const int inputVectorSize = 1024;
+    const int outputVectorSize = 1024;
+    const double minXA = 0.0;
+    const double maxXA = Math.PI;
+    const int lengthA = 500;
+    const double deltaA = (maxXA - minXA) / (double)(lengthA - 1);
+
+    const double minXB = minXA + 2.0 * deltaA;
+    const double maxXB = maxXA - 2.0 * deltaA;
+    const double weightB = 0.15;
+    const int lengthB = 500;
+    const double deltaB = (maxXB - minXB) / (double)(lengthB - 1);
+
+    //ref values 
+    double[] A, B, input, output;
+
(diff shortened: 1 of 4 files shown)


CarolEidt:
@adamsitnik @DrewScoggins Can you advise how to specify that the BilinearInterpol benchmark should only run on 3.0 and higher?

DrewScoggins:
Since it is the entire file and all of the tests that need to only be on 3.0+ and not just one test, I think you should be able to change the MicroBenchmarks.csproj to remove the file when it would not work.

My guess is that you would want to put it in this Item Group
```
  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0')">
    <Compile Remove="coreclr\Math\Functions\Double\FusedMultiplyAdd.cs" />
    <Compile Remove="coreclr\Math\Functions\Double\ILogB.cs" />
    <Compile Remove="coreclr\Math\Functions\Double\Log2.cs" />
    <Compile Remove="coreclr\Math\Functions\Double\ScaleB.cs" />
    <Compile Remove="coreclr\Math\Functions\Single\FusedMultiplyAdd.cs" />
    <Compile Remove="coreclr\Math\Functions\Single\ILogB.cs" />
    <Compile Remove="coreclr\Math\Functions\Single\Log2.cs" />
    <Compile Remove="coreclr\Math\Functions\Single\ScaleB.cs" />
    <Compile Remove="corefx\System.Memory\SequenceReader.cs" />
    <Compile Remove="corefx\System.Text.Json\**\*.cs" />
  </ItemGroup>
```

But @adamsitnik might want to chime in here with best practices as well.

adamsitnik:
@DrewScoggins is right, we typically modify the project file if entire source file needs to be excluded and use `#if` defines when we want to exclude only selected methods

CarolEidt:
@adamsitnik @DrewScoggins - Thanks! Actually, the first two versions of the benchmark (scalar and the one that uses `Vector<T>`) can run on all targets, so I've modified to use an `#if`.

src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
```diff
@@ -0,0 +1,315 @@
...
+                Vector256<double> newZ = Avx.Add(Avx.Multiply(vWeightA, Avx.Add(AVector, Avx.Multiply(lambdaA, Avx.Subtract(AVectorPlusOne, AVector)))),
+                                             Avx.Multiply(vWeightB, Avx.Add(BVector, Avx.Multiply(lambdaB, Avx.Subtract(BVectorPlusOne, BVector)))));
+                Avx.Store(pZ, newZ);
+            }
+        }
+        return z;
+    }
+#endif // !NETCOREAPP2_1 && !NETCOREAPP2_2 && !NETFRAMEWORK
+
+    public static bool CheckResult(double[] output, double[] vectorOutput)
```
adamsitnik: nit: is this a dead code that should be removed?

src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
```diff
@@ -0,0 +1,315 @@
...
+            if (Math.Abs(output[i] - vectorOutput[i]) > eps)
+            {
+                Console.WriteLine("Failed at " + i + ": output is " + output[i] + ", vectorOutput is " + vectorOutput[i]);
+                return false;
+            }
+        }
+        return true;
+    }
+
+    [Benchmark(Description = "BilinearInterpol_Scalar")]
```
adamsitnik: we use `Description` to set the display name for benchmarks which can not be renamed (method name is part of the [benchmark id](https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable))

in this case we could just name the  benchmark `Interpol_Vector` (keyword `Bilinear` is already used in the type name which is also part of benchmark id)

```suggestion
    [Benchmark]
```

src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
```diff
@@ -0,0 +1,315 @@
...
+    [BenchmarkCategory(Categories.CoreCLR)]
+    public double[] Test0()
```
adamsitnik:
```suggestion
    public double[] Interpol_Scalar()
```

src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
```diff
@@ -0,0 +1,315 @@
...
+    {
+        output = BilinearInterpol(input, A, minXA, maxXA, B, minXB, maxXB, weightB);
+        return output;
+    }
+
+    [Benchmark(Description = "BilinearInterpol_Vector")]
+    [BenchmarkCategory(Categories.CoreCLR)]
+    public double[] Test1()
```
adamsitnik:
```suggestion
    public double[] Interpol_Vector()
```

src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
```diff
@@ -0,0 +1,315 @@
...
+
+    [Benchmark(Description = "BilinearInterpol_Scalar")]
+    [BenchmarkCategory(Categories.CoreCLR)]
+    public double[] Test0()
+    {
+        output = BilinearInterpol(input, A, minXA, maxXA, B, minXB, maxXB, weightB);
+        return output;
+    }
+
+    [Benchmark(Description = "BilinearInterpol_Vector")]
```
adamsitnik:
```suggestion
    [Benchmark]
```

src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
```diff
@@ -0,0 +1,315 @@
...
+    [BenchmarkCategory(Categories.CoreCLR)]
+    public double[] Test1()
+    {
+        double[] vectorOutput = BilinearInterpol_Vector(input, A, minXA, maxXA, B, minXB, maxXB, weightB);
+        return vectorOutput;
+    }
+
+#if !NETCOREAPP2_1 && !NETCOREAPP2_2 && !NETFRAMEWORK
+    [BenchmarkCategory(Categories.CoreCLR)]
+    [Benchmark(Description = "BilinearInterpol_AVX")]
```
adamsitnik:
```suggestion
    [Benchmark]
```

src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
```diff
@@ -0,0 +1,315 @@
...
+    public double[] Test2()
```
adamsitnik:
```suggestion
    public double[] Interpol_AVX()
```

src/benchmarks/micro/coreclr/BilinearInterpol/BilinearInterpol.cs
CarolEidt: Sort of; I'll add a comment. This was used to check that the vectorized versions gave the same result, but I couldn't figure out a good way to do that without including it in the benchmark timing. I'd prefer to leave it in so that, when adding new versions in future (e.g. perhaps an SSE* version and an Arm64 Intrinsics version), I can use this to make sure I've done it correctly.

CarolEidt:
@adamsitnik @DrewScoggins I think this is ready to merge.

DrewScoggins:
I had one other small question. What do these tests look like run over run on your machine? Are they fairly stable? Looking at the test this is more than just a microbenchmark. We have been doing work to make sure that we are reducing the noise of performance tests in the repo, and so having some data on that would be great.

I don't think that it would be a blocker, unless the tests are just wildly noisy, but it is good information to have.

CarolEidt:
@DrewScoggins - I'm not sure what constitutes a microbenchmark, but these are designed to test the performance of pretty low-level performance of vector operations.

Here are the results from 3 runs of `dotnet run -c Release --framework netcoreapp5.0 --filter *Bilinear*` on my machine:
```
|          Method |      Mean |     Error |    StdDev |    Median |       Min |       Max |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |----------:|----------:|----------:|----------:|----------:|----------:|-------:|------:|------:|----------:|
| Interpol_Scalar | 14.528 us | 0.1751 us | 0.1638 us | 14.552 us | 14.212 us | 14.776 us | 1.9266 |     - |     - |   8.02 KB |
| Interpol_Vector | 11.363 us | 0.2233 us | 0.2571 us | 11.311 us | 11.034 us | 11.970 us | 1.9479 |     - |     - |   8.08 KB |
|    Interpol_AVX |  3.057 us | 0.0612 us | 0.0654 us |  3.044 us |  2.962 us |  3.189 us | 1.9520 |     - |     - |   8.02 KB |

|          Method |      Mean |     Error |    StdDev |    Median |       Min |       Max |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |----------:|----------:|----------:|----------:|----------:|----------:|-------:|------:|------:|----------:|
| Interpol_Scalar | 14.593 us | 0.2821 us | 0.2639 us | 14.558 us | 14.256 us | 15.144 us | 1.9585 |     - |     - |   8.02 KB |
| Interpol_Vector | 10.559 us | 0.1999 us | 0.1870 us | 10.504 us | 10.356 us | 10.920 us | 1.9678 |     - |     - |   8.08 KB |
|    Interpol_AVX |  3.017 us | 0.0747 us | 0.0860 us |  3.004 us |  2.883 us |  3.183 us | 1.9520 |     - |     - |   8.02 KB |

|          Method |      Mean |     Error |    StdDev |    Median |       Min |       Max |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |----------:|----------:|----------:|----------:|----------:|----------:|-------:|------:|------:|----------:|
| Interpol_Scalar | 14.631 us | 0.2046 us | 0.1914 us | 14.649 us | 14.347 us | 15.010 us | 1.9567 |     - |     - |   8.02 KB |
| Interpol_Vector | 10.546 us | 0.1801 us | 0.1597 us | 10.522 us | 10.310 us | 10.897 us | 1.9404 |     - |     - |   8.08 KB |
|    Interpol_AVX |  3.033 us | 0.0545 us | 0.0510 us |  3.038 us |  2.925 us |  3.109 us | 1.9589 |     - |     - |   8.02 KB |
```

DrewScoggins:
That looks great! I am just more used to microbenchmarks that look like a single call to `String.Split()` 😄 

LGTM, I'll go ahead and merge this.

CarolEidt:
Thanks!

PR #889, opened 2019-09-19
Add test for AnalyzerConfigSet

agocke:
This is the first test for some Roslyn APIs. We may want to add some more, like measuring GetTypeByMetadata, in the future.

### DIFF
--- a/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
+++ b/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.1.0-beta1-final" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.*" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
+++ b/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.*" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.3.1" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/benchmarks/real-world/Roslyn/RoslynApis.cs
+++ b/src/benchmarks/real-world/Roslyn/RoslynApis.cs
@@ -1,4 +1,8 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
 
+using System.Collections.Immutable;
 using System.IO;
 using BenchmarkDotNet.Attributes;
 using Microsoft.CodeAnalysis;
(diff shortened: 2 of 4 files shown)


adamsitnik:
The Linux build is red because of invalid directory separator character in `CodeAnalysisReproWithAnalyzers/MetadataReader\PEAssembly.cs`

agocke:
Heh, of course. Thanks!

src/benchmarks/real-world/Roslyn/RoslynApis.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace CompilerBenchmarks
+{
+    /// <summary>
+    /// A collection of benchmarks for Roslyn APIs
+    /// </summary>
+    [BenchmarkCategory("Roslyn")]
+    public class RoslynApis
+    {
+        [Benchmark]
+        public object BuildAnalyzerConfigs()
```
adamsitnik: could you please return `ImmutableArray<AnalyzerConfigOptionsResult>` instead of an `object`? Casting struct to an object introduces boxing and adds some extra memory allocation which originally is not part of the measured APIs and should be avoided.

```suggestion
        public ImmutableArray<AnalyzerConfigOptionsResult> BuildAnalyzerConfigs()
```

src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
```diff
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.1.0-beta1-final" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.*" />
```
adamsitnik: I am not sure about this change. The goal of having the Roslyn benchmarks here was making sure that .NET Core does not regress the performance of Roslyn. With hardcoded Roslyn version it was easy to ensure (every regression could be only .NET Core regression). With variable Roslyn version it could be both: either a Roslyn or .NET Core regression.

Have you discussed this change with @billwert ?

src/benchmarks/real-world/Roslyn/RoslynApis.cs
```diff
@@ -0,0 +1,26 @@
+
```
adamsitnik: I am sorry, I forgot to include new using statement in my previous GH suggestion which breaks the CI.

```suggestion
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

using System.Collections.Immutable;
```

src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
agocke: Not yet -- I thought that the intention was for our benchmarks to track ongoing perf, not just one build. I was actually thinking we should test against the latest nightly out of our preview branch. Is that not what most other projects are doing? How do you catch regressions otherwise?

adamsitnik: > I thought that the intention was for our benchmarks to track ongoing perf, not just one build

@agocke Our initial idea was to make sure that the daily builds of .NET Core don't regress the performance. The real-world benchmarks (Roslyn and ML.NET) were supposed to be Contract Tests - something like "this is what we care about, please don't regress our perf".

I agree that it does not solve the problem of detecting regressions in the tested products (Roslyn, ML.NET and more in the future).

Could you please revert this particular change and create a separate issue where we could discuss what would be the best way to approach this problem?

/cc @billwert

billwert: I agree with @adamsitnik that we should revert this change and come up with a better plan. The intent of having these tests is exactly as stated.

I've opened #896 to track this.

agocke: pign @billwert

PR #892, opened 2019-09-20
Redesign benchmarks for culture-specific string operations

### LINKED ISSUE #885: Implement more realistic string benchmarks
The benchmarks that we have as of today for methods like `IndexOf`, `StartsWith` etc use not very realistic test cases and moreover typically test a single culture.

/cc @jkotas

adamsitnik:
The culture-specific string benchmarks we had so far were using a very small and simple input and were testing very few cultures and `CompareOptions`. This is why we have missed issues like https://github.com/dotnet/corefx/issues/40674

This is my proposal for fixing it.

Instead of using some made-up text, we are using part of "Alice's Adventures in Wonderland" book. It  contains mostly simply ASCII characters, but also some "high" chars that get special treatment and hit the slow path.

The tested matrix now is:

```cs
(new CultureInfo("en-US"), CompareOptions.Ordinal)
(new CultureInfo("en-US"), CompareOptions.OrdinalIgnoreCase)
(new CultureInfo("en-US"), CompareOptions.None)
(new CultureInfo("en-US"), CompareOptions.IgnoreCase)
(new CultureInfo("en-US"), CompareOptions.IgnoreSymbols)
(CultureInfo.InvariantCulture, CompareOptions.None)
(CultureInfo.InvariantCulture, CompareOptions.IgnoreCase)
(new CultureInfo("pl-PL"), CompareOptions.None) // as an example of complex language hitting the slow path on Unix
```

I've removed the old benchmarks that would now be duplicated.

Moreover I've realized that `Perf_CompareInfo` had a serious bug inside - the strings were always identical because `source` argument was never used..

https://github.com/dotnet/performance/blob/0c626c0de72e9e018bb02d0f92b97117098f1fee/src/benchmarks/micro/corefx/System.Globalization/Perf.CompareInfo.cs#L14-L24

Fixes #885


src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
```diff
@@ -0,0 +1,94 @@
...
+        public bool IsPrefix_DifferentFirstChar() => CultureOptions.CultureInfo.CompareInfo.IsPrefix(_value, _diffAtFirstChar, CultureOptions.CompareOptions);
+
+        [Benchmark]
+        public bool IsSuffix_SecondHalf() => CultureOptions.CultureInfo.CompareInfo.IsSuffix(_value, _secondHalf, CultureOptions.CompareOptions);
+
+        [Benchmark] // this should return quickly
+        public bool IsSuffix_DifferentLastChar() => CultureOptions.CultureInfo.CompareInfo.IsSuffix(_value, _diffAtLastChar, CultureOptions.CompareOptions);
+
+        [Benchmark]
+        public int IndexOf_SecondHalf() => CultureOptions.CultureInfo.CompareInfo.IndexOf(_value, _secondHalf, CultureOptions.CompareOptions);
```
jkotas: The typical IndexOf use searches for small string (e.g. single world) in medium amount of text (e.g. single line).  We should have a coverage for a case like this.

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            // we are using part of Alice's Adventures in Wonderland text as test data
+            // it contains mostly simply ascii characters, but also some "high" chars that get special treatment and hit the slow path
+            char[] characters = File.ReadAllText(CompressedFile.GetFilePath("alice29.txt")).ToCharArray().Take(Count).ToArray();
+            _value = new string(characters);
+            _same = new string(characters);
+            _firstHalf = new string(characters.Take(Count / 2).ToArray());
```
jkotas: Is this always going to hit the slow paths?

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
```diff
@@ -0,0 +1,94 @@
...
+            // Polish language has a lot of special characters, for example 'ch', 'rz', 'sz', 'cz' use two chars to express one ;)
+            // it also has a lot of characters with accent so we use it as an example of "complex" language
+            yield return (new CultureInfo("pl-PL"), CompareOptions.None);
+        }
+
+        [ParamsSource(nameof(GetCultureOptions))]
+        public (CultureInfo CultureInfo, CompareOptions CompareOptions) CultureOptions;
+
+        [Params(
+            128, // small input that fits into stack-allocated array
```
jkotas: Why does it matter that it fits into stack-allocated array? None of the methods tested here use stack-allocated arrays.

adamsitnik: `GetHashCode` does

https://github.com/dotnet/coreclr/blob/c6675ef2e22474d6222d054ae3d022c01eda9b6d/src/System.Private.CoreLib/shared/System/Globalization/CompareInfo.Unix.cs#L824

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
adamsitnik: good point, I am going to ensure this

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
adamsitnik: another good point, I am going to add such case

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
```diff
@@ -0,0 +1,94 @@
...
+        private string _value, _same, _diffAtFirstChar, _diffAtLastChar, _firstHalf, _secondHalf;
+
+        public static IEnumerable<(CultureInfo CultureInfo, CompareOptions CompareOptions)> GetCultureOptions()
+        {
+            // Ordinal and OrdinalIgnoreCase use single execution path for all cultures, so we test it only for "en-US"
+            yield return (new CultureInfo("en-US"), CompareOptions.Ordinal);
+            yield return (new CultureInfo("en-US"), CompareOptions.OrdinalIgnoreCase);
+
+            yield return (new CultureInfo("en-US"), CompareOptions.None);
+            yield return (new CultureInfo("en-US"), CompareOptions.IgnoreCase);
```
tarekgh: >CultureInfo("en-US"), CompareOptions.IgnoreCase); [](start = 30, length = 49)

could you please add a case for ignore non space option too?

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
```diff
@@ -0,0 +1,94 @@
...
+            128, // small input that fits into stack-allocated array
+            1024 * 128)] // medium size input that fits into an array rented from ArrayPool.Shared without allocation
+        public int Count { get; set; }
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            // we are using part of Alice's Adventures in Wonderland text as test data
+            // it contains mostly simply ascii characters, but also some "high" chars that get special treatment and hit the slow path
+            char[] characters = File.ReadAllText(CompressedFile.GetFilePath("alice29.txt")).ToCharArray().Take(Count).ToArray();
```
tarekgh: >alice29.txt [](start = 77, length = 11)

This will take me some time to read and enjoy :-) ...kidding

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
```diff
@@ -0,0 +1,94 @@
...
+            // https://github.com/dotnet/coreclr/blob/cd6bc26bdc4ac06fe2165b283eaf9fb5ff5293f4/src/System.Private.CoreLib/shared/System/Globalization/CompareInfo.Unix.cs#L35
+
+            // so we need one more culture to test the "slow path"
+            // Polish language has a lot of special characters, for example 'ch', 'rz', 'sz', 'cz' use two chars to express one ;)
+            // it also has a lot of characters with accent so we use it as an example of "complex" language
+            yield return (new CultureInfo("pl-PL"), CompareOptions.None);
+        }
+
+        [ParamsSource(nameof(GetCultureOptions))]
+        public (CultureInfo CultureInfo, CompareOptions CompareOptions) CultureOptions;
```
gfoidl: Should this be a property (instead of a public field)?

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
adamsitnik: I've changed the benchmark to search for "word" in ".NET Conf provides a wide selection of live sessions streaming here that feature speakers from the community and .NET product teams. It is a chance to learn, ask questions live, and get inspired for your next software project."

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
adamsitnik: I've moved the `GetHashCode` benchmark to a dedicated type as it was the only one that has a size-dependent execution path.

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
adamsitnik: good point, done!

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
adamsitnik: I've modified the benchmark and to make ensure this I put `0x81` at the beginning and end of the string.

src/benchmarks/micro/corefx/System.Globalization/Perf.StringCultureSpecific.cs
adamsitnik: A property would be translated to a backing field with an inlinable setter and getter. I personally prefer to use fields in benchmarks (WYSIWYG).

gfoidl: The same applies to `Count` (below). My suggestion was for consistency in the code.

adamsitnik: @gfoidl thanks, now I see the point. I've just pushed a commit to use fields for all `params`

PR #893, opened 2019-09-20
Add compilationwithanalyzers test

agocke:
This is based on https://github.com/dotnet/performance/pull/889 and requires that PR to be merged first


billwert:
Looks like this one needs to be cleaned up @agocke - it wasn't immediately obvious to me how to handle the merge in the .cs file so can you take a look? (I suspect the branch is just correct.)

agocke:
@billwert Should be ready. I've rebased it onto master

PR #900, opened 2019-09-26
Minor fixes in the StageBenchmarks

adamsitnik:
I've missed #893 review and this PR are just some minor fixes:

* remove boxing (to not include extra allocations in the results)
* remove "double await" - if given benchmark just calls one async method there is no need to await it, we can just return it and BDN will await it. This avoids creating an async state machine twice. Minor detail, but adds some overhead ;)


billwert:
Good catch!

PR #904, opened 2019-09-30
Startup scenario

billwert:
Startup scenario support.

Includes lots of associated script changes etc.


DrewScoggins:
Comments for this given offline. LGTM

PR #909, opened 2019-10-02
fix the high char benchmark for en-US culture

adamsitnik:
While working on https://github.com/dotnet/coreclr/pull/26962 I've used the `IsSuffix_DifferentLastChar` benchmark and I got an x100 improvement for non `en-US` culture and a very small one for `en-US`. Which got me thinking that sth must be wrong.

It turns out that on Linux where we use ICU the following code prints true:

```cs
string a = new string((char)0x81, 1);
string b = new string((char)0x82, 1);

Console.WriteLine(CultureInfo.GetCultureInfo("en-US").CompareInfo.IsSuffix(a, b));
```

So to get a real difference for given string that starts|ends with `(char)0x81` we should not just increment the char. I decided to double it.

The change is safe for non-high char code path because the sentence starts and ends with `.` which values is `46` and even doubled it's still not a high char.

### DIFF
--- a/src/benchmarks/micro/corefx/System.Globalization/StringSearch.cs
+++ b/src/benchmarks/micro/corefx/System.Globalization/StringSearch.cs
@@ -69,10 +69,10 @@ namespace System.Globalization.Tests
             char[] copy = characters.ToArray();
             // to get a different char we can not just increment the first|last char because for the HighChars=true
             // CultureInfo.GetCultureInfo("en-US").CompareInfo.IsSuffix(new string((char)0x81, 1), new string((char)0x82, 1)) returns true
-            copy[0] = (char)(copy[0] * 2);
+            copy[0] = (char)(Options.highChars ? copy[0] * 2 : copy[0] + 1);
             _diffAtFirstChar = new string(copy);
             copy = characters.ToArray();
-            copy[characters.Length - 1] = (char)(copy[characters.Length - 1] * 2);
+            copy[characters.Length - 1] = (char)(Options.highChars ? copy[characters.Length - 1] * 2 : copy[characters.Length - 1] + 1);
             _diffAtLastChar = new string(copy);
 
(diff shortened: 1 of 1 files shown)


jkotas:
> the sentence starts and ends with . which values is 46 and even doubled it's still not a high char.

This is not obvious from the code. If the sentence is modified for some reason, this can break again easily.

adamsitnik:
> This is not obvious from the code. If the sentence is modified for some reason, this can break again easily.

Very good point. I've added the necessary checks which actually found another bug: the sentence was starting and ending with a dot, and for `IgnoreSymbols` the `IsSuffix_DifferentLastChar` and `IsPrefix_DifferentFirstChar` benchmarks were not exiting after first char ;)

src/benchmarks/micro/corefx/System.Globalization/StringSearch.cs
```diff
@@ -90,5 +105,139 @@ public void Setup()
 
         [Benchmark]
         public int LastIndexOf_Word_NotFound() => Options.CultureInfo.CompareInfo.LastIndexOf(_value, "word", Options.CompareOptions);
+
+        private static bool ContainsHighChar(string text) => text.Any(c => c > 0x80 || HighChars[c]);
+
+        // source: https://github.com/dotnet/coreclr/blob/68ff240063fc2ddb9b03275ae5d5063a09d38ace/src/utilcode/util_nodependencies.cpp#L856-L993
```
jkotas: This array gets tweaked from time to time, and it is unfortunate to have it hardcoded in yet another place.

It may be better to implement the sanity check against "known safe" range that is unlikely to change, e.g. `a-zA-Z0-9 .`.

adamsitnik: good point, I've implemented that

adamsitnik:
@jkotas thanks for the review!

Issue #189, opened 2018-12-12
Loop-unrolling


EgorBo:
Lots of benchmarks under micro/coreclr use inner loops (e.g. https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/coreclr/Bytemark/ByteMark.cs#L445-L455) - are they really needed there?
The problem is - when I run them using new mono-llvm (inprocess) - LLVM unrolls those loops 🙂 and probably .NET Core will also be able to unroll them soon

adamsitnik:
I ported these benchmarks 1:1 to make sure the results we report to BenchView don't change.

I think that we should keep it for the existing ones but avoid when adding new benchmarks to the repo.

@billwert @brianrob what do you guys think?

jorive:
@adamsitnik I thought we already removed those inner loops for some of the benchmarks.

adamsitnik:
@jorive for some yes, but not for all

billwert:
@adamsitnik If the benchmarks are left unmodified do they return the same results as they did using xunit.performance? If so I think it is correct to leave them as they are. If they are returning different results they aren't comparable and I think it is reasonable to make changes that have them looking like we want.

If we are leaving them perhaps we should drop a comment in the ones that aren't unrolled to explain why?

adamsitnik:
> If the benchmarks are left unmodified do they return the same results as they did using xunit.performance?

yes

> If so I think it is correct to leave them as they are.

I agree. We should keep the old ones "as is" unless they cause some troubles or don't make sense, but don't add any inner loops for the new benchmarks.

> If we are leaving them perhaps we should drop a comment in the ones that aren't unrolled to explain why?

I tried to leave the comments where I could, example:

https://github.com/dotnet/performance/blob/3941f11e000edd7a404599fc2e401129c25bc4ad/src/benchmarks/micro/coreclr/perflab/CastingPerf.cs#L95

danmosemsft:
Can we close this then?

PR #922, opened 2019-10-08
Added a few benchmarks for casting performance. Primarily dealing with variance.

VSadov:
It looks like there is a good number of casting benchmarks already.

Just adding a few to cover variance and array conversions.

### DIFF
--- a/src/benchmarks/micro/coreclr/perflab/CastingPerf.cs
+++ b/src/benchmarks/micro/coreclr/perflab/CastingPerf.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using BenchmarkDotNet.Attributes;
 using MicroBenchmarks;
@@ -114,6 +115,7 @@ namespace PerfLabTests
 
         public static Object[] myClass1Arr;
         public static Object[] myClass2Arr;
+        public static Object myObj;
 
         static CastingPerf()
         {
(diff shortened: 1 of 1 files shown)


VSadov:
CC: @jkotas , @davidwrighton , @sergiy-k

src/benchmarks/micro/coreclr/perflab/CastingPerf.cs
```diff
@@ -373,5 +375,93 @@ public bool CheckArrayIsInterfaceNo()
...
+
+        [GlobalSetup(Target = nameof(CheckArrayIsNonvariantGenericInterface))]
+        public void SetupCheckArrayIsNonvariantGenericInterface() => myObj = new MyClass2[5];
+
+        [Benchmark]
+        public bool CheckArrayIsNonvariantGenericInterface()
+        {
+            bool res = false;
+            for (int i = 0; i < InnerIterationCount; i++)
+                res = myObj is ICollection<MyClass2>;
```
jkotas: This does not look like a very robust benchmark pattern. What does prevent the JIT from CSEing this whole thing in front of the loop?

VSadov: I guess nothing. We will see an improvement on the benchmark then. 
I asked myself this question the same question, but this is a pattern used in basically all other casting benchmarks.

I suppose if we really get smart enough to CSE this, we would need to update most of them.

VSadov: The other concern is that JIT may one day move `is` outside of the loop (since `is` is idempotent).

@adamsitnik - any concerns about this?

VSadov: I guess, calling some noninlineable method in the loop will prevent such optimizations. - As long as JIT does not look into that method and believes there could be sideeffects.

That will add overhead though and will certainly have impact on existing benchmarks if we do this to them.

VSadov: I have added calls to `FakeSideEffects()` - a method that does nothing. That should prevent potential CSE.
It looks like the calls do not add a lot of cost compared to casts (at least on my machine), but they do seem add some noise.

I only modified the new scenarios since there is no history of results yet.

src/benchmarks/micro/coreclr/perflab/CastingPerf.cs
```diff
@@ -114,6 +116,11 @@ public class CastingPerf
 
         public static Object[] myClass1Arr;
         public static Object[] myClass2Arr;
+        public static Object myObj;
+
+        // to make
```
gfoidl: Should be updated (https://github.com/dotnet/performance/pull/922/files#r332812461?)

VSadov: yes, thanks!

gfoidl: Would `MethodImplOptions.NoOptimization` (maybe in conjunction with `NoInlining`) be even more safe?

VSadov: `NoOptimization` would have effect on the method body, which is here empty. 
We do not need to go too far with this. In theory an all-knowing perfect optimizer can see through tricks like this and would be hard to defeat. In practice non-inlinable call should be sufficient for the foreseeable future.

src/benchmarks/micro/coreclr/perflab/CastingPerf.cs
VSadov: @jkotas - does this address your concerns?

VSadov: @jkotas - doe this address your concerns

jkotas: It does address the problem, but it does not follow the best practices for writing micro-benchmarks using Benchmark.NET: https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md

- Microbenchmarks should not have loops. Benchmark.NET does it run in a loop for you.
- The optimization of side effects should be avoided by returning the result from the benchmark method

VSadov: From existing tests in this file I assumed that casts are exception from the "no loops" rule. It is plausible because casts are cheap (compared to a delegate invoke from the outer loop).   

I will try removing loops and see if results are similar and not too noisy.

jkotas: > casts are exception 

I do not see why casts should be an exception. AFAIK, the only exception are ref-types where Benchmark.NET cannot do caching around the callsite efficiently today.

> I only modified the new scenarios since there is no history of results yet.

I do not think we should be keeping poorly written benchmarks around. I think we should fix it and rename the benchmark to avoid false alarm from automated trend analysis (e.g. like here: https://github.com/dotnet/performance/pull/892/)

billwert: When benchmarks were ported to BDN from xunit.performance, we made the choice (and have since agreed to keep) the existing benchmarks. New benchmarks should not follow this style, however. #189 has some discussion of a set of these.

Please modify the new cases to use the best practices Jan linked to. If doing so results in benchmarks that are unstable in some way, we'll dig into that.

VSadov: I have converted the new benchmarks to the "right style" and rerun on mu machine.

I see some "smoothing" of diffs. - I.E. where before one implementation was taking 1.8x longer than another it now takes 1.6x longer. This mostly impacts cases where differences are small and in a way is a good thing.

I did not see additional noise.

VSadov:
I think I have addressed all the concerns so far.

src/benchmarks/micro/coreclr/perflab/CastingPerf.cs
```diff
@@ -373,5 +374,66 @@ public bool CheckArrayIsInterfaceNo()
                 res = myClass2Arr is IMyInterface1[];
             return res;
         }
+
+        [GlobalSetup(Target = nameof(CheckArrayIsNonvariantGenericInterface))]
```
billwert: There's a fair bit of repetition with these setup methods. BDN has a different overload of `GlobalSetup` which takes an array:
 ```suggestion
        [GlobalSetup(Targets = new[] {  nameof(CheckArrayIsNonvariantGenericInterface), nameof(CheckArrayIsNonvariantGenericInterfaceNo)})]
```

**Comment by VSadov** [comment]
Also -  it looks like I do not have write access to this. 
If this meets merge requirements, can someone merge the PR?

File: src/benchmarks/micro/coreclr/perflab/CastingPerf.cs (hunk 3 above)
Comment (VSadov) [reply src/benchmarks/micro/coreclr/perflab/CastingPerf.cs]: I did not know such overload exists. Will change of course. 
It was a bit annoying to repeat the same thing. Thanks!

File: src/benchmarks/micro/coreclr/perflab/CastingPerf.cs [hunk 4]
```diff
@@ -373,5 +374,59 @@ public bool CheckArrayIsInterfaceNo()
...
+            return myObj is IReadOnlyCollection<Exception>;
+        }
+
+        [GlobalSetup(Target = nameof(AssignArrayElementByVariance))]
+        public void SetupAssignArrayElementByVariance() => myClass2Arr = new IReadOnlyCollection<object>[5];
+
+        [Benchmark]
+        public void AssignArrayElementByVariance()
+        {
+            // no return is needed. The cast is potentially throwing and thus is not optimizable.
```
adamsitnik: thanks for adding the comment! 👍

PR #934, opened 2019-10-13
Allow ResultsComparer to run without specifying target framework

svick:
This allows running ResultsComparer [as documented](https://github.com/dotnet/performance/blob/master/docs/benchmarking-workflow-corefx.md#preventing-regressions) with just `dotnet run`. Without this change, doing that produces the following error:

> Unable to run your project
> Your project targets multiple frameworks. Specify which framework to run using '--framework'.

And to fix that without changing the csproj, it's required to use `dotnet run -f netcoreapp2.1`.


adamsitnik:
The CI failure seems unrelated, I am going to re-run it.

/azp run

PR #936, opened 2019-10-13
Added serializing object property to JSON

svick:
Related to https://github.com/dotnet/corefx/pull/41753 and https://github.com/dotnet/corefx/issues/41638.

This is a case where System.Text.Json currently has a performance issue, so I think it makes sense to add a benchmark for it, independently of whether the PR https://github.com/dotnet/corefx/pull/41753 is accepted into corefx.

### DIFF
--- a/src/benchmarks/micro/corefx/System.Text.Json/Serializer/WriteJson.cs
+++ b/src/benchmarks/micro/corefx/System.Text.Json/Serializer/WriteJson.cs
@@ -28,6 +28,7 @@ namespace System.Text.Json.Serialization.Tests
     {
         private T _value;
         private MemoryStream _memoryStream;
+        private object _objectWithObjectProperty;
 
         [GlobalSetup]
         public async Task Setup()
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/corefx/System.Text.Json/Serializer/WriteJson.cs
```diff
@@ -54,6 +54,10 @@ public async Task SerializeToStream()
             await JsonSerializer.SerializeAsync(_memoryStream, _value);
         }
 
+        [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
+        [Benchmark]
+        public string SerializeObjectProperty() => JsonSerializer.Serialize(new { Prop = (object)_value });
```
adamsitnik: Would it be possible to move the creation of the `new { Prop = (object)_value }` out of the benchmark?

In the current version, the benchmark includes the cost of creating and serializing the object, while if possible it should be the cost of serialization only.

svick: Good point, fixed.

ahsonkhan:
With your fix from https://github.com/dotnet/corefx/issues/41638, is the performance of `SerializeToString` and `SerializeObjectProperty` almost identical now or is there still some gap here?

svick:
@ahsonkhan There is still some gap, though considering that `SerializeObjectProperty` does more, I'd say it's fairly small. For example, for the `Dictionary<string, string>` case, `SerializeToString` takes ~44 us, while `SerializeObjectProperty` takes now ~51 us (for comparison, previously it was over 2200 us).

But I haven't looked further into it, so it's possible there are still other optimizations to be had.

PR #940, opened 2019-10-15
Update BlobClient for new SDK libs

DrewScoggins:
The Python libs that we are using had a breaking change that renamed the
named parameters that we were using. This updates to the correct names.


DrewScoggins:
This broke uploads starting a week ago.

PR #941, opened 2019-10-16
add matching command lines for parsers

ooooolivia:
OS is recycling pids more aggressively than expected and it is very easy to have some iteration setup process (specifically, `dotnet build-server shutdown`) that shares the same process name and pid with the test process. This will add a false result to the result list. We want to avoid it by also matching the command line of the process when finding the right events.


src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -247,7 +249,6 @@ private static ProcessHelper CreateProcHelper(string command, string args, strin
                 ProcessWillExit = true,
                 Executable = command,
                 Arguments = args,
-                WorkingDirectory = workingDir,
```
ooooolivia: Removing setting the working directory here because the setup process cannot delete the folder that's its working directory.

billwert: If you're going to remove setting it, you can remove taking it as a parameter.

ooooolivia: good catch!

PR #943, opened 2019-10-16
Adding some perf test for System.Xml.Linq

buyaa-n:


buyaa-n:
Somehow i cannot add reviewers CC @adamsitnik @krwq

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XDocument.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Linq;
```
krwq: does this need license header?

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XDocument.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace System.Xml.Linq
+{
+    [BenchmarkCategory(Categories.CoreFX)]
+    public class Perf_XDocument
+    {
+        private XDocument _doc;
+
+        [Benchmark]
+        public XDocument Create() => new XDocument();
```
krwq: not sure if it makes sense to have this. Are you testing anything specific change with this?

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XDocument.cs
```diff
@@ -0,0 +1,36 @@
...
+        [Benchmark]
+        public XDocument Parse()
+        {
+            return XDocument.Parse("<elem1 child1='' child2='duu' child3='e1;e2;' child4='a1' child5='goody'> text node two e1; text node three </elem1>");
+        }
+
+        [Benchmark]
+        public XDocument CreateWithRootElement()
+        {
+            return new XDocument(new XElement("Root", "text node two e1; text node three"));
```
krwq: should you pre-create XElement in the setup? Presumably you're testing the XDocument constructor and not XElement

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XElement.cs
```diff
@@ -0,0 +1,56 @@
...
+        private XElement _element;
+
+        [Benchmark]
+        public XElement CreateElement() => new XElement("Root", "text node");
+
+        [Benchmark(OperationsPerInvoke = 8)]
+        public XElement CreateWithElements()
+        {
+            XElement doc = new XElement("Root",
+                new XElement("elem1", "text node two e1; text node three"),
```
krwq: I think the elements should be pre-created in the setup to isolate what you're testing

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XElementList.cs
```diff
@@ -0,0 +1,33 @@
...
+            _list = doc.DescendantsAndSelf();
+        }
+
+        [Benchmark]
+        public void Enumerator()
+        {
+            IEnumerable<XElement> list = _list;
+            Consumer consumer = _consumer;
+
+            foreach (var element in list)
```
krwq: I'd expand the foreach into the while loop to make sure you are using `IEnumerable<XElement>.GetEnumerator()` and not `IEnumerable.GetEnumerator()`

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XDocument.cs
buyaa-n: Sure it does, thanks

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XElementList.cs
krwq: (or at least confirm you are using the right one with sharp lab or some other tool)

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XDocument.cs
buyaa-n: My change wouldn't affect XDocument, but I assume it would be helpful for measuring any future changes could happen, but agree that XDocument creation logic not  much likely to be changed. Followed test cases from existing XmlDocument test

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XElement.cs
krwq: sounds good, please add a comment that this is typical user scenario

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XElementList.cs
krwq: actually if this is what customers normally do then disregard

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XName.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        [Benchmark]
+        public string GetEmptyNameSpaceName() => _noNamespace.NamespaceName;
+
+        [Benchmark]
+        public XName GetLocalNameFromExpandedName() => _hasNamespace.LocalName;
+
+        [Benchmark]
+        public string GetNonEmptyNameSpaceName() => _hasNamespace.NamespaceName;
+    }
```
adamsitnik: which of these properties perform some actual work (not just returning a field value)?

EDIT: I've run these benchmarks and I can see that two of them just return a field value and report `0ns` as a result.

|                       Method |       Mean |
|----------------------------- |-----------:|
|                 GetLocalName | 29.4854 ns |
|        GetEmptyNameSpaceName |  0.0000 ns |
| GetLocalNameFromExpandedName | 33.4970 ns |
|     GetNonEmptyNameSpaceName |  0.0000 ns |

Please remove `GetEmptyNameSpaceName` and `GetNonEmptyNameSpaceName`

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XElementList.cs
```diff
@@ -0,0 +1,33 @@
...
+using BenchmarkDotNet.Engines;
+using MicroBenchmarks;
+using System.Collections.Generic;
+
+namespace System.Xml.Linq
+{
+    [BenchmarkCategory(Categories.CoreFX)]
+    public class Perf_XElementList
+    {
+        private readonly Consumer _consumer = new Consumer();
```
adamsitnik: thanks for using the `Consumer`! 👍

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XElement.cs
```diff
@@ -0,0 +1,56 @@
...
+        public XElement GetElement() => _element.Element("Child4");
+        
+        [Benchmark]
+        public XAttribute GetAttribute() => _element.Attribute("id");
+
+        [Benchmark]
+        public string GetValue() => _element.Value;
+
+        [Benchmark]
+        public XName GetXName() => _element.Name;
```
adamsitnik: I've run these benchmarks and it looks like `.Name` just returns a field value and does not perform any actual work:

|             Method |       Mean |
|------------------- |-----------:|
|      CreateElement | 48.9110 ns |
| CreateWithElements | 87.9020 ns |
|         GetElement | 40.9684 ns |
|       GetAttribute | 30.7949 ns |
|           GetValue | 54.9559 ns |
|           GetXName |  0.0000 ns |

Please remove the `GetXName` benchmark

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XElement.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        [Benchmark]
+        public XElement CreateElement() => new XElement("Root", "text node");
+
+        [Benchmark(OperationsPerInvoke = 8)]
+        public XElement CreateWithElements()
+        {
+            XElement doc = new XElement("Root",
+                new XElement("elem1", "text node two e1; text node three"),
+                new XElement("elem1", "text node two e1; text node three"),
```
danmosemsft: > "text node two e1; text node three"

This sounds like it makes several nodes but it simply creates `<elem1>text node two e1; text node three</elem1>`

A string like "some xml element content" would be clearer.

src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XName.cs
buyaa-n: I see, seems better to check ToString() instead

PR #944, opened 2019-10-16
Putting this back the way it was

billwert:
This didn't fix the problem the way I'd hoped


billwert:
@DrewScoggins stood over my shoulder and approved this.

PR #945, opened 2019-10-16
add two tests for NetworkInterface

wfurt:
GetAllNetworkInterfaces simply get list of all network interfaces. 
GetAllNetworkInterfacesProperties iterates through most of interesting properties to cover any cases where we throw or possibly use lazy initialization.

|                            Method |     Mean |     Error |    StdDev |   Median |      Min |      Max |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|---------------------------------- |---------:|----------:|----------:|---------:|---------:|---------:|---------:|--------:|------:|----------:|
|           GetAllNetworkInterfaces | 3.539 ms | 0.0284 ms | 0.0237 ms | 3.528 ms | 3.515 ms | 3.589 ms | 450.0000 | 12.5000 |     - |   1.81 MB |
| GetAllNetworkInterfacesProperties | 7.388 ms | 0.0261 ms | 0.0232 ms | 7.385 ms | 7.355 ms | 7.436 ms | 515.1515 | 30.3030 |     - |   2.12 MB |

### DIFF
--- a/src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
+++ b/src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
@@ -12,7 +12,7 @@ namespace System.Net.NetworkInformation.Tests
     public class NetworkInterfaceTests
     {
         [Benchmark]
-        public void GetAllNetworkInterfaces() => NetworkInterface.GetAllNetworkInterfaces();
+        public NetworkInterface[] GetAllNetworkInterfaces() => NetworkInterface.GetAllNetworkInterfaces();
 
         [Benchmark]
         public void GetAllNetworkInterfacesProperties()
--- a/src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
+++ b/src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
@@ -1,0 +1,61 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.NetworkInformation;
+using System.Reflection;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.Net.NetworkInformation.Tests
+{
+    [BenchmarkCategory(Categories.CoreFX)]
+    public class NetworkInterfaceTests
+    {
+        private PropertyInfo[] NetworkInterfaceInfo = typeof(System.Net.NetworkInformation.NetworkInterface).GetProperties(BindingFlags.Instance | BindingFlags.Public);
+        private PropertyInfo[] IPInterfaceStatisticsInfo = typeof(System.Net.NetworkInformation.IPInterfaceStatistics).GetProperties(BindingFlags.Instance | BindingFlags.Public);
+        private PropertyInfo[] IPInterfacePropertiesInfo = typeof(System.Net.NetworkInformation.IPInterfaceProperties).GetProperties(BindingFlags.Instance | BindingFlags.Public);
+
+        [Benchmark]
+        public void GetAllNetworkInterfaces() => NetworkInformation.NetworkInterface.GetAllNetworkInterfaces();
+
+        [Benchmark]
+        public void GetAllNetworkInterfacesProperties()
+        {
+            foreach (NetworkInterface ni in NetworkInterface.GetAllNetworkInterfaces())
+            {
+                IPInterfaceStatistics ips = ni.GetIPStatistics();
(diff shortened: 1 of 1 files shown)


wfurt:
cc: @stephentoub @adamsitnik

src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using MicroBenchmarks;
+
+namespace System.Net.NetworkInformation.Tests
+{
+    [BenchmarkCategory(Categories.CoreFX)]
+    public class NetworkInterfaceTests
+    {
+        private PropertyInfo[] NetworkInterfaceInfo = typeof(System.Net.NetworkInformation.NetworkInterface).GetProperties(BindingFlags.Instance | BindingFlags.Public);
+        private PropertyInfo[] IPInterfaceStatisticsInfo = typeof(System.Net.NetworkInformation.IPInterfaceStatistics).GetProperties(BindingFlags.Instance | BindingFlags.Public);
+        private PropertyInfo[] IPInterfacePropertiesInfo = typeof(System.Net.NetworkInformation.IPInterfaceProperties).GetProperties(BindingFlags.Instance | BindingFlags.Public);
```
stephentoub: I think we should be explicit about the properties we're accessing, ideally with one benchmark per property we care about, rather than using reflection to enumerate all of them as part of a single test.

src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        [Benchmark]
+        public void GetAllNetworkInterfaces() => NetworkInformation.NetworkInterface.GetAllNetworkInterfaces();
```
stephentoub: Is the `NetworkInformation` necessary here?

adamsitnik: You should return the benchmark result to prevent from possible [dead code elimination](https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#dead-code-elimination)

```suggestion
        public NetworkInterface[] GetAllNetworkInterfaces() => NetworkInterface.GetAllNetworkInterfaces();
```

src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
adamsitnik: I agree with Stephen, this should be explicit without reflection usage. Reflection can add some performance overhead and definitely introduce boxing which would be visible in the "allocated" memory column.

Example:

```cs
public long GetAllNetworkInterfacesProperties()
{
    long consume = 0;

    foreach (NetworkInterface ni in NetworkInterface.GetAllNetworkInterfaces())
    {
        consume += ni.Id.Length;
        consume += ni.SupportsMulticast ? 1 : 0;

        IPInterfaceStatistics ips = ni.GetIPStatistics();

        consume += ips.BytesReceived;
        consume += ips.BytesSent;

        IPInterfaceProperties ipp = ni.GetIPProperties();

        consume += ipp.UnicastAddresses.Count;
        consume += ipp.DnsSuffix.Length;
    }

    return consume;
}
```

wfurt:
test with and without reflection:

|                                      Method |     Mean |     Error |    StdDev |   Median |      Min |      Max |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------------------------------------- |---------:|----------:|----------:|---------:|---------:|---------:|---------:|--------:|------:|----------:|
| GetAllNetworkInterfacesPropertiesReflection | 7.481 ms | 0.0184 ms | 0.0154 ms | 7.479 ms | 7.459 ms | 7.509 ms | 500.0000 | 31.2500 |     - |   2.12 MB |
|           GetAllNetworkInterfacesProperties | 5.404 ms | 0.0278 ms | 0.0247 ms | 5.397 ms | 5.374 ms | 5.458 ms | 500.0000 |       - |     - |   2.08 MB |

src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
wfurt: done.

src/benchmarks/micro/corefx/System.Net.NetworkInformation/NetworkInterfaceTests.cs
wfurt: updated.

adamsitnik:
Are the try & catch blocks necessary? When these operations can throw?

wfurt:
yes, they are there for reason. Individual properties can throw depending on driver and OS configuration. 
I made some fixes to throw less  but I wanted to make it safe for older versions.

adamsitnik:
the CI failure is not related, so I am merging

PR #951, opened 2019-10-17
Create benchmarks for System.Threading.Tasks.Dataflow

### LINKED ISSUE #950: Dataflow needs performance tests
There is no benchmark for the System.Threading.Tasks.Dataflow library.
Given that Dataflow is often used in code that is performance critical, I think it should have at least a few performance tests.

See https://github.com/dotnet/corefx/issues/14049

manandre:
I propose multiple benchmarks for the `System.Threading.Tasks.Dataflow` library.
They cover some main performance-related use cases for each type of Dataflow block (BufferBlock, ActionBlock, TransformBlock…)

Fixes #950

### DIFF
--- a/src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
+++ b/src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
@@ -5,25 +5,53 @@
 using BenchmarkDotNet.Attributes;
 using MicroBenchmarks;
 
-//#pragma warning disable CS1998 // async methods without awaits
-
 namespace System.Threading.Tasks.Dataflow.Tests
 {
-    [BenchmarkCategory(Categories.CoreFX)]
-    public class Perf_Dataflow
+    public class UnboundedBufferBlockPerfTests : DefaultPropagatorPerfTests
     {
-        [Benchmark]
-        public async Task BufferBlock_Completion()
+        public override IPropagatorBlock<int, int> CreateBlock() => new BufferBlock<int>();
+    }
+
+    public class BoundedBufferBlockPerfTests : DefaultBoundedPropagatorPerfTests
+    {
+        public override IPropagatorBlock<int, int> CreateBlock() => new BufferBlock<int>(new DataflowBlockOptions { BoundedCapacity = 100 });
+    }
+
+    public class ActionBlockPerfTests : DefaultTargetPerfTests
+    {
+        public override ITargetBlock<int> CreateBlock() => new ActionBlock<int>(i => { });
+    }
+
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class PerfTests<T> where T : IDataflowBlock
+    {
+        protected T block;
+
+        public abstract T CreateBlock();
+
+        [GlobalSetup]
+        public void BlockSetup()
         {
-            var block = new BufferBlock<int>();
+            block = CreateBlock();
+        }
+
+        [Benchmark]
+        public async Task Completion()
+        {
             block.Complete();
             await block.Completion;
         }
+    }
 
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class DefaultTargetPerfTests : PerfTests<ITargetBlock<int>> { }
+
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class TargetPerfTests<T> : PerfTests<T> where T : ITargetBlock<int>
+    {
         [Benchmark(OperationsPerInvoke = 100_000)]
-        public void BufferBlock_Post()
+        public void Post()
         {
-            var block = new BufferBlock<int>();
             for (int i = 0; i < 100_000; i++)
             {
                 block.Post(i);
@@ -31,19 +59,27 @@ namespace System.Threading.Tasks.Dataflow.Tests
         }
 
         [Benchmark(OperationsPerInvoke = 100_000)]
-        public async Task BufferBlock_SendAsync()
+        public async Task SendAsync()
         {
-            var block = new BufferBlock<int>();
             for (int i = 0; i < 100_000; i++)
             {
                 await block.SendAsync(i);
             }
         }
+    }
 
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class DefaultBoundedPropagatorPerfTests : BoundedPropagatorPerfTests<IPropagatorBlock<int, int>> { }
+
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class DefaultPropagatorPerfTests : PropagatorPerfTests<IPropagatorBlock<int, int>> { }
+
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class PropagatorPerfTests<T> : BoundedPropagatorPerfTests<T> where T : IPropagatorBlock<int, int> 
+    {
         [Benchmark(OperationsPerInvoke = 100_000)]
-        public void BufferBlock_PostReceiveSequential()
+        public void PostReceiveSequential()
         {
-            var block = new BufferBlock<int>();
             for (int i = 0; i < 100_000; i++)
             {
                 block.Post(i);
@@ -56,9 +92,8 @@ namespace System.Threading.Tasks.Dataflow.Tests
         }
 
         [Benchmark(OperationsPerInvoke = 100_000)]
-        public async Task BufferBlock_SendReceiveAsyncSequential()
+        public async Task SendReceiveAsyncSequential()
         {
-            var block = new BufferBlock<int>();
             for (int i = 0; i < 100_000; i++)
             {
                 await block.SendAsync(i);
@@ -69,22 +104,25 @@ namespace System.Threading.Tasks.Dataflow.Tests
                 await block.ReceiveAsync();
             }
         }
+    }
 
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class BoundedPropagatorPerfTests<T> : PerfTests<T> where T : IPropagatorBlock<int, int>
+    {
         [Benchmark(OperationsPerInvoke = 100_000)]
-        public async Task BufferBlock_PostReceiveParallel()
+        public async Task PostReceiveParallel()
         {
-            var block = new BufferBlock<int>();
             await Task.WhenAll(Post(), Receive());
-            
-            Task Post() => Task.Run(()=>
+
+            Task Post() => Task.Run(() =>
             {
                 for (int i = 0; i < 100_000; i++)
                 {
-                    block.Post(i);
+                    while (!block.Post(i));
                 }
             });
 
-            Task Receive() => Task.Run(()=>
+            Task Receive() => Task.Run(() =>
             {
                 for (int i = 0; i < 100_000; i++)
                 {
@@ -94,11 +132,10 @@ namespace System.Threading.Tasks.Dataflow.Tests
         }
 
         [Benchmark(OperationsPerInvoke = 100_000)]
-        public async Task BufferBlock_SendReceiveAsyncParallel()
+        public async Task SendReceiveAsyncParallel()
         {
-            var block = new BufferBlock<int>();
             await Task.WhenAll(SendAsync(), ReceiveAsync());
-            
+
             async Task SendAsync()
             {
                 for (int i = 0; i < 100_000; i++)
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
```diff
@@ -0,0 +1,384 @@
...
+                i => { },
+                new ExecutionDataflowBlockOptions
+                {
+                    MaxDegreeOfParallelism = Environment.ProcessorCount,
+                    EnsureOrdered = false
+                }
+            );
+    }
+
+    public class SPCActionBlockPerfTests : DefaultTargetPerfTests
```
adamsitnik: could you please use full name instead of the abbreviation here?

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
```diff
@@ -0,0 +1,384 @@
...
+    {
+        [Benchmark(OperationsPerInvoke = 100_000)]
+        public Task Post() => Post(block);
+
+        [Benchmark(OperationsPerInvoke = 100_000)]
+        public Task SendAsync() => SendAsync(block);
+    }
+
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class DefaultBoundedPropagatorPerfTests : BoundedPropagatorPerfTests<IPropagatorBlock<int, int>, int> { }
```
adamsitnik: could you please remove this class to reduce the class hierarchy depth and use `BoundedPropagatorPerfTests<IPropagatorBlock<int, int>, int>` instead of `DefaultBoundedPropagatorPerfTests` when defining inheritor classes?

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
```diff
@@ -0,0 +1,384 @@
...
+
+    [BenchmarkCategory(Categories.CoreFX)]
+    public abstract class DefaultPropagatorPerfTests : PropagatorPerfTests<IPropagatorBlock<int, int>, int> { }
```
adamsitnik: could you please remove this class to reduce the class hierarchy depth and use `PropagatorPerfTests<IPropagatorBlock<int, int>, int>` instead of `DefaultPropagatorPerfTests ` when defining inheritor classes?

manandre:
@adamsitnik Are there so many issues? 😄 
Do you already have some first remarks to share?

adamsitnik:
@stephentoub could you please take a look?

adamsitnik:
> Are there so many issues?

@manandre no, appologies for the delay. It looks good to me from the benchmarking perspective, but I would like to know @stephentoub opinion before we merge it

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.ActionBlock.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    public class UnorderedParallelActionBlockPerfTests : TargetPerfTests<ITargetBlock<int>>
+    {
+        public override ITargetBlock<int> CreateBlock() =>
+            new ActionBlock<int>(
+                i => { },
+                new ExecutionDataflowBlockOptions
+                {
+                    MaxDegreeOfParallelism = Environment.ProcessorCount,
+                    EnsureOrdered = false
```
stephentoub: EnsureOrdered has no impact on ActionBlock. This isn't worth measuring.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.BroadcastBlock.cs
```diff
@@ -0,0 +1,41 @@
...
+    [BenchmarkCategory(Categories.CoreFX)]
+    public class BroadcastBlockPerfTests : ReceivablePropagatorPerfTests<BroadcastBlock<int>, int>
+    {
+        public override BroadcastBlock<int> CreateBlock() =>
+            new BroadcastBlock<int>(
+                i => i,
+                new ExecutionDataflowBlockOptions
+                {
+                    MaxDegreeOfParallelism = Environment.ProcessorCount
+                });
```
stephentoub: BroadcastBlock takes DataflowBlockOptions, not ExecutionDataflowBlockOptions, meaning it pays no attention to MaxDegreeOfParallelism.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.BroadcastBlock.cs
```diff
@@ -0,0 +1,41 @@
...
+        private async Task MultiParallel(Func<Task> doTask)
+        {
+            var options = new DataflowLinkOptions { PropagateCompletion = true };
+            var action1 = new ActionBlock<int>(i => { });
+            var action2 = new ActionBlock<int>(i => { });
+            block.LinkTo(action1, options);
+            block.LinkTo(action2, options);
+
+            await doTask();
+            block.Complete();
```
stephentoub: It looks like CreateBlock is called from the GlobalSetup, which means it's only going to be invoked once per test, regardless of the number of times the benchmark method is invoked.  Which means after the first time, the block will already have been completed.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
```diff
@@ -0,0 +1,219 @@
...
+        public void BlockSetup()
+        {
+            block = CreateBlock();
+        }
+
+        [Benchmark]
+        public async Task Completion()
+        {
+            block.Complete();
+            await block.Completion;
```
stephentoub: I'm struggling to understand the setup here.  BlockSetup creates the block, and then this Completion benchmark is going to be invoked a bunch of times, but all times other than the first, it'll already be complete. No?  What is this measuring then?

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.ActionBlock.cs
manandre: I agree. Fixed.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.BroadcastBlock.cs
manandre: I agree. Fixed.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.BroadcastBlock.cs
manandre: Moved GlobalSetup to IterationSetup.
Does it solve the issue ?

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
manandre: Ditto.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.BroadcastBlock.cs
stephentoub: I believe it addresses the issue, but potentially creates another. @adamsitnik, this is going to result in insufficient iteration counts by default, right?

adamsitnik: @stephentoub is right. In such case we should most probably call `CreateBlock` from the benchmark itself.

manandre: Fixed.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
```diff
@@ -0,0 +1,218 @@
...
+        protected const int MessagesCount = 100_000;
+        protected T block;
+
+        public abstract T CreateBlock();
+
+        [GlobalSetup]
+        public void BlockSetup()
+        {
+            block = CreateBlock();
+        }
```
stephentoub: Why is this GlobalSetup still needed when we're calling BlockSetup on each benchmark invocation?

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
```diff
@@ -0,0 +1,218 @@
...
+            for (int i = 0; i < MessagesCount / receiveSize; i++)
+            {
+                while (!source.TryReceive(out _) && retry) ;
+            }
+        });
+
+        protected static Task TryReceiveAll<U>(IReceivableSourceBlock<U> source) => Task.Run(() =>
+        {
+            source.TryReceiveAll(out _);
+        });
```
stephentoub: Why is this queueing a task to invoke the method?  Why wouldn't callers just invoke the method synchronously?

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        protected static Task Receive<U>(ISourceBlock<U> source, int receiveSize = 1) => Task.Run(() =>
+        {
+            for (int i = 0; i < MessagesCount / receiveSize; i++)
+            {
+                source.Receive();
+            }
+        });
+
+        protected static Task TryReceive<U>(IReceivableSourceBlock<U> source, int receiveSize = 1, bool retry = false) => Task.Run(() =>
```
stephentoub: Same comment as below for all of these.  The extra task seems superfluous, or if it's necessary for some reason, it should be done by the caller where it's necessary.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.TransformManyBlock.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace System.Threading.Tasks.Dataflow.Tests
+{
+    public class TransformManyBlockPerfTests : ReceivablePropagatorPerfTests<TransformManyBlock<int, int>, int>
+    {
+        public override TransformManyBlock<int, int> CreateBlock() =>
+            new TransformManyBlock<int, int>(
+                i => new int[] { i },
+                new ExecutionDataflowBlockOptions
+                {
+                    MaxDegreeOfParallelism = Environment.ProcessorCount
```
stephentoub: Why do other blocks test for non-parallel and this one only tests parallel?

stephentoub:
At a high-level this seems fine.  Personally I think the abstraction is unnecessary and detracts rather than adds value, but I'll leave that up to Adam.  I'm also a little skeptical about the setup because I suspect many of the tests are going to be measuring costs that aren't actually that relevant, e.g. a given test here creates a block, does stuff, and then waits for the block to complete, and that's going to include the overhead of creating the blocks, which in many cases would be done much less frequently. Regardless, thanks for writing these and contributing.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
manandre: We only call `BlockSetup` from the benchmark code when we are forced to call `Complete` on the tested block to have a valid test case. This concerns only two benchmarks.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
manandre: Good remark! Fixed.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.Dataflow.cs
manandre: Good remark! Fixed.

src/benchmarks/micro/corefx/System.Threading.Tasks.Dataflow/Perf.TransformManyBlock.cs
manandre: Good remark! Fixed.

manandre:
@adamsitnik Do we need to go further ? If yes, please indicate which points to address 🎯

billwert:
@adamsitnik Should this be merged?

PR #954, opened 2019-10-18
Mistakenly stopped setting perf lab arguments.

billwert:


billwert:
@DrewScoggins signed off over my shoulder.

PR #955, opened 2019-10-18
Remove double single quotes around package name

DrewScoggins:


DrewScoggins:
For some reason pip does not like the way the quoting is happening for the package name. Since they are not required I am just going to remove them to unblock the lab.

PR #960, opened 2019-10-19
Fixing Perf_Single.IsNaN and Perf_Double.IsNaN to do `result |=` rather than `result &=`

tannergooding:
This resolves an issue where LLVM was optimizing the code to just `xor eax, eax`: https://github.com/dotnet/performance/pull/952#issuecomment-544185927

CC. @EgorBo, @adamsitnik


tannergooding:
### LLVM IR
Before: https://godbolt.org/z/Crz1o7
After: https://godbolt.org/z/vTvkBq

### C++
Before: https://godbolt.org/z/huc8am
After: https://godbolt.org/z/7XHiUa

PR #968, opened 2019-10-23
fix System.Memory.ReadOnlySpan.IndexOfString benchmark logical bug

adamsitnik:
The setup method was supposed to use `source` char `count`-many times and put `replaceChar` at the `replacePos` index.

I found this bug when looking for Windows vs Linux performance comparison


adamsitnik:
since this is a trivial bug fix and I really need to re-run these benchmarks for Windows vs Linux investigation I am going to merge it right now

PR #977, opened 2019-10-28
add sdk scenarios

ooooolivia:
- added backup to precommands
- added sdk-iteration-setup.py script
- added netcoreapp scenario
- changed runner and startup to accomodate sdk tests


src/scenarios/shared/runner.py
```diff
@@ -62,11 +70,39 @@ def run(self):
...
-        #     startup.runtests(**self.traits._asdict(),
-        #         scenariotypename='Build No Changes')
-        #     # fix some other traits
-        #     startup.runtests(**self.traits._asdict(),
-        #         scenariotypename='Rebuild')
\ No newline at end of file
+        elif self.testtype == const.SDK:
+            startup = StartupWrapper()
+            envlistbuild = 'DOTNET_MULTILEVEL_LOOKUP=0'
+            envlistcleanbuild= envlistbuild+';'+'MSBUILDDISABLENODEREUSE=1'
```
billwert: nit: to make this easier in the future, if you make the environment variables a list you can just join them together:
```
>>> ';'.join(['a','b','c'])
'a;b;c'
```
If we add one it's just adding to the list, then.

src/scenarios/shared/runner.py
```diff
@@ -62,11 +70,39 @@ def run(self):
...
+            # clean build
+            startup.runtests(scenarioname=self.traits.scenarioname,
+                             exename=self.traits.exename,
+                             guiapp=self.traits.guiapp,
+                             startupmetric=const.STARTUP_PROCESSTIME,
+                             appargs='build',
+                             timeout=self.traits.timeout,
+                             warmup='false',
```
billwert: why should we not warm up the clean build case? It's still going to run many iterations so logically the first will be "warmup" anyway.

src/scenarios/shared/sdk_iteration_setup.py
```diff
@@ -0,0 +1,20 @@
...
+from shared import const
+from dotnet import shutdown_server
+from performance.logger import setup_loggers
+from logging import getLogger
+
+
+def main():
+    setup_loggers(True)
+    shutdown_server(verbose=True)
+    getLogger().info("Removing project directory...") # set up logger would create new log files, which is not necessary
```
billwert: What's the comment here mean?

src/scenarios/shared/startup.py
```diff
@@ -53,16 +53,28 @@ def runtests(self, apptorun: str, **kwargs):
             '--app-exe', apptorun,
             '--metric-type', kwargs['startupmetric'], 
             '--scenario-name', "%s - %s" % (kwargs['scenarioname'], kwargs['scenariotypename']),
-            '--trace-file-name', '%s_startup.etl' % kwargs['exename'],
+            '--trace-file-name', '%s_%s_startup.etl' % (kwargs['exename'], kwargs['scenariotypename']),
```
billwert: this potentially will introduce spaces into the file name. I guess that means everything else already handles spaces? Nice! :)

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -314,5 +327,19 @@ private static void WriteResultTable(IEnumerable<Counter> counters, Logger logge
...
         }
+
+        private static Dictionary<string, string> ParseStringToDictionary(string s)
+        {
+            Dictionary<string, string> dict = new Dictionary<string, string>();
+            foreach (string substring in s.Split(';', StringSplitOptions.RemoveEmptyEntries))
+            {
+                string key = substring.Split('=').First();
+                string value = substring.Split('=').Last();
+                if(!String.IsNullOrWhiteSpace(key) && !String.IsNullOrWhiteSpace(value)){
```
billwert: I would not require a value here. `SET FOO=` is valid to unset `FOO`. Instead of the value is null, set it to empty string?

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
```diff
@@ -52,6 +53,16 @@ public ProcessHelper(Logger logger)
             psi.Arguments = Arguments;
             psi.WorkingDirectory = WorkingDirectory;
             psi.UseShellExecute = GuiApp;
+
+            if (EnvironmentVariables != null)
+            {
+                foreach (KeyValuePair<string, string> pair in EnvironmentVariables)
```
billwert: good spot for `var`.
```suggestion
                foreach (var pair in EnvironmentVariables)
```

eng/performance/scenarios.proj
```diff
@@ -25,4 +25,14 @@
       <Timeout>4:00</Timeout>
     </HelixWorkItem>
   </ItemGroup>
-</Project>
\ No newline at end of file
+
+  <ItemGroup>
+    <HelixWorkItem Include="SDK .NET 2.0 Console Template">
+      <PayloadDirectory>$(ScenariosDir)netcoreapp</PayloadDirectory>
```
billwert: Can the directory name be something like `netcoreapp2.0` since it is hardcoded?

src/scenarios/netcoreapp/test.py
```diff
@@ -0,0 +1,21 @@
...
+
+SCENARIO_NAME = '.NET Core 2.0 Console Template'
+EXE_NAME = 'NetCoreApp'
+
+def main():
+    traits = TestTraits(scenarioname=SCENARIO_NAME,
+                        exename=EXE_NAME,
+                        guiapp='false',  # string passed through to tool
+                        sdk=True,
+                        startup=True,
```
billwert: This should not be a startup test since we never run a `netcoreapp2.0` compatible framework.

src/scenarios/netcoreapp/test.py
```diff
@@ -0,0 +1,21 @@
...
+                        )
+    runner = Runner(traits)
+    runner.parseargs()
```
billwert: calling `run` is sufficient. It calls `parseargs`.

src/scenarios/shared/const.py
```diff
@@ -14,3 +16,14 @@
 PUBDIR = 'pub'
 APPDIR = 'app'
 TRACEDIR = 'traces'
+TMPDIR = 'tmp'
+
+DOTNET = 'dotnet'
+PYTHON = 'py'
```
billwert: You should pull this out of the environment. We set it into an environment variable called `Python`.

src/scenarios/shared/sdk_iteration_setup.py
```diff
@@ -0,0 +1,20 @@
...
+    if os.path.isdir(const.TMPDIR):
+        shutil.rmtree(const.TMPDIR)
+    getLogger().info("Copying clean project directory...")
+    shutil.copytree(const.APPDIR, const.TMPDIR)
```
billwert: This code is duplicated in `precommands.py`. It's basically the same as the `backup` function. Given that the harness will run this every time, the `tmp` directory will be created as part of `pre.py` and then again each iteration. There's a wasted step there. It's fine if SDK tests don't have `pre.py` invocations if they don't need them.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -83,8 +86,17 @@ static void checkArg(string arg, string name)
                     Directory.CreateDirectory(traceDirectory);
                 }
             }
+
+            Dictionary<string, string> envDict = null;
```
billwert: nit: we don't need to know if it's a dict as part of the name.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -314,5 +327,19 @@ private static void WriteResultTable(IEnumerable<Counter> counters, Logger logge
                 logger.Log($"{counter.Name,-15}|{average,-15}|{min,-15}|{max,-15}");
             }
         }
+
+        private static Dictionary<string, string> ParseStringToDictionary(string s)
+        {
+            Dictionary<string, string> dict = new Dictionary<string, string>();
```
billwert: another good spot for `var`.
```suggestion
            var dict = new Dictionary<string, string>();
```

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -314,5 +327,19 @@ private static void WriteResultTable(IEnumerable<Counter> counters, Logger logge
...
+            foreach (string substring in s.Split(';', StringSplitOptions.RemoveEmptyEntries))
+            {
+                string key = substring.Split('=').First();
+                string value = substring.Split('=').Last();
```
billwert: Instead of calling split twice, store the array and index into it.
```
var values = substring.Split('=');
var key = values[0];
var value = values[1];
```
`Split` will also naturally give you the empty string, so nothing has to be done to handle the case I called out below:
```
> "foo=".Split('=')
string[2] { "foo", "" }
```

src/scenarios/shared/runner.py
ooooolivia: good point!

src/scenarios/shared/sdk_iteration_setup.py
ooooolivia: oops should've been deleted

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
ooooolivia: because 1) the variable is only used locally and  2) it's more convenient to do declare the implicit type?

src/scenarios/netcoreapp/test.py
ooooolivia: do you mean startup=True implies running the app's executable under netcoreapp2.0 framework?

src/scenarios/Directory.Build.props
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
DrewScoggins: Why does this file and the Directory.Build.targets exist?

src/scenarios/shared/const.py
```diff
@@ -14,3 +16,14 @@
...
+
+BUILD_CLEAN = 'Clean Build'
+BUILD_NO_CHANGES = 'Build No Changes'
```
DrewScoggins: Do we really need these to be defined in const? We seem to only use them in one place.

src/scenarios/netcoreapp/test.py
billwert: Not precisely. `startup=True` will attempt to run the startup harness using the generated app as the target. That will fail because we have no `netcoreapp2.0` runtime on the test machine (unless we are running `netcoreapp2.0` tests, which we don't measure.)

src/scenarios/shared/startup.py
DrewScoggins: Why will this introduce spaces? I would assume that is scenariotypename is None that it is just ""

src/tools/ScenarioMeasurement/Startup/Startup.cs
DrewScoggins: But do we want that semantic enforced here? It feels like we want to use this just to set new environment variables for when we want to run, not use it to arbitrarily shape the environment space

billwert: I can pretty easily imagine I want to run scenarios where I ensure a particular variable isn't set. Say I want to be very sure tiered compilation isn't disabled, etc.

However the semantics don't quite work like I thought they would. To make this work we'd have to actually remove the variable from the collection if it had an empty string. What we have here is sufficient.

src/scenarios/shared/sdk_iteration_setup.py
ooooolivia: The current test order is build(no change) runs after clean build. However, if we flip the order or run only the build(no change) scenario, pre.py will still be needed.

src/tools/ScenarioMeasurement/Startup/Startup.cs
ooooolivia: nice! Didn't know it could gently handle index out of bounds.

src/scenarios/Directory.Build.props
ooooolivia: it's placed there to override the arcade settings

src/scenarios/shared/const.py
ooooolivia: u r right. no good reason to make it constant for now

src/scenarios/shared/const.py
ooooolivia: Weird I didn't see the ```Python``` environment variable in the machine-setup.py dump? If that's already in the PATH we can just invoke ```py``` or ```python```. Or do you mean it should call ```%Python%```.

src/tools/ScenarioMeasurement/Startup/Startup.cs
ooooolivia: great solution below!

src/scenarios/shared/runner.py
ooooolivia: true. The first-ever iteration could be much slower and a warmup iteration could make all the iterations consistent.

src/scenarios/shared/startup.py
ooooolivia: scenariotypename has space in it, like ```SDK (Clean Build)```

src/scenarios/shared/runner.py
```diff
@@ -62,11 +70,39 @@ def run(self):
...
-        #     startup.runtests(**self.traits._asdict(),
-        #         scenariotypename='Build No Changes')
-        #     # fix some other traits
-        #     startup.runtests(**self.traits._asdict(),
-        #         scenariotypename='Rebuild')
\ No newline at end of file
+        elif self.testtype == const.SDK:
+            startup = StartupWrapper()
+            envlistbuild = 'DOTNET_MULTILEVEL_LOOKUP=0'
+            envlistcleanbuild= ';'.join(['MSBUILDDISABLENODEREUSE=1', envlistbuild])
```
billwert:
```suggestion
            envlistcleanbuild= ';'.join(envlistbuild)
```

src/scenarios/shared/sdk_iteration_setup.py
billwert: Okay.

src/scenarios/shared/startup.py
DrewScoggins: Of course

src/scenarios/shared/runner.py
```diff
@@ -62,11 +70,39 @@ def run(self):
...
-        #     startup = StartupWrapper()
-        #     startup.runtests(**self.traits._asdict(),
-        #         scenariotypename='Build No Changes')
-        #     # fix some other traits
-        #     startup.runtests(**self.traits._asdict(),
-        #         scenariotypename='Rebuild')
\ No newline at end of file
+        elif self.testtype == const.SDK:
+            startup = StartupWrapper()
+            envlistbuild = 'DOTNET_MULTILEVEL_LOOKUP=0'
```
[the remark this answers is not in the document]
ooooolivia: hmm why would we need to turn off node reuse for build (no changes)? Leaving it on seems closer to the customer scenario.

Issue #896, opened 2019-09-24
Create a plan for tracking moving product builds of non-platform libraries/tools


billwert:
Currently we run ML.NET and Roslyn benchmarks in this repo. They use pinned versions of those products and test against live .NET Core bits. This is a great defense mechanism for ensuring future .NET Core products don't have regressions. It does not help us track the performance of ML.NET and Roslyn as they change, however. We'd like to do that as well.

The challenge is that in a given perf run we'd like there to be a single variable. Moving both the underlying .NET Core version as well as the library version makes it very hard to reason about what caused a regression. This is complicated by the fact that it is reasonable to assume that future versions of libraries and tools will take dependencies on new .NET Core features in prerelased versions. We can't expect a given Roslyn build to just work on the previously RTMed .NET Core.

We need a scheme that allows for:

* Rolling forward the non-platform bits frequently
* Pinning a .NET Core platform version as long as possible
* Has a mechanism for moving the pinned version with fixed versions of the other libraries

There may be other considerations but those are the critical ones to me.

BruceForstall:
A specific example: based on the discussion in https://github.com/dotnet/performance/issues/894, ML.NET in this repo is 0.10.0, which is quite old compared to the current version, 1.4.0-preview2 (https://www.nuget.org/packages/Microsoft.ML/). In particular, it sounds like one important difference is that the current version uses managed SIMD / hardware intrinsics instructions instead of P/Invoke to a native library for these calculations. It's important to the JIT team, at least, to have these newer implementations be in the benchmark suite.

adamsitnik:
One of the reasons why we did not update ML.NET version for a while was the fact that it was very frequently changing the public API. Since ML.NET 1.0 no breaking changes were introduced. 

ML.NET has it's own copy of the benchmarks, which are kept up-to-date. To update, we could just copy the source code from https://github.com/dotnet/machinelearning/tree/master/test/Microsoft.ML.Benchmarks

adamsitnik:
Maybe we could have two jobs defined for ML.NET and Roslyn benchmarks: one using hardcoded and one latest version. I could extend the `.yml` micro-benchmark template with something optional like:

```yml
      versions: 
        - hardcoded
        - latest
```

And handle it properly from the MSBuild level using some Env Vars and conditions:

```xml
  <ItemGroup>
    <PackageReference Condition="'$(LibVersion)' == 'Hardcoded' Include="ML.NET" Version="1.4" />
[...]

PR #988, opened 2019-10-31
ML.NET rolling and fixed version CI legs

adamsitnik:
In #896 we have agreed that it would be good to have two different configs for ML.NET:

* one that uses latest .NET Core and fixed ML.NET to detect regressions in .NET Core
* one that uses latest version of ML.NET and LTS .NET Core to detect regressions in ML.NET

This PR tries to implement that by doing the following:

* extending the benchmarks `.yml` template with an optional `libraryVersion` parameter
* defining two jobs for ML.NET:
    * fixed ML.NET, latest .NET Core
    * latest .NET Core, fixed ML.NET
* extending the python script used by CI to set env var for controlling the version

I could not read latest ML.NET version from Maestro because ML.NET is not using Arcade, so I've used an asterisk to obtain latest preview version. 

```xml
<LibraryVersion Condition="'$(PERFLAB_LIBRARY_VERSION)' == 'rolling'">1.5.0-*</LibraryVersion>
```

This solution is simple, but requires a version bump after ML.NET increases major|minor version number (few times a year) due to https://github.com/NuGet/home/issues/912


adamsitnik:
@DrewScoggins could you please take a look?

eng/performance/benchmark_jobs.yml
```diff
@@ -92,9 +93,9 @@ jobs:
             ${{ each framework in parameters.frameworks }}:
               ${{ framework }}:
                 _Framework: ${{ framework }}
-                _Configs: CompilationMode=Tiered RunKind="${{ parameters.kind }}"
+                _Configs: CompilationMode=Tiered RunKind="${{ parameters.kind }}${{ parameters.libraryVersion }}"
```
adamsitnik: I don't know how the new uploader works so I've just extended the `RunKind` string with `fixed|rolling` This parameter is optional and null by default, so the "id" should change only for ML.NET runs

DrewScoggins: This seems like the right way to go about this.

kunalspathak:
@DrewScoggins - after our today's discussion, should we re-evaluate this?

Also - Is it possible to update the ML.NEt benchmarks to the latest https://github.com/dotnet/machinelearning/tree/main/test/Microsoft.ML.PerformanceTests?

danmoseley:
cc @michaelgsharp

kunalspathak:
@adamsitnik can chime in, but I want to make sure that we remove all the [project references ](https://github.com/dotnet/machinelearning/blob/52ddbcd0a57b440e60e43e6cb39b49fe55bdfe55/test/Microsoft.ML.PerformanceTests/Microsoft.ML.PerformanceTests.csproj#L20-L27), add back nuget references and then make sure that all the benchmarks run. If you can get to that, I can verify if they run on windows/arm64 as well.

adamsitnik:
Existing ML.NET benchmarks that we have in this [repo](https://github.com/dotnet/performance/blob/b7537def45f972d107b25e00b8200ab8837ef73a/src/benchmarks/real-world/Microsoft.ML.Benchmarks/Microsoft.ML.Benchmarks.csproj#L10) are using a hardcoded version of ML.NET in order to catch regressions in .NET itself.

The purpose of this PR was to add another CI leg that would be using latest ML.NET version to detect regressions in ML.NET itself.

@kunalspathak I assume you would be interested in updating the benchmarks, ML.NET version and using the benchmarks as part of our automated regression detection suite?

kunalspathak:
> @kunalspathak I assume you would be interested in updating the benchmarks, ML.NET version and using the benchmarks as part of our automated regression detection suite?

That's right. It is part of our ongoing goal of having real-world benchmarks to measure the .NET performance. I also am working on adding ImageSharp to this suite in https://github.com/dotnet/performance/pull/2149. So help me understand, with the current state, if we just copy the latest code from https://github.com/dotnet/machinelearning/tree/main/test/Microsoft.ML.PerformanceTests and used nuget references to the latest ML.NET, we should be good and this PR is not blocking it.

Of course, we certainly want to track ML.NET performance and in order to do that, we need this PR.

adamsitnik:
> we should be good and this PR is not blocking it, yes?

Yes, that is correct.

PR #991, opened 2019-10-31
Add more scenarios

billwert:


billwert:
this failure was due to the mismatch we where 'netcoreapp3.0' meant 'netcoreapp3.1', I think.

.gitignore
```diff
@@ -52,6 +52,12 @@ bld/
 # Visual Studio 2017 auto generated files
 Generated\ Files/
 
+# scenario intermediate directories
+src/scenarios/*/pub
+src/scenarios/*/app
+src/scenarios/*/bin
+src/scenarios/*/traces
```
ooooolivia: add obj folder?

src/scenarios/emptyvbconsoletemplate/test.py
```diff
@@ -5,8 +5,8 @@
 import os
 from shared.runner import TestTraits, Runner
 
-SCENARIONAME = '.NET Core 5.0 Console Template'
-EXENAME = '50consoletemplate'
+SCENARIONAME = 'Empty VB Console Template'
+EXENAME = 'emptycsconsoletemplate'
```
ooooolivia: did u mean ```emptyvbconsoletemplate```?

src/scenarios/init.ps1
```diff
@@ -1,2 +1,2 @@
-$scripts = Join-Path $PSScriptRoot '..\..\scripts\' -Resolve
+$scripts = Join-Path $PSScriptRoot '..\..\scripts' -Resolve
```
ooooolivia: just wondering is it more like a nit thing to remove the slash or otherwise it would break

billwert: just a nit thing. I don't think it breaks anything.

src/scenarios/emptyvbconsoletemplate/test.py
billwert: I did! good catch.

.gitignore
billwert: do we have anything that actually makes an obj folder in this place? I haven't seen one yet.

ooooolivia: oops  I meant bin and obj in the app folder and they are already covered, my mistake.

src/scenarios/shared/precommands.py
```diff
@@ -50,16 +52,27 @@ def __init__(self):
         args = parser.parse_args()
         self.configuration = args.configuration
```
ooooolivia: It breaks here when operation is RESTORE because it doesnt have configuration as arg

billwert: Good catch! As discussed I'm just going to remove the `restore` verb.

src/scenarios/shared/runner.py
```diff
@@ -104,5 +107,6 @@ def run(self):
                              iterationsetup=None,
                              setupargs=None,
                              workingdir=const.TMPDIR,
-                             environmentvariables=envlistbuild
+                             environmentvariables=envlistbuild,
+                             processwillexit=self.traits.processwillexit
```
ooooolivia: should add ```measurementdelay=None```

src/scenarios/staticconsoletemplate/src/Program.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace _50consoletemplate
+namespace staticconsoletemplate
```
ooooolivia: is staticconsoletemplate same as netcoreapp2.0 src ?  i dont see any diff between them except their names

ooooolivia: maybe we could just combine them?

src/scenarios/staticconsoletemplate/test.py
```diff
@@ -0,0 +1,18 @@
+'''
+C# Console app
+'''
+import sys
+import os
```
ooooolivia: these imports are not used. should we delete them?

src/scenarios/staticvbconsoletemplate/pre.py
```diff
@@ -0,0 +1,14 @@
+'''
+pre-command
+'''
+import os
+import sys
+from performance.logger import setup_loggers
+from shared import const
+from shared.precommands import PreCommands
+from test import EXENAME
```
ooooolivia: same, EXENAME is not used.

src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs
```diff
@@ -40,7 +40,7 @@ public IEnumerable<Counter> Parse(string mergeTraceFile, string processName, ILi
 
                 source.Kernel.ProcessStart += evt =>
                 {
-                    if (processName.Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase) && pids.Contains(evt.ProcessID) && evt.CommandLine == commandLine)
+                    if (processName.Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase) && pids.Contains(evt.ProcessID) && evt.CommandLine.Trim() == commandLine.Trim())
```
ooooolivia: Would it break here without trim?  If so, can we do trim in Startup.cs so we don't need to add this code to every parser?

billwert: The erroneous space is on the event payload, so it should be here. I'm trimming both out of paranoia. I'll clean this up when I refactor this code.

src/scenarios/staticconsoletemplate/src/Program.cs
billwert: sure looks like it. I'll leave it to you in the SDK test refactoring.

src/scenarios/staticconsoletemplate/test.py
billwert: done

src/scenarios/staticvbconsoletemplate/pre.py
billwert: made a change to use it.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -215,6 +215,10 @@ static void checkArg(string arg, string name)
                 }
                 TraceEventSession.Merge(files.ToArray(), traceFileName);
 
+                if (guiApp)
+                {
+                    appExe = Path.Join(workingDir, appExe);
```
ooooolivia: wondering why we need to join the paths here; seems evt.commandLine only takes whatever appExe is

billwert: in the GUI app case the event payload includes the full path to the exe. We happen to know (by convention) that the specified working dir is always the root to the app to run. We could change this to always pass in the full path also if this becomes problematic.

ooooolivia: "We happen to know (by convention) that the specified working dir is always the root to the app to run." not true for sdk tests but it's fine as long as guiApp=false for sdk tests...

billwert: fair point. though maybe we should rethink that. if SDK tests are so different we'll have trouble sharing templates etc..

PR #992, opened 2019-11-01
Adding supporting performance test for System.Linq.Enumerable.OrderBy enhancement

manofstick:
This adds some performance tests for OrderBy, which are then used to support the PR ["Special case strings in System.Linq.Enumerable.OrderBy"](https://github.com/dotnet/corefx/pull/42286)

### DIFF
--- a/src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
+++ b/src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
@@ -1,0 +1,151 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Engines;
+using MicroBenchmarks;
+
+namespace System.Linq.Tests
+{
+    class FavourColourComparer
+        : IComparer<System.Drawing.Color>
+    {
+        public static IComparer<System.Drawing.Color> Instance => new FavourColourComparer();
+        static int TurnIntoValue(System.Drawing.Color x) => (x.R + x.G + x.B) * x.A;
+        public int Compare(System.Drawing.Color lhs, System.Drawing.Color rhs) => TurnIntoValue(lhs).CompareTo(TurnIntoValue(rhs));
+    }
+
+    // Bit of a mix of some value type fields and string
+    public class Person
+    {
+        public string FirstName { get; set; }
+        public string MiddleName { get; set; }
+        public string LastName { get; set; }
+
+        public string FullName => $"{LastName}, {FirstName} {MiddleName[0]}.";
+
+        public DateTime DOB { get; set; }
+        public System.Drawing.Color FavouriteColour { get; set; }
+        public int IQ { get; set; }
+    }
+
+    class PersonData
+    {
+        public static Person[] Generate(int count)
+        {
+            var r = new Random(42); // Same seed for consistent dataset
+
+            var dataset = new Person[count];
+
+            for(var i=0; i < count; ++i)
+            {
+                dataset[i] = new Person
+                {
+                    FirstName = FirstNames[r.Next(0, FirstNames.Length)],
+                    MiddleName = FirstNames[r.Next(0, FirstNames.Length)],
+                    LastName = Surnames[r.Next(0, Surnames.Length)],
+
+                    DOB = new DateTime(1919, 1, 1).AddDays(r.Next(0, 100 * 365)),
+                    FavouriteColour = System.Drawing.Color.FromArgb(r.Next(0, 256), r.Next(0, 256), r.Next(0, 256)),
+                    IQ = 100 + r.Next(-30, 30),
+                };
+            }
+
+            return dataset;
+        }
+
+        private static readonly string[] FirstNames = new[]
+        {
+"James", "John", "Robert", "Michael", "William", "David", "Richard", "Charles", "Joseph", "Thomas", "Christopher", "Daniel", "Paul", "Mark", "Donald", "George", "Kenneth", "Steven", "Edward", "Brian",
+"Mary", "Patricia", "Linda", "Barbara", "Elizabeth", "Jennifer", "Maria", "Susan", "Margaret", "Dorothy", "Lisa", "Nancy", "Karen", "Betty", "Helen", "Sandra", "Donna", "Carol", "Ruth", "Sharon", "Michelle",
+"Ronald", "Anthony", "Kevin", "Jason", "Matthew", "Gary", "Timothy", "Jose", "Larry", "Jeffrey", "Frank", "Scott", "Eric", "Stephen", "Andrew", "Raymond", "Gregory", "Joshua", "Jerry", "Dennis", "Walter",
+"Laura", "Sarah", "Kimberly", "Deborah", "Jessica", "Shirley", "Cynthia", "Angela", "Melissa", "Brenda", "Amy", "Anna", "Rebecca", "Virginia", "Kathleen", "Pamela", "Martha", "Debra", "Amanda", "Stephanie",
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Engines;
+using MicroBenchmarks;
+
+namespace System.Linq.Tests
+{
+    class FavourColourComparer
+        : IComparer<System.Drawing.Color>
+    {
+        public static IComparer<System.Drawing.Color> Instance => new FavourColourComparer();
```
adamsitnik: this should be a readonly field to make sure that the allocation of comparer is not included in the benchmark results

```suggestion
        public static readonly IComparer<System.Drawing.Color> Instance = new FavourColourComparer();
```

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+        public DateTime DOB { get; set; }
+        public System.Drawing.Color FavouriteColour { get; set; }
+        public int IQ { get; set; }
+    }
+
+    class PersonData
+    {
+        public static Person[] Generate(int count)
+        {
+            var r = new Random(42); // Same seed for consistent dataset
```
adamsitnik: thanks for using a const seed! :+1: 

https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#always-the-same-input-data

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+
+    [BenchmarkCategory(Categories.CoreFX, Categories.LINQ)]
+    public class Perf_OrderBy
+    {
+        private readonly Consumer _consumer = new Consumer();
+
+        [Params(0, 1, 10, 100, 1000)]
+        public int NumberOfPeople;
+
+        private Person[] People;
```
adamsitnik: nit: private fields should start with `_`

```suggestion
        private Person[] _people;
```

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+        [GlobalSetup]
+        public void GlobalSetup()
+        {
+            People = PersonData.Generate(NumberOfPeople);
+        }
+
+        [Benchmark]
+        public void OrderByFirstName() => People.OrderBy(p => p.FirstName).Consume(_consumer);
+        [Benchmark]
+        public void OrderByFullName() => People.OrderBy(p => p.FullName).Consume(_consumer);
```
adamsitnik: this benchmark also tests ordering by a string type property, but the difference with `OrderByFirstName` is that when we call `FullName` it always allocates memory. I don't think that we need it. (`OrderByFirstName` should be enough). Please remove it

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+        [Benchmark]
+        public void OrderByDOB() => People.OrderBy(p => p.DOB).Consume(_consumer);
```
adamsitnik: Ordering by a value-type property is NOT covered by our our existing benchmarks [1](https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/corefx/System.Collections/Sort.cs), [2](https://github.com/dotnet/performance/blob/5a0d9d7a858a1fecc4ef5f5383c7be8dc1b9a515/src/benchmarks/micro/corefx/System.Linq/Perf.Enumerable.cs#L169) so adding this benchmark adds value, however I think that we need a better name that tells us that we are ordering by a value type property

```suggestion
        public void OrderByValueType() => People.OrderBy(p => p.DOB).Consume(_consumer);
```

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+        [Benchmark]
+        public void OrderByFavouriteColour() => People.OrderBy(p => p.FavouriteColour, FavourColourComparer.Instance).Consume(_consumer);
```
adamsitnik: this is something we were lacking so far :+1:  but we need a better name

```suggestion
        public void OrderByCustomComparer() => People.OrderBy(p => p.FavouriteColour, FavourColourComparer.Instance).Consume(_consumer);
```

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+        [Benchmark]
+        public void OrderByIQ() => People.OrderBy(p => p.IQ).Consume(_consumer);
```
adamsitnik: I think that this test case (ordering by an integer field) is covered by https://github.com/dotnet/performance/blob/5a0d9d7a858a1fecc4ef5f5383c7be8dc1b9a515/src/benchmarks/micro/corefx/System.Linq/Perf.Enumerable.cs#L169 Please remove it

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+    }
+}
```
adamsitnik: nit: could you please add new line between every benchmark?

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        [Params(0, 1, 10, 100, 1000)]
+        public int NumberOfPeople;
+
+        private Person[] People;
+
+        [GlobalSetup]
+        public void GlobalSetup()
+        {
+            People = PersonData.Generate(NumberOfPeople);
```
adamsitnik: nit: this could could use `=>` and save some space

```cs
[GlobalSetup]
public void Setup() => People = PersonData.Generate(NumberOfPeople);
```

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
```diff
@@ -0,0 +1,151 @@
...
+"Larsen", "Hoover", "Beasley", "Glenn", "Petersen", "Whitehead", "Meyers", "Keith", "Garrison", "Vincent", "Shields", "Horn", "Savage", "Olsen", "Schroeder", "Hartman", "Woodard", "Mueller", "Kemp", "Deleon",
+        };
+    }
+
+    [BenchmarkCategory(Categories.CoreFX, Categories.LINQ)]
+    public class Perf_OrderBy
+    {
+        private readonly Consumer _consumer = new Consumer();
+
+        [Params(0, 1, 10, 100, 1000)]
```
adamsitnik: We don't need 5 test cases. I think that it's great to present all of them in your CoreFX PR, but in this repo we track the performance of .NET Core over time and due to the amount of benchmarks we have (thousands), we want to have 1 benchmark test case per execution path:

https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#Test-Cases

```suggestion
        [Params(Utils.DefaultCollectionSize)]
```

src/benchmarks/micro/corefx/System.Linq/Perf.OrderBy.cs
manofstick: The reason for including it was just to ensure that key fields were only read once from the original object - i.e. existing sorting collects the set of keys once into a seperate aerray and then performs sorts on those (i.e. so the concatenation I included currently only is execute once). But I'm happy to remove it if you wish.

adamsitnik: I think that it's very unlikely that we would remove this optimization from CoreFX. Please remove it.

Thank you for the explanation!

adamsitnik:
@manofstick thank you!

PR #997, opened 2019-11-02
Fix Bilinear Interpolation benchmark

CarolEidt:
- Fix two bugs in HW intrinsic implementation
- Add copyright


CarolEidt:
@adamsitnik @tannergooding PTAL
I found these issues while creating a test for the coreclr repo that includes checking of the result.

PR #1004, opened 2019-11-05
update the profiling docs, mention that VTune works great on Linux

adamsitnik:


adamsitnik:
Since this is a small doc update and I want to share the link with Team, I am merging the PR right now

PR #1007, opened 2019-11-05
split sdk workitem

ooooolivia:
- split sdk workitem into clean build workitem and build (no change) workitem to avoid sharing the same artifacts folder
- remove spaces in artifacts file name so the artifacts could be uploaded


src/scenarios/shared/startup.py
```diff
@@ -78,7 +78,8 @@ def runtests(self, apptorun: str, **kwargs):
         RunCommand(startup_args, verbose=True).run()
 
 
-        if runninginlab() and uploadtokenpresent():
-            import upload
+        if runninginlab():
```
billwert: @DrewScoggins in particular this seem okay to you? I think this is closer to the behavior I'd like.

src/scenarios/shared/runner.py
```diff
@@ -46,20 +46,29 @@ class Runner:
...
         '''
         Parses input args to the script
         '''
         parser = ArgumentParser()
-        parser.add_argument('testtype', choices=testtypes, type=str.lower)
+        subparsers = parser.add_subparsers(title='subcommands for sdk tests', required=True, dest='testtype')
+        startupparser = subparsers.add_parser(const.STARTUP)
+        sdkparser = subparsers.add_parser(const.SDK)
+        sdkparser.add_argument('sdktype', choices=[const.CLEAN_BUILD, const.BUILD_NO_CHANGE], type=str.lower)
+
```
billwert: This is great! I was going to need something like this for the crossgen tests anyway.

src/scenarios/shared/startup.py
DrewScoggins: Yeah, this seems better. We always want to save artifacts, but only do the upload if we are in an internal run.

src/scenarios/shared/runner.py
```diff
@@ -85,7 +94,7 @@ def run(self):
                              timeout=self.traits.timeout,
                              warmup='true',
                              iterations=self.traits.iterations,
-                             scenariotypename='%s (%s)' % (const.SCENARIO_NAMES[const.SDK], 'Clean Build'),
+                             scenariotypename='%s_%s' % (const.SCENARIO_NAMES[const.SDK], const.CLEAN_BUILD),
```
billwert: this is going to be 'clean_build', not 'Clean Build'. Which do you intend?

ooooolivia: clean_build bc we want to remove the space in filenames...

billwert: I think we should remove spaces more generically (in our upload function) and leave the scenario names as we want them to be. This string also is what is used on the webpage for display - I'd rather those are human readable. What do you think?

ooooolivia:
oh no the merge has overrode scenarios.proj

ooooolivia:
The merge was messed up. Will update it.

src/scenarios/shared/runner.py
ooooolivia: Sounds good to me! Could make change in the following PR.

billwert: That's fine but FYI - test names are their identity. So changing this later will break the trend. It's fine to merge this so that we can make sure everything works as we want it to. We'll basically get new trends when we make the next change. This is one of the few real invariants in our system so we need to be careful around it.

ooooolivia: Actually removing the spaces won't thoroughly fix the problem. The artifacts' url is not valid bc it contains '\' which is not accepted by helix. Still keeping in touch with the engineering services folks...

PR #1011, opened 2019-11-06
[WIP] net461 -> net48

adamsitnik:
I want to check if our CI images do have 4.8 SDK installed.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -81,19 +81,19 @@ jobs:
         - netcoreapp2.2
         - netcoreapp2.1
 
-  # Windows x64 net461 micro benchmarks
+  # Windows x64 net48 micro benchmarks
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: windows
       osVersion: RS4
-      kind: micro_net461
+      kind: micro
       architecture: x64
       pool: Hosted VS2017
       queue: Windows.10.Amd64.ClientRS4.DevEx.15.8.Open
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
       runCategories: 'coreclr corefx'
       frameworks: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
-        - net461
+        - net48
 
   # Windows x86 micro benchmarks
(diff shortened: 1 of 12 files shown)


adamsitnik:
```log
 The reference assemblies for .NETFramework,Version=v4.8 were not found
```

azure-pipelines.yml
```diff
@@ -81,19 +81,19 @@ jobs:
...
+  # Windows x64 net48 micro benchmarks
   - template: /eng/performance/benchmark_jobs.yml
     parameters:
       osName: windows
       osVersion: RS4
-      kind: micro_net461
+      kind: micro
       architecture: x64
       pool: Hosted VS2017
       queue: Windows.10.Amd64.ClientRS4.DevEx.15.8.Open
```
adamsitnik: @ViktorHofer @safern do you know which queue|pool has .NET 4.8 SDK installed?

billwert:
I wonder if this about 4.8 or the reference assemblies? I missed the comments on dotnet/benchmarkdotnet#1197. I don't have time to try that again this week but perhaps that's also a useful path.

azure-pipelines.yml
ViktorHofer: right, I assume you are talking about the ref assembly pack? That is available here: https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies.net48/1.0.0-preview.2

adamsitnik: I want to build and run as 4.8. Is using ref assembly pack going to be enough to achieve that?

ViktorHofer: The ref assembly pack is for compiling on both windows and Unix. If you only care about windows here, you probably don't need that pack as windows always has the latest netframework installed which is updated via windows update. Targeting net48 should just work...

safern: In corefx for package testing, since we can't guarantee all our build/test machines have the targeting packs we need installed, we get it via NuGet, to ensure it always works. https://github.com/dotnet/corefx/blob/master/src/pkg/test/frameworkSettings/net/settings.targets#L11
-- even though we only run in Windows.

PR #1013, opened 2019-11-06
Add all sdk scenarios

ooooolivia:
Too many placeholder files in the large projects are added. Please see main changes in ```shared``` folder and any ```pre.py```, ```test.py``` and ```post.py``` in project folders.


src/scenarios/mvcapptemplate/pre.py
```diff
@@ -10,7 +10,7 @@
 setup_loggers(True)
 precommands = PreCommands()
 precommands.new(template='mvc',
-                output_dir=const.APPDIR,
+                output_dir=const.SRCDIR,
```
billwert: Why? I think the intent is that apps always run from "app" and if there is existing source code it should be checked into "src". If you use the `existing` path from `precommands.py` it copies the `src` to `app` so that build and other verbs work out. Is there a flaw with this scheme?

src/scenarios/mvcapptemplate/pre.py
```diff
@@ -0,0 +1,17 @@
+'''
+pre-command
+'''
+import sys
+from performance.logger import setup_loggers
+from shared.precommands import PreCommands
+from shared import const
+from test import EXE_NAME
```
billwert: I think I called this `EXENAME` in the ones I added. Can you change either yours or mine to be consistent?

src/scenarios/mvcapptemplate/test.py
```diff
@@ -0,0 +1,20 @@
+import os
+from shared.runner import TestTraits, Runner
+from shared import const
+
+SCENARIO_NAME = 'ASP.NET Core MVC App Template'
```
billwert: Same comment - consistency

src/scenarios/mvcapptemplate/pre.py
ooooolivia: Thought that's the naming convention in python... I'll change mine to EXENAME

src/scenarios/mvcapptemplate/pre.py
ooooolivia: I think this is doing what you said if what ```dotnet new``` generates could be considered as "source code" for SDK scenarios? SDK iteration setup always copies things from ```src``` to ```app``` and do ```dotnet build``` in ```app```. So we need the template goes to ```src``` instead of ```app```.

src/scenarios/mvcapptemplate/pre.py
billwert: My take on PEP8 is that it's optional if it improves readability. (I do use that sometimes.)

src/scenarios/mvcapptemplate/pre.py
billwert: Ah - I see the confusion. My intent is that _static_ source code, that's checked in, goes under `src`. Anything dynamic can go in `app`. The automatic copy from whatever you say the project dir is to `app` happens only if you call [`precommands.existing`.](https://github.com/dotnet/performance/blob/master/src/scenarios/shared/precommands.py#L97)

ooooolivia: yes wherever the src code is from we just can't have ```dotnet new``` and ```dotnet build``` in the same folder.

ooooolivia: Also it seems the backup command in ```precommands.py``` is not necessary any more since it's basically ```precommands.existing```...

ooooolivia:
Split the large commit to smaller commits. ```52850a732b40f36ec86a4645e11f46a788f2b681``` is the commit that contains only static source code.

src/scenarios/mvcapptemplate/pre.py
billwert: one other point in "generate to `const.APPDIR`"'s favor: we specifically **don't** exclude `const.SRCDIR` from the `.gitignore`, so if there are mistaken changes while someone is working they're discovered.

ooooolivia: Good to know! This won't affect our case though because ```src``` is not in ```.gitignore``` anyway. The extra overhead is manually not checking in the ```src``` folder on commit for dynamic template tests.

DrewScoggins:
https://github.com/dotnet/performance/compare/e67f28d882fc624e7316a1d47a4cc41fa6713079..1ec00db44400d1ebb370755e9ec077e65a54c7b1

This link is the comparison minus the 20k+ static files

ooooolivia:
> https://github.com/dotnet/performance/compare/e67f28d882fc624e7316a1d47a4cc41fa6713079..1ec00db44400d1ebb370755e9ec077e65a54c7b1
> 
> This link is the comparison minus the 20k+ static files

Guess you can also select the range of commits in Files Changed tab.

DrewScoggins:
I didn't know this

src/scenarios/weblarge2.0/test.py
```diff
@@ -0,0 +1,21 @@
...
+
+SCENARIONAME = 'Web Large 2.0'
+EXENAME = 'weblarge20'
+
+def main():
+    traits = TestTraits(scenarioname=SCENARIONAME,
+                        exename=EXENAME,
+                        guiapp='false',
+                        workingdir='mvc',
+                        timeout='1000',  # increase timeout for the large project
```
billwert: 1000 seconds is a weird value here. Typically we do this sort of thing in multiples of 60 seconds. Maybe 900 or 1200 here? (Bonus points: something like `f'{60*15}'`, or `f'{MINUTE*15}'` and define `MINUTE` somewhere.)

src/scenarios/mvcapptemplate/pre.py
billwert: I still think this should be `APPDIR`. (As with all the other places you're calling `.new`. Let's chat Monday.

src/scenarios/weblarge2.0/test.py
ooooolivia: Thanks for the suggestion! Added the constant to ```const.py```

PR #1015, opened 2019-11-06
Enable 3.0 runs for micros for baseline purposes

billwert:


billwert:
@DrewScoggins signed off over my shoulder.

PR #1016, opened 2019-11-06
benchmark for Thread.GetCurrentProcessorId

VSadov:
`Thread.GetCurrentProcessorId` has nontrivial implementation. 
There are tradeoffs between precision and performance, which may be worth revisiting and tuning. It would be nice to have perf history for this API.

### DIFF
--- a/src/benchmarks/micro/corefx/System.Threading/Perf.Thread.cs
+++ b/src/benchmarks/micro/corefx/System.Threading/Perf.Thread.cs
@@ -12,5 +12,12 @@ namespace System.Threading.Tests
     {
         [Benchmark]
         public Thread CurrentThread() => Thread.CurrentThread;
+
+#if !NETFRAMEWORK
+        public int ThreadGetCurrentProcessorId()
+        {
+            return System.Threading.Thread.GetCurrentProcessorId();
+        }
+#endif
     }
 }


VSadov:
Please let me know if there are already tests for this or if they should go to a different location.

adamsitnik:
Hi @VSadov 

We have [TreatWarningsAsErrors=true](https://github.com/dotnet/performance/blob/0c82eb806ded059e71b6ec801daeb558c716137d/src/Directory.Build.props#L15-L16) and the CI fails with:

```log
[2019/11/07 00:08:33][INFO] coreclr/perflab/LowLevelPerf.cs(238,31): error CS0649: Field 'LowLevelPerf.t_object' is never assigned to, and will always have its default value null [/home/helixbot/work/e3b27947-eb87-4424-ab4b-9b54495df9de/Work/712ba3d1-4390-43e6-8f02-2856d936f004/Exec/src/benchmarks/micro/MicroBenchmarks.csproj]
[2019/11/07 00:08:33][INFO] coreclr/perflab/LowLevelPerf.cs(229,28): error CS0649: Field 'LowLevelPerf.t_value' is never assigned to, and will always have its default value 0 [/home/helixbot/work/e3b27947-eb87-4424-ab4b-9b54495df9de/Work/712ba3d1-4390-43e6-8f02-2856d936f004/Exec/src/benchmarks/micro/MicroBenchmarks.csproj
```

> Please let me know if there are already tests for this or if they should go to a different location.

I think that https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/corefx/System.Threading/Perf.Thread.cs would be a better place for it

VSadov:
Fyi: @stephentoub, @jkotas

src/benchmarks/micro/coreclr/perflab/LowLevelPerf.cs
```diff
@@ -224,7 +224,25 @@ public void StaticIntPlus()
...
+        }
+
+        [ThreadStatic]
+        private static object t_object = "q";
+
+        [Benchmark]
+        public object ReadThreadStaticObject()
+        {
+            return t_object;
+        }
```
stephentoub: We already have:
https://github.com/dotnet/performance/blob/b142ff9c57d3807677bf228635460e40be676591/src/benchmarks/micro/corefx/System.Threading/Perf.ThreadStatic.cs

VSadov:
@stephentoub  oh, good!! 
 I’ll do only ProcessorId then.

src/benchmarks/micro/corefx/System.Threading/Perf.Thread.cs
```diff
@@ -12,5 +12,13 @@ public class Perf_Thread
     {
         [Benchmark]
         public Thread CurrentThread() => Thread.CurrentThread;
+
+#if !NETFRAMEWORK
+        [BenchmarkCategory(Categories.CoreFX)]
```
jkotas: The same attribute is on the type already. Does it need to be repeated here again?

VSadov: Not after i moved it here from  LowLevel\Perf.cs where i put it originally.

PR #1026, opened 2019-11-10
Add benchmark data for testing non-vectorised paths

Gnbrkm41:
Related: https://github.com/dotnet/coreclr/pull/27798

The first commit is related to the PR mentioned above; I think it could be worth adding this even if the PR above doesn't make it in, in case there could be a similar change.

The second commit is per https://github.com/dotnet/corefx/pull/39173#issuecomment-516728961.

### DIFF
--- a/src/benchmarks/micro/corefx/System.Collections/Perf.BitArray.cs
+++ b/src/benchmarks/micro/corefx/System.Collections/Perf.BitArray.cs
@@ -13,7 +13,8 @@ namespace System.Collections.Tests
     {
         private const int DefaultShiftCount = 17;
         private const bool BooleanValue = true;
-
+        // 4 - Small size to test non-vectorised paths
+        // DefaultCollectionSize - Big enough size to go through the vectorised paths
         [Params(4, Utils.DefaultCollectionSize)]
         public int Size { get; set; }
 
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/corefx/System.Collections/Perf.BitArray.cs
```diff
@@ -14,7 +14,7 @@ public class Perf_BitArray
         private const int DefaultShiftCount = 17;
         private const bool BooleanValue = true;
 
-        [Params(Utils.DefaultCollectionSize)]
+        [Params(4, Utils.DefaultCollectionSize)]
```
adamsitnik: could you please add a comment explaining why we have two test cases? Something similar to:

https://github.com/dotnet/performance/blob/cc73a0126ae9056382fe23a9dee183b18b1b67de/src/benchmarks/micro/corefx/System.Globalization/StringHash.cs#L33-L36

Gnbrkm41:
I'll need to add some more cases to follow up with changed implementations.

adamsitnik:
> I'll need to add some more cases to follow up with changed implementations.

Please tag me when the PR is ready to merge (I would prefer to merge 1 final PR)

Gnbrkm41:
Is the performance repo being merged into runtime or somewhere else as well? Thought this was going to stay.

adamsitnik:
> Is the performance repo being merged into runtime or somewhere else as well?

Gnbrkm41:
Actually... I think this is good as-is. I was going to add a slightly longer string, thinking it would matter since I changed the implementation to process more at once; but it is covered by the current data as well.

PR #1033, opened 2019-11-13
Refactor Startup.cs

ooooolivia:
- Wrap up setup, test and cleanup into one ```RunIteration``` method so it could be called in warmup, test and profile iterations without too much repetitive code
- Better logic for running setup and cleanup (details in comments)
- Exit early if 1) warmup iteration failed so no test iterations 2) test iterations failed so no profile iteration

### DIFF
--- a/src/tools/ScenarioMeasurement/Startup/Startup.cs
+++ b/src/tools/ScenarioMeasurement/Startup/Startup.cs
@@ -17,6 +17,7 @@ namespace ScenarioMeasurement
     }
     class Startup
     {
+        static int FAILED_PID = -1;
         /// <summary>
         /// 
         /// </summary>
@@ -42,6 +43,7 @@ namespace ScenarioMeasurement
         /// <param name="traceDirectory">Directory to put files in (defaults to current directory)</param>
         /// <param name="environmentVariables">Environment variables set for test processes (example: var1=value1;var2=value2)</param>
         /// <returns></returns>
+
         static int Main(string appExe,
                         MetricType metricType,
                         string scenarioName,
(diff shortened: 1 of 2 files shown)


src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -165,42 +169,13 @@ static void checkArg(string arg, string name)
                     for (int i = 0; i < iterations; i++)
                     {
                         logger.Log($"=============== Iteration {i} ================ ");
-                        // set up iteration
-                        if (setupProcHelper != null)
+                        int pid = RunIteration(setupProcHelper, procHelper, cleanupProcHelper, logger);
+                        if (pid < 0)
```
billwert: Why `< 0` instead of checking against `FAILED_PID`?

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -319,6 +270,48 @@ private static ProcessHelper CreateProcHelper(string command, string args, Logge
...
+                {
+                    // cleanup failed
+                    return FAILED_PID;
+                }
+            }
+
+            return pid;
+        }
+
+        private static int RunProcess(ProcessHelper helper, Logger logger)
```
billwert: I think a more .NET way of doing this would be either to return `(bool, int)` or to return `bool` and have the PID as an out param. Sentinel values like `FAILED_PID` aren't common.

ooooolivia: Good to know! I'll make it return a tuple.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -270,47 +269,46 @@ private static ProcessHelper CreateProcHelper(string command, string args, Logge
             return procHelper;
         }
 
-        private static int RunIteration(ProcessHelper setupHelper, ProcessHelper testHelper, ProcessHelper cleanupHelper, Logger logger)
+        private static (bool success, int pid) RunIteration(ProcessHelper setupHelper, ProcessHelper testHelper, ProcessHelper cleanupHelper, Logger logger)
```
billwert: nit: capitalize return value names

PR #1039, opened 2019-11-15
Merge netcoreapp2.0(sdk) asset and staticconsoletemplate(startup)

ooooolivia:
- The test assets were originally designed for share among scenarios. Currently ```staticconsoletemplate``` and ```netcoreapp2.0``` basically share the same static code (unexpectedly this is the only asset we are able to merge for now). However, we need to consider which target framework should be used. Ideally we want to create a matrix of scenarios that has different sdks vs. different runtimes.  

- refactor post commands


src/scenarios/staticconsoletemplate/test.py
```diff
@@ -11,6 +11,7 @@
                         exename=EXENAME, 
                         startupmetric='TimeToMain',
                         startup=True,
+                        sdk=True,
```
DrewScoggins: Will this now also be true for when we run the staticconsoleempty tests not eh SDK tests?

ooooolivia: resolved offline :)

ooooolivia:
Just pushed the changes of refactoring post commands too

src/scenarios/emptyvbconsoletemplate/post.py
```diff
@@ -2,10 +2,6 @@
...
-from performance.common import remove_directory
-from shared.const import BINDIR, PUBDIR, APPDIR, TRACEDIR
- 
-remove_directory(BINDIR)
-remove_directory(PUBDIR)
-remove_directory(APPDIR)
-remove_directory(TRACEDIR)
+from shared.postcommands import clean_directories
+
+clean_directories()
```
DrewScoggins: Aren't we changing the semantics here? Before we did not delete the TMPDIR, but now we are. My assumption is that these tests never used TMPDIR so this is unlikely to be a problem. remove_directory works fine if they directory does not exists, right?

ooooolivia: yes the nice part of remove_directory is it silently ignores the directories that don't exist

src/scenarios/shared/postcommands.py
```diff
@@ -0,0 +1,7 @@
+from shared.const import APPDIR, TMPDIR, TRACEDIR, PUBDIR, BINDIR
+from performance.common import remove_directory
+
+def clean_directories():
```
billwert: Thought: Is there a scenario where we do not want this to run in someone's `post.py`? If we always want it to run, we could leave this as the top level content of the script instead of inside a function, and then `post.py` is just `import shared.postcommands`.

ooooolivia: some very tentative thoughts...  1) keeping this means it could be expanded to add other functions like terminating servers, disconnecting the database...? (purely guess)
2) if this could be true, then maybe it's possible to just selectively do certain things in ```post.py``` and not deleting the folders 3) what's the advantage of having a light ```import shared.postcommands``` except it's only one line

PR #1042, opened 2019-11-18
Support .NET desktop in GCPerfSim

andy-ms:
Modify GCPerfSim to support building in desktop .NET.
It can be used with the infra using a coreclr of `self_contained: true`.

Another miscellaneous change is to make `cpu_group_number` truly optional;
before `NumaNodeInfo` could be written correctly on linux but not parsed back.

Other formatting changes are due to running the linter.


billwert:
@andy-ms can you fix the conflict?

andy-ms:
@billwert done

src/benchmarks/gc/src/exec/GCPerfSim/tests_to_be_adapted/cwt/Program.cs
```diff
@@ -75,9 +75,10 @@ static void Main2(string[] args)
                     
                     SomeKey sk = new SomeKey(i.ToString(), objSize);
                     SomeSet ss = new SomeSet(i.ToString(), objSize);
-                    s_cwt.AddOrUpdate(sk, ss);
+                    // s_cwt.AddOrUpdate(sk, ss); // Not available in .net 4.7.2
```
sergiy-k: Is it OK to not add the values here at all?

andy-ms: This is dead code right now -- see the exception below. These tests need to be adapted so the problem could be fixed then -- possibly this could only be available in a coreclr build.

PR #1056, opened 2019-11-22
Minor fixes to GC infra

andy-ms:
* Removed most of `generate_tests.py`.
  This contained (hidden) commands to generate certain tests,
  that we haven't needed to re-generate in a long time.
  They could probably be written by hand without much trouble.
  The exception is `generate-coreclr-unit-tests`,
  which finds all the GC unit tests in coreclr and create a benchfile for them.

  - We no longer need `run_single_test_temporary` as that was only used here.

* Coreclr recently moved from its own repository to the `runtime` repository.
  Updated code in `get_built.py` to use the new repository and handle the new binary output paths.

* Numpy package isn't used anymore so removing. Same for 'pywin32'.

### DIFF
--- a/src/benchmarks/gc/README.md
+++ b/src/benchmarks/gc/README.md
@@ -92,7 +92,8 @@ Navigate to `src/analysis/managed-lib` and run `dotnet publish`.
 
 ### Windows-Only Building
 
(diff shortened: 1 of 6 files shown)


andy-ms:
@billwert I don't have write access, could you merge?

PR #1067, opened 2019-11-28
GC infra: Improve documentation

andy-ms:
* `Microsoft.Build.Utilities` error message is often buried, so mention what comes first.
* Mention bug involving running shebang scripts on Windows.
* Fix typo: AMD64 -> ARM.
* Clarify how to use join info.
* Remove now-empty file `jupyter_notebook_utils.py`.


src/benchmarks/gc/README.md
```diff
@@ -29,65 +31,24 @@ On other systems it’s better to use your system’s package manager.
...
-* WARN: The instructions there tell you to run `pip install --no-index --find-links=.\dist\ pythonnet`.
-  This may "succeed" saying `Requirement already satisfied: pythonnet in /path/to/pythonnet`.
-  INSTEAD, go to the *parent* directory and use `sudo python3.7 -m pip install --no-index --find-links=./pythonnet/dist/` which circumvents this bug.
-* Run `import clr` in the python interpreter to verify that installation worked.
+### Install pythonnet
 
+Pythonnet is only needed to analyze test results, not to run tests.
+If you just want to run tests on this machine, you can skip installing pythonnet,
+copy bench output to a different machine, and do analysis there.
+When running tests, you must provide the `--no-check-runs` option to avoid using pythonnet.
```
Maoni0: if Pythonnet is not needed anyway, why does one need to specify anything to avoid using it? sounds confusing.

src/benchmarks/gc/README.md
```diff
@@ -179,6 +140,21 @@ Each trace file can be opened in PerfView if you need to.
...
 
+## Test status files
+
+Each trace has (at least) two files associated with it, `x.etl` (or `x.nettrace`) and `x.yaml`.
+The `.yaml` file is called a test status file. It provides, among other things, the process ID to focus on.
+A minimal test status file looks like:
+
+    # this file: `x.yaml`
+    success: true
+    trace_file_name: x.etl  # Should generally match the name of this file
```
Maoni0: does trace_file_name take a full path of x.etl or just x.etl? if it's the latter does it mean the .yaml always needs to be the in same dir as x.etl?

billwert:
@andy-ms let me know when you're ready with this.

src/benchmarks/gc/README.md
andy-ms: There was code `_check_test_run` for verifying that a test run had the number of heaps expected; this hasn't caught any bugs in a very long itmes so I've simply removed it and you no longer have to pass `--no-check-runs`.

src/benchmarks/gc/README.md
andy-ms: Added clarifying comment.

andy-ms:
@billwert Should be good to go.

PR #1072, opened 2019-12-02
Add a few regex perf tests

stephentoub:


billwert:
the CI failure is weird - going to spin it and see what happens.

/azp run

billwert:
How should I think of these in contrast to https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/corefx/System.Text.RegularExpressions/Perf.Regex.cs? Should we retire the older test in favor of more granular tests like this?

stephentoub:
> How should I think of these

Good question. I'm not entirely sure what the value of those is.  They're explicitly turning off the regex cache and evaluating each regex once, which means it's really a test for the cost of constructing and parsing regular expressions, rather than matching.  I don't know if that was the original intent, but if it was, that's reasonable, albeit poorly named tests.  In contrast, these are about testing the actual matching speed.

I don't believe these few I'm adding are sufficient to accurately cover everything about Regex matching throughput, but it's a start.

adamsitnik:
I agree that the old tests don't provide a lot of value. They look more like unit tests that were ported to a benchmarking tool. I am going to remove them in a separate PR.

adamsitnik:
https://github.com/dotnet/performance/pull/1077

PR #1077, opened 2019-12-04
Remove old regex benchmarks

adamsitnik:
`Perf.RegexRedux.cs` was duplicating [RegexRedux_5](https://github.com/dotnet/performance/blob/3941f11e000edd7a404599fc2e401129c25bc4ad/src/benchmarks/micro/coreclr/BenchmarksGame/regex-redux-5.cs)

`Perf_Regex` and `Perf_Regex_Cache` were more like unit tests, not providing too much value


src/benchmarks/micro/corefx/System.Text.RegularExpressions/Perf.Regex.Cache.cs
```diff
@@ -1,118 +0,0 @@
...
-
-using System.Collections.Generic;
-using System.Threading.Tasks;
-using BenchmarkDotNet.Attributes;
-using MicroBenchmarks;
-
-namespace System.Text.RegularExpressions.Tests
-{
-    [BenchmarkCategory(Categories.CoreFX)]
-    public class Perf_Regex_Cache
```
stephentoub: I think we should keep the cache ones, or at least someone that fits that purpose.

src/benchmarks/micro/corefx/System.Text.RegularExpressions/Perf.RegexRedux.cs
```diff
@@ -1,66 +0,0 @@
...
-
-using System.IO;
-using System.Threading.Tasks;
-using BenchmarkDotNet.Attributes;
-using MicroBenchmarks;
-
-namespace System.Text.RegularExpressions.Tests
-{
-    [BenchmarkCategory(Categories.CoreFX)]
-    public class RegexRedux
```
stephentoub: Both this and RegexRedux_5 are creating new compiled Regex instances on every invocation; that adds non-trivial overhead:
https://github.com/dotnet/runtime/pull/109#issuecomment-555670867

adamsitnik: great catch! I've extended the regex-redux-5 benchmark to run for both RegexOptions.Compiled and RegexOptions.None

src/benchmarks/micro/corefx/System.Text.RegularExpressions/Perf.Regex.Cache.cs
adamsitnik: I reverted the change

src/benchmarks/micro/coreclr/BenchmarksGame/regex-redux-5.cs
```diff
@@ -26,78 +25,57 @@ namespace BenchmarksGame
     [BenchmarkCategory(Categories.CoreCLR, Categories.BenchmarksGame)]
     public class RegexRedux_5
```
stephentoub: Not specifically related to this change, but how are we thinking about these benchmark categories in light of dotnet/runtime?  It was already a little strange having this benchmark categorized as "coreclr" when it's more about Regex, and now that we have dotnet/runtime instead of dotnet/coreclr + dotnet/corefx, the distinction seems even more odd.

adamsitnik: I agree that it seems odd now. However, I think that Runtime devs are still mostly interested in running the "CoreCLR" benchmarks only. I am going to wait to hear more feedback. Most probably we are going to remove it within the next few months.

billwert: Yeah. The categories really represent "what repo the test used to live in."

I think we should ditch the CoreCLR/CoreFX distinction and come up with categories that are meaningful to the test. There's no reason we can't have "Serializers", "Regex", etc, etc.

I also think that whatever we put in the attribute should be a specific descriptor of the test and what functionality area it targets, and nothing about grouping or organization. That way when we inevitably re-organize again in three years we don't have to go touch a ton of places.

For example, if we want "metacategories" like CoreCLR so that folks can run groups of tests, I think we should build a feature into either our argument parser or BDN directly to allow grouping those. We change the categories on a given test to better match what it does, and then if someone specifies a metacategory we transform it to everything it maps to.

src/benchmarks/micro/coreclr/BenchmarksGame/regex-redux-5.cs
```diff
@@ -26,78 +25,57 @@ namespace BenchmarksGame
...
-        public int RunBench()
+        [Arguments(RegexOptions.Compiled)]
+        [Arguments(RegexOptions.None)]
+        public int RunBench(RegexOptions options)
         {
             using (var inputStream = new FileStream(helpers.InputFile, FileMode.Open))
             using (var input = new StreamReader(inputStream))
             {
-                return Bench(input, false);
+                return Bench(input, options);
```
stephentoub: @adamsitnik, how much of the time in this benchmark is spent reading the input file?  This was all separated out of the benchmark into GlobalSetup in the version you're deleting.

adamsitnik: I thought that the original benchmark used by The Computer Language Benchmarks Game read it from the file, but it turns out that the input is read from... Console ;)

https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/regexredux-csharpcore-5.html

```cs
var sequences = Console.In.ReadToEnd();
```

I am going to move it to Setup ;)

PR #1079, opened 2019-12-05
Added instructions to run with .NET Desktop to the README.

ivdiazsa:
The GC Benchmarking Infra tool has the functionality to run tests using .NET Desktop apart from .NET Core. Here we add documented instructions on how to build and run tests this way.

### DIFF
--- a/src/benchmarks/gc/README.md
+++ b/src/benchmarks/gc/README.md
@@ -13,8 +13,6 @@ The general workflow when using the GC infra is:
 * Run the benchfile and collect traces.
 * Run analysis on the output.
 
-
-
 # Setup
 
 ### Install python 3.7+
@@ -28,13 +26,11 @@ It's recommended to install a 64-bit version if possible, but not required.
 
 On other systems, it’s better to use your system’s package manager.
 
-
 ### Install Python dependencies
 
 ```sh
 py -m pip install -r src/requirements.txt
 ```
-
 
 ### Install pythonnet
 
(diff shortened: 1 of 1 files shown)


ivdiazsa:
@andy-ms @Maoni0 @VSadov

src/benchmarks/gc/README.md
```diff
@@ -179,6 +179,31 @@ Each trace file can be opened in PerfView if you need to.
...
+```
+
+In this example, we are adding version 4.7.2 to the already existing ones of .NET Core.
+
+Next, you have to rebuild the binaries like before. Issue `dotnet build -c release` again. This will generate the new _self contained_ executable of _GCPerfSim_ back in the `artifacts` directory path at the root of the _performance_ repo. By default, it is located in `performance/artifacts/bin/GCPerfSim/release/net472/GCPerfSim.exe`, supposing you are targeting 4.7.2 as in this example. Otherwise, replace that folder with the version you selected.
+
+After this is completed, we will tell our _bench_ files to use this executable. Open your favorite one in your favorite editor.
+
+There, under the `coreclrs` section, replace the `core_root` property with `self_contained` set to `true`.
+
```
VSadov: An example here could help more than descripiton. Should it look like this? :

myBench.yml
```yml
coreclrs:
  a:
    self_contained: true
. . .

benchmarks:
  a:
    executable: 
      performance/artifacts/bin/GCPerfSim/release/net472/GCPerfSim.exe
    arguments:
      tc: 10
      tagb: 40
      tlgb: 5
      lohar: 0
      sohsi: 30
      lohsi: 0
      sohpi: 50
      lohpi: 0
      sohfi: 0
      lohfi: 0
      allocType: reference
      testKind: time
```

Maoni0: agreed. in fact it would be great to add an example for running with coreclr as well, if you don't mind, @ivdiazsa. other than this, LGTM!

ivdiazsa: I can add it as well. I believe it would be along the lines of "For CoreCLR, the bench files generated by `py . suite-create` are ready to be used out of the box. Here is an example."

ivdiazsa:

PR #1081, opened 2019-12-05
Update/add more meaningful performance tests

buyaa-n:


src/benchmarks/micro/corefx/System.Xml.Linq/Perf.XName.cs
```diff
@@ -13,22 +13,25 @@ public class Perf_XName
...
+
         [GlobalSetup]
         public void Setup()
         {
             _noNamespace = XName.Get("Root");
             _hasNamespace = XName.Get("{http://www.example.test}Root");
         }
 
-        [Benchmark]
-        public XName GetLocalName() => _noNamespace.LocalName;
```
buyaa-n: XName.LocalName not return XName object it return plain string, mistakenly set return type to XName which would run implicit constructor using the LocalName string. So this test has side effect and we don't need to test XName.LocalName as that will just return _localName field. So just removed it

krwq:
@buyaa-n this generally looks good to me but I'm unsure if this won't affect perf of existing benchmarks which will likely mess up the stats. @adamsitnik thoughts?

PR #1086, opened 2019-12-09
Support for process flag and updated documentation for GC Infra

ivdiazsa:
When analyzing and comparing trace files, one requires to be able to select which process to work with. This PR adds this functionality via a `--process` flag, which works with either the name or the id of the process. Some refactoring was also made.

Additionally, the documentation received a significant overhaul, giving _Pythonnet_ its own doc, adding more detailed explanations on scenarios and working with trace files, and some metrics were updated as well.

### DIFF
--- a/src/benchmarks/gc/src/analysis/process_trace.py
+++ b/src/benchmarks/gc/src/analysis/process_trace.py
@@ -8,11 +8,17 @@ from typing import cast, Dict, Iterable, Sequence
 
 from result import Err, Ok, Result
 
-from ..commonlib.bench_file import is_trace_path, load_test_status, TestResult, TestRunStatus
+from ..commonlib.bench_file import (
+    is_trace_path,
+    load_test_status,
+    ProcessQuery,
+    TestResult,
+    TestRunStatus,
+)
 from ..commonlib.collection_util import indices, map_to_mapping, repeat, zip_check, zip_check_3
 from ..commonlib.option import map_option, non_null
 from ..commonlib.result_utils import map_err, map_ok, match, option_to_result, unwrap
-from ..commonlib.util import change_extension, show_size_bytes
+from ..commonlib.util import show_size_bytes
 
 from .clr import Clr, get_clr
 from .clr_types import (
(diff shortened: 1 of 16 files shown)


ivdiazsa:
@Maoni0 @VSadov Here are the changes Andy last worked on, as well as the process support we worked with him in his last week.

src/benchmarks/gc/README.md
```diff
@@ -20,7 +20,9 @@ The general workflow when using the GC infra is:
 ### Install python 3.7+
 
 On Windows, just go to https://www.python.org/downloads/ and run the installer.
-On other systems it’s better to use your system’s package manager.
+It's recommended to install a 64-bit version if possible, but not required.
```
VSadov: This is already in the upstream. I think you need to rebase to pick documentation changes.

src/benchmarks/gc/src/analysis/jupyter_notebook_utils.py
```diff
@@ -1,3 +0,0 @@
-# Licensed to the .NET Foundation under one or more agreements.
-# The .NET Foundation licenses this file to you under the MIT license.
-# See the LICENSE file in the project root for more information.
```
VSadov: License header was removed. Is this on purpose?

src/benchmarks/gc/src/analysis/process_trace.py
```diff
@@ -37,39 +43,36 @@
...
+        )
 
 
 def get_processed_trace(
-    clr: Clr,
-    test_result: TestResult,
-    process: ProcessQuery,
-    need_mechanisms_and_reasons: bool,
-    need_join_info: bool,
+    clr: Clr, test_result: TestResult, need_mechanisms_and_reasons: bool, need_join_info: bool
```
VSadov: I assume this was done by the linter. Parameter list got shorter and was reformatted.

src/benchmarks/gc/src/commonlib/bench_file.py
```diff
@@ -1137,6 +1153,14 @@ def __str__(self) -> str:
         return str(self.test_status_or_trace_path)
 
 
+# @with_slots
```
VSadov: Is this unused? Probably should delete or comment why this is here.

src/benchmarks/gc/src/commonlib/document.py
```diff
@@ -139,6 +139,9 @@ def _pad(width: int, s: str, align: Align) -> str:
     return switch[align]()
 
 
+_SUPPORTED_TERMINALS = {"ConEmuC64.exe", "WindowsTerminal.exe"}
```
VSadov: did we actually figure why there was no color in any of the terminals? Do we need a bug logged on that?

VSadov:
"process" change looks good. Left few minor comments. 
For the document part this needs to be rebased. Many changes seem to be already in the repo.

src/benchmarks/gc/src/commonlib/bench_file.py
```diff
@@ -1145,6 +1153,14 @@ def __str__(self) -> str:
...
+# @dataclass(frozen=True)
+# class TestResultWithTuples:
+#     test_status_path: Optional[Path] = None
+#     trace_path: Optional[Path] = None
+#     process: ProcessQuery = None
+
```
Maoni0: can be deleted?

src/benchmarks/gc/src/commonlib/document.py
Maoni0: I'm curious about the same thing

---
In reply to: [355727790](https://github.com/dotnet/performance/pull/1086#discussion_r355727790) [](ancestors = 355727790)

src/benchmarks/gc/src/commonlib/bench_file.py
Maoni0: this is used in this file, which I presume was the reason why it got move here from analysis.

---
In reply to: [355727310](https://github.com/dotnet/performance/pull/1086#discussion_r355727310) [](ancestors = 355727310)

Maoni0:
what kind of testing was done on this? we should make sure we don't break the test status file case. 

please also make sure the doc is updated to reflect this.

src/benchmarks/gc/src/commonlib/document.py
ivdiazsa: Actually, I should either remove this and leave the color fix on its own branch and PR, as it is right now, or I could merge it in this one. I believe the former is the better way to proceed, but I would like to hear your thoughts.

src/benchmarks/gc/src/commonlib/bench_file.py
ivdiazsa: That was just an experiment Andy and I did at some point. I have deleted it now.

ivdiazsa:
The `diff` command should be now working properly with `.yaml` test status files, as well as `.etl` trace files with the `--process` flag. I tested other examples, such as charting gcs and analyzing single metrics, and everything seems to be just fine.

src/benchmarks/gc/src/commonlib/document.py
ivdiazsa: Taking a closer look, this set was added to simplify the color function, but does not affect that functionality. That fix will be added in another PR.

ivdiazsa:

PR #1088, opened 2019-12-13
Add missing Benchmark attribute and simplify the benchmark name

adamsitnik:
Despite 3 reviewers we have forgot about... `[Benchmark]` attribute in #1016

I've also simplified the name and benchmark code itself.

/cc @VSadov


adamsitnik:
since it's a simple bug fix I merging it without review

PR #1092, opened 2019-12-16
use channels in pipeline

ooooolivia:
- updated pipeline to use channels for downloding dotnet 
- added channel to tfm, channel to branch mapping in the channel_map.py script
- updated our newest testing branch from ```release/3.1.1xx``` (broken channel)  to ```release/3.1.2xx```


eng/performance/benchmark_jobs.yml
```diff
@@ -90,16 +90,18 @@ jobs:
...
                 _PerfLabArguments: $(PerfLabArguments)
-                _BuildConfig: ${{ parameters.architecture }}_$(_Framework)_${{ parameters.kind }} # needs to be unique to avoid logs overwriting in mc.dot.net
+                _BuildConfig: ${{ parameters.architecture }}_$(_Channel)_${{ parameters.kind }} # needs to be unique to avoid logs overwriting in mc.dot.net
         steps:
         - checkout: self
           clean: true
-        - script: $(Python) scripts/ci_setup.py --frameworks $(_Framework) --architecture ${{parameters.architecture}} --perf-hash $(Build.SourceVersion) --queue ${{parameters.queue}} --build-number $(Build.BuildNumber) --build-configs $(_Configs)
+        - script: $(Python) scripts/channel_map.py --channel $(_Channel) --setup-pipeline # find the channel's matching tfm 
+          displayName: Look up TFM
+        - script: $(Python) scripts/ci_setup.py --channels $(_Channel) --architecture ${{parameters.architecture}} --perf-hash $(Build.SourceVersion) --queue ${{parameters.queue}} --build-number $(Build.BuildNumber) --build-configs $(_Configs)
```
billwert: This isn't a problem with your PR but since you're changing this line can you add a `displayName` here also? I dislike that it shows up as "CmdLine" in AzDo

billwert:
> from argparse import ArgumentParser

Unused

---
Refers to: scripts/micro_benchmarks.py:9 in 24f22db. [](commit_id = 24f22dbc044b1079e2e300052814b5e6848f0eb9, deletion_comment = False)

eng/performance/benchmark_jobs.yml
ooooolivia: yah np

billwert:
> from argparse import ArgumentError

this is just a test comment

---
Refers to: scripts/micro_benchmarks.py:8 in 24f22db. [](commit_id = 24f22dbc044b1079e2e300052814b5e6848f0eb9, deletion_comment = False)

scripts/micro_benchmarks.py
```diff
@@ -4,8 +4,6 @@
 Builds the Benchmarks
 '''
 
-from argparse import Action
-from argparse import ArgumentError
 from argparse import ArgumentParser
 from argparse import ArgumentTypeError
```
billwert: testing again

azure-pipelines.yml
```diff
@@ -92,8 +89,8 @@ jobs:
       queue: Windows.10.Amd64.ClientRS4.DevEx.15.8.Open
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
       runCategories: 'coreclr corefx'
-      frameworks: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
-        - net461
+      channels: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
+        - LTS # use LTS channel for net461 framework
```
adamsitnik: 👍 thanks for adding the comment!

azure-pipelines.yml
```diff
@@ -74,12 +74,9 @@ jobs:
       queue: Windows.10.Amd64.ClientRS5.Open
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
       runCategories: 'coreclr corefx' 
-      frameworks: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
-        - netcoreapp5.0
-        - netcoreapp3.1
-        - netcoreapp3.0
-        - netcoreapp2.2
-        - netcoreapp2.1
+      channels:
```
adamsitnik: with this change, we test all frameworks on Ubuntu and only two on Windows. Is this intentional?

https://github.com/dotnet/performance/blob/8b9c22bb7d75e966260c89eccf02ae015ea87b21/azure-pipelines.yml#L151-L156

eng/performance/benchmark_jobs.yml
```diff
@@ -90,16 +90,19 @@ jobs:
...
-            ${{ each framework in parameters.frameworks }}:
-              ${{ framework }}:
-                _Framework: ${{ framework }}
+            ${{ each channel in parameters.channels }}:
+              ${{ channel }}:
+                _Channel: ${{ channel }}
                 _Configs: CompilationMode=Tiered RunKind="${{ parameters.kind }}"
                 _PerfLabArguments: $(PerfLabArguments)
-                _BuildConfig: ${{ parameters.architecture }}_$(_Framework)_${{ parameters.kind }} # needs to be unique to avoid logs overwriting in mc.dot.net
+                _BuildConfig: ${{ parameters.architecture }}_$(_Channel)_${{ parameters.kind }} # needs to be unique to avoid logs overwriting in mc.dot.net
```
adamsitnik: this is changing CI build ID used in the Build Status Page in the README.md file. It would be nice to update the Build Status Page or just remove it and keep the two most important buttons (Public Build Status and Internal Build Status)

scripts/channel_map.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def __main():
+    parser = ArgumentParser()
+    parser.add_argument('--channel', type=str, required=True, dest='channel', help='channel name or status')
+    parser.add_argument('--setup-pipeline', action='store_true', dest='setup_pipeline', help='channel required to set up pipeline variables')
+    args = parser.parse_args()
+
+    if args.setup_pipeline:
+        print('##vso[task.setvariable variable=_Framework]%s' % ChannelMap.get_target_framework_moniker(args.channel))
```
adamsitnik: it would be nice to add a comment here explaining what it is doing

eng/performance/benchmark_jobs.yml
```diff
@@ -90,16 +90,19 @@ jobs:
...
         steps:
         - checkout: self
           clean: true
-        - script: $(Python) scripts/ci_setup.py --frameworks $(_Framework) --architecture ${{parameters.architecture}} --perf-hash $(Build.SourceVersion) --queue ${{parameters.queue}} --build-number $(Build.BuildNumber) --build-configs $(_Configs)
+        - script: $(Python) scripts/channel_map.py --channel $(_Channel) --setup-pipeline # find the channel's matching tfm 
+          displayName: Look up TFM
+        - script: $(Python) scripts/ci_setup.py --channels $(_Channel) --architecture ${{parameters.architecture}} --perf-hash $(Build.SourceVersion) --queue ${{parameters.queue}} --build-number $(Build.BuildNumber) --build-configs $(_Configs)
```
adamsitnik: just curious: why did you decide to keep the logic from `channel_map.py` in a separate file? as far I understand it translates the channel to a tfm and then stores the result in some AZDO specific way (so if I run it locally it won't work?)? would it be better to just have it as part of `ci_setup.py`?

scripts/micro_benchmarks.py
```diff
@@ -160,10 +60,9 @@ def __dotnet_configuration(configuration: str) -> str:
 
     parser.add_argument(
         '-f', '--frameworks',
-        required=True,
+        required=False,
```
adamsitnik: big thanks for keeping the old feature working (I use it very frequently) 👍 

One funny side effect of this change is that if we run the script locally without specyfing the tfm or channel, we get a very strange python error:

```log
PS C:\Projects\performance> py .\scripts\benchmarks_ci.py
[2020/01/06 16:47:27][INFO] ----------------------------------------------
[2020/01/06 16:47:27][INFO] Initializing logger 2020-01-06 16:47:27.512296
[2020/01/06 16:47:27][INFO] ----------------------------------------------
  File ".\scripts\benchmarks_ci.py", line 262, in <module>
    __main(sys.argv[1:])
  File ".\scripts\benchmarks_ci.py", line 199, in __main
    .get_target_framework_monikers(args.frameworks)
  File "C:\Projects\performance\scripts\dotnet.py", line 79, in get_target_framework_monikers
    for framework in frameworks
TypeError: 'NoneType' object is not iterable
```

Previously it would complain about the lack of mandatory argument. Now we have two optional arguments. It would be nice to tell the user that tfm or channel has to be provided.

ooooolivia:
> Thanks for keeping the `-f` working, I use it very frequently! I left some comments, mostly nits.
> 
> Just curious: how do we know that a new channel has been created?

Thanks for your comment! After discussion with sdk team, we decide that sdk team will notify us for new branches&channels, and we will update the map accordingly.

azure-pipelines.yml
ooooolivia: Good catch! This is a mistake. I'll update it asap.

eng/performance/benchmark_jobs.yml
ooooolivia: Currently ```channel_map.py``` can be used by other scripts such as ```benchmarks_ci.py```, ```micro_benchmarks.py``` and ```ci_setup.py```, to get the right tfm for a channel; it is also used independently in the pipeline. Because we'd like to support local benchmark runs triggered by ```benchmarks_ci.py```, putting them in ```ci_setup.py``` would make it only available for ci runs. What do you think about this?

scripts/channel_map.py
ooooolivia: Thanks for your suggestion! Will add it in the following PR.

scripts/micro_benchmarks.py
ooooolivia: Thanks for pointing it out! Will update this.

eng/performance/benchmark_jobs.yml
ooooolivia: Thanks! I wasn't aware of this dependency. Will update this.

eng/performance/benchmark_jobs.yml
ooooolivia: New plan based on offline discussion: moving the content of ```channel_map.py``` to ```ci_setup.py``` sounds right, because 1) it's high-coupling for ```channel_map.py``` to know anything about the pipeline 2) if there's no need to run ```channel_map.py```, we can totally move the ```ChannelMap``` class to ```ci_setup.py``` and let ```benchmarks_ci.py``` use this module in local runs

PR #1096, opened 2019-12-19
Added support for WSL Terminal and show colors on the Summary Section.

ivdiazsa:
Added support for WSL Terminal on Windows. Also, fixed the issue that was preventing colors from showing in the Summary Section when diffing traces.

### DIFF
--- a/src/benchmarks/gc/src/commonlib/document.py
+++ b/src/benchmarks/gc/src/commonlib/document.py
@@ -149,7 +149,7 @@ def _shell_supports_color() -> bool:
         if parent.name() in ("Code.exe", "jupyter-notebook.exe"):
             return True
         else:
-            assert parent.name() in ("powershell.exe", "cmd.exe")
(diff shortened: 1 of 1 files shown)


ivdiazsa:
@Maoni0 @VSadov Right now, only large percentage differences (I think it's 20%+) in the Summary stats show color. Would we want to also add color to the detailed tables of `py . diff`?

ivdiazsa:
Adding @cshung as well.

VSadov:
Re: color in `py . diff` - I guess, why not. 
How would it look like?

We probably want a separate PR for that.

ivdiazsa:
> Re: color in py . diff - I guess, why not.
> How would it look like?
> We probably want a separate PR for that.

If you run `py . diff` and compare any two trace files, there are at least two sections in the output:
- Summary of important stats
- Each test's detailed stats

Right now, colors are only shown for bigger deltas in the summary section. We might later work on another PR to show color in the detailed section as well.

ivdiazsa:

PR #1097, opened 2019-12-20
Added support to capture Thread Time Stacks when running GC Benchmarks.

ivdiazsa:
When running GC Benchmark tests, it might be necessary to also capture Thread Time stacks for further analysis. This functionality was missing from GC Infra and this PR adds it.


ivdiazsa:
@Maoni0 @VSadov @cshung

Maoni0:
ThreadTime already includes CSwitch events. I don't think you need to add yet another option. when I said ThreadTime I just meant you needed to collect CSwitch events - ThreadTime is the perfview way to collect them. you could change the current cswitch to collect ThreadTime though - it would make analysis in perfview better.

ivdiazsa:
> ThreadTime already includes CSwitch events. I don't think you need to add yet another option. when I said ThreadTime I just meant you needed to collect CSwitch events - ThreadTime is the perfview way to collect them. you could change the current cswitch to collect ThreadTime though - it would make analysis in perfview better.

I understand. So if ThreadTime includes CSwitch events as well, then my question is which one is better known to users. I agree we should just keep one option then, but would it be better to rename it to "thread_times" or leaving it at "cswitch" would be more understandable?

ivdiazsa:
The `thread_times` support is done. I have run several scenarios to ensure it works correctly and didn't break anything else, and updated the documentation throughout the docs and other source files. Also, I withdrew the color support change because it's causing trouble under native CMD and the fix doesn't seem to be jumping out easily.
@Maoni0 @VSadov @cshung

src/benchmarks/gc/src/exec/run_single_test.py
```diff
@@ -306,27 +306,29 @@ def _get_perfview_start_or_stop_cmd(
...
+    CollectKind.cpu_samples: [
+        "-OnlyProviders:ClrPrivate:1:5,Clr:1:5",
+        "-ClrEvents:GC+Stack",
+        "-ClrEventLevel:Verbose",
+        "-KernelEvents:Default",
+    ],
+    CollectKind.thread_times: [
+        "-OnlyProviders:ClrPrivate:1:5,Clr:1:5",
+        "-ClrEvents:GC+Stack",
+        "-ClrEventLevel:Verbose",
```
Maoni0: >"-ClrEventLevel:Verbose", [](start = 8, length = 25)

do you actually need this? we are already saying 5 for the level in OnlyProviders

src/benchmarks/gc/src/exec/run_single_test.py
```diff
@@ -306,27 +306,29 @@ def _get_perfview_start_or_stop_cmd(
...
+        "-KernelEvents:Default,ThreadTime",
```
Maoni0: >"-KernelEvents:Default,ThreadTime" [](start = 8, length = 34)

I don't think you need both Default and ThreadTime, ThreadTime alone is sufficient

src/benchmarks/gc/src/exec/run_single_test.py
ivdiazsa: I forgot to remove that line. I'll change it.

src/benchmarks/gc/src/exec/run_single_test.py
ivdiazsa: Are we certain about this? I left it that way since Andy originally had it as `Default,ContextSwitch`. How can I know whether default events were recorded? I ask to run tests and compare to make sure we can leave just `ThreadTime`.

Maoni0: see [KernelTraceEventParser.cs#L178](https://github.com/microsoft/perfview/blob/master/src/TraceEvent/Parsers/KernelTraceEventParser.cs#L178). feel free to run tests and collect with both cmdline args to make sure.

ivdiazsa: Thanks for pointing me out to the docs, Maoni. I also ran a couple tests and indeed we seem to have the same behavior. I have now removed those parameters.

ivdiazsa:

PR #1109, opened 2020-01-02
recursively change framework version for weblarge3.0 project

ooooolivia:
Weblarge3.0 project has more than one .csproj files that need to change, so we need to replace the frameworks recursively.

### DIFF
--- a/src/scenarios/shared/precommands.py
+++ b/src/scenarios/shared/precommands.py
@@ -91,12 +91,16 @@ class PreCommands:
                             metavar='/p:Foo=Bar;/p:Baz=Blee;...')
         parser.set_defaults(configuration=RELEASE)
 
-    def existing(self, projectdir: str, projectfile: str):
-        'create a project from existing project file'
+    def existing(self, projectdir: str, projectfiles):
+        'create a project for each existing project file'
         self._backup(projectdir)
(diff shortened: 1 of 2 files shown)


src/scenarios/shared/precommands.py
```diff
@@ -91,12 +91,16 @@ def add_common_arguments(self, parser: ArgumentParser):
...
+        'create a project for each existing project file'
         self._backup(projectdir)
-        csproj = CSharpProjFile(os.path.join(const.APPDIR, projectfile), sys.path[0])
-        self.project = CSharpProject(csproj, const.BINDIR)
-        self._updateframework(csproj.file_name)
+        if type(projectfiles) is str:
+            projectfiles = [projectfiles]
+        for projectfile in projectfiles:
+            csproj = CSharpProjFile(os.path.join(const.APPDIR, projectfile), sys.path[0])
+            self.project = CSharpProject(csproj, const.BINDIR)
```
billwert: Storing each project into `self.project` here doesn't make sense. Does this mean that by convention the project that is going to get built/published by the other verbs has to be the last one in the list?

ooooolivia:
This PR will be closed since we don't need this functionality for now. There will be an upcoming PR for making the library projects target netstandard2.0.

PR #1117, opened 2020-01-07
Expand Regex perf tests

stephentoub:
cc: @adamsitnik, @billwert

### DIFF
--- a/src/benchmarks/micro/corefx/System.Text.RegularExpressions/Perf.Regex.Common.cs
+++ b/src/benchmarks/micro/corefx/System.Text.RegularExpressions/Perf.Regex.Common.cs
@@ -13,35 +13,68 @@ namespace System.Text.RegularExpressions.Tests
     [BenchmarkCategory(Categories.CoreFX)]
     public class Perf_Regex_Common
     {
-        private const string EmailPattern = @"^([a-zA-Z0-9_\-\.]+)@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.)|(([a-zA-Z0-9\-]+\.)+))([a-zA-Z]{2,12}|[0-9]{1,3})(\]?)$";
-        private const string DatePattern = @"\b(?<month>\d{1,2})/(?<day>\d{1,2})/(?<year>\d{2,4})\b";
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/corefx/System.Text.RegularExpressions/Perf.Regex.Common.cs
```diff
@@ -13,35 +13,68 @@ namespace System.Text.RegularExpressions.Tests
...
+            _searchSet = new Regex(@"\w{10,}", Options);
+            _searchBoundary = new Regex(@"\b\w{10,}\b", Options);
         }
 
-        [Benchmark] public void Email() => _email.IsMatch("yay.performance@dot.net");
-        [Benchmark] public void Date() => _date.IsMatch("Today is 11/18/2019");
-        [Benchmark] public void IP() => _ip.IsMatch("012.345.678.910");
-        [Benchmark] public void Uri() => _uri.IsMatch("http://example.org");
+        [Benchmark] public void Email_IsMatch() => _email.IsMatch("yay.performance@dot.net");
+        [Benchmark] public void Email_IsNotMatch() => _email.IsMatch("yay.performance@dot.net#");
```
adamsitnik: How big is the difference in the reported time for `IsMatch` and `IsNotMatch`? I wonder if the execution path is different enough to make it worth to "duplicate" the benchmarks (https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#Code-Paths)

stephentoub: > Another disadvantage of adding the "IsNotMatch" test case is changing the IDs of the existing benchmarks

I don't think that's important.  We should ensure we have the benchmarks we want to have.  If we're not measuring the right things, we should stop doing so and measure the right things.

> How big is the difference in the reported time for IsMatch and IsNotMatch?

Big.

> I wonder if the execution path is different enough to make it worth to "duplicate" the benchmarks

It is.  The match case is the happy path: every check as part of the regex succeeds.  The not-match case highlights the quality of the regex engine, since that's where backtracking potentially kicks in.

> I wonder if instead of this we should have a single benchmark where the mismatched character is the first char.

That'd be a fine additional case, but it's not the most important case.

stephentoub:
Thanks.

PR #1129, opened 2020-01-15
pass workitem name as scenario name

ooooolivia:
- make scenario name a parameter to ```test.py``` so it can be set equal to workitem name, which includes framework and is enough to distinguish tests
- set trace file name to ```exename+scenariotype``` for customer runs and ```scenarioname``` for CI runs
- refactor ```scenarios.proj```

### DIFF
--- a/eng/performance/sdk_scenarios.proj
+++ b/eng/performance/sdk_scenarios.proj
@@ -82,10 +82,10 @@
 
   <ItemGroup>
     <HelixWorkItem Include="@(SDKWorkItem -> '%(Identity) Clean Build')">
-      <Command>%(Command) clean_build --scenario-name %(Identity)</Command>
+      <Command>%(Command) clean_build --scenario-name &quot;%(Identity)&quot;</Command>
     </HelixWorkItem>
     <HelixWorkItem Include="@(SDKWorkItem -> '%(Identity) Build(No Change)')">
-      <Command>%(Command) build_no_change --scenario-name %(Identity)</Command>
+      <Command>%(Command) build_no_change --scenario-name &quot;%(Identity)&quot;</Command>
     </HelixWorkItem>
   </ItemGroup>
 </Project>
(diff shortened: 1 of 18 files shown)


eng/performance/sdk_scenarios.proj
```diff
@@ -82,10 +82,10 @@
 
   <ItemGroup>
     <HelixWorkItem Include="@(SDKWorkItem -> '%(Identity) Clean Build')">
-      <Command>%(Command) clean_build</Command>
+      <Command>%(Command) clean_build --scenario-name &quot;%(Identity)&quot;</Command>
```
billwert: Does `%(Identity)` here become `%(Identity) Clean Build` automatically or should it be explicitly included?

ooooolivia: It includes ```Clean Build```

src/scenarios/shared/runner.py
```diff
@@ -60,9 +60,10 @@ def parseargs(self):
         parser = ArgumentParser()
         subparsers = parser.add_subparsers(title='subcommands for sdk tests', required=True, dest='testtype')
         startupparser = subparsers.add_parser(const.STARTUP)
+        startupparser.add_argument('--scenario-name', dest='scenarioname')
```
billwert: Can this go on the root parser? Does each sub-parser always have to include this?

ooooolivia: It could but this would make it mandatory to position the optional argument before the required sub-parsers. For example, ```python test.py ---scenario-name 'empty console template' startup``` is a bit counter-intuitive.

billwert: Fair. Could we build a little factory or something? We will have more parsers in the future I suspect.

ooooolivia:
> LGTM, but why are we keeping the SCENARIONAME variable in the test.py files? We don't use it anywhere?

I kept it because it might be useful at some point in the future? It will not be used in generating the test name any more, but could be used as a definition of the test for customer runs, at least that's what I think.

PR #1136, opened 2020-01-17
Remove 2.2

### LINKED ISSUE #1000: Remove 2.2 after 23rd of December 2019
According to https://dotnet.microsoft.com/platform/support/policy/dotnet-core .NET Core 2.2 is going to reach the end of support at 23rd of December 2019

Since we have a LOT of CI legs and there is a lot of `#if` in the code I believe that it would be good to remove all 2.2 stuff **after** it becomes

- [ ]  snap a branch with the existing state
- [ ]  update `.yml` file and build matrix page
- [ ]  update all `#if NETCOREAPP2_2` usages
- [ ]  update all docs that mention 2.2

adamsitnik:
I've removed everything related to 2.2 which is not supported since 23rd of December 2019

As requested by @billwert  I've also created 2.2 branch: https://github.com/dotnet/performance/tree/2.2

I've also noticed that 3.1 was not on the list, so I've added it. I've also updated the docs to show 3.1 vs 5.0 as the example comparison since this is what we care about the most for the next release.

fixes #1000

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -92,6 +92,8 @@ jobs:
       channels:
         - master
         - release/3.1.2xx
+        - 3.0
+        - 2.1
 
   # Windows x64 net461 micro benchmarks
   - template: /eng/performance/benchmark_jobs.yml
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -167,7 +167,6 @@ jobs:
         - master
         - release/3.1.2xx
         - 3.0
-        - 2.2
         - 2.1
 
   # Ubuntu 1804 x64 ML.NET benchmarks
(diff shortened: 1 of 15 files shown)


billwert:
@ooooolivia Can you confirm if this impacts the SDK team's matrix? As 2.2 is out of support we should probably make sure they know we're going to turn it off.

Please hold off merging this until we get that confirmation, @adamsitnik. It LGTM but I want to make sure we aren't impacting something else.

ooooolivia:
> @ooooolivia Can you confirm if this impacts the SDK team's matrix? As 2.2 is out of support we should probably make sure they know we're going to turn it off.
> 
> Please hold off merging this until we get that confirmation, @adamsitnik. It LGTM but I want to make sure we aren't impacting something else.

Currently SDK tests do not have 2.2 in the matrix so it should be fine.

adamsitnik:
Thank you for the reviews! #merging

PR #1132, opened 2020-01-16
Update docs after repos merge

adamsitnik:
Fixes #990

PR #1142, opened 2020-01-22
Fix microbenchmarks input files

adamsitnik:
In #1132 I've forgotten to change the folder names used by input file helpers.

Yesterday @wfurt has tried to run the networking benchmarks and they have all failed due to missing files.

This PR fixes that. What worries me is that CI was green despite the benchmarks were failing. I am going to file a separate issue for that.


wfurt:
thanks @adamsitnik  for a quick fix.

PR #1143, opened 2020-01-22
Fix broken benchmarks and make sure we catch the problems next time

adamsitnik:


src/harness/BenchmarkDotNet.Extensions/Extensions.cs
```diff
@@ -17,7 +17,7 @@ public static int ToExitCode(this IEnumerable<Summary> summaries)
                 return 1;
 
             // if anything has failed, it's an error
-            if (summaries.Any(summary => summary.HasCriticalValidationErrors || summary.Reports.Any(report => !report.BuildResult.IsBuildSuccess || !report.ExecuteResults.Any())))
+            if (summaries.Any(summary => summary.HasCriticalValidationErrors || summary.Reports.Any(report => !report.BuildResult.IsBuildSuccess || !report.AllMeasurements.Any())))
```
adamsitnik: @billwert `report.ExecuteResults.Any()` was returning true if there was a result with exit code != 0 with no measurements (everything built, but exception thrown during execution). Now we are just checking if there are any measurements.

PR #1148, opened 2020-01-25
Add crossgen tests

billwert:
This is the harness for measuring crossgen perf against runtime builds. I'll add the tests to scenario ymls in a separate PR.


DrewScoggins:
Looking at this now

DrewScoggins:
LGTM

PR #1157, opened 2020-01-30
Added basic support for ARM/ARM64 to GC Benchmarking Infra

ivdiazsa:
This PR adds support to be able to setup the GC Benchmarking Infrastructure and run basic tests on _ARM/ARM64_. Also, I updated the documentation regarding the steps to take when setting up on these architectures, as well as the limitations currently existing within the tool.

As an additional improvement, I changed `del` to `rd` in a build script to avoid potential issues due to a directory already existing.

### DIFF
--- a/src/benchmarks/gc/README.md
+++ b/src/benchmarks/gc/README.md
@@ -463,10 +463,10 @@ This is the build output of coreclr that is used to run benchmarks.
 
 These are specified in the `coreclrs` section of a benchfile.
 
-This can be found in a directory like `bin/tests/Windows_NT.x64.Release/Tests/Core_Root` (adjust for different OS or architecture) of a coreclr repository. (The Core_Root can be moved anywhere and doesn't need to remain inside the coreclr repository).
+This can be found in a directory like `runtime/artifacts/bin/coreclr/Windows_NT.x64.Release/` (adjust for different OS or architecture) of a runtime repository. (The Core_Root can be moved anywhere and doesn't need to remain inside the runtime repository).
 
-A clone of https://github.com/dotnet/coreclr,  which may be on an arbitrary commit (including one not checked in).
-When you make a change to coreclr, you will generally make two clones, one at master and one at your branch (which may be on your fork).
+A clone of https://github.com/dotnet/runtime, which may be on an arbitrary commit (including one not checked in).
+When you make a change to coreclr within runtime, you will generally make two clones, one at master and one at your branch (which may be on your fork).
 Alternately, you may have only one checkout, build multiple times, copy the builds to somewhere, and specify coreclrs using `core_root` instead of `path`.
 
 ### Config
--- a/src/benchmarks/gc/README.md
+++ b/src/benchmarks/gc/README.md
@@ -463,7 +463,7 @@ This is the build output of coreclr that is used to run benchmarks.
 
 These are specified in the `coreclrs` section of a benchfile.
 
-This can be found in a directory like `runtime/artifacts/bin/coreclr/Windows_NT.x64.Release/` (adjust for different OS or architecture) of a runtime repository. (The Core_Root can be moved anywhere and doesn't need to remain inside the runtime repository).
+This can be found in a directory like `runtime/artifacts/bin/coreclr/Windows_NT.x64.Release/` (adjust for different OS or architecture) of a runtime repository. (The Core_Root can be moved anywhere and doesn't need to remain inside the runtime repository). However, it cannot be used as is after building. Move/Copy all files within `crossgen2` to the Core_Root directory as the benchmark executable currently does not find assemblies or binaries deeper in the directory tree.
 
 A clone of https://github.com/dotnet/runtime, which may be on an arbitrary commit (including one not checked in).
 When you make a change to coreclr within runtime, you will generally make two clones, one at master and one at your branch (which may be on your fork).
--- a/src/benchmarks/gc/README.md
+++ b/src/benchmarks/gc/README.md
@@ -13,6 +13,8 @@ The general workflow when using the GC infra is:
 * Run the benchfile and collect traces.
 * Run analysis on the output.
 
+NOTE: If running under ARM/ARM64, the program's functionalities are limited to running certain benchmarks only and the setup process is slightly different. This is pointed out as necessary throughout this document. Look out for the _ARM NOTE_ labels. As for the other necessary tools without official ARM/ARM64 downloads (e.g. python, cmake), you can install and run the x86 versions.
+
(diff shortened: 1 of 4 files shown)


src/benchmarks/gc/README.md
```diff
@@ -429,7 +463,7 @@ This is the build output of coreclr that is used to run benchmarks.
 
 These are specified in the `coreclrs` section of a benchfile.
 
-This can be found in a directory like `bin/tests/Windows_NT.x64.Release/Tests/Core_Root` (adjust for different OS or architecture) of a coreclr repository. (The Core_Root can be moved anywhere and doesn't need to remain inside the coreclr repository.)
+This can be found in a directory like `bin/tests/Windows_NT.x64.Release/Tests/Core_Root` (adjust for different OS or architecture) of a coreclr repository. (The Core_Root can be moved anywhere and doesn't need to remain inside the coreclr repository).
```
cshung: Does this seem to imply Windows only?

src/benchmarks/gc/README.md
```diff
@@ -429,7 +463,7 @@ This is the build output of coreclr that is used to run benchmarks.
...
 
 A clone of https://github.com/dotnet/coreclr,  which may be on an arbitrary commit (including one not checked in).
```
cshung: Can you also update the docs to refer to the runtime repo instead of the CoreCLR repo?

src/benchmarks/gc/src/commonlib/get_built.py
```diff
@@ -51,13 +51,13 @@ def _get_os_name() -> str:
...
 
 
 def _get_platform_name() -> str:
     if is_arm():
-        # TODO: detect arm32 vs 64
-        return "arm64"
+        # On ARM, the machine() function returns the exact name we need here.
+        return machine().lower()
     else:
         p = processor()
```
cshung: How about x86?

src/benchmarks/gc/src/commonlib/setup.py
```diff
@@ -11,15 +11,19 @@
...
 def setup() -> None:
     if os_is_windows():
         _download_perfview_and_sigcheck()
-    # Need sigcheck to do this
-    write_host_info()
+
+    # Need sigcheck to do this. Also, since the host info tools are not
+    # available on ARM, skip when on said architecture.
+    if not is_arm():
+        write_host_info()
```
cshung: I imagine people (like me :p) would forget doing the manual host info specification.
What would happen if the host info were not manually specified?
Can we emit a warning somewhere?

src/benchmarks/gc/README.md
```diff
@@ -65,6 +69,34 @@ Finally, run `py . setup` from the same directory as this README.
...
+    n_caches:
+    total_bytes:
+  l3:
+    n_caches:
+    total_bytes:
+clock_ghz:
+total_physical_memory_mb:
+```
+
+Most (if not all) of these fields can be retrieved from your machine's _Task Manager_ and under _System_ within _Control Panel_.
```
cshung: ARM specifics could be moved to a separate doc to keep this one short.
Some brief instruction on how to do this would be nice.
How about Linux ARM devices?

src/benchmarks/gc/README.md
ivdiazsa: That's a change worth doing. At the time Andy wrote this documentation, runtime was not around yet.

src/benchmarks/gc/src/commonlib/get_built.py
ivdiazsa: To add another architecture here, we would need to see how Python shows the output of `processor()` and `machine()` on said machine. I'm not sure whether Andy tested this and got `x86_64` as a result. This change would probably be better off in another PR.

src/benchmarks/gc/src/commonlib/setup.py
ivdiazsa: That's an interesting observation worth adding. Will work on it.

src/benchmarks/gc/README.md
ivdiazsa: Windows is just an example in this case. You need a Core_Root for Linux testing as well.

ivdiazsa: I think we could add two examples, one for Windows and one for Linux?

src/benchmarks/gc/README.md
ivdiazsa: Wouldn't this cause redundant information lying around? Most steps are the same with a few deviations here and there. Hence I thought adding it to the README would be more clean.

cshung: My apologies if I wasn't clear. I meant if we wanted to expand on this (e.g. more detailed description on how to author the yaml file), then we might up those information in a new page to avoid cluttering this one.

ivdiazsa: Ok got it! If we add more information on this on a future PR, we could move it to another file. Maybe one named `docs/host_info_setup.md` or something like that.

src/benchmarks/gc/README.md
```diff
@@ -65,6 +69,34 @@ Finally, run `py . setup` from the same directory as this README.
 This will read information about your system that's relevant to performance analysis (such as cache sizes) and save to `bench/host_info.yaml`.
 It will also install some necessary dependencies on Windows.
 
+_ARM NOTE_: Since build tools do not work on ARM/ARM64, `py . setup` will automatically skip reading and writing the system's information. You will have to get the machine's specs and write `bench/host_info.yaml` independently. This file follows the format shown below (it might vary depending on your machine's NUMA nodes and caches).
```
Maoni0: >independently [](start = 213, length = 13)

probably wanna say "manually" instead of "independently" here

src/benchmarks/gc/README.md
Maoni0: could you please do what @cshung suggested since coreclr repo is obsolete now?

src/benchmarks/gc/src/commonlib/get_built.py
Maoni0: I think the older version of python would return x86_64. 
agree that checking for x86 should be done in a separate PR.

Maoni0:
LGTM minus the couple of comments above.

src/benchmarks/gc/README.md
ivdiazsa: I have changed it now to runtime. I also built stuff on the runtime repo and updated this entire section.

ivdiazsa:
@billwert @DrewScoggins This PR is ready for merge.

PR #1166, opened 2020-02-04
Add client env setup for scenario tests

ooooolivia:
- Extend the functionality of ```init.ps1``` and ```init.sh``` so it can use a specific version of dotnet when running on client desktops
- In ```dotnet.py```, get supported channels from ChannelMap instead of harcoding


DrewScoggins:
LGTM

PR #1171, opened 2020-02-06
Change syntax of QueueClient to match new libs

DrewScoggins:


DrewScoggins:
The syntax of calls to QueueClient had changed since we last used this.

DrewScoggins:
The failed CI leg is not related to this change. It seems to be a transient issue that we hit during cleanup

Issue #727, opened 2019-08-02
Error libgdiplus is missing


noahfalk:
Repro:
1) Ubuntu 18.04 x64
2) Install .Net Core 2.2 SDK
3) git clone https://github.com/dotnet/performance
4) cd performance/src/benchmarks/micro
5) sudo apt-get install libgdiplus
6) Run dotnet run -c Release -f netcoreapp2.2 --filter *Burgers*
Expected: Burgers benchmark runs
Actual: Error message is printed:

```
libgdiplus is missing, you can install it by running 'apt-get install libgdiplus'

Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> SYstem.TypeInitializationException: The type initializer for 'Gdip' threw an exception. ---> System.DllNotGoundException: Unable to load shared library 'libdl' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: liblibdl: cannot open shared object file: No such file or directory
    at Interop.Libdl.dlopen(String filename, Int32 flag)
    at System.Drawing.SafeNativeMethods.Gdip.LoadNativeLibrary()
…
```

This error occurred both before and after running sudo apt-get install libgdiplus. Enabling LD_DEBUG shows that the actual library the code is trying to load and not finding just prior to the error message printing is 'libdl'

billwert:
Historically installing libgdiplus has fixed this for me. I wonder if the package has changed in some way.

@adamsitnik you added this error message. Do you have further context?

adamsitnik:
Anytime I was using a fresh Ubuntu VM I was always hitting the problem of missing `libgdiplus` and `sudo apt-get install libgdiplus` was always enough to solve the problem. I've never hit the missing `libdl`.

@noahfalk is there anything special about your Ubuntu 18.04 machine? Is it latest Ubuntu 18.04 or some very old version? How did you solve the problem?

Also your log contains a suprising typo:

System.DllNot**G**oundException

is it some copy-paste error?

noahfalk:
it was the latest ubuntu 18.04 downloaded ~1 week ago

> How did you solve the problem?

I did something like 'echo Burgers | dotnet run -c Release -f netcoreapp2.2'. The test with the failure wasn't one I wanted to run and doing it this way appears to avoid the issue.

> is it some copy-paste error?

[...]

PR #1173, opened 2020-02-06
make tests runnable on OSX with default .NET dependencies

wfurt:
OSX by default does not have any package manager and .NET itself requires no additional dependencies. 

Drawing tests are only one reason why performance tests do not run on macOS out of the box. 
Since gdi performance is probably not that interesting on macOS, this change will skip tests and make everything else easily runnable. 

The test will still run if proper dependencies are installed.

cc: @safern @adamsitnik

### DIFF
--- a/src/benchmarks/micro/libraries/System.Drawing/Perf_Image_Load.cs
+++ b/src/benchmarks/micro/libraries/System.Drawing/Perf_Image_Load.cs
@@ -50,7 +50,7 @@ namespace System.Drawing.Tests
         {
             try
             {
-                return new [] 
+                return new []
                 {
                     new ImageTestData(ImageFormat.Bmp),
                     new ImageTestData(ImageFormat.Jpeg),
@@ -65,6 +65,11 @@ namespace System.Drawing.Tests
                 Console.ResetColor();
 
                 throw;
+            }
+            catch (Exception) when (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                // skip tests on macOS instead of failing
+                return new ImageTestData[0];
             }
         }
 


wfurt:
contributes to https://github.com/dotnet/performance/issues/727

src/benchmarks/micro/libraries/System.Drawing/Perf_Image_Load.cs
```diff
@@ -66,6 +66,11 @@ private static ImageTestData[] CreateTestCases()
 
                 throw;
             }
+            catch (Exception) when (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
safern: Would it be better to just do a check at the beginning of the test and return the first element if in OSPlatform.OSX?

wfurt: What exactly do you suggest? Replicate detection from runtime? This is essentially constructor for the tests so will publish no tests if this fails without need to understand runtime dependencies here. Unlike XUnit, there is no easy way how to skip method based on runtime detection.

safern: I was suggesting early detection in this constructor: https://github.com/dotnet/performance/pull/1173/files/d47ddc061dea94c6f801e59cdd2f9893c45ea4a0#diff-18cd5ab72627c912954386cf254f95a6R51

```cs
if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
    return new ImageTestData[0];

try
{
....
}
```

wfurt: With that, we could _never_ run the test on macOS. The proposed change allows running the test if you get libgdi via brew/mono-libgdiplus.

safern: Ah that's true. Ok then disregard 😄

PR #1174, opened 2020-02-07
Fixed the missing GC's from trace analysis.

ivdiazsa:
Originally, the tool was always skipping the first two GC's under the assumption they would always have incomplete data. This is not always the case, and makes the analysis fail for some scenarios, like for example, traces with lots of events that are stopped at the first long GC. This is fairly common and removing GC's can entirely break the analysis.

Also, in this PR, I added a safeguard to make the tool fail gracefully under the case a bad trace comes across and has no GC's to analyze.

@Maoni0 @cshung

### DIFF
--- a/src/benchmarks/gc/src/analysis/core_analysis.py
+++ b/src/benchmarks/gc/src/analysis/core_analysis.py
@@ -224,7 +224,7 @@ def _get_gcs_from_mang(
     mang: AbstractTraceLoadedDotNetRuntime, name: str
 ) -> Sequence[AbstractTraceGC]:
     # Skip the first two GCs which often have incomplete events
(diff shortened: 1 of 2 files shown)


ivdiazsa:
@billwert @DrewScoggins This PR is ready for merging.

src/benchmarks/gc/src/analysis/core_analysis.py
```diff
@@ -224,7 +224,7 @@ def _get_gcs_from_mang(
     mang: AbstractTraceLoadedDotNetRuntime, name: str
 ) -> Sequence[AbstractTraceGC]:
     # Skip the first two GCs which often have incomplete events
```
cshung: How about removing the comment as well?

PR #1182, opened 2020-02-11
change manual run parameters

ooooolivia:
Because pipeline variables are processed at runtime and job display name is generated while the yaml is compiled, the job name will be something like ```windows RS5 $(architecture) sdk_scenarios```, which is not allowed by Azure pipeline. 

The workaround is to have a job for each scenario and use the variable ```$(manual_test_name)``` to choose which job to run.


ooooolivia:
The failures are caused by broken dotnet-install scripts. I'll check in the code first to test manual runs.

PR #1192, opened 2020-02-14
Cleanup benchview refs

billwert:
Mirroring some work I just did in the runtime repo, I wanted to excise as many remaining references to benchview as I could.

I've deleted the doc for porting from xunit-performance as I think that project is officially done. @adamsitnik let me know if you disagree.

I've similarly removed all the comments on specific fields about not changing benchmark identity in Benchview. We cover this specifically in our [docs](https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable) and a handful of comments talking about BenchView don't help.


DrewScoggins:
LGTM

PR #1202, opened 2020-02-23
[Utf8JsonWriter] Testing with Memory from MemoryManager

artelk:
Currently the tests for Utf8JsonWriter only used the IBufferWriter implementation that works on top of managed arrays. But the Memory performance characteristics may vary depending on what underlying mechanism is used in the Memory internals. And that could be useful to know what is changed for different Memory types with the newer .Net versions in the future.

Utf8JsonWriter is currently implemented using the IBufferWriter.GetMemory(). That potentially could be changed to use IBufferWriter.GetSpan(). The tests could be useful to check how the changes affect the performance for both buffer writer types.

I've created a simple buffer writer that shares the memory via the MemoryManager.
Also I've created a simple buffer writer that uses a managed array and replaced the current ArrayBuffer with that. That is needed to be able to compare the results from both types of the Memory.

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -97,4 +97,9 @@
     <None Include="**/*.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="libraries\Common\SimpleArrayBufferWriter.cs" />
+    <None Remove="libraries\Common\SimpleMemoryManagerBufferWriter.cs" />
+  </ItemGroup>
+
 </Project>
--- a/src/benchmarks/micro/libraries/Common/SimpleArrayBufferWriter.cs
+++ b/src/benchmarks/micro/libraries/Common/SimpleArrayBufferWriter.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using System.Buffers;
+using System.Runtime.CompilerServices;
+
+namespace MicroBenchmarks.libraries.Common
+{
+    public sealed class SimpleArrayBufferWriter<T> : IBufferWriter<T>
+    {
+        private readonly T[] _array;
+        private int _index;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public SimpleArrayBufferWriter(int length)
+        {
+            _array = new T[length];
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Advance(int count)
+        {
+            var index = _index + count;
+            if (index >= _array.Length)
+                ThrowInvalidOperationException();
+            _index = index;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public Memory<T> GetMemory(int sizeHint = 0)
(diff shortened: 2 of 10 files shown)


billwert:
Hello @artelk!

It's not clear to me that these are appropriate benchmarks for this repo at this time. These aren't adding additional coverage of new library surface, nor are they covering a very common pattern (I think.) Can you help me understand what the overall goal of these benchmarks is?

In addition, if we took such benchmarks, we would not want to change any existing benchmarks in any way. Benchmarks are [immutable in our system](https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable).

artelk:
> It's not clear to me that these are appropriate benchmarks for this repo at this time. These aren't adding additional coverage of new library surface, nor are they covering a very common pattern (I think.) Can you help me understand what the overall goal of these benchmarks is?

I would like to prepare a pool request to dotnet runtime changing the Utf8JsonWriter (use GetSpan instead of GetMemory). I expect the changes won't affect the performance characteristics for managed-array-based buffer writers but they could visibly improve the performance when unmanaged memory is used as the underlying memory provider. Currently there are no tests that would show the performance changes in that case.

Ordinary people play with unsafe unmanaged things while they try to achieve better performance. To be able to use the unmanaged memory with the framework classes like Utf8JsonWriter, Socket etc. they need to wrap it into the Memory type using the MemoryManager functionality.
The problem is that such memory values become the slowest. The Memory.Span property is implemented like this: https://github.com/dotnet/runtime/blob/fff8f5ab763f3c1eea95ab201bfde9bd4b6e62fb/src/libraries/System.Private.CoreLib/src/System/Memory.cs#L346
so it is at the end of the if/else chain (look at the code above). Moreover the MemoryManager.GetSpan() in virtual.

I understand the new tests formally don't increase the coverage. But they increase the knowledge on the performance of the Utf8Writer for the situation when people wanted to achieve the maximum performance (so they use unmanaged memory, pointers etc.)

> In addition, if we took such benchmarks, we would not want to change any existing benchmarks in any way. Benchmarks are [immutable in our system](https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable).

What you would suggest?

billwert:
Thanks for the context.

The primary purpose of benchmarks in this repo is to enable developers who are making changes to .NET Core. It's not our goal here to have examples of the absolute fastest way to do things; it's our goal to capture what we think are the "common" uses of types and track their performance. This benchmark, to me, is more about answering questions about how a user might choose to use Utf8Writer for ultimate performance. As you show with this PR, a user today can replace the memory manager and use a different allocation strategy. I think this would make an excellent post somewhere to share this knowledge with the community.

I'm going to close this PR for now. If we make significant changes (either to code or guidance) around `Utf8Writer` we can revisit this.

> What you would suggest?

First, changing the existing tests to use your `SimpleArrayBufferWriter` isn't a change we would take. This would disrupt our trends. Second, adding a new `Params` field to the type changes the identity of the test which breaks reporting. If we were to take these tests I think they would need to be in a new type.

PR #1206, opened 2020-02-25
update init scripts

ooooolivia:
The new version of ```init.ps1``` and ```init.sh``` would work in the following way:

- To use the default dotnet: ```./init.ps1```
- To install dotnet from a specific channel and use the installed dotnet: ```./init.ps1 -InstallDotnetFromChannel <channel> ```
- To use a specific dotnet source: ```./init.ps1 -DotnetDirectory <path>```

All three modes will set up basic environment variables such as ```PYTHONPATH```, ```DOTNET_MULTILEVEL_LOOKUP```

### DIFF
--- a/src/scenarios/init.ps1
+++ b/src/scenarios/init.ps1
@@ -1,9 +1,23 @@
+[CmdletBinding(PositionalBinding=$false)]
 Param(
-    [string] $InstallDotnetFromChannel,
+    [string] $Channel,
     [string] $DotnetDirectory
 )
 
-If ($DotnetDirectory -eq ""){
+function Print-Usage(){
+    Write-Host "Invalid argument. Usage:"
+    Write-Host ".\init.ps1"
+    Write-Host ".\init.ps1 -DotnetDirectory <custom dotnet directory>"
+    Write-Host ".\init.ps1 -Channel <channel to download new dotnet>"
+    Exit 1
+}
+
+# Parse arguments
+If (($Channel -ne "") -and ($DotnetDirectory -ne "")) {
+    Print-Usage
+}
+
+If ($DotnetDirectory -eq "" ){
     $DotnetDirectory = Join-Path $PSScriptRoot '..\..\tools\dotnet\x64'
 }
 
@@ -11,13 +25,11 @@ If ($DotnetDirectory -eq ""){
 $scripts = Join-Path $PSScriptRoot '..\..\scripts' -Resolve
 $env:PYTHONPATH="$scripts;$PSScriptRoot"
 
-$channel = $InstallDotnetFromChannel
-
-If (($channel -ne "")){
+If (($Channel -ne "")){
     # Download dotnet from the specified channel
     Write-Host "Downloading dotnet from channel $channel"
     $dotnetScript= Join-Path "$scripts" 'dotnet.py' -Resolve
-    python $dotnetScript install --channels $channel
+    python $dotnetScript install --channels $channel -v
     If (!$?){
         Write-Host "Dotnet installation failed."
         Exit 1
--- a/src/scenarios/init.ps1
+++ b/src/scenarios/init.ps1
@@ -1,24 +1,45 @@
 Param(
-    $Channel = 'master' # default is master
+    [string] $InstallDotnetFromChannel,
+    [string] $DotnetDirectory
 )
 
+# Add scripts and current directory to PYTHONPATH
 $scripts = Join-Path $PSScriptRoot '..\..\scripts' -Resolve
 $env:PYTHONPATH="$scripts;$PSScriptRoot"
-$dotnetScript= Join-Path "$scripts" 'dotnet.py' -Resolve
-$dotnetDirectory = Join-Path $PSScriptRoot '..\..\tools\dotnet\x64'
 
-Write-Host "Downloading dotnet from channel $Channel"
-python $dotnetScript install --channels $Channel -v
+If (($InstallDotnetFromChannel -ne "") -and ($DotnetDirectory -eq "")){
+    # Download dotnet from the specified channel
+    
+    $installDirectory= Join-Path $PSScriptRoot 'dotnet'
+    # Remove existing dotnet directory to make sure we only have one version of dotnet
+    If (Test-Path $installDirectory){
+        Write-Host "Removing $installDirectory ..."
+        Remove-Item  $installDirectory -Recurse
+    }
 
-if (!$?){
-    Write-Host "Dotnet installation failed."
-    Exit 1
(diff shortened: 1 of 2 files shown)


src/scenarios/init.ps1
```diff
@@ -1,24 +1,45 @@
 Param(
-    $Channel = 'master' # default is master
+    [string] $InstallDotnetFromChannel,
```
billwert: I think you can leave this as `Channel` - lot to type otherwise? (Though autocomplete helps some.)

src/scenarios/init.ps1
```diff
@@ -1,24 +1,45 @@
...
+        Write-Host "Removing $installDirectory ..."
+        Remove-Item  $installDirectory -Recurse
+    }
 
-if (!$?){
-    Write-Host "Dotnet installation failed."
-    Exit 1
+    Write-Host "Downloading dotnet from channel $InstallDotnetFromChannel"
+    $dotnetScript= Join-Path "$scripts" 'dotnet.py' -Resolve
+    python $dotnetScript install --channels $InstallDotnetFromChannel --install-dir $installDirectory
```
billwert: I would not pass `--install-dir` here. Not passing it just uses the default location in the repo (which is `tools\dotnet\<arch>`. Having a separate copy here means a) it shows up as added files in git and b) we have random executable things that need cleanup outside of `tools` or `artifacts`.

src/scenarios/init.ps1
```diff
@@ -1,24 +1,45 @@
...
+
+    If (!$?){
+        Write-Host "Dotnet installation failed."
+        Exit 1
+    }
+}
+
+If ($DotnetDirectory -ne ""){
```
billwert: You could simplify this by setting `$DotnetDirectory` based on what the user chose and then just add that to the path always.

src/scenarios/init.ps1
```diff
@@ -1,24 +1,45 @@
...
+    $env:DOTNET_ROOT=$installDirectory
 }
 
 $env:DOTNET_CLI_TELEMETRY_OPTOUT='1'
 $env:DOTNET_MULTILEVEL_LOOKUP='0'
 $env:UseSharedCompilation='false'
-$env:DOTNET_ROOT=$dotnetDirectory
-$env:Path="$dotnetDirectory;$env:Path"
+
+dotnet --info
```
billwert: What's the value in this output?

src/scenarios/init.ps1
ooooolivia: Sounds right I'll keep it

billwert:
All the comments in the ps1 apply to the .sh also.

src/scenarios/init.ps1
ooooolivia: Cool I didn't know it can download directly to ```tools``` folder.

src/scenarios/init.ps1
ooooolivia: It's intended to verify if the right ```dotnet``` is used.

src/scenarios/init.ps1
ooooolivia: I'll still keep ```-InstallDotnetFromChannel``` as a command line argument? This better emphasizes we need to download new dotnet.

billwert: If you think that's the better command line argument that's fine. I was suggesting that the argument could just be $Channel. Simply adding a new $Channel variable here isn't super useful.

src/scenarios/init.ps1
```diff
@@ -1,24 +1,34 @@
...
+$channel = $InstallDotnetFromChannel
 
-if (!$?){
-    Write-Host "Dotnet installation failed."
-    Exit 1
+If (($channel -ne "")){
+    # Download dotnet from the specified channel
+    Write-Host "Downloading dotnet from channel $channel"
+    $dotnetScript= Join-Path "$scripts" 'dotnet.py' -Resolve
+    python $dotnetScript install --channels $channel
```
billwert: Can you add back the -v here? I think this is a more informative way to a) let the user know something happened and b) show which version of .NET is being used.

DrewScoggins:
What should the behavior of the scripts be if I pass it both a channel name and a dotnet path? As it stands it will just use the path, but also install dotnet. Is that the behavior that we want?

ooooolivia:
> What should the behavior of the scripts be if I pass it both a channel name and a dotnet path? As it stands it will just use the path, but also install dotnet. Is that the behavior that we want?

Good point! I think the script should only parse the first argument and ignore the rest, because using 2 arguments are way more confusing. What do you think ? @billwert

src/scenarios/init.ps1
ooooolivia: I'm not against using $Channel since I also assume the user knows what it means.

billwert:
I think if both are specified you just print an error. Be unambiguous.

src/scenarios/init.ps1
```diff
@@ -1,24 +1,46 @@
...
 }
 
+$env:Path="$DotnetDirectory;$env:Path"
+$env:DOTNET_ROOT=$DotnetDirectory
 $env:DOTNET_CLI_TELEMETRY_OPTOUT='1'
 $env:DOTNET_MULTILEVEL_LOOKUP='0'
 $env:UseSharedCompilation='false'
-$env:DOTNET_ROOT=$dotnetDirectory
-$env:Path="$dotnetDirectory;$env:Path"
+
```
billwert: I think the script should only manipulate the environment if we installed a dotnet or if the user specified a dotnet directory. It's reasonable in lots of cases (such as when we are doing development) to just use the ambient state. Or the user wants to just use the thing that came with VS and check it out.

ooooolivia: reasonable

src/scenarios/init.ps1
```diff
@@ -1,24 +1,48 @@
...
+If (($Channel -ne "") -and ($DotnetDirectory -ne "")) {
+    Print-Usage
+}
+ElseIf ($DotnetDirectory -ne ""){
+    Setup-Env -directory $DotnetDirectory
+}
+ElseIf ($Channel -ne "") {
+    $DotnetDirectory = Join-Path $PSScriptRoot '..\..\tools\dotnet\x64'
+    Setup-Env -directory $DotnetDirectory
+    Download-Dotnet -channel $Channel
```
billwert: I would reverse these two lines. I suspect it's fine but I don't know what weird things might happen if we set the variables while the install is happening? Might be too paranoid on my part but..

ooooolivia: Good catch I intended the reversed one.

PR #1211, opened 2020-02-26
Add serializer tests for value types

steveharter:


src/benchmarks/micro/libraries/System.Text.Json/Serializer/ReadJson.cs
```diff
@@ -24,6 +24,9 @@ namespace System.Text.Json.Serialization.Tests
     [GenericTypeArguments(typeof(HashSet<string>))]
     [GenericTypeArguments(typeof(ArrayList))]
     [GenericTypeArguments(typeof(Hashtable))]
+    [GenericTypeArguments(typeof(SimpleStructWithProperties))]
+    [GenericTypeArguments(typeof(LargeStructWithProperties))]
+    [GenericTypeArguments(typeof(int))]
```
adamsitnik: the code looks good to me, but I just wonder if it is a real use case: (de)serializing a single integer.

steveharter: It is there primarily to detect general front-end overhead of the serializer without any extra baggage.

However, one [potential scenario](https://github.com/dotnet/runtime/issues/1562#issuecomment-581512105) is enumerating over a list of elements and calling serialize on each one.

However I don't recommend re-entering the serializer for each element -- first there is overhead of a dictionary lookup in the front-end to get the converter for each element and second because currently the "JsonPath" exception functionality doesn't work on re-entry (we need to add a new API to pass that state).

The better way is to write a custom converter for that collection which in turn would forward to a cached element converter. However, it may be that the list of element types are not known (base type of System.Object for example), so the dictionary lookup may need to occur per element anyway in order to find the converter for each element.

adamsitnik: Thank you for the explanation. If this provides value to you then I am merging it.

src/benchmarks/micro/Serializers/DataGenerator.cs
```diff
@@ -106,6 +109,16 @@ private static IndexViewModel CreateIndexViewModel()
...
             };
 
+        private static LargeStructWithProperties CreateLargeStructWithProperties()
+            => new LargeStructWithProperties
+            {
+                String1 = "1",
+                String2 = "2",
+                String3 = "3",
+                String4 = "4",
+                String5 = "5",
```
ahsonkhan: no initialization of the int properties?

PR #1217, opened 2020-03-02
Make high_memory GC tests stable and more consistent

ivdiazsa:
These changes help the _high_memory_ scenario from the GC Benchmarking Infra to be more stable and consistent between runs. The main idea of this scenario is to load the machine's memory to the specified percent in the _yaml_ file, and then run the GC tests.

The problem was that memory loads tend to be quite variable and the tool had almost no tolerance for these changes and would fail. Now, if the required memory load is not achieved but we are within an acceptable tolerance threshold (currently set to &#177;5%), the tool will just issue a warning and allow the tests to run.

Also, some refactoring was made and messages were updated to be more informative and easier to read (both for the tool and debugging).

### DIFF
--- a/src/benchmarks/gc/src/exec/env/make_memory_load.c
+++ b/src/benchmarks/gc/src/exec/env/make_memory_load.c
@@ -403,6 +403,8 @@ int main(const int argc, char** argv)
         if (ak == adjust_kind_no_adjust)
         {
             // No need to adjust as we got the memory load we required.
+            // Do NOT change this message, as run_single_test.py is expecting it
+            // to know everything went fine here.
             fprintf(stderr, "make_memory_load finished starting up\n");
             break;
         }
@@ -414,7 +416,7 @@ int main(const int argc, char** argv)
             // which is later parsed by run_single_test.py.
             fprintf(
                 stderr,
-                "threshold memory achieved, %.5f\n",
+                "threshold memory achieved,%.5f\n",
                 get_memory_used_fraction_from_mem(get_memory_status()) * 100.0
             );
             break;
--- a/src/benchmarks/gc/src/exec/env/make_memory_load.c
+++ b/src/benchmarks/gc/src/exec/env/make_memory_load.c
@@ -93,7 +93,7 @@ static void print_memory_status()
     puts("\n=======================");
     puts("Memory Status Achieved:");
     puts("=======================");
-    printf("Memory load achieved: %d%%\n",
+    printf("Memory Load Achieved: %d%%\n",
            (int) round(100 * get_memory_used_fraction_from_mem(mem)));
     printf("Total Physical Memory: %f GB\n",     bytes_to_gb(mem.total_physical_bytes));
     printf("Available Physical Memory: %f GB\n", bytes_to_gb(mem.available_bytes));
--- a/src/benchmarks/gc/src/exec/env/make_memory_load.c
+++ b/src/benchmarks/gc/src/exec/env/make_memory_load.c
@@ -276,7 +276,7 @@ static AdjustKind adjust(Mem* mem, int iteration_num)
     if (to_allocate != 0 && iteration_num == MEMORY_LOAD_NUM_ATTEMPTS)
     {
         double acc_mem_delta =   ((double) mem->desired_available_bytes)
-                               * ((double) ACCEPTABLE_MEMORY_DELTA_PCT / 100);
+                               * ((double) ACCEPTABLE_MEMORY_DELTA_PCT / 100.0);
         double current_delta = fabs(
             ((double) current_available_bytes) - ((double) mem->desired_available_bytes)
         );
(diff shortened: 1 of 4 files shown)


ivdiazsa:
@Maoni0 @VSadov @cshung

src/benchmarks/gc/src/exec/env/make_memory_load.c
```diff
@@ -105,7 +106,9 @@ typedef struct Args
     bool no_readjust;
 } Args;
 
-const char* USAGE = "Usage: make_memory_load.exe -percent 50 [-neverRelease]\n";
+// TODO: Add Verbose/Debug option to show all the stats while running without
```
VSadov: If you think it is worth to follow up on this, there should be a bug/workitem instead of TODO. 
If you change your mind later, the bug could be closed without touching this file.

src/benchmarks/gc/src/exec/env/make_memory_load.c
```diff
@@ -329,38 +377,68 @@ int main(const int argc, char** argv)
...
+            // No need to adjust as we got the memory load we required.
+            // Do NOT change this message, as run_single_test.py is expecting it
+            // to know everything went fine here.
+            fprintf(stderr, "make_memory_load finished starting up\n");
             break;
         }
-        else if (attempts == 5)
+        else if (ak == adjust_kind_within_threshold)
         {
+            // No need to adjust as we are within the acceptable memory load threshold.
```
VSadov: >// No need to ad [](start = 12, length = 16)

Should we attempt to get closer to the target if we still have attempts left? 
It is more likely that we will get closer to the target than that we get out of the threshold range, so seems like a good gamble to try.

Maoni0:
> The problem was that memory loads tend to be quite variable

can you please elaborate more? is it just because we are a few bytes off when we adjust? 'cause we shouldn't be using the # of bytes, we should really just be using the percentage which is more coarse and is what GC is looking at anyway.

src/benchmarks/gc/src/exec/env/make_memory_load.c
ivdiazsa: I had not thought about it, but I agree it's the better option. I've removed this comment and open a new issue to track this task: https://github.com/dotnet/performance/issues/1225

src/benchmarks/gc/src/exec/env/make_memory_load.c
ivdiazsa: This case can only occur in such a case. The function `adjust()` can only return `adjust_kind_within_threshold` if we are at the last attempt.

ivdiazsa:
> can you please elaborate more? is it just because we are a few bytes off when we adjust? 'cause we shouldn't be using the # of bytes, we should really just be using the percentage which is more coarse and is what GC is looking at anyway.

Looking at the code of `adjust()`, the tool defines whether memory is ready by doing some sort of rounding against the machine's _allocation granularity_:

``` c
const ptrdiff_t to_allocate =
    round_to_nearest((ptrdiff_t) current_available_bytes - (ptrdiff_t) mem->desired_available_bytes,
                      mem->allocation_granularity);
```

I require more explanation as to how _allocation granularity_ works. From what I've read, it exists to keep memory allocations uniform? It is defined to be 64KB on most machines, but this function will only return 0 (i.e. what we need to stop adjusting), when we are at 32KB or less from the desired available bytes (e.g. 20% in the case of the 80% load test).

Back to the topic at hand, depending on the value of `to_allocate`, the tool defines whether it needs to commit more memory, or reset some and continue from there.

I do not know if Andy had specific reasons to go this route, rather than always checking the memory percentage (maybe to keep memory allocations uniform in size?). That's the reason I took the cautious approach and only allowed the test to run by checking percentage on the last attempt.

src/benchmarks/gc/src/exec/env/make_memory_load.c
ivdiazsa: This only occurs in such a case. The `adjust()` function can only return `adjust_kind_within_threshold` if we are at the last attempt.

Maoni0:
we don't need to go to the alloc granularity. this tool takes a memory percentage and using that to check would be sufficient. you can safely change it to checking for percentage instead. that should reduce the volatility you saw. unless we are in some sort of wildly changing environment (which is almost never the case when you are running a benchmark) you should rarely ever need to retry.

ivdiazsa:
I see now why Andy developed _make_memory_load.c_ this way. He is writing memory by pages and through the use of VirtualAlloc. The following code snippet illustrates this:

``` c
const size_t size = min(mem->total_memory - mem->total_memory_committed, (size_t)to_allocate);

byte* const start = mem->memory + mem->total_memory_committed;
VirtualAlloc(start, size, MEM_COMMIT, PAGE_READWRITE);
write_to_every_page(start, size, mem->page_size);
mem->total_memory_committed += size;
```

This is executed in the case where we need to load more memory. Now, here is how pages are written:

```c
static void write_to_every_page(byte* start, size_t size, size_t page_size)
{
    for (size_t i = 0; i < size; i += page_size)
    {
        start[i] = (byte)(i % 256);
    }
}
```

I believe keeping this would be for the best since it allows for more precision and uniformity in the tool and with my changes, we are already doing the initial run by checking percentage.

Maoni0:
> What we could do is to let the tool keep writing memory this way, but determine the memory loading success entirely by checking percentage.

that was exactly what I meant. there's no need to change how the memory is committed/released. just the way to check. the way he was checking is unnecessarily granular.

ivdiazsa:

PR #1218, opened 2020-03-03
Benchmarks for Math.Min/Max double/float overloads

gfoidl:
See https://github.com/dotnet/runtime/issues/33057#issuecomment-593564840 for reference.


src/benchmarks/micro/runtime/Math/Functions/Double/MaxDouble.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        private const double maxDelta = 0.0004;
+        private const double maxExpectedResult = -1.0;
+
+        [Benchmark]
+        public void Max() => MaxTest();
+
+        public static void MaxTest()
+        {
+            var result = 0.0; var val1 = -1.0; var val2 = -1.0 - maxDelta;
```
adamsitnik: I know that the existing benchmarks use `var` a lot, but I think that it would be better to use explicit type.

```suggestion
            double result = 0.0, val1 = -1.0, val2 = -1.0 - maxDelta;
```

src/benchmarks/micro/runtime/Math/Functions/Double/Acosh.cs
```diff
@@ -35,4 +35,4 @@ public static void AcoshTest()
             }
         }
     }
-}
```
adamsitnik: nit: please try to avoid changing end of the file (it adds a noise to the code review, but does not provide too much value)

adamsitnik:
> Could you please apply the suggested fixes before we merge this PR?

@gfoidl nvm, I was able to do it myself

src/benchmarks/micro/runtime/Math/Functions/Double/Acosh.cs
gfoidl: Sorry, will keep that it mind the next time.

gfoidl:
> nvm, I was able to do it myself

Great that this works.

I used `var` as the other benchmarks also use `var` instead of the concrete type (which I would prefer). Shall I update the other benchmarks to not use `var` (in a different PR) or leave them as they are?

adamsitnik:
> Shall I update the other benchmarks to not use var (in a different PR)

Yes, please!

gfoidl:
> Yes, please!

Will do during the rest of the week.

PR #1224, opened 2020-03-05
explain possible side effects of using var and implicit casting in the docs

adamsitnik:


docs/microbenchmark-design-guidelines.md
```diff
@@ -562,3 +563,29 @@ By relying on the BDN mechanism you are going to avoid loop alignment issues. Be
...
+```
+
+In extreme cases, we might be measuring the performance of a wrong method. In the following example, we measure the performance of `Math.Max` overload that accepts doubles, not floats (this is what we wanted).
+
+```cs
+var x = 1.0f; var y = 2.9; // x is float, y is double (missing f)
+var result = Math.Max(x, y); // we use double overload because float has implicit cast to double
+```
+
+It's our responsibility to ensure that the benchmarks do exactly what we want. This is why using explicit types over `var` is preffered.
```
billwert:
```suggestion
It's our responsibility to ensure that the benchmarks do exactly what we want. This is why using explicit types over `var` is preferred.
```

PR #1240, opened 2020-03-17
Enable Linux tracing in Startup 

ooooolivia:
This is a prototype of the new Startup which enables both Windows and Linux tracing. The following components are added:
- ```TraceSessionManager.cs```: a static class that emits platform-specific trace session
- ```ITraceSession```: an interface that specifies the abstraction of platform-specific trace sessions
- ```LinuxTraceSession.cs```: linux trace session powered by perfcollect
- ```WindowsTraceSession.cs```: windows trace session powered by TraceEvent
- ```Perfcollect.cs```: wrapper around perfcollect
The status of the PR is till


src/tools/ScenarioMeasurement/Util/Logger.cs
```diff
@@ -15,6 +15,16 @@ public void Log(string message)
...
+        public void LogHeader1(string message)
+        {
+            Console.WriteLine($"=============== {message} ================ ");
+        }
+
+        public void LogHeader2(string message)
+        {
+            Console.WriteLine($"***{message}***");
+        }
+
```
billwert: 1) These names are not great, doesn't help explain what they do.
2) Instead of methods like this that hard code some number of = and *, why not a method that takes the string, defaults to a char to surround with, and a width? 
3) Do we really even need headers like this in the output? Maybe we ditch them.

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
```diff
@@ -0,0 +1,119 @@
+﻿using Microsoft.Diagnostics.Tracing.Parsers;
+using ScenarioMeasurement;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+
+namespace ScenarioMeasurement
+{
+    public class Perfcollect : IDisposable
```
billwert: nit: should be PerfCollect

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        private readonly string filepath = Path.Combine(Directory.GetParent(Directory.GetCurrentDirectory()).FullName, @"Startup/perfcollect");
+        private ProcessHelper perfcollectProcess;
```
billwert: Can you guarantee all uses of this tool will be in the directory you expect here?

billwert: nit: perfCollectProcess

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
```diff
@@ -0,0 +1,119 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+
+namespace ScenarioMeasurement
+{
+    public class Perfcollect : IDisposable
+    {
+        private readonly string filepath = Path.Combine(Directory.GetParent(Directory.GetCurrentDirectory()).FullName, @"Startup/perfcollect");
```
billwert: nit: instead of ```filepath```, which should be ```filePath```, maybe ```perfCollectScript```

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
```diff
@@ -0,0 +1,119 @@
...
+            return perfcollectProcess.Run().Result;
+        }
+
+        public void Dispose()
+        {
+            Stop();
+        }
+
+        public enum KernelKeyword{ 
+            Empty,
```
billwert: nit: space before ```{```

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        public ProcessHelper.Result Start()
+        {
+            string arguments = $"start {TraceName} -events ";
+            
+            foreach (var keyword in KernelEvents)
+            {
+                arguments += Enum.GetName(typeof(KernelKeyword), keyword)+",";
+            }
```
billwert: Why not just call keyword.ToString()?

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            foreach (var keyword in ClrEvents)
+            {
+                arguments += Enum.GetName(typeof(ClrKeyword), keyword)+",";
+            }
+
+            arguments.TrimEnd(',');
+
+            perfcollectProcess.Arguments = arguments;
+            return ProcessHelper.Result.Success;
```
billwert: I assume it is point in time that this does not actually start the process.

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
```diff
@@ -0,0 +1,77 @@
...
+        private Dictionary<TraceSessionManager.KernelKeyword, KernelTraceEventParser.Keywords> winKwMapKernel;
+        private Dictionary<TraceSessionManager.ClrKeyword, ClrPrivateTraceEventParser.Keywords> winKwMapClr;
+
+        public WindowsTraceSession(string sessionName, string traceName, string traceDirectory, Logger logger)
+        {
+            string fileName = Path.ChangeExtension(traceName, "perflab.etl");
+            string filePath = Path.Combine(traceDirectory, fileName);
+            // Currently kernel and user share the same session
+            session = new TraceEventSession(sessionName, filePath);
+            InitWindowsKeywordMaps();
```
billwert: I believe we decided to split these apart, yes?

src/tools/ScenarioMeasurement/Util/Logger.cs
ooooolivia: Good suggestion! 
1. I agree, but I couldn't think of a better naming. Any suggestion?
2. It seems easier to me to call ```logger.LogHeader1(<message>)```  than to call ```logger.LogHeader(<number of symbols>, <width>, <message>)``` multiple times?? 
3. Maybe it's just my personal feeling but the headers help me identify important info out of a great deal of texts very well, just like how BenchmarkDotNet renders its layout. I hope it's not a personal preference thing so I'll ask @DrewScoggins  what do you think?

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
ooooolivia: Hmm I'm not sure.. Currently only Startup tool downloads and uses it, so the script is in that project folder. Are you suggesting we should move it out so other projects can use it? Or the script should go to wherever the working directory is set to in Startup?

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
ooooolivia: Great.

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
ooooolivia: Didn't know that. Definitely better than GetName

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
ooooolivia: Good catch! I left it there for testing and forgot to revert it.

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
ooooolivia: Yes. Update is coming soon.

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
ooooolivia: hmmm just got what you mean. Maybe the path should be passed in.

src/tools/ScenarioMeasurement/Util/Logger.cs
billwert: Can use default parameters for width and character. Or just hard code it to be some fixed width and put the string in the middle.

ooooolivia: Just want to clarify before I make the change, it is to get rid of the confusing ```LogHeader``` methods?

billwert: Yes. I think we should not have `LogHeader1` and `LogHeader2`. Either define what makes them different (`LogHeader`, `LogWhatever`) or collapse them to one method. What do we use `=` and `*` to indicate? Do we need both?

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace ScenarioMeasurement
+{
+    class LinuxTraceSession : ITraceSession
+    {
+        private PerfCollect perfCollect;
+        private Dictionary<TraceSessionManager.KernelKeyword, PerfCollect.KernelKeyword> linKwMapKernel;
+        private Dictionary<TraceSessionManager.ClrKeyword, PerfCollect.ClrKeyword> linKwMapClr;
```
billwert: We know these are for Linux because they're in the `LinuxTraceSession` and we know it's a map because of the type. In general we do not abbreviate variables like this, and we avoid encoding type information.
```suggestion
        private Dictionary<TraceSessionManager.KernelKeyword, PerfCollect.KernelKeyword> kernelKeywords;
        private Dictionary<TraceSessionManager.ClrKeyword, PerfCollect.ClrKeyword> clrKeywords;
```

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
```diff
@@ -0,0 +1,116 @@
...
+        public WindowsTraceSession(string sessionName, string traceName, string traceDirectory, Logger logger)
+        {
+            this.traceName = traceName;
+            this.traceDirectory = traceDirectory;
+            this.logger = logger;
+
+            string kernelFileName = Path.ChangeExtension(traceName, "perflabkernel.etl");
+            string userFileName = Path.ChangeExtension(traceName, "perflabuser.etl");
+
+            // Currently kernel and user share the same session
```
billwert: Remove the stale comment?

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
```diff
@@ -0,0 +1,65 @@
...
+            linKwMapKernel[TraceSessionManager.KernelKeyword.ContextSwitch] = PerfCollect.KernelKeyword.ContextSwitch;
+
+            // initialize linux clr keyword map
+            linKwMapClr = new Dictionary<TraceSessionManager.ClrKeyword, PerfCollect.ClrKeyword>();
+            linKwMapClr[TraceSessionManager.ClrKeyword.Startup] = PerfCollect.ClrKeyword.DotNETRuntimePrivate_StartupKeyword;
+        }
+
+        public void EnableUserProvider(string provider)
+        {
+        }
```
billwert: How does enabling an `EventSource` work on Linux? I recall there's something about them getting funneled through the CLR provider somehow but I'm not sure what enables them. @brianrob may have advice here.

ooooolivia: My understanding is, you set the environment variable ```COMPlus_PerfMapEnabled=1``` and ```COMPlus_EnableEventLog=1``` to enable tracing config in CoreCLR. And enable the corresponding user-space events in lttng, which could be specified as command line arguments. I assume the provider information is already taken care by coreclr, no need to specify it in the tool. @brianrob  correct me if I'm wrong.

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
ooooolivia: good catch!

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
ooooolivia: Yes I struggled with coming up a descriptive but concise name. Good to know!

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            string traceName = "test-windows-trace";
+            using(var session = new WindowsTraceSession(sessionName, traceName, traceDirectory, logger))
+            {
+                var parser = new ProcessTimeParser();
+                session.EnableProviders(parser);
+                RunTestIteration();
+            }
+        }
+
+        [Fact]
```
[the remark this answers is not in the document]
ooooolivia: Good point! Will add this.

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
brianrob: For `EventSources` specifically, there are a few things that you need to know:

1. If you're not capturing `perf` profiles, you don't need to set `COMPlus_PerfMapEnabled=1`.  This will just do some extra work that you won't need.
2. You do need to set `COMPlus_EnableEventLog=1`.  This turns on the ability for the runtime to emit events to `LTTng`.
3. `EventSource` events are special when it comes to `LTTng`, because `LTTng` doesn't allow for new `providers` or `tracepoints` to be added on-the-fly.  Instead they all must be added at compile time.  So the runtime adds a `tracepoint` called `DotNETRuntime:EventSource` that is the single `tracepoint` for all `EventSource` events.  This also means that if you turn it on, then every `EventSource` event is enabled.  To filter this, there are a couple of environment variable you can set: `COMPlus_EventSourceFilter` will filter by `EventSource` name and `COMPlus_EventNameFilter` will filter by event name.  See https://github.com/dotnet/runtime/blob/70bfd2ac9b844701c361b90b90ef399ae6fec483/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/TraceLogging/XplatEventLogger.cs#L162 and https://github.com/dotnet/runtime/blob/70bfd2ac9b844701c361b90b90ef399ae6fec483/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/TraceLogging/XplatEventLogger.cs#L177 for how these are implemented.
4. Once you're able to collect these events, they will show up in `TraceEvent` with an event name of `EventSource` and the payload is JSON.  One thing that would be really nice is to have `TraceEvent` do the conversion so that they look like every other event in the system.  Note that I'm not 100% sure how complicated this will be, as there isn't a manifest, so we'd basically have to make up per-event metadata on-the-fly.  I completely understand if this turns out to be too much work and you want to punt on it.

src/tools/ScenarioMeasurement/Startup/Startup.csproj
```diff
@@ -18,4 +25,17 @@
...
+    </PropertyGroup>
+    <DownloadFile SourceUrl="$(PerfcollectUrl)" 
+                  DestinationFolder="$(MSBuildProjectDirectory)"
+                  Retries="3"
+                  SkipUnchangedFiles="true">
+      <Output TaskParameter="DownloadedFile" ItemName="Content"/>
+    </DownloadFile>
+    <Copy SourceFiles="perfcollect" DestinationFolder="$(PublishDir)" />
+    <Copy SourceFiles="perfcollect" DestinationFolder="$(OutDir)" />
+  </Target>
```
DrewScoggins: Can we also make this target dependent on non-Windows?

ooooolivia: I was worrying the publish won't always happen on Linux platform for Linux runs, so I didn't set the condition. But it seems the script task that runs the publish is platform-dependent already.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -151,34 +150,27 @@ static void checkArg(string arg, string name)
...
+                        failed = true;
+                        break;
                     }
+                    pids.Add(iterationResult.Pid);
                 }
             }
 
-            if (!failed)
+
+   /*         if (!failed)
```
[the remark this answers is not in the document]
ooooolivia: No, it was commented for test and is uncommented now.

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
```diff
@@ -0,0 +1,119 @@
...
+        public List<KernelKeyword> KernelEvents { get; set; }
+        public List<ClrKeyword> ClrEvents { get; set; }
+        public Perfcollect(string traceName, Logger logger) : this (traceName, Environment.CurrentDirectory, logger)
+        {
+        }
+
+        public Perfcollect(string traceName, string traceDirectory, Logger logger)
+        {
+            TraceName = traceName;
+/*            if (!File.Exists(filepath))
```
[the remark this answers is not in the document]
ooooolivia: Uncommented now.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            var processTimeParser = new ProcessTimeParser();
+            var profileParser = new ProfileParser(processTimeParser);
+            using(var profileSession = TraceSessionManager.CreateProfileSession(sessionName, traceName, traceDirectory, logger))
+            {
+                profileSession.EnableProviders(profileParser);
+                RunTestIteration();
+            }
+        }
+
+        private void RunTestIteration()
```
billwert: There's no assertion so I suppose this only fails if there is an exception?

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -0,0 +1,90 @@
...
+using Reporting;
+using ScenarioMeasurement;
+using System;
+using System.IO;
+using System.Resources;
+using Xunit;
+
+namespace Startup.Tests
+{
+    public class StartupTests
```
billwert: We talked about tests to validate that the dictionaries representing the keyword maps were properly filled out. Did you think about that?

src/tools/ScenarioMeasurement/Startup/Perfcollect.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                ProcessWillExit = true,
+                Executable = perfCollectScript,
+                Timeout = 300
+            };
+        }
+
+        public ProcessHelper.Result Start()
+        {
+            string arguments = $"start {TraceName} -events ";
```
billwert: It doesn't meaningfully matter here, but this is a good pattern to think about using `StringBuilder` for. Every time you append the string you end up allocating a new string. In high performance code that's definitely a pattern to avoid. Just FYI.

src/tools/ScenarioMeasurement/Startup/ProfileParser.cs
```diff
@@ -13,19 +13,21 @@ internal class ProfileParser : IParser
...
         }
 
-        public void EnableUserProviders(TraceEventSession user)
+        public void EnableUserProviders(ITraceSession user)
         {
-            // make sure we turn on whatever the user wanted so the start/stops are findable.
             other.EnableUserProviders(user);
-            user.EnableProvider(ClrTraceEventParser.ProviderGuid, Microsoft.Diagnostics.Tracing.TraceEventLevel.Verbose, (ulong)ClrTraceEventParser.Keywords.Default);
-            user.EnableProvider(ClrPrivateTraceEventParser.ProviderGuid, Microsoft.Diagnostics.Tracing.TraceEventLevel.Verbose, (ulong)(ClrPrivateTraceEventParser.Keywords.GC
+            TraceEventSession userSession = ((WindowsTraceSession)user).UserSession;
```
billwert: Does this work on Linux? I don't see where the profile iteration gets skipped or anything.

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
billwert: I would like to know the plan to get this implemented. I know that most of our current tests don't use event sources, but some of them do/will.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: Oops just found out it's not even running the iteration process. I'll add assertion to check if the etl file is generated.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: Almost forgot. Will add the validation to this PR.

src/tools/ScenarioMeasurement/Startup/ProfileParser.cs
ooooolivia: It is skipped in ```TraceSessionManager.CreateProfileSession()```. It might be safer if we check the platform again here, but I'm not sure if an ```IParser``` should know that.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -0,0 +1,99 @@
...
+            string traceFilePath = "";
+            using (session)
+            {
+                session.EnableProviders(parser);
+                result = RunTestIteration();
+                traceFilePath = session.GetTraceFilePath();
+            }
+            if (result != ProcessHelper.Result.Success)
+            {
+                throw new Exception("Test iteration failed.");
```
billwert: I think it is bad practice to throw here. Why not `Assert.True` the condition?

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+            Assert.True(!String.IsNullOrEmpty(traceFilePath) && File.Exists(traceFilePath));
```
billwert: You can have more than one assertion. In this formulation you don't know which failed. I would use an `Assert.False` on the first condition and make the second its own `Assert.True`

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: Can you elaborate on why we shouldn't have a ```throw``` here? My thinking is, the failure of the child process should not be equivalent as the failure of the test, because the child process is not we are testing and we should only assert on what we are testing.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: Not blaming it but I saw this use of combined logic in TraceEvent library so I copied that. Learned something bad.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: Thanks for your advice! Just knew we don't usually throw exceptions in tests.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
billwert: This is specific advice for tests. Having assertions be very granular makes the diagnosis of failures easier because the failure can only mean one thing. In normal code this would have been just fine.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
billwert: IMO anything the test does is "part of the test", so anything it does should be a failure. I would make very sure the child process here is basically never going to fail when I wrote the test and then have this assertion just to protect myself if that became not true for some reason.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        private void TestSession(ITraceSession session, IParser parser)
+        {
+            ProcessHelper.Result result;
+            string traceFilePath = "";
+            using (session)
+            {
+                session.EnableProviders(parser);
+                result = RunTestIteration();
```
DrewScoggins: Do we even need to make a call to a sub-process here? We are not checking to see if that process shows up in the file, just to see if enabling providers works and we get something. Why not just put a sleep? That seems like it will make the test more robust and still test the same thing.

billwert: Totally reasonable. We're indirectly getting test coverage of the process starting goo, which we could move to being an explicit test instead if we wanted.

I doubt we'll ever write tests that look for specific events - that's what the perf tests will ultimately do. So just ensuring that files get produced is not a bad idea.

ooooolivia: Cool! I was thinking ```dotnet --info``` is a bit closer to the real scenario... but less reasonable if we are just testing the trace sessions. Checking individual events will probably be implemented in tests of the kernel parser wrapper that will be added to Startup soon.

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
ooooolivia: Let me clarify, you mean ultimately we'd like to enable other custom event sources just like what we do in [GenericStartupParser](https://github.com/dotnet/performance/blob/afdf541e36060c0e38af3bd7737a9c977be09b71/src/tools/ScenarioMeasurement/Startup/GenericStartupParser.cs#L11-L16) ? I think this suggests we need to make significant changes to perfcollect. Currently it only allows Kernel and Clr providers.

src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
```diff
@@ -0,0 +1,136 @@
...
+                arguments.Append(",");
+            }
+
+            foreach (var keyword in ClrEvents)
+            {
+                arguments.Append(keyword.ToString());
+                arguments.Append(",");
+            }
+
+            string args = arguments.ToString().TrimEnd(',');
```
billwert: just FYI: Since you went to the trouble to use `StringBuilder`, the final `.TrimEnd()` here will end up allocating an extra string you don't need. Instead I would so something like `.Remove(arguments.Length-1,1)`  before you call `ToString()`. When you're using a `StringBuilder` you should do all your editing and manipulation on the `StringBuilder` instance and then think of calling `ToString()` as "finalizing" the string.

src/tools/ScenarioMeasurement/Startup/Startup.csproj
```diff
@@ -8,6 +8,13 @@
     <LangVersion>Preview</LangVersion>
   </PropertyGroup>
 
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
+      <_Parameter1>Startup.Tests</_Parameter1>
+    </AssemblyAttribute>
```
billwert: what's left that's internal that we need this for? Should we just make it public? What about using Moq or something similar?

src/tools/ScenarioMeasurement/Startup/Startup.csproj
```diff
@@ -18,4 +25,17 @@
     <ProjectReference Include="..\..\Reporting\Reporting\Reporting.csproj" />
   </ItemGroup>
 
+  <Target Name="DownloadPerfCollect" BeforeTargets="CoreCompile" Condition=" '$(OS)' != 'Windows_NT' ">
```
billwert: >'$(OS)' != 'Windows_NT' [](start = 77, length = 23)

Is this a standard msbuild property for .NET Core or is this set by Helix somewhere?

src/tools/ScenarioMeasurement/Startup/TraceSessionManager.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                return new WindowsTraceSession(sessionName, traceName, traceDirectory, logger);
+            }
+            else
+            {
+                return new LinuxTraceSession(sessionName, traceName, traceDirectory, logger);
+            }
+        }
+
+        public static ITraceSession CreateProfileSession(string sessionName, string traceName, string traceDirectory, Logger logger)
```
billwert: >CreateProfileSession [](start = 36, length = 20)

Three questions:
1) Why does this need to throw? Why can't it just print a message? It doesn't seem like we should fail a helix run (or show a user a nasty exception) in this case if we specify something wrong.
2) What is actually stopping this from getting called on Linux so it won't throw? Is the plan that the script will be special cased to pass in `--skip-profile-iteration` on Linux?
3) Do we have this extra method only to be able to special case the platform? What else makes a "profile" session different from a normal session?

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public void EnableUserProvider(string provider)
+        {
+            UserSession.EnableProvider(provider);
+        }
+
+        public string GetTraceFilePath()
+        {
+            return traceFilePath;
+        }
```
billwert: Why is this a method instead of a property?

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
```diff
@@ -0,0 +1,116 @@
...
+            var files = new List<string>();
+            if (File.Exists(kernelTraceFile))
+            {
+                files.Add(kernelTraceFile);
+            }
+            if (File.Exists(userTraceFile))
+            {
+                files.Add(userTraceFile);
+            }
+            if (files.Count != 0)
```
billwert: This construct suggests that one of the kernel or user files may be missing. That makes sense to me as there are cases (like the `ProcessTimeParser`) where there will be no user session. There will ALWAYS be a kernel session, though. Nothing really works without it. I think it's reasonable to fail if there's not a kernel session.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using Microsoft.Diagnostics.Tracing.Parsers;
-using Microsoft.Diagnostics.Tracing.Session;
 using Reporting;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Runtime.CompilerServices;
```
billwert: What uses these new usings? Not seeing it in the code immediately..

src/tools/ScenarioMeasurement/Startup/Startup.csproj
ooooolivia: This is a standard msbuild property so it works on client runs as well.

src/tools/ScenarioMeasurement/Startup/Startup.cs
ooooolivia: Yes it should be removed.

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
ooooolivia: Okay, so I will change it to throw exception upon no kernel trace file, and remove the merge condition.

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
ooooolivia: Misused some Java convention here. Will get rid of this method.

src/tools/ScenarioMeasurement/Startup/Startup.csproj
ooooolivia: All the parser classes are internal. Sure we can make it public. My understanding of them being set "internal" originally was to minimize their scope and prevent external assemblies from using them?

billwert: Eh, I don't think it's that principled. Ultimately there's no great reason for those types to be internal.

src/tools/ScenarioMeasurement/Startup/TraceSessionManager.cs
ooooolivia: Great questions!

1. Print does sound better, so it can just ignore the profile iteration without wasting the previous results of normal iterations.
2. Yes using ```--skip-profile-iteration``` is what I intended. If not specified in Linux runs, as mentioned about the program will print the message and exit 0.
3. Yes the profile session is separated just for the special case of Linux, and I think it is good to have checking platforms only in ```TraceSessionManager.cs```.

billwert: Got it.

I don't know how I feel about this, honestly. I think that having a special case for the profile iteration so we can flag linux is kind of strange. It's also true that some user mode traces don't work on linux (eventsource) right now, but we aren't doing magic about that. It's also hopefully true that at some point we WILL have Linux profiling support, and then we have this strange API that makes no sense. 

I think it's better to just put this in Startup.cs and set `skipProfileIteration=true` if we're on Linux, and not have a second API here. What do you think about that?

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            var parser = new TimeToMainParser();
+            TestSession(session, parser);
+        }
+
+        [WindowsOnly]
+        public void TestProfileIteration()
+        {
+            string sessionName = "test-profile-iteration-session";
+            string traceName = "test-profile-iteration-trace";
+            var timeToMainParser = new TimeToMainParser();
```
billwert: looks like this `timeToMainParser` is unused?

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
ooooolivia: Resolved offline. "Punting" this issue for now.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: It's passed in ```ProfileParser(IParser other)``` as a required parameter.

billwert: oh, duh. sorry. Merge!

ooooolivia:
Merging it now. Thanks for all the valuable comments!

PR #1241, opened 2020-03-18
Demo sdk new

ooooolivia:
To run sdk new console scenarios: 

- download this branch
- start powershell
- change to working directory: ```src\scenarios```
- run ```init.ps1``` to set up the environment
- run ```python test.py sdk sdk new_console``` to run the test

It will run ```dotnet new console``` and create console app in ```app``` folder with default number of iterations ( supposedly you can change the number of iterations in ```test.py```). For each iteration, it creates ```app``` folder before the iteration and delete that folder after the iteration. 

This is just a quick working draft for demo only.

### DIFF
--- a/src/scenarios/shared/sdk_iteration_setup.py
+++ b/src/scenarios/shared/sdk_iteration_setup.py
@@ -6,10 +6,10 @@ from performance.logger import setup_loggers
 from logging import getLogger
 from argparse import ArgumentParser
 
-BUILD_SETUP = 'setup_build'
-NEW_SETUP = 'setup_new'
+SETUP_BUILD = 'setup_build'
+SETUP_NEW = 'setup_new'
 CLEANUP = 'cleanup'
-operations = (BUILD_SETUP, NEW_SETUP, CLEANUP)
+operations = (SETUP_BUILD, SETUP_NEW, CLEANUP)
 
 def main():
     parser = ArgumentParser()
@@ -18,7 +18,7 @@ def main():
 
     setup_loggers(True)
 
-    if args.operation == BUILD_SETUP:
+    if args.operation == SETUP_BUILD:
         shutdown_server(verbose=True)
         if not os.path.isdir(const.TMPDIR):
             if not os.path.isdir(const.APPDIR):
@@ -31,7 +31,7 @@ def main():
             getLogger().info("Copying clean project directory...")
             shutil.copytree(const.TMPDIR, const.APPDIR) # use the copy
 
-    if args.operation == NEW_SETUP:
+    if args.operation == SETUP_NEW:
         if not os.path.isdir(const.APPDIR):
             getLogger().info("Creating new project directory...")
             os.mkdir(const.APPDIR)
--- a/src/scenarios/shared/sdk_iteration_setup.py
+++ b/src/scenarios/shared/sdk_iteration_setup.py
@@ -6,9 +6,10 @@ from performance.logger import setup_loggers
 from logging import getLogger
 from argparse import ArgumentParser
 
-SETUP = 'setup'
+BUILD_SETUP = 'setup_build'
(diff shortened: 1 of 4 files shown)


src/scenarios/shared/startup.py
```diff
@@ -16,12 +16,14 @@ class StartupWrapper(object):
...
-        elif os.path.exists(workitem_startup):
-            self._setstartuppath(workitem_startup)
+        if helixpayload():
+            payload_startup = os.path.join(helixpayload(), 'Startup')
+            if os.path.exists(payload_startup):
+                self._setstartuppath(payload_startup)
+        elif helixworkitempayload():
+            workitem_startup = os.path.join(helixworkitempayload(), 'Startup')
+            if os.path.exists(workitem_startup):
+                self._setstartuppath(workitem_startup)
```
billwert: Did you change this or did you not merge the previous change into whatever branch this is working off of?

ooooolivia: Lemme check. I did pull from the upstream master branch before working on it.

ooooolivia: The current code on master is not checking if the helixpayload or helixworkitempayload is null and breaks on client runs.

billwert: Aha. It only breaks locally, I guess, not in CI. I thought I tested it but I must not have. Sorry.

Either way, this change reintroduces the logic problem I was fixing. This will always take the first branch in Helix, because the variable is always set. 

```suggestion
        if helixpayload() and os.path.exists(os.path.join(helixpayload(), 'Startup'):
            self._setstartuppath(os.path.join(helixpayload(), 'Startup')
        elif helixworkitempayload() and os.path.exists(os.path.join(helixworkitempayload(), 'Startup')):
            self._setstartuppath(os.path.join(helixworkitempayload(), 'Startup'))
```

ooooolivia: Oh you're right ! I'll commit the suggestion right away.

billwert: I only know this because I wrote the bug the first time. :)

src/scenarios/shared/startup.py
```diff
@@ -16,12 +16,10 @@ class StartupWrapper(object):
...
-        payload_startup = os.path.join(helixpayload(), 'Startup')
-        workitem_startup = os.path.join(helixworkitempayload(), 'Startup')
-        if os.path.exists(payload_startup):
-            self._setstartuppath(payload_startup)
-        elif os.path.exists(workitem_startup):
-            self._setstartuppath(workitem_startup)
+        if helixpayload() and os.path.exists(os.path.join(helixpayload(), 'Startup'):
+            self._setstartuppath(os.path.join(helixpayload(), 'Startup')
+        elif helixworkitempayload() and os.path.exists(os.path.join(helixworkitempayload(), 'Startup')):
+            self._setstartuppath(os.path.join(helixworkitempayload(), 'Startup'))
```
billwert: Please test this. :) I wrote this in the github textbox.. I think it's right but worth making sure.

src/scenarios/shared/sdk_iteration_setup.py
```diff
@@ -6,9 +6,10 @@
 from logging import getLogger
 from argparse import ArgumentParser
 
-SETUP = 'setup'
+BUILD_SETUP = 'setup_build'
+NEW_SETUP = 'setup_new'
```
billwert: why are the values inverse of the variable names? Kind of confusing.

billwert:
And pending passing CI.

src/scenarios/shared/startup.py
ooooolivia: You're right I shouldn't have tested it with just my eyes :)

PR #1243, opened 2020-03-19
Fix build and execution for windows/arm64

kunalspathak:
When building performance for 

Currently micro benchmarks fails to build on windows/arm64 because of https://github.com/dotnet/roslyn/issues/42393. As a workaround,  we need to pass `/p:DebugType=portable` to build the micro benchmark project. Also, while running the micro benchmarks on windows/arm64 pass this parameter to build the autogenerated project. 

Ref - https://github.com/dotnet/BenchmarkDotNet/issues/1395

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -9,12 +9,11 @@
 
     <OutputType>Exe</OutputType>
     <PlatformTarget>AnyCPU</PlatformTarget>
-    <DebugType>pdbonly</DebugType>
+    <DebugType>portable</DebugType>
     <DebugSymbols>true</DebugSymbols>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 
     <IsARM Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().StartsWith(`arm`, System.StringComparison.OrdinalIgnoreCase))'">true</IsARM>
-    <DebugType Condition="'$(IsARM)' == 'true' AND '$(OS)' == 'Windows_NT'">portable</DebugType>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


billwert:
I wonder if we should just switch to portable PDBs entirely and not conditionalize this. @noahfalk - can you opine about about "classic" vs "portable" PDBs and how we should think about them?

kunalspathak:
Yeah, I was not sure about the compatibility with older netcoreapp / net framework versions if given portable. If @noahfalk  or someone else can confirm, I will go ahead with that.

noahfalk:
At this point I think of portable pdb format as a reasonable default for most use cases, and then use the classic format if you need to be back compatible with specific tools that have never added portable pdb support. There is [a symbol conversion tool](https://github.com/dotnet/symreader-converter) so that you can generate only portable pdb at build time and then later create the classic pdb when/if your use-case requires it. Last I recall .Net Core produces and uploads both formats to symbol server attempting to facilitate the largest number of use-cases. I'm not sure if symbol upload is something you bother with for these benchmark apps though.

There is a [doc](https://github.com/dotnet/core/blob/master/Documentation/diagnostics/portable_pdb.md) that describes use cases that are lacking portable pdb support but I believe it is out of date. @tmat usually has the most up-to-date info on support and maybe he could update the doc? Giving it a quick read I think the MSDIA case is now supported, and .Net reflector might be as well?

kunalspathak:
Alright, I will make the `MicroBenchmarks.csproj` build with `portable` symbols unless @billwert  or @adamsitnik think otherwise.

adamsitnik:
> Alright, I will make the MicroBenchmarks.csproj build with portable symbols unless @billwert or @adamsitnik think otherwise.

I don't mind switching to portable symbols

billwert:
Love it. Thanks!

Thanks,
-Bill

PR #1247, opened 2020-03-24
Feed cleanup

mmitche:
Switch away from blob feeds

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -5,8 +5,9 @@
   </solution>
   <packageSources>
     <clear />
-    <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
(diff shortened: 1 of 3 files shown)


adiaaida:
@DrewScoggins  any idea why this change would cause commandline.dragonfruit to not be found?

mmitche:
It's probably a random package that should get pushed to another feed. I'll figure it out.

billwert:
Do servicing branch packages for 3.1.x and 2.whatever appear in these "dotnet5" feeds you're adding?

mmitche:
> Do servicing branch packages for 3.1.x and 2.whatever appear in these "dotnet5" feeds you're adding?

No, but I can add those separate feeds.

DrewScoggins:
Is the dotnet-eng feed the 2.x feed?

mmitche:
> Is the dotnet-eng feed the 2.x feed?

No, the 2.x feed is dotnet-core. However, it's only ever pushed on release day, which means the non-transport packages are on nuget.org.

PR #1248, opened 2020-03-25
Added support for btl trace files to GC Benchmarking Infra

ivdiazsa:
Btl is another file format for traces that is fully supported by PerfView. GC Benchmarking Infra did not recognize it and therefore failed. This PR adds the necessary support to be able to analyze these traces as well.

### DIFF
--- a/src/benchmarks/gc/src/analysis/managed-lib/Analysis.cs
+++ b/src/benchmarks/gc/src/analysis/managed-lib/Analysis.cs
@@ -509,7 +509,7 @@ namespace GCPerf
             Util.Assert(source != null, $"PrintEventsWithoutTraceLog: Bad path {tracePath}.");
 
             uint n = 0;
-            source.AllEvents += (TraceEvent te) =>
+            source!.AllEvents += (TraceEvent te) =>
             {
                 if ((maxEvents == null || n < maxEvents.Value) && filter(te))
                 {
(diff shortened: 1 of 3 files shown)


ivdiazsa:
@Maoni0 @VSadov @cshung

src/benchmarks/gc/src/analysis/managed-lib/Analysis.cs
```diff
@@ -505,18 +505,20 @@ public class TracedProcesses
 
         public static void PrintEventsWithoutTraceLog(string tracePath, Func<TraceEvent, bool> filter, uint? maxEvents)
         {
-            using TraceEventDispatcher source = getSource(tracePath);
+            using TraceEventDispatcher source = TraceEventDispatcher.GetDispatcherFromFileName(tracePath);
```
Maoni0: >using TraceEventDispatcher source = TraceEventDispatcher.GetDispatcherFromFileName(tracePath); [](start = 12, length = 94)

you could do
`using TraceEventDispatcher source = TraceEventDispatcher.GetDispatcherFromFileName(tracePath)!;`

that way you don't need to do source! at all the other places.

Maoni0:
otherwise LGTM! also CC @brianrob

brianrob:
Awesome.  LGTM as well.

ivdiazsa:
Thanks! I had to leave the '!' in a few statements because, in spite of adding '!' to the declaration of `source`, dotnet still complained about those cases, giving an error saying `Dereference of possible null reference`.

brianrob:
@ivdiazsa, @Maoni0 let me know when you're happy with this and I will merge it.

ivdiazsa:
Thanks Brian! I see that all the checks have passed, so it's now ready for merging. @brianrob

PR #1251, opened 2020-03-25
fix doc for gc-where

Maoni0:
there needs to be quotes around the args otherwise some operations do not work (since some involve special characters which cause the args to not be read in correctly).

### DIFF
--- a/src/benchmarks/gc/docs/commands syntax.md
+++ b/src/benchmarks/gc/docs/commands syntax.md
@@ -44,15 +44,15 @@ py . diff --argsfile bench/diff_low_memory_container.yaml
 ```
 
 
-## Where arguments
+## gc-where arguments
 
-Many commands have arguments ending in `where` which take the following syntax:
+Many commands have arguments ending in `gc-where` which take the following syntax:
 
-    py . analyze-single foo.etl --gc-where Generation=2 PauseDurationMSec>100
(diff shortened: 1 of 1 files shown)


Maoni0:
@ivdiazsa PTAL

src/benchmarks/gc/docs/commands syntax.md
```diff
@@ -44,14 +44,14 @@ py . diff --argsfile bench/diff_low_memory_container.yaml
...
-Many commands have arguments ending in `where` which take the following syntax:
+Many commands have arguments ending in `gc-where` which take the following syntax:
 
-    py . analyze-single foo.etl --gc-where Generation=2 PauseDurationMSec>100
+    py . analyze-single foo.etl --gc-where "Generation=2" "PauseDurationMSec>100"
 
 In this case the argument filters it so we only print GCs that are gen 2 and took over 100ms.
-The value of `where` consists of a number of space-separated filters.
+The value of `gc-where` consists of a number of space-separated filters.
 Each filter is of the form “name operator value”, e.g.,
```
ivdiazsa: I would add a small note regarding the quotes here. Something like:

> Each filter is of the form "name operator value" (don't forget to enclose them in quotes), e.g.,

ivdiazsa:
Added a small suggestion but otherwise, LGTM! Thanks!

Maoni0:
thanks! @brianrob this is ready for merge.

PR #1256, opened 2020-04-01
Updated the PerfView and .NET Core versions in GC Benchmarking Infra

ivdiazsa:
The GC Benchmarking Infrastructure remained using _PerfView_ version _2.0.44_ and _.NET Core_ version _3.0_ (with support for _2.2_) for the GCPerfSim test. This PR updates to PerfView 2.0.52 (the most recent one) and adds 3.1 to GCPerfSim.

Also, it includes a new mechanism to automatically select the latest .NET Core version available for GCPerfSim. Let me elaborate on this. When GCPerfSim is built, usually three versions are deployed, one for each of the supported versions of .NET Core. Should an error prevent any of these binaries from being generated, the other ones can still be built just fine. So, this new mechanism allows the benchmarking tests to run with previous versions of Core should any issues arise with the most recent one.

@Maoni0 @VSadov @cshung

### DIFF
--- a/src/benchmarks/gc/src/commonlib/get_built.py
+++ b/src/benchmarks/gc/src/commonlib/get_built.py
@@ -220,10 +220,21 @@ def get_built_gcperf() -> Sequence[Path]:
     return [assert_file_exists(p) for p in (gcperf_dll_path, traceevent_dll_path)]
 
 
+def _get_latest_testbin_path(test_name: str) -> Path:
+    base_bin_path = _ARTIFACTS_BIN_PATH / test_name / "release"
+    bin_build_dirs = [str(f.absolute()).split('\\')[-1]
+                  for f in base_bin_path.iterdir() if f.is_dir()
+                  and 'netcoreapp' in str(f)]
+
+    bin_versions = list(map(lambda d: float(d.split('netcoreapp')[-1]),
+                            bin_build_dirs))
+    return base_bin_path / f"netcoreapp{bin_versions[-1]}" / f"{test_name}.dll"
+
+
 def _get_built_test(name: str, build_kind: BuildKind) -> Path:
     # Apparently, built files go to the root of the performance repo instead of next to the source.
     test_dir = _get_test_path(name)
-    out_path = _ARTIFACTS_BIN_PATH / name / "release" / "netcoreapp3.0" / f"{name}.dll"
+    out_path = _get_latest_testbin_path(name)
     test_cs = test_dir / f"{name}.cs"
     assert_file_exists(test_cs)
     msg = _is_build_is_out_of_date(_get_cs_files(test_dir), out_path, build_kind)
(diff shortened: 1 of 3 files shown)


Maoni0:
this actually brings up a good question - we could be adding things that use the new features, like right now, @VSadov is adding the capability of allocating on POH to GCPerfSim, should we have defines for those so we could build a version that still runs on 3.1 and prior? what are folks' thoughts on this?

VSadov:
I think diagnostics tools should keep compatibility (PerfView should be able to parse old logs, etc ). 
On the other hand tools used for product development do not need to support frameworks that are no longer in development. GCPerfSim can go forward and take dependency on new features.

In an unlikely case that we need to simulate something for 3.1 in the future, it will be possible to go back in history and build an old version.
It may be handy to have a fork for 3.1 - so that it is easy to find which version works with 3.1, also in case if we want to back-port fixes.

Maoni0:
I was proposing that instead of a fork, which would be hard to keep up with other features we add that don't rely on new features that only work in new releases (such as we have a better way to measure time), we simply include the block of code under #if defined (VERSION_X). this way if you want to build a version that runs on an older ver you just don't define VERSION_X. not sure what you meant by "simulate" but it's very likely we might want to run GCPerfSim with older versions of the runtime to compare.

VSadov:
I assumed that we would not keep up the fork with every new feature in the sim. Basically just backport fixes for serious bugs if found. 

We can start with  `#if defined (VERSION_X)` to see how that works. 

I think targeting multiple runtime versions may actually be a supported scenario (multitargeting?). 
If there is an officially supported way to do this in the build system, we should consider that as well.

ivdiazsa:
I agree with Maoni on it being more difficult than it's worth to keep multiple forks at this time. I believe the preprocessor defines would be a good way to start adding new features unsupported on previous versions of .NET Core. Now, I'm assuming we would add these defines on another PR?

That said, what are your thoughts on these updates I added? From the version-comparison perspective, I'm guessing we might want to add some sort of switch to choose which `netcoreapp` version to use on a specific run?

VSadov:
Yes, I think introducing `#if defined (VERSION_X)` should be a separate PR from this one. 

It looks like something that I will have to deal with prior to adding POH scenarios to the perf sim.

Maoni0:
@ivdiazsa sorry to reply late...I just haven't had the time to look at this. I'm a little confused about choosing a build for running GCPerfSim, isn't this specified via

`py . suite-create bench/suite --coreclrs path_to_coreclr0 path_to_coreclr1`

this is the only way I ran it.

how are you running GCPerfSim that requires it to automatically select a runtime build?

Maoni0:
ahh I see; I misunderstood. this is where GCPerfSim is run from,not where the runtime it runs with comes from. makes sense.

ivdiazsa:

PR #1258, opened 2020-04-01
Add parsing of version.props to get the current CLI build

DrewScoggins:


eng/performance/benchmark_jobs.yml
```diff
@@ -99,7 +99,9 @@ jobs:
         steps:
         - checkout: self
           clean: true
-        - script: $(Python) scripts/ci_setup.py --channel $(_Channel) --architecture ${{parameters.architecture}} --perf-hash $(Build.SourceVersion) --queue ${{parameters.queue}} --build-number $(Build.BuildNumber) --build-configs $(_Configs)
+        - script: $(Python) scripts/parse_props.py
+          displayName: Parse Version.props
+        - script: $(Python) scripts/ci_setup.py --dotnet-versions $(DotnetVersion) --channel $(_Channel) --architecture ${{parameters.architecture}} --perf-hash $(Build.SourceVersion) --queue ${{parameters.queue}} --build-number $(Build.BuildNumber) --build-configs $(_Configs)
```
billwert, line 104: If for some reason the parse script fails (say they change the format?) and does not parse something, what happens? I assume `$(DotnetVersion)` will end up being an empty string so this will turn into a noisy failure?

DrewScoggins: I just made a change that puts the whole string in the environment variable. I will also put some checks in place to make sure that file exists and that the node we parse it not null. IF we fail in either of those ways we will just fall back to the old method, which I think is best.

DrewScoggins: I have thought about this more and want to stick with the noisy failure. This will make it clear when this method fails and should also protect us from the CI side of the file ever moving or changing structure.

brianrob: Agreed - make it fail, because it should fail in the CI and we'll know exactly what happened.

DrewScoggins:
Any other comments on this? It turned out to be pretty straight-forward.

scripts/parse_props.py
```diff
@@ -0,0 +1,20 @@
...
+parser.add_argument('--branch-name', type=str, dest='branch',
+                   help='Name of the branch we are running')
+
+args = parser.parse_args()
+
+if not args.branch == "master":
+    print("##vso[task.setvariable variable=DotnetVersion;isSecret=false;isOutput=false]")
+else:
+    if not os.path.exists('eng/Versions.props'):
+        raise  ValueError("Versions.props does not exist")
```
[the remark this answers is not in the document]
ooooolivia: Btw I didn't realize my comment/review was not submitted until now.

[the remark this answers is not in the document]
DrewScoggins: Yeah, that is a good idea.

PR #1259, opened 2020-04-02
Add sod

billwert:
There are 3 main changes in this PR:
1) Move writing report tables to the Reporting library so we can use them from more than one place, and clean it up a bit so it will variably size if necessary. This also necessitated fixing / adding some tests.
2) Change Startup to use the central version of report tables.
3) Add SOD tool.

### DIFF
--- a/src/tools/Reporting/Reporting/Reporter.cs
+++ b/src/tools/Reporting/Reporting/Reporter.cs
@@ -10,7 +10,6 @@ using System.Globalization;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
-using System.Xml.Schema;
 using RuntimeEnvironment = Microsoft.DotNet.PlatformAbstractions.RuntimeEnvironment;
 
 namespace Reporting
--- a/src/tools/Reporting/Reporting/Reporter.cs
+++ b/src/tools/Reporting/Reporting/Reporter.cs
@@ -9,6 +9,8 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
+using System.Xml.Schema;
 using RuntimeEnvironment = Microsoft.DotNet.PlatformAbstractions.RuntimeEnvironment;
 
 namespace Reporting
@@ -39,12 +41,11 @@ namespace Reporting
         {
             var ret = new Reporter();
             ret.environment = environment == null ? new EnvironmentProvider() : environment;
-            if (!ret.CheckEnvironment())
+            if (ret.CheckEnvironment())
             {
-                return null;
+                ret.Init();
             }
                         
-            ret.Init();
             return ret;
         }
 
@@ -89,6 +90,10 @@ namespace Reporting
         }
         public string GetJson()
         {
+            if (!CheckEnvironment())
+            { 
+                return null;
+            }
             var jsonobj = new
             {
                 build,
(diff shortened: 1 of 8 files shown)


billwert:
@ooooolivia I also cleaned up these warnings in the Startup tests and things seem fine. Please let me know if there was a reason you had explicit package refs:
```
1>C:\Program Files\dotnet\sdk\3.1.300-preview-015048\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.DefaultItems.targets(257,5): warning NETSDK1023: A PackageReference for 'Microsoft.NET.Test.Sdk' was included in your project. This package is implicitly referenced by the .NET SDK and you do not typically need to reference it from your project. For more information, see https://aka.ms/sdkimplicitrefs
1>C:\Program Files\dotnet\sdk\3.1.300-preview-015048\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.DefaultItems.targets(257,5): warning NETSDK1023: A PackageReference for 'xunit' was included in your project. This package is implicitly referenced by the .NET SDK and you do not typically need to reference it from your project. For more information, see https://aka.ms/sdkimplicitrefs
1>C:\Program Files\dotnet\sdk\3.1.300-preview-015048\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.DefaultItems.targets(257,5): warning NETSDK1023: A PackageReference for 'xunit.runner.visualstudio' was included in your project. This package is implicitly referenced by the .NET SDK and you do not typically need to reference it from your project. For more information, see https://aka.ms/sdkimplicitrefs
```

src/tools/ScenarioMeasurement/SizeOnDisk/SizeOnDisk.cs
```diff
@@ -0,0 +1,134 @@
...
+                if (!String.IsNullOrEmpty(reportJsonPath))
+                {
+                    var json = reporter.GetJson();
+                    if(json != null)
+                    {
+                        File.WriteAllText(reportJsonPath, json);
+                    }
+                }
+            }
+            Console.WriteLine(reporter.WriteResultTable());
```
ooooolivia, line 82: Will we have a logger for this class?

billwert: Good question. I did forget to use it here. 

I'm sort of torn about `Logger` at this point. Originally my goal was to use that abstraction to ensure that text went to console and some other persistent store, such as a log file. Given that our overall model doesn't allow for long term persistence I don't know if it is very useful to wrap up console output and do nothing else. I'm inclined to give up on it. What do you think?

billwert: We sorted this offline. `Logger` is gonna go away eventually since it didn't end up being a thing we need.

src/tools/ScenarioMeasurement/SizeOnDisk/SizeOnDisk.cs
```diff
@@ -0,0 +1,134 @@
...
+                    Console.WriteLine($"Directory {dir} does not exist");
+                    return -1;
+                }
+            }
+            var directories = new Dictionary<string, Dictionary<string, long>>();
+            foreach(var dir in dirs)
+            {
+                FindVersionNumbers(dir);
+                directories.Add(dir, GetDirSize(dir));
+            }
```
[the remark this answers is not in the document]
billwert: I wanted to do all the validation before I did any actual work. I can pull it together if you think that's not the right principle.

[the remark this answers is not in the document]
DrewScoggins: It looks odd to me to loop through a collection twice when nothing is changing in the first loop. Also since we are tearing down the process on any failure we don't need to worry about torn state. All that to say, I don't see anything actually wrong with this.

src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -103,6 +107,46 @@ public string GetJson()
...
+        }
+        private string Print(Counter counter, int counterWidth, int resultWidth)
+        {
+            string average = $"{counter.Results.Average():F3} {counter.MetricName}";
+            string max = $"{counter.Results.Max():F3} {counter.MetricName}";
+            string min = $"{counter.Results.Min():F3} {counter.MetricName}";
+            return $"{Justify(counter.Name, counterWidth)}|{Justify(average, resultWidth)}|{Justify(min, resultWidth)}|{Justify(max, resultWidth)}";
+        }
+
+        private string Justify(string str, int width)
```
DrewScoggins, line 145: nit: Left Justify

src/tools/ScenarioMeasurement/SizeOnDisk/SizeOnDisk.cs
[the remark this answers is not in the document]
billwert: Sure.

src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -39,12 +40,11 @@ public static Reporter CreateReporter(IEnvironment environment = null)
         {
             var ret = new Reporter();
             ret.environment = environment == null ? new EnvironmentProvider() : environment;
-            if (!ret.CheckEnvironment())
+            if (ret.CheckEnvironment())
             {
-                return null;
+                ret.Init();
```
DrewScoggins, line 45: So before we would fail to make a reporter object if we were not in the lab, but now we just do the Init on an empty EnvironmentProvider. If I am running a test in a window that does not have those environment variables initialized will that cause problems for the report writing? I assume that all of those values will now be the empty string.

billwert: Note the change in `GetJson` that is the other half of this.

I needed to do this so tables could always be printed to console regardless of being in the lab or not.

src/tools/ScenarioMeasurement/SizeOnDisk/SizeOnDisk.cs
```diff
@@ -0,0 +1,130 @@
...
+            // the Templates dir seems to have the same version as the SDK version, but with a slightly different format
+            versions.Add(Regex.Replace(sdkVersion, @"(\d+\.\d+\.\d)00", "$1"));
+
+            versions.Add(new DirectoryInfo(Directory.GetDirectories(templateDir).Single()).Name);
+            
+            foreach(var dir in Directory.GetDirectories(sharedDir))
+            {
+                versions.Add(new DirectoryInfo(Directory.GetDirectories(dir).Single()).Name);
+            }
+            foreach (var version in versions) Console.WriteLine(version);
```
DrewScoggins, line 112: nit: Braces

src/tools/Reporting/Reporting/Reporter.cs
DrewScoggins: Ahh, I see. And that is why we make a check to see if GetJson is null later in the write function

src/tools/ScenarioMeasurement/SizeOnDisk/SizeOnDisk.cs
billwert: Good catch. This was actually just debug code I forgot to remove.

DrewScoggins:
One other question, why did we add the RootNamespace property in the Startup csproj?

billwert:
It was always supposed to be there; note that everything in that project is already in that namespace. I just noticed it was missing.

billwert:
@ooooolivia any other thoughts?

ooooolivia:
LGTM!

brianrob:
I'm very excited for this.  Thanks @billwert for doing this!

PR #1263, opened 2020-04-02
turn off weblarge projects

ooooolivia:


DrewScoggins:
Should we have an issue to track turning this scenario back on?

PR #1264, opened 2020-04-03
Add benchmarks for BitOperations

kunalspathak:
Add benchmarks for following APIs in `System.Numerics.BitOperations `:
- LeadingZeroCount
- TrailingZeroCount
- Log2

### DIFF
--- a/src/benchmarks/micro/libraries/System.Numerics.BitOperations/Perf_BitOperations.cs
+++ b/src/benchmarks/micro/libraries/System.Numerics.BitOperations/Perf_BitOperations.cs
@@ -1,0 +1,96 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Numerics;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+using System.Runtime.CompilerServices;
+
+namespace System.Numerics.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.SIMD)]
+    public class Perf_BitOperations
+    {
+        private static Random s_random = new Random(5);
+        private static uint[] input_uint = GenerateRandomValues<uint>();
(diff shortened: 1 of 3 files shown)


src/benchmarks/micro/libraries/System.Numerics.BitOperations/Perf_BitOperations.cs
```diff
@@ -0,0 +1,96 @@
...
+using MicroBenchmarks;
+using System.Runtime.CompilerServices;
+
+namespace System.Numerics.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.SIMD)]
+    public class Perf_BitOperations
+    {
+        private static Random s_random = new Random(5);
+        private static uint[] input_uint = GenerateRandomValues<uint>();
```
adamsitnik, line 16: We have a dedicated type that allows for generating [always the same input data](https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#Always-the-same-input-data):

https://github.com/dotnet/performance/blob/dd9884c879def10a8a96dd3cefc5d6c080c32253/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs#L44

Could you please use it (and extend with new method if needed)?

src/benchmarks/micro/libraries/System.Numerics.BitOperations/Perf_BitOperations.cs
```diff
@@ -0,0 +1,96 @@
...
+            return randomValues;
+        }
+
+        [Benchmark]
+        public int LeadingZeroCount_uint()
+        {
+            int sum = 0;
+            for (int i = 0; i < input_uint.Length; i++)
+            {
+                sum += BitOperations.LeadingZeroCount(input_uint[i]);
```
adamsitnik, line 36: nit: JIT is typically producing a better codegen when using local variables rather than fields, it's then 100% sure that nobody can change the value of a field in the meantime.

```suggestion
            int sum = 0;
            uint[] local = input_uint;
            for (int i = 0; i < local.Length; i++)
            {
                sum += BitOperations.LeadingZeroCount(local[i]);
```

Issue #1222, opened 2017-03-16
Expand the Perf test coverage for the System.Numerics.Vector types


tannergooding:
The current perf test coverage for the System.Numerics.Vector types is fairly limited (this extends to types such as Matrix4x4, Quaternion, and Plane as well).

Given that these are meant to be fairly core high-performance types (used in things such as Multimedia-based applications), it is important that we have a high amount of perf-test coverage over the current implementations.

adamsitnik:
> should this be moved to dotnet/performance?

@tannergooding good idea! #done

PR #1267, opened 2020-04-06
Add Vector<T> conversion benchmarks

saucecontrol:
Adding these in preparation for work on https://github.com/dotnet/runtime/issues/9766

Contributes to https://github.com/dotnet/performance/issues/1222

### DIFF
--- a/src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
+++ b/src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
@@ -1,0 +1,232 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Extensions;
+using MicroBenchmarks;
+
+namespace System.Numerics.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.SIMD)]
+    public class Perf_VectorConvert
+    {
+        private const int Iterations = 1000;
+
+        // Vector.ConvertToXXX
+        [Benchmark]
+        public Vector<int> Convert_float_int() => Convert<float, int>(s_valuesFloat);
+
+        [Benchmark]
+        public Vector<uint> Convert_float_uint() => Convert<float, uint>(s_valuesFloat);
+
+        [Benchmark]
+        public Vector<long> Convert_double_long() => Convert<double, long>(s_valuesDouble);
+
+        [Benchmark]
+        public Vector<ulong> Convert_double_ulong() => Convert<double, ulong>(s_valuesDouble);
+
+        [Benchmark]
+        public Vector<float> Convert_int_float() => Convert<int, float>(s_valuesInt32);
+
+        [Benchmark]
+        public Vector<float> Convert_uint_float() => Convert<uint, float>(s_valuesUInt32);
+
+        [Benchmark]
+        public Vector<double> Convert_long_double() => Convert<long, double>(s_valuesInt64);
+
+        [Benchmark]
+        public Vector<double> Convert_ulong_double() => Convert<ulong, double>(s_valuesUInt64);
+
+        // Vector.Narrow
+        [Benchmark]
+        public Vector<float> Narrow_double() => Narrow<double, float>(s_valuesDouble);
+
+        [Benchmark]
+        public Vector<sbyte> Narrow_short() => Narrow<short, sbyte>(s_valuesInt16);
+
+        [Benchmark]
+        public Vector<byte> Narrow_ushort() => Narrow<ushort, byte>(s_valuesUInt16);
+
+        [Benchmark]
+        public Vector<short> Narrow_int() => Narrow<int, short>(s_valuesInt32);
+
+        [Benchmark]
+        public Vector<ushort> Narrow_uint() => Narrow<uint, ushort>(s_valuesUInt32);
+
+        [Benchmark]
+        public Vector<int> Narrow_long() => Narrow<long, int>(s_valuesInt64);
+
+        [Benchmark]
+        public Vector<uint> Narrow_ulong() => Narrow<ulong, uint>(s_valuesUInt64);
+
+        // Vector.Widen
+        [Benchmark]
+        public Vector<double> Widen_float() => Widen<float, double>(s_valuesFloat);
+
+        [Benchmark]
+        public Vector<short> Widen_sbyte() => Widen<sbyte, short>(s_valuesSByte);
+
+        [Benchmark]
+        public Vector<ushort> Widen_byte() => Widen<byte, ushort>(s_valuesByte);
+
+        [Benchmark]
+        public Vector<int> Widen_short() => Widen<short, int>(s_valuesInt16);
+
+        [Benchmark]
+        public Vector<uint> Widen_ushort() => Widen<ushort, uint>(s_valuesUInt16);
+
+        [Benchmark]
+        public Vector<long> Widen_int() => Widen<int, long>(s_valuesInt32);
+
+        [Benchmark]
+        public Vector<ulong> Widen_uint() => Widen<uint, ulong>(s_valuesUInt32);
+
+
+        // These arrays are used for the Narrow benchmarks, so they need 2 vectors per iteration
+        private static readonly double[] s_valuesDouble = ValuesGenerator.Array<double>(Vector<double>.Count * 2 * Iterations);
+        private static readonly short[] s_valuesInt16 = ValuesGenerator.Array<short>(Vector<short>.Count * 2 * Iterations);
+        private static readonly ushort[] s_valuesUInt16 = ValuesGenerator.Array<ushort>(Vector<ushort>.Count * 2 * Iterations);
+        private static readonly int[] s_valuesInt32 = ValuesGenerator.Array<int>(Vector<int>.Count * 2 * Iterations);
+        private static readonly uint[] s_valuesUInt32 = ValuesGenerator.Array<uint>(Vector<uint>.Count * 2 * Iterations);
+        private static readonly long[] s_valuesInt64 = ValuesGenerator.Array<long>(Vector<long>.Count * 2 * Iterations);
+        private static readonly ulong[] s_valuesUInt64 = ValuesGenerator.Array<ulong>(Vector<ulong>.Count * 2 * Iterations);
+
+        // The remainder are used only for Widen and same-size Convert benchmarks
+        private static readonly float[] s_valuesFloat = ValuesGenerator.Array<float>(Vector<float>.Count * Iterations);
+        private static readonly byte[] s_valuesByte = ValuesGenerator.Array<byte>(Vector<byte>.Count * Iterations);
+        private static readonly sbyte[] s_valuesSByte = ValuesGenerator.Array<sbyte>(Vector<sbyte>.Count * Iterations);
+
+        private static Vector<TTo> Convert<TFrom, TTo>(TFrom[] values) where TFrom : struct where TTo : struct
+        {
+            var input = Unsafe.As<Vector<TFrom>[]>(values);
+            var accum = Vector<TTo>.Zero;
+
+            ref Vector<TFrom> ptr = ref input[0];
+            for (int i = Iterations; i >= 0; --i)
+            {
+                accum ^= ConvertVector<TFrom, TTo>(ptr);
+                ptr = ref Unsafe.Add(ref ptr, 1);
+            }
+
+            return accum;
+        }
+
+        private static Vector<TTo> Narrow<TFrom, TTo>(TFrom[] values) where TFrom : struct where TTo : struct
+        {
+            var input = Unsafe.As<Vector<TFrom>[]>(values);
+            var accum = Vector<TTo>.Zero;
+
+            ref Vector<TFrom> ptr = ref input[0];
+            for (int i = Iterations; i >= 0; --i)
+            {
+                accum ^= NarrowVector<TFrom, TTo>(ptr, Unsafe.Add(ref ptr, 1));
+                ptr = ref Unsafe.Add(ref ptr, 2);
+            }
+
+            return accum;
+        }
+
+        private static Vector<TTo> Widen<TFrom, TTo>(TFrom[] values) where TFrom : struct where TTo : struct
+        {
+            var input = Unsafe.As<Vector<TFrom>[]>(values);
+            var accum = Vector<TTo>.Zero;
+
+            ref Vector<TFrom> ptr = ref input[0];
+            for (int i = Iterations; i > 0; --i)
+            {
+                accum ^= WidenVector<TFrom, TTo>(ptr);
+                ptr = ref Unsafe.Add(ref ptr, 1);
+            }
+
+            return accum;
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,232 @@
...
+            if (typeof(TFrom) == typeof(int) && typeof(TTo) == typeof(float))
+                return (Vector<TTo>)(object)Vector.ConvertToSingle((Vector<int>)(object)value);
+            if (typeof(TFrom) == typeof(uint) && typeof(TTo) == typeof(float))
+                return (Vector<TTo>)(object)Vector.ConvertToSingle((Vector<uint>)(object)value);
+            if (typeof(TFrom) == typeof(long) && typeof(TTo) == typeof(double))
+                return (Vector<TTo>)(object)Vector.ConvertToDouble((Vector<long>)(object)value);
+            if (typeof(TFrom) == typeof(ulong) && typeof(TTo) == typeof(double))
+                return (Vector<TTo>)(object)Vector.ConvertToDouble((Vector<ulong>)(object)value);
+
+          throw new NotSupportedException("Type combination unsupported for Vector.ConvertToXXX");
```
adamsitnik, line 166: nit: spacing

```suggestion
            throw new NotSupportedException("Type combination unsupported for Vector.ConvertToXXX");
```

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,232 @@
...
+          if (typeof(TFrom) == typeof(int) && typeof(TTo) == typeof(short))
+              return (Vector<TTo>)(object)Vector.Narrow((Vector<int>)(object)value1, (Vector<int>)(object)value2);
+          if (typeof(TFrom) == typeof(uint) && typeof(TTo) == typeof(ushort))
+              return (Vector<TTo>)(object)Vector.Narrow((Vector<uint>)(object)value1, (Vector<uint>)(object)value2);
+          if (typeof(TFrom) == typeof(long) && typeof(TTo) == typeof(int))
+              return (Vector<TTo>)(object)Vector.Narrow((Vector<long>)(object)value1, (Vector<long>)(object)value2);
+          if (typeof(TFrom) == typeof(ulong) && typeof(TTo) == typeof(uint))
+              return (Vector<TTo>)(object)Vector.Narrow((Vector<ulong>)(object)value1, (Vector<ulong>)(object)value2);
+
+          throw new NotSupportedException("Type combination unsupported for Vector.Narrow");
```
adamsitnik, line 187: nit: spacing

```suggestion
        private static Vector<TTo> NarrowVector<TFrom, TTo>(Vector<TFrom> value1, Vector<TFrom> value2) where TFrom : struct where TTo : struct
        {
            if (typeof(TFrom) == typeof(double) && typeof(TTo) == typeof(float))
                return (Vector<TTo>)(object)Vector.Narrow((Vector<double>)(object)value1, (Vector<double>)(object)value2);
            if (typeof(TFrom) == typeof(short) && typeof(TTo) == typeof(sbyte))
                return (Vector<TTo>)(object)Vector.Narrow((Vector<short>)(object)value1, (Vector<short>)(object)value2);
            if (typeof(TFrom) == typeof(ushort) && typeof(TTo) == typeof(byte))
                return (Vector<TTo>)(object)Vector.Narrow((Vector<ushort>)(object)value1, (Vector<ushort>)(object)value2);
            if (typeof(TFrom) == typeof(int) && typeof(TTo) == typeof(short))
                return (Vector<TTo>)(object)Vector.Narrow((Vector<int>)(object)value1, (Vector<int>)(object)value2);
            if (typeof(TFrom) == typeof(uint) && typeof(TTo) == typeof(ushort))
                return (Vector<TTo>)(object)Vector.Narrow((Vector<uint>)(object)value1, (Vector<uint>)(object)value2);
            if (typeof(TFrom) == typeof(long) && typeof(TTo) == typeof(int))
                return (Vector<TTo>)(object)Vector.Narrow((Vector<long>)(object)value1, (Vector<long>)(object)value2);
            if (typeof(TFrom) == typeof(ulong) && typeof(TTo) == typeof(uint))
                return (Vector<TTo>)(object)Vector.Narrow((Vector<ulong>)(object)value1, (Vector<ulong>)(object)value2);

            throw new NotSupportedException("Type combination unsupported for Vector.Narrow");
        }
```

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,232 @@
...
+
+        [Benchmark]
+        public Vector<long> Widen_int() => Widen<int, long>(s_valuesInt32);
+
+        [Benchmark]
+        public Vector<ulong> Widen_uint() => Widen<uint, ulong>(s_valuesUInt32);
+
+
+        // These arrays are used for the Narrow benchmarks, so they need 2 vectors per iteration
+        private static readonly double[] s_valuesDouble = ValuesGenerator.Array<double>(Vector<double>.Count * 2 * Iterations);
```
adamsitnik, line 88: nit: style: we try to keep all the fields at the top of the file

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,232 @@
...
+        [Benchmark]
+        public Vector<uint> Widen_ushort() => Widen<ushort, uint>(s_valuesUInt16);
+
+        [Benchmark]
+        public Vector<long> Widen_int() => Widen<int, long>(s_valuesInt32);
+
+        [Benchmark]
+        public Vector<ulong> Widen_uint() => Widen<uint, ulong>(s_valuesUInt32);
+
+
```
adamsitnik, line 86: nit: extra line

```suggestion
```

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,232 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Extensions;
+using MicroBenchmarks;
+
+namespace System.Numerics.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.SIMD)]
```
adamsitnik, line 12: Is https://github.com/dotnet/runtime/issues/9766 going to need a change in JIT? If so, could you please change the category to `Runtime`?

```suggestion
    [BenchmarkCategory(Categories.Runtime, Categories.SIMD)]
```

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,232 @@
...
+
+        [Benchmark]
+        public Vector<double> Convert_long_double() => Convert<long, double>(s_valuesInt64);
+
+        [Benchmark]
+        public Vector<double> Convert_ulong_double() => Convert<ulong, double>(s_valuesUInt64);
+
+        // Vector.Narrow
+        [Benchmark]
+        public Vector<float> Narrow_double() => Narrow<double, float>(s_valuesDouble);
```
adamsitnik, line 44: is it possible to narrow `double` only to `float`? If not, we should add the "to" type name to method name.

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,232 @@
...
+        private static readonly double[] s_valuesDouble = ValuesGenerator.Array<double>(Vector<double>.Count * 2 * Iterations);
+        private static readonly short[] s_valuesInt16 = ValuesGenerator.Array<short>(Vector<short>.Count * 2 * Iterations);
+        private static readonly ushort[] s_valuesUInt16 = ValuesGenerator.Array<ushort>(Vector<ushort>.Count * 2 * Iterations);
+        private static readonly int[] s_valuesInt32 = ValuesGenerator.Array<int>(Vector<int>.Count * 2 * Iterations);
+        private static readonly uint[] s_valuesUInt32 = ValuesGenerator.Array<uint>(Vector<uint>.Count * 2 * Iterations);
+        private static readonly long[] s_valuesInt64 = ValuesGenerator.Array<long>(Vector<long>.Count * 2 * Iterations);
+        private static readonly ulong[] s_valuesUInt64 = ValuesGenerator.Array<ulong>(Vector<ulong>.Count * 2 * Iterations);
+
+        // The remainder are used only for Widen and same-size Convert benchmarks
+        private static readonly float[] s_valuesFloat = ValuesGenerator.Array<float>(Vector<float>.Count * Iterations);
```
adamsitnik, line 97: nit: you can totally ignore it: the naming is inconsistent: sometimes you are using `C#` type names like `int` for method names, sometimes .NET type names like `Int32` for fields (but also `float` for fields, instead `single`). IMHO it would be good to choose one naming and use it everywhere

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -84,19 +84,26 @@ private static T GenerateValue<T>(Random random)
...
             if (typeof(T) == typeof(int))
                 return (T)(object)random.Next();
+            if (typeof(T) == typeof(uint))
+                return (T)(object)(uint)random.Next();
+            if (typeof(T) == typeof(long))
+                return (T)(object)(long)random.Next();
+            if (typeof(T) == typeof(ulong))
+                return (T)(object)(ulong)random.Next();
+            if (typeof(T) == typeof(float))
+                return (T)(object)(float)random.NextDouble();
```
adamsitnik, line 100: thank you for extending the existing logic with new features!

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,232 @@
...
+            return accum;
+        }
+
+        private static Vector<TTo> Widen<TFrom, TTo>(TFrom[] values) where TFrom : struct where TTo : struct
+        {
+            var input = Unsafe.As<Vector<TFrom>[]>(values);
+            var accum = Vector<TTo>.Zero;
+
+            ref Vector<TFrom> ptr = ref input[0];
+            for (int i = Iterations; i > 0; --i)
```
adamsitnik, line 137: why two of the loops are limited by `i >= 0` while this one by `i > 0` (missing equal operator?) ?

adamsitnik:
@tannergooding could you please take a look?

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
saucecontrol: The goal would actually be to remove the code for these methods from the JIT and build them out using HWIntrinsics only.  Not sure if that changes the category choice or not...

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
saucecontrol: Yep, that's the only double conversion supported by `Narrow`.  The `ConvertTo` methods have multiple overloads, so I qualified those names with both from and to types.  Could do that here as well for consistency if you think it helps.

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
saucecontrol: Just an oversight.  Thanks for catching!

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
adamsitnik: In such case `Libraries` is better. Thanks for explanation!

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -84,19 +84,26 @@ private static T GenerateValue<T>(Random random)
         {
```
tannergooding, line 84: Would it be better to make some of these return their full range (at least the unsigned ones since `Random.Next()` doesn't return negatives)?

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
```diff
@@ -0,0 +1,231 @@
...
+        [Benchmark]
+        public Vector<uint> Widen_ushort() => Widen<ushort, uint>(s_valuesUShort);
+
+        [Benchmark]
+        public Vector<long> Widen_int() => Widen<int, long>(s_valuesInt);
+
+        [Benchmark]
+        public Vector<ulong> Widen_uint() => Widen<uint, ulong>(s_valuesUInt);
+
+        private static Vector<TTo> Convert<TFrom, TTo>(TFrom[] values) where TFrom : struct where TTo : struct
```
tannergooding, line 100: @adamsitnik, is there an issue on BDN that tracks adding better support for measuring operations that take less time than the timer frequency?
Likewise is there something that would more readily allow "random" inputs within some domain for those operations (for both primitive and structured data)?

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
saucecontrol: I was going for consistency with the existing `uint` and `ulong` ones, but I see they were just added last week.  Probably wouldn't be bad to change those now.  One other issue I noticed is that the `double` (and now `float`) one return values between 0 and 1, which is great for most floating point tests, but when you're converting those to integer values, it doesn't give you much coverage.

tannergooding: Yeah. I would think ideally we'd have some level of configuration that allows returning `negative values` and values within the entire `floating-point` domain.
The latter would likely still be random, but you wouldn't have "even distribution" (since `floating-point` isn't evenly distributed).

src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorConvert.cs
adamsitnik: > is there an issue on BDN that tracks adding better support for measuring operations that take less time than the timer frequency?

I think that we are doing quite OK with things that take less time than the timer frequency, the problem might be benchmarks that take less than 1ns. Do you have any concrete example of an operation that is hard to benchmark properly as of today? We could create an issue and start a discussion with @AndreyAkinshin

BTW for this benchmark [OperationsPerInvoke](https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#operationsperinvoke) could be handy.

> Likewise is there something that would more readily allow "random" inputs within some domain for those operations (for both primitive and structured data)?

No. We rather expect the users to generate such data and put it into an array during a setup, and then just iterate over it during benchmark.

PR #1274, opened 2020-04-09
Enable Linux Kernel Parsing in Startup

ooooolivia:
- add ```TraceSourceManager.cs``` which identifies which (Linux or Windows) source the file is
- update things in ```scenarios``` folder to accommodate ubuntu runs
- add ubuntu to sdk tests in the pipeline


scripts/ci_setup.py
```diff
@@ -273,8 +273,8 @@ def __main(args: list) -> int:
 
     # On non-windows platforms, delete dotnet, so that we don't have to deal with chmoding it on the helix machines
     # This is only necessary for netcoreapp3.0 and netcoreapp5.0
-    if sys.platform != 'win32' and remove_dotnet:
-        dotnet.remove_dotnet(architecture)
+    #if sys.platform != 'win32' and remove_dotnet:
+        #dotnet.remove_dotnet(architecture)
```
billwert, line 277: Can we just remove this instead of comment it out?

billwert, line 277: Can we remove this instead of comment it?

src/scenarios/shared/precommands.py
```diff
@@ -49,6 +49,10 @@ def __init__(self):
 
         args = parser.parse_args()
 
+        if not args.operation:
+            getLogger().error("Please specify an operation: %s" % list(OPERATIONS))
+            sys.exit(1)
```
billwert, line 54: Why this instead of the previous just requiring an operation?

src/scenarios/shared/runner.py
```diff
@@ -9,38 +9,69 @@
...
 # scenarios should try and run a given test type.
 testtypes = {const.STARTUP: False,
              const.SDK: False,
              const.CROSSGEN: False}
 
-TestTraits = namedtuple('TestTraits',
-                        reqfields  + tuple(testtypes.keys()) + optfields,
-                        defaults=tuple(testtypes.values()) + (None,) * len(optfields))
+
+class TestTraits:
```
billwert, line 23: Should this go into its own file now? When it was just a tuple it seemed reasonable to have here but now that it has code etc maybe it should move?

src/scenarios/shared/runner.py
```diff
@@ -61,7 +92,7 @@ def parseargs(self):
         Parses input args to the script
         '''
         parser = ArgumentParser()
-        subparsers = parser.add_subparsers(title='subcommands for scenario tests', required=True, dest='testtype')
+        subparsers = parser.add_subparsers(title='subcommands for scenario tests', dest='testtype')
```
billwert, line 95: Same question as above; why not rely on the argument parser to enforce required parameters?

src/scenarios/shared/runner.py
```diff
@@ -9,38 +9,69 @@
...
+    def add_trait(self, key: str, value: str, overwrite=True):
+        if not self.is_valid_trait(key):
+            raise Exception("%s is not a valid trait." % key)
+        if not self.get_trait(key) or overwrite:
+            self.traits[key] = value
+
+    # add multiple traits if not present or overwrite existing traits if overwrite=True
+    def add_traits(self, overwrite=True, **kwargs):
+        for keyword in kwargs:
+            self.add_trait(keyword, kwargs[keyword], overwrite=overwrite)
```
billwert, line 61: `add_traits` feels weird because it reverses the overwrite param from the other one. I get that's necessary because of using `kwargs`. I wonder if we should just collapse these into a single method and just call `add_traits` always, even if there's only one trait? That would also be a nicer pattern when writing code, as you would unify on `trait=value` instead of `key=trait, value=value`.

src/scenarios/shared/util.py
```diff
@@ -44,6 +44,12 @@ def uploadtokenpresent():
 def runninginlab():
     return environ.get('PERFLAB_INLAB') is not None
 
+def startupdir():
+    if sys.platform == 'win32':
+        return 'Startup'
+    else:
+        return 'startup'
```
billwert, line 51: Instead of adding this here can we just fix the yml to build to `Startup`? I intended this but apparently got it wrong.

src/scenarios/shared/runner.py
```diff
@@ -9,38 +9,69 @@
...
+                 'warmup',
+                 'workingdir',
+                 'iterationsetup',
+                 'setupargs',
+                 'iterationcleanup',
+                 'cleanupargs',
+                 'processwillexit',
+                 'measurementdelay',
+                 'environmentvariables'
+                 )
```
billwert, line 43: I would make these part of the class that you can fill in later. Put differently: now that `TestTraits` is a class it really is a class with the semantic that it holds some required fields and some optional fields. Make the class do that, and then runner can set it up as it sees fit.

src/scenarios/shared/startup.py
```diff
@@ -48,7 +48,7 @@ def __init__(self):
 
     
     def _setstartuppath(self, path: str):
-        self.startupexe = os.path.join(path, 'Startup.exe')
+        self.startuppath = os.path.join(path, 'Startup.exe' if sys.platform == 'win32' else 'Startup')
```
billwert, line 51: We have a helper method for this: https://github.com/dotnet/performance/blob/master/src/scenarios/shared/util.py#L29-L31

```suggestion
        self.startuppath = os.path.join(path, 'Startup%s' % extension())
```

(Or rewrite `extension` to do the string manipulation and pass in the exename.

src/scenarios/shared/util.py
```diff
@@ -64,4 +70,13 @@ def getruntimeidentifier():
...
     return rid
 
+def pythoncommand():
+    if sys.platform == 'win32':
+        return 'py'
+    else:
+        return 'python3'
+
+def iswin():
+    return sys.platform == 'win32'
```
billwert, line 80: If you're adding this here why not use it above in `pythoncommand` and other places?

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
```diff
@@ -1,10 +1,12 @@
 ﻿using System.Collections.Generic;
+using System;
 
 namespace ScenarioMeasurement
 {
     public class LinuxTraceSession : ITraceSession
     {
-        public string TraceFilePath { get { return perfCollect?.TraceFilePath; } }
+        public string TraceFilePath { 
+            get { return perfCollect?.TraceFilePath;} }
```
billwert, line 9: Why this line break?

src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
```diff
@@ -77,26 +85,37 @@ public ProcessHelper.Result Stop()
...
             {
-                throw new FileNotFoundException("Trace file not found.");
+                throw new FileNotFoundException($"Trace file not found at {Path.GetFullPath(TraceFileName)}.");
             }
-            string destinationFile = Path.Combine(TraceDirectory, traceFile);
-            if (File.Exists(destinationFile))
-            {
-                File.Delete(destinationFile);
+            // Don't move file if destination directory is current directory
+            if (Path.GetFullPath(Path.GetDirectoryName(TraceFilePath)) != Environment.CurrentDirectory){
```
billwert, line 94: nit: line break
```suggestion
            if (Path.GetFullPath(Path.GetDirectoryName(TraceFilePath)) != Environment.CurrentDirectory)
            {
```

src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
```diff
@@ -77,26 +85,37 @@ public ProcessHelper.Result Stop()
...
+            checkLttngProcess.WaitForExit();        
+            // checkLttngProcess.StartInfo.FileName = "lttng";
+            // checkLttngProcess.StartInfo.Arguments = ">/dev/null 2>&1";
+            //checkLttngProcess.StartInfo.UseShellExecute = true;
+            if (checkLttngProcess.ExitCode != 0)
+            {
+                perfCollectProcess.Arguments = "install -force";
+                return perfCollectProcess.Run().Result;
+            }
+            return ProcessHelper.Result.Success;*/
```
billwert, line 118: Either uncomment or delete?

src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
```diff
@@ -119,9 +138,9 @@ public void AddKernelKeyword(KernelKeyword keyword)
         public enum KernelKeyword
         {
             Empty,
-            ProcessLifetime,
-            Thread,
-            ContextSwitch
+            LTTng_Kernel_ProcessLifetimeKeyword,
+            LTTng_Kernel_ThreadKeyword,
+            LTTng_Kernel_ContextSwitchKeyword
```
billwert, line 143: Do we need to put `LTTng` in these names for some reason? If we can keep our view a little more generic that seems better to me.

src/tools/ScenarioMeasurement/Util/Util.cs
```diff
@@ -6,19 +6,24 @@ namespace ScenarioMeasurement
...
-        
-        
+        // [DllImport("User32.Dll")]
+        // private static extern long SetCursorPos(int x, int y);
+
+
         public static void Init()
         {
-            SetCursorPos(0, 0);
+            // SetCursorPos(0, 0);
```
billwert, line 15: This still needs to work on Windows. Please don't comment this out. We may need to split this into `Util.Windows.cs` and `Util.cs` and make them partial classes?

scripts/ci_setup.py
ooooolivia: Yes I submitted another PR #1282  for this issue and will leave it as original for now. Created another issue for now because I'd like to investigate in another seemingly unnecessary download of dotnet cli along with that.

src/scenarios/shared/precommands.py
ooooolivia: ```required``` parameter in ArgumentParser.add_subparser is only supported beginning in python3.7.  Doing this workaround because currently [we have trouble updating the python version in VMs](https://github.com/dotnet/core-eng/issues/9647).  Before the python version thing get fixed I think it's good to just avoid python3.7 completely.

src/scenarios/shared/runner.py
ooooolivia: Yeah I also prefer having it in another file.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -43,6 +46,72 @@ public void TestProfileIteration()
...
+            var parser = new TimeToMainParser();
+            var pids = new List<int>() { 24352, 24362, 24371, 24380, 24389 };
+            IEnumerable<Counter> counters = parser.Parse(ctfFile, "emptycsconsoletemplate", pids, "\"pub\\emptycsconsoletemplate.exe\"");
+            int count = 0;
+            foreach (var counter in counters)
+            {
+                Assert.True(counter.Results.Count == pids.Count, $"Counter {counter.Name} is expected to have {pids.Count} results.");
+                count++;
+            }
+            Assert.True(count == 2, "Only Time To Main counter should be present.");
```
DrewScoggins, line 94: This message and assert don't seem to add up. Is this supposed to `count == 1`?

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -43,6 +46,72 @@ public void TestProfileIteration()
...
+            var pids = new List<int>() { 32752, 6352, 16876, 10500, 17784 };
+            IEnumerable<Counter> counters = parser.Parse(etlFile, "dotnet", pids, "\"dotnet\" build");
+            int count = 0;
+            foreach (var counter in counters)
+            {
+                Assert.True(counter.Results.Count == pids.Count, $"Counter {counter.Name} is expected to have {pids.Count} results.");
+                count++;
+            }
+            Assert.True(count==2, "Both Process Time and Time To Main counter should be present.");
+        }
```
DrewScoggins, line 79: Why does the Linux test only find the process time counter, while the Windows one finds process time and time to main?

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
```diff
@@ -43,6 +46,72 @@ public void TestProfileIteration()
...
+            int count = 0;
+            foreach (var counter in counters)
+            {
+                Assert.True(counter.Results.Count == pids.Count, $"Counter {counter.Name} is expected to have {pids.Count} results.");
+                count++;
+            }
+            Assert.True(count == 2, "Both Time To Main and Time To Thread counter should be present.");
+        }
+
+
```
DrewScoggins, line 114: This also seems from the name like it should only be testing the time to main parsing, but again we check for two counters.

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            if (source.IsWindows)
+            {
+                if (((string)GetPayloadValue(evt, "CommandLine")).Trim() != commandLine)
+                {
+                    return false;
+                }
+            }
+            // Match the command line as well because pids might be reused during the session
+            return true;
```
DrewScoggins, line 32: This seems to always return true on Linux. Is this intended?

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
```diff
@@ -0,0 +1,94 @@
...
+            return MatchCommandLine(evt, source, commandLine) &&
+                   MatchProcessName(evt, source, processName) &&
+                   MatchProcessID(evt, source, pids);
+        }
+
+        public static bool MatchCommandLine(TraceEvent evt, TraceSourceManager source, string commandLine)
+        {
+            if (source.IsWindows)
+            {
+                if (((string)GetPayloadValue(evt, "CommandLine")).Trim() != commandLine)
```
DrewScoggins, line 26: We should make this compare look like all the other ones that we do in this file.

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
```diff
@@ -0,0 +1,94 @@
...
+                        return false;
+                    }
+                }
+                else if (!processName.Substring(0, 15).Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase))
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
```
DrewScoggins, line 60: If using an if/else block I tend to like having each block do it's own returning so that I can look at each block and know what will happen before moving on to the next one. In this case I have to look all the way at the end of the function to know what the code will do.

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                // 15 characters is the maximum length of a process name in Linux kernel event payload
+                if (processName.Length < 15)
+                {
+                    if (!processName.Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return false;
+                    }
+                }
+                else if (!processName.Substring(0, 15).Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase))
```
DrewScoggins, line 54: Should we be alerting the user of the library in some way that they are not getting the behavior they are expecting. Not sure how likely we are to have collisions on process names, but this feels like a place an insidious bug could get in.

src/scenarios/shared/runner.py
ooooolivia: Totally reasonable!

src/tools/ScenarioMeasurement/Startup/TraceSourceManager.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public sealed class LinuxKernelParser : IKernelParser
+    {
+        LinuxKernelEventParser parser;
+        public event Action<TraceEvent> ProcessStart { add { parser.ProcessStart += value; } remove { parser.ProcessStart -= value; } }
+        public event Action<TraceEvent> ProcessStop { add { parser.ProcessStop += value; } remove { parser.ProcessStop -= value; } }
+        public event Action<TraceEvent> ContextSwitch { add { } remove { } } // not implemented
+        public LinuxKernelParser(CtfTraceEventSource source) {
+            parser = new LinuxKernelEventParser(source);
+        }
```
DrewScoggins, line 66: nit: Line break
```suggestion
        public event Action<TraceEvent> ContextSwitch { add { } remove { } } // not implemented
        
        public LinuxKernelParser(CtfTraceEventSource source)
        {
            parser = new LinuxKernelEventParser(source);
        }
```

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -29,7 +29,7 @@ class Startup
         /// <param name="logFileName">optional log file. Default is appExe.startup.log</param>
         /// <param name="workingDir">optional working directory</param>
         /// <param name="warmup">enables/disables warmup iteration</param>
-        /// <param name="traceFileName">trace file name</param>
+        /// <param name="traceName">trace file name</param>
```
DrewScoggins, line 32: Did the semantics of this variable change or is this just a cosmetic change?

src/scenarios/shared/runner.py
ooooolivia: I also thought about having the traits as fields of the class but it seems not as flexible as just having a dictionary ```traits={ }```. So if we make them class fields I assume it will 
1) call ```setattr(traits_object, key, kwargs[key])``` to assign value to the field if the field name is a variable
2) call ```vars(traits_object)``` to get all the trait fields 

Another concern is it'd be very easy to do ```trait_object.field=<some random value>``` (but easier access as trade-off)

src/scenarios/shared/util.py
ooooolivia: Will replace other ifs with this method

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
ooooolivia: In what way? I assume the null-conditional operator will return null if perfCollect is null. Do you mean we can remove the operator because we always initialize ```TraceFilePath``` in ```PerfCollect( )``` constructor?

DrewScoggins: I believe he means why
```
public string TraceFilePath { 
            get { return perfCollect?.TraceFilePath;} }
```
instead of 
```
public string TraceFilePath { get { return perfCollect?.TraceFilePath;} }
```
or 
```
public string TraceFilePath
{
    get { return perfCollect?.TraceFilePath;}
}
```

src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
ooooolivia: hmm these are keywords defined in the new perfcollect script and we are just tostring() these enums to pass them as arguments to the script. These keywords exist only in ```PerfCollect.cs``` and lttng is indispensable part of perfcollect script. In other words, the concreteness of the naming seems equivalent to the scope of the context where it's used...

src/tools/ScenarioMeasurement/Util/Util.cs
ooooolivia: Sure! Why do we need to reset the cursor position to (0,0) each time? Does the same functionality apply to Linux as well? hmm not sure how partial classes can help here, if all files haves to be present at compile time even for Linux.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: Because on Linux we haven't found a proper way to catch the context switch event, which time to main counter depends on.

ooooolivia: hmm the error message should actually be changed to ```"Both Process Time and Thread Time counter should be present."```

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: Good catch! It should be 1.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
ooooolivia: Uh I think it should check 2 counters here on Windows: "time to main" and "time on thread".  (shocked by how many mistakes I made here)

src/tools/ScenarioMeasurement/Startup/LinuxTraceSession.cs
ooooolivia: I literally read it as "Will this line break?"

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
ooooolivia: Yes because there's no commandline field in Linux kernel event payload.

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
ooooolivia: Oh do you mean using "Equals"?

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
ooooolivia: Yes, it is unlikely we will have another process name that has the same 15 characters but we can avoid the bug by checking the filenames. Do you mean the expected behavior is it should get the full length of the process name?

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
ooooolivia: hmm not sure how the first approach works? My logic is return false upon mismatch, and return true in the end when everything passed the match.

src/tools/ScenarioMeasurement/Startup/Startup.cs
ooooolivia: The semantics changed because we can end up having different extensions for Windows and Linux trace files, so having a trace name but not file name here makes more sense. Yeah I should change the its description as well.

ooooolivia:
@DrewScoggins  Good question regarding checking in the binary files! They also add a non-negligible extra size to Linux payload since it copies the entire repo.... not to say these files are not used very often since they don't run in our CI. What would be the approach to store the test asset somewhere else and download them when needed? I'm thinking about NugetPackage/another git repo/Azure storage...

src/scenarios/shared/precommands.py
billwert: Ah, got it. :(

src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
billwert: Got it. Knowing we just pass these through is a good enough reason. Presumably we won't move off of LTTng anytime soon.

src/tools/ScenarioMeasurement/Util/Util.cs
billwert: This is only necessary for GUI apps which at this point we only support on Windows. We do this to ensure the cursor won't mistakenly be on a window or control for some iterations. By default Windows will shift the position of new windows down and to the right every time the app is launched. If you leave the cursor in its default position in the middle of the screen eventually you'll end up with an iteration of the measurement that has the mouse over it. This can cause any number of things, including mouse events, which can run different code. This can lead to noise. By moving it to 0,0, we avoid the problem entirely.

We can change the project file to only include the `Windows` or `Linux` file conditionally. That way we only try to find `SetCursorPos` on Windows.

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
DrewScoggins: Yeah, seems like it would be nice to do the compares similarly

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
DrewScoggins: It would look like the change below, but I think this is probably more of a personal preference. Keep it in a way that makes sense to you.
```suggestion
        public static bool MatchProcessName(TraceEvent evt, TraceSourceManager source, string processName)
        {
            if (source.IsWindows)
            {
                if (!processName.Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase))
                {
                    return false;
                }
                else
                {
                    return true;
                }
            }
            else
            {
                // 15 characters is the maximum length of a process name in Linux kernel event payload
                if (processName.Length < 15)
                {
                    if (!processName.Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase))
                    {
                        return false;
                    }
                }
                else if (!processName.Substring(0, 15).Equals(evt.ProcessName, StringComparison.OrdinalIgnoreCase))
                {
                    return false;
                }
                else
                {
                    return true;
                }
            }
        }
```

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
DrewScoggins: I mean that if my filename is over fifteen characters I might not be aware of the 15 character limit on Linux. You could imagine someone making filenames like ThisIsFifteenCharactersOrMore1, 2, 3, etc... Then that person would be quite confused why there tests are not working the way the expected. I don't think we need to fail or anything, but just writing to the log that we had to truncate the file name seems good, as we are not using the full input that we were given.

ooooolivia: I see! So far the user's not able to input a filename to our program, and our console output does not print out the event payload either. We are just using the truncated filename to do the matching, and full filename anywhere else. But it's good to keep in mind in case we run into it in the future.

src/tools/ScenarioMeasurement/Startup.Tests/StartupTests.cs
DrewScoggins: You are good here. I mis-read the test names and didn't realize it was testing different parsers. I thought it was supposed to be testing different counters we could get out of the parsers.

src/scenarios/shared/util.py
ooooolivia: Weird, I saw both having ```startup``` as publish output directory in yaml

DrewScoggins:
I'll be honest. I don't quite understand git well enough to know the full ramifications of doing this binary check-in, but I know that it will at the very least effect clone speed, which is something that we do all of the time. Using another git repo that we just clone we want to run the tests to me seems the easiest way forward.

src/scenarios/shared/util.py
billwert: It is both `startup`. As Windows is case insensitive by default it hasn't mattered. So either a) change the yml to `Startup` or change this to `startup` for both platforms. I don't care which.

src/scenarios/shared/startup.py
```diff
@@ -60,10 +61,10 @@ def runtests(self, apptorun: str, **kwargs):
         reportjson = os.path.join(TRACEDIR, 'perf-lab-report.json')
         defaultiterations = '1' if runninginlab() and not uploadtokenpresent() else '5' # only run 1 iteration for PR-triggered build
         startup_args = [
-            self.startupexe,
+            self.startuppath,
             '--app-exe', apptorun,
             '--metric-type', kwargs['startupmetric'], 
-            '--trace-file-name', '%s_startup.etl' % (kwargs['scenarioname'] or '%s_%s' % (kwargs['exename'],kwargs['scenariotypename'])),
+            '--trace-name', '%s_startup' % (kwargs['scenarioname'] or '%s_%s' % (kwargs['exename'],kwargs['scenariotypename'])),
```
billwert, line 67: Since we're losing `.etl` here shouldn't we re-append it somewhere else (in startup.cs perhaps?)

ooooolivia: The PR will move it to ```WindowsTraceSession.cs```

billwert: Ahh that already happened in another PR, got it.

src/scenarios/shared/runner.py
```diff
@@ -89,114 +66,100 @@ def parseargs(self):
...
+                    cleanupargs='%s %s cleanup' % ('-3' if iswin() else '', const.ITERATION_SETUP_FILE),
+                    workingdir=const.APPDIR,
+                    environmentvariables=envlistcleanbuild,
+                )
+                self.traits.add_traits(overwrite=True, startupmetric=const.STARTUP_PROCESSTIME)
+                startup.runtests(**self.traits.get_all_traits(),
+                                 scenarioname=self.scenarioname,
+                                 scenariotypename='%s_%s' % (const.SCENARIO_NAMES[const.SDK], const.CLEAN_BUILD),
+                                 apptorun=const.DOTNET
+                                 )
```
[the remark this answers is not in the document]
ooooolivia: Yes this is the same question that constantly bothers me. My understanding is ```TestTrait``` should be attributes that are asset-specific, and anything scenario-specific should go to ```runtests()```. The reason for this classification is based on our design --- we create ```TestTraits``` in ```test.py```, which exists individually in each asset, and we override them in ```runner.py``` for each scenario . Thus, we want to put attributes that not only belong to this asset but also are MOST SHARED across different scenario in ```TestTraits```, so we don't have to override them in different scenarios, otherwise they are just a group of kwargs. For example, I think ```appargs``` is a bad test trait, because we very possibly have different ```appargs``` across different scenarios, so is ```workingdir```, ```iterationsetup```, etc...

src/scenarios/shared/startup.py
```diff
@@ -48,7 +49,7 @@ def __init__(self):
 
     
     def _setstartuppath(self, path: str):
-        self.startupexe = os.path.join(path, 'Startup.exe')
+        self.startuppath = os.path.join(path, "Startup%s" % extension()) 
 
     def runtests(self, apptorun: str, **kwargs):
```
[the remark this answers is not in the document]
ooooolivia: hmm you mean we could do something like ```runtests(self, traits: TestTraits, scenarioname: str, scenariotypename: str, apptorun: str)```?   I'm not sure if startup should know ```TestTraits```.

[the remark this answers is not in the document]
billwert: it basically already does, but it's a bunch of strings instead of a type. :)

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
ooooolivia: We can probably avoid getting a false positive by doing this platform-specific check in the parser classes, but I don't feel confident that 1) we'd have a bunch of if-else again in the parser classes or 2)encounter exception when we forget to have the conditionals. Maybe we can create another child class in ```ParserUtility.cs``` that  extends the functionality of boolean ---```Match:true```, ```NotMatch:false```, ```NotApplicable:true``` ?

src/scenarios/shared/startup.py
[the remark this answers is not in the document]
ooooolivia: hmm I'm fine with having a bunch of string params for ```runtests()```, just not sure how does that solve the problem of using ```**self.traits.get_all_traits( )```? I assume we still need to do ```startup.runtests(**self.traits.get_all_traits( ), <non-trait attributes>)``` for convenience... or you mean use ```startup.runtests(guiapp=self.traits.guiapp, ...)```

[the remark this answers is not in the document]
billwert: I mean that the current implementation of `runtests` looks like this (in this PR):
```
	        startup_args = [
            self.startupexe,	            self.startuppath,
            '--app-exe', apptorun,	            '--app-exe', apptorun,
            '--metric-type', kwargs['startupmetric'], 	            '--metric-type', kwargs['startupmetric'], 
            '--trace-file-name', '%s_startup.etl' % (kwargs['scenarioname'] or '%s_%s' % (kwargs['exename'],kwargs['scenariotypename'])),	            '--trace-name', '%s_startup' % (kwargs['scenarioname'] or '%s_%s' % (kwargs['exename'],kwargs['scenariotypename'])),
This conversation was marked as resolved by billwert
            '--process-will-exit', (kwargs['processwillexit'] or 'true'),	            '--process-will-exit', (kwargs['processwillexit'] or 'true'),
            '--iterations', '%s' % (kwargs['iterations'] or defaultiterations),	            '--iterations', '%s' % (kwargs['iterations'] or defaultiterations),
            '--timeout', '%s' % (kwargs['timeout'] or '50'),	            '--timeout', '%s' % (kwargs['timeout'] or '50'),
            '--warmup', '%s' % (kwargs['warmup'] or 'true'),	            '--warmup', '%s' % (kwargs['warmup'] or 'true'),
            '--gui-app', kwargs['guiapp'],	            '--gui-app', kwargs['guiapp'],
            '--working-dir', '%s' % (kwargs['workingdir'] or sys.path[0]),	            '--working-dir', '%s' % (kwargs['workingdir'] or sys.path[0]),
            '--report-json-path', reportjson,	            '--report-json-path', reportjson,
            '--trace-directory', TRACEDIR	            '--trace-directory', TRACEDIR
```

So the contract is already that `runtests` knows a bunch of strings which are properties on `TestTraits`. It knows about `TestTraits` whether we tell it the name in a param because it's already tighly coupled. So instead of passing `kwargs` we could pass a `TestTraits`, and for the few required things that aren't on `TestTratis` we have actual required parameters to `runtests`. Does that make sense?

src/scenarios/shared/runner.py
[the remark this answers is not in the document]
billwert: Ah. Got it. Now I understand the confusion.

My intent with the traits was to have it be the holding area for everything required to run a test. That includes app args, working directories, everything. If you go back to the first [checkin](https://github.com/dotnet/performance/commit/b8483afd517af2eeb0e0c45018162ecdc6695f88#diff-c701c2cae0458d92a0b5cdab39708506R60-R64) you can kind of see that.

If we want to establish a more precise set of principles that's fine. As we evolved things and the complexity grew perhaps the original concept didn't make sense any longer. Maybe the whole idea of the traits type is bad?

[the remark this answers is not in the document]
ooooolivia: Got it. If this is true, kwargs like ```scenariotypename```, ```scenarioname``` and ```apptorun``` seem no less like a trait than the existing traits. Maybe we can just put everything in test traits, let ```startup.runtests()``` choose what it wants, and make ```test.py``` be an access point that's asset-specific and can override the default scenario-specific values?  Also maybe it's good to differentiate which is accessible by ```test.py``` and which is not.

src/scenarios/shared/startup.py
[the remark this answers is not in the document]
ooooolivia: Sounds great! And we don't have to type string every time we access the kwargs dict.

PR #1275, opened 2020-04-10
React to branding change

cshung:
As of https://github.com/dotnet/runtime/pull/33694, the product name changes from `b"Microsoft \xae .NET Core"` to `b"Microsoft \xae .NET"`. This broke the replacement logic. 

To avoid future failures, fixing it by simply replacing away that offending character. We don't need to be that specific. Any occurrence of that character is going to break the decoding anyway.


ivdiazsa:
LGTM! Thanks for updating this.

PR #1276, opened 2020-04-11
Add sod python

billwert:


eng/performance/scenarios.proj
```diff
@@ -5,41 +5,84 @@
...
+  </ItemGroup>
+  <!-- This is useful for testing -->
+  <!-- <Target Name="t">
+        <Message Text="@(HelixWorkItem -> 'name: %(HelixWorkItem.Identity)
+     dir: %(HelixWorkItem.PayloadDirectory)
+     pre: %(HelixWorkItem.PreCommands)
+     command: %(HelixWorkItem.Command)
+     post: %(HelixWorkItem.PostCommands)
+     timeout: %(HelixWorkItem.Timeout)  '"/>
+  </Target> -->
```
DrewScoggins, line 86: Should we just uncomment this?

src/scenarios/shared/sod.py
```diff
@@ -0,0 +1,73 @@
...
+    Wraps sod.exe, building it if necessary.
+    '''
+    def __init__(self):
+        if helixpayload() and os.path.exists(os.path.join(helixpayload(), 'SOD')):
+            self._setsodpath(os.path.join(helixpayload(), 'SOD'))
+        elif helixworkitempayload() and os.path.exists(os.path.join(helixworkitempayload(), 'SOD')):
+            self._setsodpath(os.path.join(helixworkitempayload(), 'SOD'))
+        else:
+            relpath = os.path.join(get_artifacts_directory(), 'SOD')
+            startupproj = os.path.join('..',
```
DrewScoggins, line 25: nit: Rename to sodproj?

eng/performance/scenarios.proj
billwert: It turns out that even if you try to use `/t:test` (assuming it had a better name) including the Helix SDK makes a bunch of other stuff run anyway, so you have to manipulate the file before you can do local testing.

src/scenarios/shared/sod.py
```diff
@@ -0,0 +1,73 @@
...
+    def _setsodpath(self, path: str):
+        self.sodexe = os.path.join(path, 'SizeOnDisk.exe')
+
+    def runtests(self, scenarioname, dirs):
+        '''
+        Runs tests through sod tool
+        '''
+        if not os.path.exists(TRACEDIR):
+            os.mkdir(TRACEDIR)
+        reportjson = os.path.join(TRACEDIR, 'perf-lab-report.json')
```
DrewScoggins, line 59: Should the name of the json have more information about the test?

billwert: I didn't do that because they get persisted to individual locations with the job and workitem GUID which should be enough to avoid dupes, and the test info is inside the json. (This decision predates this PR.)

src/scenarios/crossgen/test.py
```diff
@@ -8,7 +8,6 @@
 if __name__ == "__main__":
     traits = TestTraits(exename=EXENAME,
                         startupmetric='ProcessTime',
-                        crossgen=True,
```
DrewScoggins, line 11: Why are we removing all of these test traits from the default? The ones that we are calling for SoD tests make sense, but why the crossgen test as well?

DrewScoggins:
As part of the name change for the startup tests it looks like we have also added " - FDD Deploy" as well. Once these get uploaded I will go back and fixup all of the results.

src/scenarios/crossgen/test.py
billwert: We're removing the concept. The idea was this would guard against trying to do "the wrong thing" with test assets that didn't make sense for a given type of measurement, but it turns out that's not really a thing we need. The crossgen test is maybe a specific counter example, but I think it is very unlikely that anyone is ever really going to try and run things randomly (which was the chief user experience I was trying to protect.)

eng/performance/scenarios.proj
DrewScoggins: Could we have that documented somewhere? I think just having this puts people who want to try and run it in a bad position. This could be in another PR.

DrewScoggins:
LGTM, but I would like to make sure we address the test code in the proj file.

eng/performance/scenarios.proj
billwert: Done.

src/scenarios/shared/runner.py
```diff
@@ -175,6 +178,7 @@ def run(self):
                                 )
 
         elif self.testtype == const.CROSSGEN:
+            startup = StartupWrapper()
```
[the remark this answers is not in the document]
billwert: Correct. More specifically, if we're running an SOD test we should not have to build other tools. We could decouple building from the ctor of this thing but that's not something I really want to do. Reasonable?

src/scenarios/shared/sod.py
```diff
@@ -0,0 +1,73 @@
...
+                                get_packages_directory(),
+                                None,
+                                getruntimeidentifier(),
+                                '--no-restore'
+                                )
+            self._setsodpath(sod.bin_path)
+
+    
+    def _setsodpath(self, path: str):
+        self.sodexe = os.path.join(path, 'SizeOnDisk.exe')
```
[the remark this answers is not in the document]
billwert: Yes. Good catch.

eng/performance/scenarios.proj
```diff
@@ -5,41 +5,84 @@
...
+      <PreCommands>$(Python) pre.py publish -f $(_Framework) -c Release</PreCommands>
+      <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
+    </HelixWorkItem>
+  </ItemGroup>
+
+  <!-- SOD SCD publish w/ RID -->
+  <ItemGroup>
+    <HelixWorkItem Include="@(Scenario -> 'SOD - %(Identity) - SCD Publish')">
+      <PreCommands>$(Python) pre.py publish -f $(_Framework) -c Release -r $(RID)</PreCommands>
+      <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)quot;</Command>
```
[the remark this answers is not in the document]
billwert: Wild. No idea how that worked.

src/scenarios/shared/sod.py
```diff
@@ -0,0 +1,73 @@
...
+        '''
+        Runs tests through sod tool
+        '''
+        if not os.path.exists(TRACEDIR):
+            os.mkdir(TRACEDIR)
+        reportjson = os.path.join(TRACEDIR, 'perf-lab-report.json')
+        sod_args = [
+            self.sodexe,
+            '--report-json-path', reportjson,
+            '--scenario-name', (scenarioname or "Empty Scenario Name"),
```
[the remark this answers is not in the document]
billwert: This feels like a different PR at this point if we want to clean this up.

billwert:
@ooooolivia @DrewScoggins I think everything is addressed. I also tested on ubuntu (in WSL) and everything seems to work fine. Let me know if you have other thoughts or I missed something. Thanks!

src/scenarios/shared/runner.py
```diff
@@ -199,4 +203,14 @@ def run(self):
...
                              cleanupargs=None,
-                             )
\ No newline at end of file
+                             )
+        elif self.testtype == const.SOD:
+            sod = SODWrapper()
+            builtdir = const.PUBDIR if os.path.exists(const.PUBDIR) else None
+            print(builtdir)
+            if not builtdir:
+                builtdir = const.BINDIR if os.path.exists(const.BINDIR) else None
```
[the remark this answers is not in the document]
ooooolivia: Not sure if this has been changed?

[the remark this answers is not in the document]
billwert: Ah, good catch. I misread this as being part of the "remove `print`" comments.

PR #1280, opened 2020-04-13
Stop consuming result of GetPinnedReference()

kunalspathak:
We were consuming the values returned from `GetPinnedReference()` API, however
`.Consume()` uses `volatile` and that introduces memory barriers for ARM64.
Since we just want to measure the performance of `GetPinnedReference()` it is
unnecessary to introduce a differentiating factor for 1 architecture and not the
other. Hence I have removed the `.Consume()` and instead just storing the returned
result in a variable.

Before:
```asm
...
G_M52573_IG23:
        79400063          ldrh    w3, [x3]
        D5033BBF          dmb     ish
        79008023          strh    w3, [x1,#64]
        D2800003          mov     x3, #0
        34000040          cbz     w0, G_M52573_IG25
                                                ;; bbWeight=1    PerfScore 15.50
G_M52573_IG24:
        AA0203E3          mov     x3, x2
                                                ;; bbWeight=0.25 PerfScore 0.13
G_M52573_IG25:
        79400063          ldrh    w3, [x3]
        D5033BBF          dmb     ish
        79008023          strh    w3, [x1,#64]
        D2800003          mov     x3, #0
        34000040          cbz     w0, G_M52573_IG27
                                                ;; bbWeight=1    PerfScore 15.50
G_M52573_IG26:
        AA0203E3          mov     x3, x2
                                                ;; bbWeight=0.25 PerfScore 0.13
...

```

After
```asm
...
G_M51552_IG23:
        79400021          ldrh    w1, [x1]
        D2800001          mov     x1, #0
        34000040          cbz     w0, G_M51552_IG25
                                                ;; bbWeight=1    PerfScore 4.50
G_M51552_IG24:
        AA0203E1          mov     x1, x2
                                                ;; bbWeight=0.25 PerfScore 0.13
G_M51552_IG25:
        79400021          ldrh    w1, [x1]
        D2800001          mov     x1, #0
        34000040          cbz     w0, G_M51552_IG27
                                                ;; bbWeight=1    PerfScore 4.50
G_M51552_IG26:
        AA0203E1          mov     x1, x2
                                                ;; bbWeight=0.25 PerfScore 0.13
...

```

This change reduces the ARM64 numbers for this benchmark from 10ns to 1ns. I am not sure if
we should just add `Consume()` method and mark it as `NoInline`. That's what is done for
`System.Memory.Span<T>.GetPinnedReference()` benchmark. I have also increased the number of calls to `GetPinnableReference()` to make sure we can a valid time measurement value.

### DIFF
--- a/src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
+++ b/src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
@@ -17,7 +17,7 @@ namespace System.Memory
         [Benchmark]
         public ReadOnlySpan<char> StringAsSpan() => _sampeString.AsSpan();
         
-        [Benchmark(OperationsPerInvoke = 16)]
+        [Benchmark(OperationsPerInvoke = 16 * 2)]
         public char GetPinnableReference()
         {
             ReadOnlySpan<char> span = _sampeString.AsSpan();
--- a/src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
+++ b/src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
@@ -13,25 +13,33 @@ namespace System.Memory
     public class ReadOnlySpan
     {
         private readonly string _sampeString = "this is a very nice sample string";
-        private readonly Consumer _consumer = new Consumer();
 
         [Benchmark]
         public ReadOnlySpan<char> StringAsSpan() => _sampeString.AsSpan();
         
         [Benchmark(OperationsPerInvoke = 16)]
-        public void GetPinnableReference()
+        public char GetPinnableReference()
         {
             ReadOnlySpan<char> span = _sampeString.AsSpan();
-            var consumer = _consumer;
-            
-            consumer.Consume(span.GetPinnableReference()); consumer.Consume(span.GetPinnableReference());
-            consumer.Consume(span.GetPinnableReference()); consumer.Consume(span.GetPinnableReference());
-            consumer.Consume(span.GetPinnableReference()); consumer.Consume(span.GetPinnableReference());
-            consumer.Consume(span.GetPinnableReference()); consumer.Consume(span.GetPinnableReference());
-            consumer.Consume(span.GetPinnableReference()); consumer.Consume(span.GetPinnableReference());
-            consumer.Consume(span.GetPinnableReference()); consumer.Consume(span.GetPinnableReference());
-            consumer.Consume(span.GetPinnableReference()); consumer.Consume(span.GetPinnableReference());
-            consumer.Consume(span.GetPinnableReference()); consumer.Consume(span.GetPinnableReference());
+            char c;
+
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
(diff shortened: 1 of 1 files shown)


BruceForstall:
@adamsitnik @billwert

src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
```diff
@@ -13,25 +13,33 @@ namespace System.Memory
...
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
+            c = span.GetPinnableReference(); c = span.GetPinnableReference();
```
adamsitnik, line 41: is there a possibility that in the future JIT is going to replace all 32 calls with a single one?

@EgorBo or perhaps it can happen as of today with Mono using LLVM?

If so, we should most probably xor the results

```suggestion
            c = span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
            c ^= span.GetPinnableReference(); c ^= span.GetPinnableReference();
```

EgorBo: @adamsitnik I've just checked - Mono-LLVM optimizes out all the dead stores (basicaly, does `span.GetPinnableReference()` only once). The "^=" trick doesn't help 🙂 (with the proposed change it optimized the whole function to `return 0`

EgorBo: Regarding Ryujit I guess it's: https://github.com/dotnet/runtime/issues/13727

adamsitnik: @EgorBo thank you! Do you know how can we trick the Mono-LLVM ? Mix `+` with `^` ?

EgorBo: No idea, `GetPinnableReference` is just too simple and returns a pointer. I guess only if you hide `span` to a non-inlineable method e.g. `GetSpan().GetPinnableReference() ^ ..` but I guess the overhead of that method will be bigger than the actual `GetPinnableReference`

kunalspathak: Weird. Surprisingly RYUJit doesn't eliminate dead stores (at least for this example). The code that I pasted above is precisely that gets generated today. I will investigate and possibly open an issue to track this.

kunalspathak: Perhaps we can do something like this? https://github.com/dotnet/performance/blob/8aed638c9ee65c034fe0cca4ea2bdc3a68d2a6b5/src/benchmarks/micro/libraries/System.Memory/Slice.cs#L293 
But yeah, that would have method call overhead too.

adamsitnik: In such case, I think that I am going to merge it as it is.

If it becomes a problem in the future, we might consider removing this benchmark because it looks like it might be impossible to properly benchmark this method (it's too simple).

/cc @billwert @DrewScoggins

billwert:
I agree with @adamsitnik that the value of this benchmark is likely to be very small. What's the real regression risk for this method? As @EgorBo points out the method is pretty trivial as is. Is that ever likely to change? (I imagine not.)

I'm fine with the change as is but if this becomes problematic in the future (or becomes impossible to filter out of our results if it is noisy) we should indeed just get rid of it.

PR #1281, opened 2020-04-14
Support POH allocations in GCPerfSim

VSadov:
Adding support for POH allocations in GCPerfSim and a few knobs to control the allocations.

Light-up of the new features requires building against  `netcoreapp 5.0` with public support for POH. 
(I do not think it shipped yet, should be in the next preview, testing was done with private builds).

The project is backwards compatible and can be compiled to target earlier versions of the runtime such as `netcoreap 3.1`. 
In such case the features related to POH will be unavailable. Attempting to use them will result in exceptions saying that `netcoreapp 5.0` is required.

### DIFF
--- a/src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
+++ b/src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
@@ -312,16 +312,16 @@ class Item : ITypeWithPayload
     static readonly uint Overhead = (3 + 2) * Util.POINTER_SIZE;
 
     //TODO: isWeakLong never used
-    public static Item New(uint size, bool isPinned, bool isFinalizable, bool isWeakLong=false)
+    public static Item New(uint size, bool isPinned, bool isFinalizable, bool isWeakLong=false, bool isPoh = false)
     {
         if (isFinalizable)
         {
             throw new Exception("TODO");
         }
-        return new Item(size, isPinned, isWeakLong);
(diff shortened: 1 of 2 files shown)


VSadov:
CC: @Maoni0   -  Support POH allocations in GCPerfSim

VSadov:
@Maoni0 - do we want to merge this or wait until the next preview is shipped?

The default build will try building against all the targeted frameworks.
If 5.0 SDK does not yet have public `GC.AllocateArray`, the `netcoreapp50` part of the build will fail.

There are many workarounds though: 
- manually specify the target (i.e. `netcoreapp31`) or 
- install prerelease build or, 
- temporarily remove `netcoreapp50` from the csproj file...

I would be ok with waiting couple weeks before merging this if the above is a hassle.

Maoni0:
how are you testing this? by installing a prerelease build? I think there should a well defined way to do this since obviously we use this during development stage.

VSadov:
Actually nevermind. I think we can merge this.

**Preview 3 of SDK has the API**. I just had something older on my machine, so I had to patch System.Runtime ref assembly to make C# compiler happy. Probably because the preview was not available at the time.

There is more natural way to achieve the same now - just install Preview 3 or later from https://github.com/dotnet/installer

Note that this is only for compiling GCPerfSim. 
When running we use development bits. I.E. in GC infra you can specify several CoreRoot folders to run with and compare.

VSadov:

PR #1285, opened 2020-04-15
Fixed setup issue when cache sizes are unknown on Linux on GC Benchmarking Infra

### LINKED ISSUE #1250: 'py . setup' fails if lscpu has 'unknown size' for caches
Some systems may not have properly configured lscpu output, where the reported value for things like L1d/L1i/L2/L3 cache size is 'unknown size'. Consider changing line: 
https://github.com/dotnet/performance/blob/bce72f868c37f6af15dc788e1162c52751f01521/src/benchmarks/gc/src/commonlib/host_info.py#L133
to `return map_option(get_opt(name), lambda s: int(remove_str_end(s, "K")) if s != 'unknown size' else None)`

ivdiazsa:
Fixes #1250.

There is a possibility that `lscpu` might return the value _"unknown size"_ when it is not configured correctly or can't retrieve the cache sizes appropriately. GC Benchmarking Infra parses the output from `lscpu` as a Dictionary. When it attempts to read the cache size values, it expects a number followed by a "K", and therefore fails when receiving this "unknown" output.

This PR fixes this by setting the cache size to `None` should this scenario occur, as it already supports dealing with this value.

@Maoni0 @cshung @VSadov

### DIFF
--- a/src/benchmarks/gc/src/commonlib/host_info.py
+++ b/src/benchmarks/gc/src/commonlib/host_info.py
@@ -116,7 +116,7 @@ def _get_host_info(built: Built) -> HostInfo:
     ](built)
 
 
-_UNKNOWN_SIZE_MSG: str = "unknown size"
+_UNKNOWN_MSG: str = "unknown"
 
 def _get_host_info_posix() -> HostInfo:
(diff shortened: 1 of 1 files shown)


cshung:
LGTM

ivdiazsa:

PR #1291, opened 2020-04-21
Remove 0ns benchmarks

adamsitnik:
@DrewScoggins I've gone through the list that you have provided and I removed benchmarks that were taking 0ns because they were trying to benchmark something extremely simple like a property that was just returning a field value.

I was not able to remove all of them. I would like to discuss it with you and @billwert first.

There are some benchmarks that are immediate to execute on .NET Core, but take some time to run on Full .NET Framework. Mostly due to the optimizations applied to .NET Core.

A good example is:

```cs
[Benchmark]
public bool CheckArrayIsVariantGenericInterfaceReflection()
{
    return typeof(IReadOnlyCollection<object>).IsAssignableFrom(typeof(MyClass2[]));
}
```

|                                        Method |       Runtime |        Mean |
|---------------------------------------------- |-------------- |------------:|
| CheckArrayIsVariantGenericInterfaceReflection |      .NET 4.8 | 141.4506 ns |
| CheckArrayIsVariantGenericInterfaceReflection | .NET Core 5.0 |   0.0000 ns |

I am not sure what we should do with these benchmarks. We want to make sure that the optimizations persist and are not removed from future versions of .NET Core, but it's also very unlikely that we would regress them due to strict code reviews in dotnet/runtime repo. @DrewScoggins perhaps you could extend the new reporting tool with a possibility to ignore all the results below `0.2ns` similarly to what I did in the [ResultsComparer](https://github.com/dotnet/performance/blob/1c160f6fd074d104e6de39b2fae48d657673a5ab/src/tools/ResultsComparer/Program.cs#L77-L78)? So they don't pop up in the results as long as there is no actual regression (somebody reverts the optimization)?

The other problematic benchmarks are the ones that take `0ns` to execute on `x64` but are non-trivial on ARM mostly due to memory model differences between both architectures.

An example:

```cs
public class Perf_Volatile
{
    private double _location = 0;
    private double _newValue = 1;
    
    [Benchmark]
    public double Read_double() => Volatile.Read(ref _location);

    [Benchmark]
    public void Write_double() => Volatile.Write(ref _location, _newValue);
}
```

```ini
 BenchmarkDotNet=v0.12.1, OS=ubuntu 18.04
 ARMv8 Processor rev 1 (v8l), 4 logical cores
 .NET Core SDK=5.0.100-preview.4.20217.5
   [Host]     : .NET Core 5.0.0 (CoreCLR 5.0.20.21702, CoreFX 5.0.20.21702), Arm64 RyuJIT
   Job-ULPSAR : .NET Core 5.0.0 (CoreCLR 5.0.20.21702, CoreFX 5.0.20.21702), Arm64 RyuJIT
```

 |       Method |     Mean |
 |------------- |---------:|
 |  Read_double | 10.34 ns |
 | Write_double | 17.53 ns |

```ini
 BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.778 (1909/November2018Update/19H2)
Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.100-preview.4.20217.2
  [Host]     : .NET Core 5.0.0 (CoreCLR 5.0.20.21611, CoreFX 5.0.20.21611), X64 RyuJIT
  Job-OXOUZI : .NET Core 5.0.0 (CoreCLR 5.0.20.21611, CoreFX 5.0.20.21611), X64 RyuJIT
```

|       Method |      Mean |
|------------- |----------:|
|  Read_double | 0.0006 ns |
| Write_double | 0.2595 ns |

I think that there is value in having them. Perhaps we could conditionally exclude them for non-ARM architectures? I could send another PR for that.

The last group are benchmarks that take `0ns` to execute on a very new hardware but produce stable, non 0 results for older hardware.

An example:

```cs
[Benchmark]
public System.ReadOnlySpan<T> ReadOnlySpanFromArrayStartLength() => new System.ReadOnlySpan<T>(_nonEmptyArray, start: 0, length: Size);
```

```ini
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.778 (1909/November2018Update/19H2)
Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.100-preview.4.20217.2
  [Host]     : .NET Core 5.0.0 (CoreCLR 5.0.20.21611, CoreFX 5.0.20.21611), X64 RyuJIT
  Job-MFWBPT : .NET Core 5.0.0 (CoreCLR 5.0.20.21611, CoreFX 5.0.20.21611), X64 RyuJIT
```

|                           Method |      Mean |
|--------------------------------- |----------:|
| ReadOnlySpanFromArrayStartLength | 0.4132 ns |

This benchmark is producing very stable results on my PC, but it's quite old (E5). I think that we should also just filter out these benchmarks from the reporting tool and treat them as noise (below 0.2ns).


DrewScoggins:
I will try and take a look at this later today.

DrewScoggins:
In the cases where we see a difference between coreclr and Desktop I think the right thing to do is leave those tests around as a canary to make sure that we never regress those. Then we can ignore those tests on the reporting side unless we see a large actual delta not a percentage difference. As part of that work it might be nice to flag those tests with additional categories, as that would seem to be the easiest way to alert the visualization layer that these tests should be treated differently.

As for the second group, I think that only running those benchmarks on ARM is the right way forward. We should keep tests if they are useful on some platforms and not on others.

The last set of tests I don't have a great intuition or principle to go on, so for now leaving them and doing the filtering on the back end seems like a fine approach.

billwert:
I agree with @DrewScoggins in principle.

I guess we can conditionally compile those benchmarks on ARM? Trying to imagine what our good mechanism for this is. Or maybe BDN already has some filtering and we can apply attributes?

adamsitnik:
Can I merge this PR as it's removing the "obviously bad" benchmarks?

billwert:
Sure. FWIW, it wasn't clear to me the first time through that this PR was only for the "obviously bad" benchmarks. Perhaps the body of this PR could have been an issue and the PR clearly state which group was being addressed?

adamsitnik:
> it wasn't clear to me the first time through that this PR was only for the "obviously bad" benchmarks. Perhaps the body of this PR could have been an issue and the PR clearly state which group was being addressed?

You are right. I should have sent a simple PR and create a separate issue with the description of what I've found and could not be fixed immediately. I apologize for the mess.

PR #1293, opened 2020-04-22
Make reporting more explicit

billwert:
A recent change to the reporting library changed the semantic of how tools should detect whether or not they should try and write results. I missed updating the exporter in the microbenchmarks project. It is only broken when you try to use more than one runtime at once.

Fix is to just expose a property with the information. I don't want to bake it just into the logic of this class because different tools have different ideas of whether or not to use this functionality at all.

### DIFF
--- a/src/tools/Reporting/Reporting/Reporter.cs
+++ b/src/tools/Reporting/Reporting/Reporter.cs
@@ -40,7 +40,7 @@ namespace Reporting
         {
             var ret = new Reporter();
             ret.environment = environment == null ? new EnvironmentProvider() : environment;
-            if (ret.CheckEnvironment())
+            if (ret.InLab)
             {
                 ret.Init();
             }
@@ -89,7 +89,7 @@ namespace Reporting
         }
         public string GetJson()
         {
-            if (!CheckEnvironment())
+            if (!InLab)
             { 
                 return null;
             }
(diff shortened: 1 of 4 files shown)


src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -146,10 +146,7 @@ private string LeftJustify(string str, int width)
         {
             return String.Format("{0,-" + width + "}", str);
         }
-        
-        private bool CheckEnvironment()
```
ooooolivia, line 150: By any chance would we want to check more than one environment variables someday, so we need to keep it as a method here?

ooooolivia: Or your intention is to make ```CheckEnvironment``` more explicit by just having individual ```InLab``` boolean, so we know what it's doing.

billwert: Yeah, I would say that the original name was not well thought out. It was really just intended to asses whether or not we were in the lab. All the other interesting work happens in `Init`.

billwert:
RE: Moving to util. I thought about that, but then we'd be introducing a dependency on Util in the Microbenchmark project which doesn't currently use it. The reporting library right now stands alone. I think of Util as "shared things for ScenarioMeasurement". Does that make sense?

PR #1298, opened 2020-04-25
Introduce MathF.CopySign and Math.CopySign micro benchmarks

john-h-k:
For https://github.com/dotnet/runtime/pull/35456

Introduce benchmarks for the CopySign methods, which just test with 2500 iters of sign being 0 and 2500 with sign being 1. Criticism welcome as I am not really sure what i am doing here 😃

### DIFF
--- a/src/benchmarks/micro/runtime/Math/Functions/Single/CopySignSingle.cs
+++ b/src/benchmarks/micro/runtime/Math/Functions/Single/CopySignSingle.cs
@@ -1,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using BenchmarkDotNet.Attributes;
+
+namespace System.MathBenchmarks
+{
+    public partial class Single
+    {
+        // Tests Math.CopySign(float) over 5000 iterations for the domain -1, +1
+
+        private const float copySignDelta = 0.0004f;
+        private const int copySignExpectedResult = 0;
+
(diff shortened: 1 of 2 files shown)


billwert:
This looks fine to me. @tannergooding any comments?

src/benchmarks/micro/runtime/Math/Functions/Single/CopySignSingle.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            float result = 0.0f, value = -1.0f;
+
+            for (int iteration = 0; iteration < MathTests.Iterations; iteration++)
+            {
+                value += copySignDelta;
+                result += MathF.CopySign(result, value);
+            }
+
+            if (value != copySignExpectedResult)
```
tannergooding, line 30: Shouldn't this be `result != copySignExpectedResult`?

src/benchmarks/micro/runtime/Math/Functions/Single/CopySignSingle.cs
```diff
@@ -0,0 +1,36 @@
...
+        public void CopySign() => CopySignTest();
+
+        public static void CopySignTest()
+        {
+            float result = 0.0f, value = -1.0f;
+
+            for (int iteration = 0; iteration < MathTests.Iterations; iteration++)
+            {
+                value += copySignDelta;
+                result += MathF.CopySign(result, value);
```
tannergooding, line 27: It might be good if `result` started off at some non-zero value.... This is going to subtract `result` 2500 times and then add it back the next `2500`
Otherwise we are just testing adding/subtracting `+/-0`

PR #1303, opened 2020-04-30
Fixes a typo

cshung:
It appears to me that the hyphens are not needed (in fact the build would fail if included).
@ivdiazsa


ivdiazsa:
I must have overlooked that. It's on the Linux script where the hyphens are needed. LGTM!

ivdiazsa:

PR #1305, opened 2020-04-30
Add "- Count" to counter names to dedupe

billwert:


billwert:
Also added the full directory name to the counter name for individual files to avoid any conflicts where the same file exists in more than one place.

DrewScoggins:
We will never end up in a place where we could have the same directory name? If that is the case this LGTM.

billwert:
Yeah, we should not end up with the same directory name.

PR #1307, opened 2020-04-30
Add exclusion filter as an extension

DrewScoggins:


src/harness/BenchmarkDotNet.Extensions/CommandLineOptions.cs
```diff
@@ -30,6 +30,29 @@ public static List<string> ParseAndRemoveIntParameter(List<string> argsList, str
...
+                if (parameterIndex + 1 < argsList.Count)
+                {
+                    // remove --partition-count args
+                    parameterValue = argsList[parameterIndex+1];
+                    argsList.RemoveAt(parameterIndex + 1);
+                    argsList.RemoveAt(parameterIndex);
+                }
+                else
+                {
+                    throw new ArgumentException(String.Format("{0} must be followed by an string", parameter));
```
ooooolivia, line 49: typo: 'a string'

src/harness/BenchmarkDotNet.Extensions/ExclusionFilter.cs
```diff
@@ -0,0 +1,31 @@
...
+using BenchmarkDotNet.Running;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BenchmarkDotNet.Extensions
+{
+    class ExclusionFilter : IFilter
+    {
+        private string filter;
```
adamsitnik, line 11: nit: it can be readonly
```suggestion
        private readonly string filter;
```

src/harness/BenchmarkDotNet.Extensions/ExclusionFilter.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public bool Predicate(BenchmarkCase benchmarkCase)
+        {
+            if(String.IsNullOrEmpty(filter))
+            {
+                return true;
+            }
+            string testName = benchmarkCase.DisplayInfo.ToLowerInvariant().Substring(0, benchmarkCase.DisplayInfo.IndexOf(':') + 1);
+            return !testName.Contains(filter);
```
adamsitnik, line 28: the "inclusive" `--filter` allows for some simple regular expressions like `System.Collections*ToArray`, IMHO it would be great to keep consistency between both to avoid the user confusion

maybe you could even use [GlobFilter](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Filters/GlobFilter.cs#L19-L29) from BDN and just negate what it is returning?
```cs
internal class ExclusionFilter : IFilter
{
    private readonly GlobFilter _inclusiveFilter;
    
    internal ExclusionFilter(string[] patterns) => _inclusiveFilter = new GlobFilter(patterns);
    
    public bool Predicate(BenchmarkCase benchmarkCase) => !_inclusiveFilter.Predicate(benchmarkCase);
}
```

src/harness/BenchmarkDotNet.Extensions/BenchmarkDotNet.Extensions.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
adamsitnik, line 5: is this change necessary? 
```suggestion
    <TargetFramework>netstandard2.0</TargetFramework>
```

src/tools/Reporting/Reporting/Reporting.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
adamsitnik, line 5: is this change necessary? 
```suggestion
    <TargetFramework>netstandard2.0</TargetFramework>
```

src/harness/BenchmarkDotNet.Extensions/ExclusionFilter.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace BenchmarkDotNet.Extensions
+{
+    class ExclusionFilter : IFilter
+    {
+        private string filter;
+
+        public ExclusionFilter(string _filter)
```
adamsitnik, line 13: the "inclusive" `--filter` allows for many "filters" (example: `--filter abc bcd cba`), perhaps this one could support it too (for consistency)?

src/harness/BenchmarkDotNet.Extensions/BenchmarkDotNet.Extensions.csproj
DrewScoggins: This was from something else I was testing that leaked into the commit. I removed the net5.0 framework, but not the Frameworks tag. I'll fix this.

src/harness/BenchmarkDotNet.Extensions/ExclusionFilter.cs
DrewScoggins: The only thing I am worried about here is that we are doing this separate from the actual arg parsing, and if we get into a situation where we are processing n number of things for the filter I am afraid of introducing bugs in other parts of the parsing. I think for now I will leave it as just one filter, but we can add this later, and pay the testing cost, if we have a scenario that needs it.

src/harness/BenchmarkDotNet.Extensions/ExclusionFilter.cs
DrewScoggins: I can switch over to the GlobFilter, I was actually wanting to use it, but didn't know where the type was :)

DrewScoggins:
Turns out that time is now for having multiple filters as we also have failures because of named pipes. I will get this in tomorrow.

PR #1313, opened 2020-05-05
Make exclusion filter expect multiple filters

DrewScoggins:


src/harness/BenchmarkDotNet.Extensions/CommandLineOptions.cs
```diff
@@ -30,25 +30,21 @@ public static List<string> ParseAndRemoveIntParameter(List<string> argsList, str
...
-            parameterValue = null;
+            parameterValue = new List<string>();
 
-            if (parameterIndex != -1)
+            if (parameterIndex + 1 < argsList.Count)
             {
-                if (parameterIndex + 1 < argsList.Count)
+                while (parameterIndex + 1 < argsList.Count && !argsList[parameterIndex + 1].Contains("-"))
                 {
                     // remove --partition-count args
```
billwert, line 42: nit: this comment is wrong now right?

DrewScoggins: Good catch

src/harness/BenchmarkDotNet.Extensions/CommandLineOptions.cs
```diff
@@ -30,25 +30,21 @@ public static List<string> ParseAndRemoveIntParameter(List<string> argsList, str
...
         {
             int parameterIndex = argsList.IndexOf(parameter);
-            parameterValue = null;
+            parameterValue = new List<string>();
 
-            if (parameterIndex != -1)
+            if (parameterIndex + 1 < argsList.Count)
             {
-                if (parameterIndex + 1 < argsList.Count)
+                while (parameterIndex + 1 < argsList.Count && !argsList[parameterIndex + 1].Contains("-"))
```
adamsitnik, line 40: nit: `StartsWith` could be potentially more future-proof in case somebody sends an argument with `-` in the value (some path for example?)

```suggestion
                while (parameterIndex + 1 < argsList.Count && !argsList[parameterIndex + 1].StartsWith("-"))
```

DrewScoggins: Good idea

PR #1314, opened 2020-05-05
Remove wait for lock

ooooolivia:
The original HelixPrecommand was added to avoid [this core-eng issue](https://github.com/dotnet/core-eng/issues/9647). Removing the commands since the issue is fixed.


DrewScoggins:
It looks like from Matt's comment on that issue that this fix has not been rolled out yet. We should wait to make this change until we know it is in prod. Other than that LGTM.

ooooolivia:
@DrewScoggins  [Turned out it's a mistake](https://github.com/dotnet/core-eng/issues/9647) and it's already fixed in prod.

PR #1318, opened 2020-05-08
Fix broken venv

ooooolivia:
Our runs intent to happen in venv but it's currently not activated. Adding these changes to make sure all the packages are installed and tests are running in venv.


eng/performance/scenarios.yml
```diff
@@ -58,10 +58,10 @@ jobs:
             value: ''
           - ${{ if eq(parameters.osName, 'windows') }}:
             - name: HelixPreCommand
-              value: 'py -3 -m venv %HELIX_WORKITEM_PAYLOAD%\.venv;call %HELIX_WORKITEM_PAYLOAD%\.venv\Scripts\activate.bat;py -3 -m pip install --user azure.storage.blob==12.0.0 --force-reinstall;py -3 -m pip install --user azure.storage.queue==12.0.0 --force-reinstall;set "PERFLAB_UPLOAD_TOKEN=$(PerfCommandUploadToken)";$(AdditionalHelixPreCommands)'
+              value: 'py -3 -m venv %HELIX_WORKITEM_PAYLOAD%\.venv;call %HELIX_WORKITEM_PAYLOAD%\.venv\Scripts\activate.bat;set PYTHONPATH=;py -3 -m pip install azure.storage.blob==12.0.0 --force-reinstall;py -3 -m pip install azure.storage.queue==12.0.0 --force-reinstall;set "PERFLAB_UPLOAD_TOKEN=$(PerfCommandUploadToken)";$(AdditionalHelixPreCommands)'
```
DrewScoggins, line 61: I am fairly certain that setting the pythonpath to empty here negates the use of venv, as this is how when you call python you get the venv copy and not the base copy.

ooooolivia: hmmm I think activating venv just appends the venv binary copy to ```$PATH``` and ```PYTHONPATH``` will not be affected? I need to double-check. The purpose of emptying PYTHONPATH here is to remove the default python paths so pip3 will not look for libraries in the default locations. Check out the error without emptying PYTHONPATH here https://helix.dot.net/api/2019-06-17/jobs/3b5b2a0a-78cd-4cf4-aa51-fc71c4090cea/workitems/SDK%20ASP.NET%20MVC%20App%20Template%20netcoreapp3.0%20Clean%20Build/console

billwert: looks like that log is gone. Can you copy/paste the error or find another link that is still around?

ooooolivia: Sorry I forgot to note you need to add your access token in the end because it's internal.

PR #1323, opened 2020-05-11
test on ps1

ooooolivia:


ooooolivia:
Closed because arcade requires the global version of dotnet toolset to be installed.

PR #1324, opened 2020-05-12
Fix 0GB tests and retro-compatibility on GC Benchmarking Infra

ivdiazsa:
When Pinned Object Heap support was added to GCPerfSim, the default POH survival and finalizable values were conflicting with the 0GB live data configuration, thus resulting in an error. This PR modifies GC Benchmarking Infra's structure when generating the test yaml files, to also include POH arguments to feed GCPerfSim the correct information, as well as teach it the existence of this new .NET 5 feature.

However, this would break compatibility with .NET Core 3.1 and older versions. To amend this, this PR changes the asserting failures in GCPerfSim to warning messages, as simply ignoring the POH flags when not in .NET Core 5 does not cause any harm.


ivdiazsa:
@Maoni0 @cshung @VSadov

.vscode/extensions.json
```diff
@@ -1,8 +0,0 @@
-{
-    "recommendations": [
-        "DavidAnson.vscode-markdownlint",
-        "ms-python.python",
-        "ms-dotnettools.csharp",
-        "ms-vscode.powershell",
-    ]
-}
```
Maoni0, line 8: I dunno why these files are deleted.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -93,6 +96,9 @@
 -lohPinningInterval/-lohpi: g_lohPinningInterval
 meaning every Nth LOH object survived will be pinned. 
 
+-pohPinningInterval/-pohpi:
+meaning every Nth POH object survived will be pinned.
```
Maoni0, line 100: now I'm puzzled to see the pohpi arg since everything on POH is pinned... @VSadov?

src/benchmarks/gc/src/suite.py
```diff
@@ -215,22 +215,46 @@ def _normal_benchmarks(proc_count: int) -> Mapping[str, Benchmark]:
...
+            min_seconds=10
         ),
         "2gb": Benchmark(
-            arguments=GCPerfSimArgs(tc=proc_count, tagb=300 * tagb_factor, tlgb=2, sohsi=50)
+            arguments=GCPerfSimArgs(
+                tc=proc_count,
+                tagb=300 * tagb_factor,
+                tlgb=2,
+                sohsi=50,
+                pohsi=5,
```
Maoni0, line 231: do we have to specify poh args here? if we don't, doesn't that just mean there's nothing on the poh?

.vscode/extensions.json
ivdiazsa: This was an accidental mistake. I've restored them now :)

src/benchmarks/gc/src/suite.py
ivdiazsa: This is an interesting one. For all flags not passed, GCPerfSim sets a default value. Due to Python class semantics, since I had to add them for the 0GB benchmarks, it was much simpler to add them to all benchmarks. I checked these default values with the debugger beforehand.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
VSadov: It is allowed to pin POH objects, so `-pohpi` this _can_ be implemented.

I do not think there will be a lot of scenarios using this though.  I can imagine there could be scenarios where we would want to compare scenarios where pinning still happens, but target objects are POH-allocated.

In short - it was easier to match LOH here, than argue that we might not need this. I do not think this will be very popular knob.

src/benchmarks/gc/src/suite.py
```diff
@@ -215,22 +215,46 @@ def _normal_benchmarks(proc_count: int) -> Mapping[str, Benchmark]:
...
+                pohpi=100
```
VSadov, line 232: I think `-pohpi` could be 0 here. 

`-pohpi` might be interesting when comparing a run using `-sohpi` and equivalent one that uses `-pohpi`. - So that you could have everything as similar as possible except the kind of allocations.

As a setting for before/after test runs, `-pohpi` is unlikely to add a lot value.

Leaving it at 100 is ok as well, bot someone might be confused later about what the purpose.

ivdiazsa: Ok! So for "default" runs '0' would be the best value? I set it to 100 because that's the value GCPerfSim sets as default when the flag is not passed in `ParseFromCommandLine()`.

VSadov: I think I copied the default from soh or poh in GCPerfSim. 0 would be a more sensible default. We should fix that.

ivdiazsa: Ok great! Then I'll include that change as well.

VSadov: Thanks!

Maoni0: does this mean there will be allocations on POH? we should not be changing the basic scenarios which have no POH allocations. we can add more that specifically test POH allocations. but none of the current scenarios should change.

ivdiazsa: I had forgot to remove those. Now all POH parameters are set to zero when creating the bench files, which respects the basic scenarios' behavior.

Maoni0: are you sure that means there will be no allocations on POH? from a cursory look, I see that `GetPohAllocWeight` seems to set `lohAllocRatio` (can you rename this to `pohAllocRatio`, please?) to 5 if no poh args are specified, ie both `pohAllocInterval` and `lohAllocRatio` are null. @vsadov can you please confirm?

VSadov: `GetPohAllocWeight` argument should be `pohAllocRatio`.  It is a copy-paste error.

Just like with LOH there are more than one ways to specify how many POH objects you may want - in terms of the interval between allocations and in terms of ratio of allocations. 
If both are specified you get an assert.

VSadov: If neither `pohAllocInterval` nor `pohAllocRatio`  are specified, then `pohAllocRatio` is assumed to be 5. 
That is the same behavior as LOH had. I am not sure why that is not 0

Maoni0: because there's always LOH allocations; but POH should be completely optional. 

but the main point here is the basic scenarios (0gb, 2gb and etc) that we generate in the suite should not be changed - if you are adding something that causes the default to change, then the args to GCPerfSim in the suite should also change to make sure the basic scenarios don't change.

VSadov: I see. Yes, if we want such behavior for POH, then it seems better if defaults for POH would be all 0. 
I.E.  `DEFAULT_POH_SURV_INTERVAL` and `DEFAULT_POH_INTERVAL` should be changed to 0.

ivdiazsa:

PR #1332, opened 2020-05-19
Added PyPI notes when using the Jupyter Notebook

ivdiazsa:
An important use case of the Jupyter Notebook in GC Benchmarking Infra, is to analyze individual GC information from a trace on the fly. However, due to how the ProcessedGC` class is implemented, it is not as straightforward to retrieve the desired values.

These fields are wrapped in PyPI's Result Type. This PR adds documentation on how this works in a nutshell and how to apply it in the Jupyter Notebook to retrieve the information you want.


ivdiazsa:
@Maoni0 @cshung

ivdiazsa:

Issue #462, opened 2019-05-07
[Discussion] How can we detect regressions in CoreFX libs that are not part of the shared sdk?


adamsitnik:
This is a follow up to https://github.com/dotnet/performance/pull/438 and to my today's conversation with @stephentoub 

As of today, the [MicroBenchmarks.csproj](https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/MicroBenchmarks.csproj) references following CoreFX packages:

https://github.com/dotnet/performance/blob/3b75c755a25dc11ae405351c7b9b476760e98755/src/benchmarks/micro/MicroBenchmarks.csproj#L21-L31

For those packages which are part of the shared sdk we are detecting regressions using our existing infrastracture. The script used by CI: [benchmarks_ci.py](https://github.com/dotnet/performance/blob/master/scripts/benchmarks_ci.py) downloads the latest dotnet cli with latest sdk and runs the benchmarks using it. If `MicroBenchmarks.csproj` references and older version of a library present in the latest shared sdk, the latest version from shared folder is loaded and used. This is the most common scenario.

The problem are libraries that are not part of the shared sdk:
- System.Drawing.Common
- System.IO.Pipelines
- System.Threading.Channels

As of today, we don't catch the regressions for these libraries because we are referencing a single, hardcoded version. This is my fault, I did miss this scenario when working on the performance repo.

# Proposal

I have few ideas how we could solve this issue:

## Maestro

Use Maestro for dependency flow (the performance repo is already using it for Arcade updates) to update those 3 packages. 

The side effect is that we would trigger the benchmark runs every time new versions are released. 

@ViktorHofer how many times a day we can expect those packages to be updated?
@billwert do you think that this is desired and our infra is ready for more runs?

## Asterisk

The alternative to Maestro is to use an asterisk in version numbers and let the NuGet client restore the latest package.

```xml
<PackageReference Include="System.Drawing.Common" Version="4.*" />
<PackageReference Include="System.IO.Pipelines" Version="4.*" />
<PackageReference Include="System.Threading.Channels" Version="4.*" />
```
It's a very simple change, however it makes it hard to find out which version is currently used. I personally don't like it and would prefer the explicit approach.

# Remaining Issues

The problem is that both solutions work only for the explicitly configured packages. When we add new benchmarks that require a new dependency, somebody can forget about the proper configuration and introduce a new dependency to a hardcoded version.

[...]

PR #1334, opened 2020-05-20
Microsoft.Extensions.Caching.Memory.MemoryCache: Initial Benchmarks

NickCraver:
Creating some benchmarks for planned MemoryCache optimizations we can upstream - first changes in https://github.com/dotnet/runtime/pull/36775. Adding these for all to use, happy to change anything for styling, etc. just tried to match best I knew first pass!

See #462 for context, the lib in here (also: should it point to non-preview?) gets replaced with a `dotnet/runtime` version if using `coreRun`, which is awesome.

### DIFF
--- a/src/benchmarks/micro/libraries/Microsoft.Extensions.Caching.Memory/MemoryCacheTests.cs
+++ b/src/benchmarks/micro/libraries/Microsoft.Extensions.Caching.Memory/MemoryCacheTests.cs
@@ -1,0 +1,33 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/libraries/Microsoft.Extensions.Caching.Memory/MemoryCacheTests.cs
```diff
@@ -0,0 +1,33 @@
...
+            _memCache = new MemoryCache(new MemoryCacheOptions());
+            for (var i = 0; i < 1024; i++)
+            {
+                _memCache.Set(i.ToString(), i.ToString());
+            }
+        }
+
+        [Benchmark]
+        public object GetHit() => _memCache.Get("256");
+        [Benchmark]
```
adamsitnik, line 27: nit: new line

```suggestion

        [Benchmark]
```

adamsitnik:
the CI failure was not related to microbenchmarks so I am merging right now

davidfowl:
Looking forward to fixes here. Lots of low hanging fruit!

PR #1335, opened 2020-05-20
Add Crossgen2 Parser and Scenario

ooooolivia:
- add crossgen2 parser to Startup
- add crossgen2 scenario to scenarios

### Commands to run crossgen2 scenarios:

for single-file compilation:
```python test.py crossgen2 --core-root <path to core-root> --single <name of the single assembly>```

for composite compilation:
```python test.py crossgen2 --core-root <path to core-root> --composite <path to .rsp file that contains a list of assemblies>```

### CODE REVIEW

File: src/tools/ScenarioMeasurement/Startup/Crossgen2Parser.cs [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+                processTimeCounter,
+                new Counter() { Name = "Loading Interval", MetricName = "ms", DefaultCounter=false, TopCounter=true, Results = loadingParser.Intervals.ToArray() },
+                new Counter() { Name = "Emitting Interval", MetricName = "ms", DefaultCounter=false, TopCounter=true, Results = emittingParser.Intervals.ToArray() },
+                new Counter() { Name = "Jit Interval", MetricName = "ms", DefaultCounter=false, TopCounter=true, Results = jittingParser.Intervals.ToArray() },
+                new Counter() { Name = "Compilation Interval", MetricName = "ms", DefaultCounter=false, TopCounter=true, Results = compilationParser.Intervals.ToArray() }
+            };
+        }
+    }
+
+    public sealed class EventParser
```
billwert, line 60: Should this go into some shared spot? Looks pretty generic.

src/tools/ScenarioMeasurement/Startup/Crossgen2Parser.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        string EventName;
+        int? PrevPid = null;
+        int? Pid = null;
+        double Start = 0;
+        double Interval = 0;
+        public Stack<double> Intervals = new Stack<double>();
```
billwert, line 67: Property?

src/tools/ScenarioMeasurement/Startup/Crossgen2Parser.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        public EventParser(string eventName)
+        {
+            EventName = eventName;
+        }
+
+        public void AddEventStartCallback(TraceSourceManager source, DynamicTraceEventParser dynamicParser, string provider, IList<int> pids)
+        {
+            dynamicParser.AddCallbackForProviderEvent(provider, $"{EventName}/Start", evt =>
```
billwert, line 76: @brianrob isn't there an easier way to do this? I thought TraceEvent had nice support already for start/stop pairs. (Maybe not for dynamic events?)

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -185,7 +189,7 @@ static void checkArg(string arg, string name)
                 {
                     commandLine = commandLine + " " + appArgs;
                 }
-
+                Console.WriteLine($"commandline: {commandLine}");
```
billwert, line 192: debug output?

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
```diff
@@ -104,7 +104,7 @@ private void InitWindowsKeywordMaps()
 
         public void EnableUserProvider(string provider)
         {
-            UserSession.EnableProvider(provider);
+            UserSession.EnableProvider(provider, TraceEventLevel.Verbose);
```
billwert, line 107: Can verbosity be a parameter to `EnableUserProvider`? Not all providers are usable at the most verbose level. Some get very chatty.

src/tools/ScenarioMeasurement/Startup/Crossgen2Parser.cs
ooooolivia: This class only finds a pair of start/stop events with the specified event name. Its matching algorithm is designed for crossgen2 events, which is a bit loose compared to the matching algorithm of kernel events. It can definitely go to a shared spot if it could apply to other parsers.

src/tools/ScenarioMeasurement/Startup/Crossgen2Parser.cs
ooooolivia: Yes are you suggesting we should add modifiers ?

src/tools/ScenarioMeasurement/Startup/Startup.cs
ooooolivia: Oops will delete this.

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
ooooolivia: Good point!

src/tools/ScenarioMeasurement/Startup/Crossgen2Parser.cs
billwert: In general public properties are not used in C#, so this should be something like
```suggestion
        public Stack<double> Intervals {get; set; } = new Stack<double>();
```

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
```diff
@@ -102,9 +102,9 @@ private void InitWindowsKeywordMaps()
             clrKeywords[TraceSessionManager.ClrKeyword.Startup] = ClrPrivateTraceEventParser.Keywords.Startup;
         }
 
-        public void EnableUserProvider(string provider)
+        public void EnableUserProvider(string provider, TraceEventLevel verboseLevel)
```
billwert, line 105: should this have a default value for `verboseLevel`?

src/scenarios/shared/runner.py
```diff
@@ -169,6 +179,45 @@ def run(self):
...
+                referencefilenames = ['System.*.dll', 'Microsoft.*.dll', 'netstandard.dll', 'mscorlib.dll']
+                referencefiles = [os.path.join(self.coreroot, filename) for filename in referencefilenames]
+                # single assembly filename: example.dll
+                filename, ext = os.path.splitext(self.singlefile)
+                outputfile = os.path.join(self.coreroot, os.path.join('out', filename+'.ni'+ext ))
+                crossgen2args = '%s -o %s -O %s' % (os.path.join(self.coreroot, self.singlefile), outputfile, ' -r '.join(['']+referencefiles))
+            
+            elif compiletype == const.CROSSGEN2_COMPOSITE:
+                # composite rsp filename: ..\example.dll.rsp
+                dllname, ext = os.path.splitext(os.path.basename(self.compositefile))
```
billwert, line 205: use a discard here since you're not using the extension? 
```suggestion
                dllname, _= os.path.splitext(os.path.basename(self.compositefile))
```

src/scenarios/crossgen2/test.py
```diff
@@ -0,0 +1,16 @@
...
+from shared.testtraits import TestTraits, testtypes
+from shared import const
+
+EXENAME = 'crossgen2'
+
+if __name__ == "__main__":
+    traits = TestTraits(exename=EXENAME,
+                        startupmetric='Crossgen2',
+                        guiapp='false',
+                        timeout=const.MINUTE*30
```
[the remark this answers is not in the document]
ooooolivia: crossgen2 composite scenario takes longer time in general, depending on how many assemblies it compiles. It will take longer for PR builds on virtual machines. It is hard to measure exactly how long it will take but 30 min should be a safe number. We will see how long it takes once it runs on a regular basis and shorten the time if needed.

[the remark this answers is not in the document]
ooooolivia: crossgen2 composite scenario takes longer time in general, depending on how many assemblies it compiles. It will take longer for PR builds on virtual machines. It is hard to measure exactly how long it will take but 30 min should be a safe number. We will see how long it takes once it runs on a regular basis and shorten the time if needed.

src/tools/ScenarioMeasurement/Startup/ParserUtility.cs
```diff
@@ -21,7 +22,14 @@ public static bool MatchCommandLine(TraceEvent evt, TraceSourceManager source, s
         {
             if (source.IsWindows)
             {
-                if (!commandLine.Equals( ((string)GetPayloadValue(evt, "CommandLine")).Trim()))
+                int bufferMax = 512; 
+                string payloadCommandLine = (string)GetPayloadValue(evt, "CommandLine");
+                if (payloadCommandLine.Length >= bufferMax && commandLine.Length >= bufferMax)
```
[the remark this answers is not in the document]
ooooolivia: Yes, the payloadCommandLine can be longer than 512 when it takes several absolute paths, as it is a frequent case for crossgen2 tests... the TraceEvent payload truncates the string when it exceeds the threshold.

src/tools/ScenarioMeasurement/Startup/WindowsTraceSession.cs
ooooolivia: sure!

ooooolivia: sure!

src/scenarios/shared/testtraits.py
```diff
@@ -44,6 +45,8 @@ def __init__(self, **kwargs):
     # add traits if not present or overwrite existing traits if overwrite=True
     def add_traits(self, overwrite=True, **kwargs):
         for keyword in kwargs:
+            if keyword == 'timeout':
+                    print(f"keyword: timeout value: {kwargs[keyword]}")
```
billwert, line 49: is this diagnostic?

PR #1341, opened 2020-06-01
Enable scenarios tests on Ubuntu

ooooolivia:
TimeToMain parser is supposed to function on Linux now with the runtime fix https://github.com/dotnet/runtime/pull/35737 and we are good to turn on Linux scenarios job on master (bug fixed in .NET 5). However, the fix is not rolled out in 3.x release branches.


billwert:
closing per offline conversation - this is superseded by #1539

PR #1343, opened 2020-06-02
Update JSON tests to use new property name

DrewScoggins:


DrewScoggins:
The changes from https://github.com/dotnet/runtime/pull/36829 caused a break in our performance tests. This was missed because we had an exclude path in the yaml that caused our tests to not run on prs that only touched library code, this is fixed by https://github.com/dotnet/runtime/pull/37316.

DrewScoggins:
Failures are because this change has not propagated to the installer repo. When tested locally these tests passed.

PR #1346, opened 2020-06-04
build-server shutdown workaround 

ooooolivia:
The issue can be found here: https://github.com/dotnet/sdk/issues/10573

### DIFF
--- a/src/scenarios/shared/sdk_iteration_setup.py
+++ b/src/scenarios/shared/sdk_iteration_setup.py
@@ -1,10 +1,11 @@
 import os
 import shutil
-from shared import const
+from shared import const, util
 from dotnet import shutdown_server
 from performance.logger import setup_loggers
 from logging import getLogger
 from argparse import ArgumentParser
+
 
 SETUP_BUILD = 'setup_build'
 SETUP_NEW = 'setup_new'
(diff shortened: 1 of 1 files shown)


src/scenarios/shared/sdk_iteration_setup.py
```diff
@@ -42,5 +43,13 @@ def main():
             shutil.rmtree(const.APPDIR)
 
 
+def shutdown_dotnet_servers():
+    # shutdown_server(verbose=True) # This is the correct way to shut down dotnet build servers, but it has been disabled due to https://github.com/dotnet/sdk/issues/10573
+    getLogger().info("Shutting down dotnet build servers...")
+    if util.iswin():
+        os.system('TASKKILL /F /T /IM dotnet.exe || TASKKILL /F /T /IM VSTest.Console.exe || TASKKILL /F /T /IM msbuild.exe')
```
DrewScoggins, line 50: The `||` syntax means to only run the commands after it if the first one fails. Is that the behavior that we want?

ooooolivia: This command is suggested by @wli3 . I suppose the rationale here is ```dotnet.exe``` could be the parent process that spawns the other two, so if ```dotnet.exe``` has terminated we'd like to kill the child processes that are still running. Otherwise by specifying '/T' will be sufficient to kill the parent and child processes together.

DrewScoggins: SGTM

PR #1351, opened 2020-06-09
Use Python3.7 for ubuntu runs

ooooolivia:
After we removed tests on perfsnakes which don't have python3.7 installed, we should be ready for updating our pipeline to use Python3.7 now.

### DIFF
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -28,6 +28,8 @@ jobs:
             value: 'call %HELIX_CORRELATION_PAYLOAD%\tools\machine-setup.cmd'
           - name: Python
             value: 'py -3'
+          - name: PythonBuild
+            value: 'py -3'
           - ${{ if ne(parameters.osVersion, 'RS1')}}:
             - name: ArtifactsDirectory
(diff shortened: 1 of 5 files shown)


eng/performance/benchmark_jobs.yml
```diff
@@ -26,7 +26,9 @@ jobs:
             value: '--dotnet-versions %DOTNET_VERSION% --cli-source-info args --cli-branch %PERFLAB_BRANCH% --cli-commit-sha %PERFLAB_HASH% --cli-repository https://github.com/%PERFLAB_REPO% --cli-source-timestamp %PERFLAB_BUILDTIMESTAMP%'
           - name: AdditionalHelixPreCommands
             value: 'call %HELIX_CORRELATION_PAYLOAD%\tools\machine-setup.cmd'
-          - name: Python
+          - name: PythonJob
```
ooooolivia, line 29: @billwert  Kindly remind me is this the variable name you'd prefer to change?  (the name's lost in our Teams chat)

billwert: yeah, I would have left this just as `Python`. `PythonBuild` is the special one for the CI build machine, and `Python` can just be everywhere else.

ooooolivia: Yup thanks for reminding me!

PR #1352, opened 2020-06-09
Allow GCInfra to work with Process Name from Test Status Files

ivdiazsa:
In order to analyze performance traces captured somewhere else (i.e. not with GCPerfSim) with GC Benchmarking Infra, one has to write a quick test status file, which contains at least the name of the trace file, a flag denoting success, and the process id:

```yaml
success: true
trace_file_name: trace.etl
process_id: 12345
```

However, since this is written separately for traces captured elsewhere, one has to get the process id either via `print-processes` or some other way, which can be a tedious procedure.

This PR adds the possibility of writing the process name to the test status file, rather than having to look for the id. Also, it is worth mentioning it accepts regular expressions to match with process names:

```yaml
success: true
trace_file_name: trace.etl
process_name: CoreRun
```

@Maoni0 @cshung


Maoni0:
thanks! couple of Qs -
* how would this behave if there are multiple instances of the same process name? 
* could you specify args with this?

ivdiazsa:
> thanks! couple of Qs -
> 
> * how would this behave if there are multiple instances of the same process name?
> * could you specify args with this?

Good questions!

For the first one, GC Infra throws an exception if more than one process matches and prints all of the matching processes. It might be worth it to just show the error and not print the processes, as it prints them with lots of stuff like cmdline arguments. It looks like a ton of garbage at first glance.

For the second one, could you elaborate more on what you are referring to with _args_?
Thanks!

Maoni0:
for the 1st one, if you look at the print-processes cmd, it has some filters that allow much less output like --clr-only and --hide-threads. you can print just the process names and their cmdline args for processes actually use the runtime.

for the 2nd one, I meant cmdline args. like asp.net processes are all named w3wp.exe but they have different cmdline args like "w3wp -ap "my_app_pool_name".

```python
class _RegexProcessPredicate(ProcessPredicate):
    def __init__(self, parts: Sequence[str]):
        name = None
        args = None
        proc_id = None
        for part in parts:
            colon_sep = part.split(":")
            assert len(colon_sep) == 2, f"Expected to split '{part}' into two by a ':'"
            key, value = colon_sep
            if key == "name":
                name = value
            elif key == "args":  ####<---- this is parsing args
                args = value
            else:
                assert key == "id"
                proc_id = int(value)
```

Maoni0:
can you please add an example of how this will work with the cmdline args?

ivdiazsa:
@billwert This PR is ready to merge.

PR #1353, opened 2020-06-10
Fixed all py . lint errors

cshung:
As of [this](https://github.com/dotnet/performance/commit/46003e5150fa86ada9ea617aa5b92641df8c2cf0#diff-bb3134dbe91ea53ff67ddbc68f1e4b7d) commit, the `py . lint` is failing. Since then, there are a few changes that broke the type checker and resulted in a pile of errors.

It is quite obvious that @andy-ms spent some good effort to make sure the code is well annotated and type check, it is a pity to live with the errors.

This change fixes it all. With this change, `py . lint` will run clean without error. However, it will suggest some package updates. Do **not** upgrade, as the linter will fail after the upgrade as the upgraded linter is more strict, we might want to also fix that in the future but that is out of the scope of this PR.

I'd love to see that it keep clean. The contribution documentation is updated to reflect that.

The majority of the change is performed by the linter itself, there are no semantics changes.

@dotnet/gc

### DIFF
--- a/src/benchmarks/gc/docs/bench_file.md
+++ b/src/benchmarks/gc/docs/bench_file.md
@@ -35,10 +35,13 @@ benchmarks:
       lohar: 0
       sohsi: 50
       lohsi: 0
+      pohsi: 0
       sohpi: 0
       lohpi: 0
+      pohpi: 0
       sohfi: 0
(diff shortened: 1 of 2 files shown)


cshung:

PR #1357, opened 2020-06-11
Add blazor scenario

ooooolivia:
Adding blazor scenario for blazor sod:
```$(Python) pre.py publish```  --- set up by creating a blazorwasm template and publishing it
```$(Python) test.py sod --scenario-name <scenario name>``` --- run the test for its size on disk

### DIFF
--- a/eng/performance/scenarios.proj
+++ b/eng/performance/scenarios.proj
@@ -80,7 +80,8 @@
   <ItemGroup>
     <HelixWorkItem Include="SOD - New Blazor 3.2 Template - Publish" Contidion="'$(_Framework)' == 'netcoreapp3.1'">
       <PayloadDirectory>$(ScenariosDir)blazor</PayloadDirectory>
-      <PreCommands>$(Python) pre.py --msbuild-static AdditionalMonoLinkerOptions=&quot;%24(AdditionalMonoLinkerOptions) --dump-dependencies&quot;</PreCommands>
+      <PreCommands>dotnet new -i Microsoft.AspNetCore.Components.WebAssembly.Templates::3.2.0-preview2.20160.5</PreCommands>
+      <PreCommands>$(PreCommands);$(Python) pre.py --msbuild-static AdditionalMonoLinkerOptions=&quot;%24(AdditionalMonoLinkerOptions) --dump-dependencies&quot;</PreCommands>
       <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
     </HelixWorkItem>
   </ItemGroup>
--- a/eng/performance/scenarios.proj
+++ b/eng/performance/scenarios.proj
@@ -77,7 +77,7 @@
   </ItemGroup>
 
   <!-- New Blazor 3.2 Template SOD -->
-  <ItemGroup Contidion="'$(_Framework)' == 'netcoreapp3.1'">
+  <ItemGroup Condition="'$(_Framework)' == 'netcoreapp3.1'">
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: src/scenarios/blazor/test.py [hunk 1]
```diff
@@ -0,0 +1,13 @@
+'''
+Empty Blazor Wasm Template
+'''
+from shared.runner import TestTraits, Runner
+
+SCENARIONAME = 'Empty Blazor Wasm Template'
```
billwert, line 6: this isn't actually used right? I know it's still hanging out in some other files but we shouldn't keep including it if it isn't doing anything.

ooooolivia: Agree. I have lost track of why it's there ( maybe originally put there as a metadata ?) but it's not used and duplicate info from the comment on the top.

DrewScoggins:
Everything LGTM, but I have one quick question. How do the arguments passed to test.py get routed into the runner?

ooooolivia:
@DrewScoggins  Do you mean ```sod``` and ```--scenario-name```? The ```ArgumentParser``` is created in ```runner.py``` and we just parse when we call ```Runner(traits).run()``` https://github.com/dotnet/performance/blob/ecd9653e5017e0c3e5c985a86dd667d83249d08a/src/scenarios/shared/runner.py#L98

src/scenarios/shared/sod.py
```diff
@@ -65,6 +65,9 @@ def runtests(self, scenarioname, dirs):
         ]
 
         RunCommand(sod_args, verbose=True).run()
+ 
+        linker_dump_file = os.path.join(APPDIR, 'obj', 'Release', 'netstandard2.1', 'blazor', 'linker', 'linker-dependencies.xml.gz')
```
billwert, line 69: Can we avoid hardcoding `Release` here? If we ever want to run a `Debug` version of this test this will be an issue.

eng/performance/scenarios.proj
```diff
@@ -75,6 +75,14 @@
...
     </HelixWorkItem>
   </ItemGroup>
+
+  <!-- Blazor 3.2 Template SOD -->
+  <ItemGroup>
+    <HelixWorkItem Include="Blazor 3.2 Template SOD" Contidion="'$(_Framework)' == 'netcoreapp3.1'">
+      <PreCommands>$(Python) pre.py --msbuild-static AdditionalMonoLinkerOptions=&quot;$(AdditionalMonoLinkerOptions) --dump-dependencies&quot;</PreCommands>
+      <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
+    </HelixWorkItem>
+  </ItemGroup>
```
billwert, line 85: Are you also automating installing the Blazor templates?

ooooolivia: Oops missed that.

ooooolivia:
Passed blazor workitem with console log and linker dump: https://helix.dot.net/api/2019-06-17/jobs/ece3886f-0cb1-4f67-97d9-3a65f7787502/workitems/SOD%20-%20New%20Blazor%203.2%20Template%20-%20Publish

src/scenarios/shared/sod.py
ooooolivia: It seems that linker dump file will not be generated on Debug.

src/scenarios/blazor/test.py
billwert, line 6:
```suggestion

```

ooooolivia: oops, missed this comment. I should probably remove it from all ```test.py```

PR #1358, opened 2020-06-12
Removed restriction forcing traces to have same number of GC's when charting

ivdiazsa:
Certain charting operations in GC Infra, such as `chart_lines_from_fields`, used to fail at an assertion preventing graphics generation of traces with different number of GC's. It is not uncommon to have this scenario, so this PR removes this assertion, as charting these traces like that is harmless.

@Maoni0 @cshung

### DIFF
--- a/src/benchmarks/gc/src/analysis/chart_utils.py
+++ b/src/benchmarks/gc/src/analysis/chart_utils.py
@@ -195,13 +195,6 @@ class BasicLineChart:
     def n_values(self) -> int:
         return self.lines[0].n_values
 
-    def __post_init__(self) -> None:
-        for line in self.lines:
-            assert line.n_values == self.n_values, (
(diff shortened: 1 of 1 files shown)


Maoni0:
this would chart however many datapoints each trace has right? like if trace 1 has 1000 datapoints and trace 2 has 2000, they both will show up in the same chart.

ivdiazsa:
> this would chart however many datapoints each trace has right? like if trace 1 has 1000 datapoints and trace 2 has 2000, all 3000 will show up in the same chart.

That's right. For example, if Trace1 has 45 points and Trace2 has 49, the chart's X-axis will go up until 50. Trace1's line will end at 45, and the remaning 4 x-values will only have points from Trace2.

Mismatched-Traces

ivdiazsa:

PR #1288, opened 2020-04-17
Adds Microsoft.Extensions.Primitives benchmarks to perf repo

maryamariyan:
Contains history of files also.

cc: @adamsitnik

PR #1360, opened 2020-06-16
Add Remaining Extensions performance tests

maryamariyan:
cc: @adamsitnik @JunTaoLuo 

PR https://github.com/dotnet/performance/pull/1288 would add Microsoft.Extensions.Primitives perf tests.

This PR adds DI, Logging and Http perf tests too.


src/benchmarks/micro/libraries/Microsoft.Extensions.DependencyInjection/ServiceProviderEngineBenchmark.cs
```diff
@@ -5,19 +5,21 @@
...
     {
-        internal ServiceProviderMode ServiceProviderMode { get; private set; }
-
-        [Params("Expressions", "Dynamic", "Runtime", "ILEmit")]
-        public string Mode {
-            set {
-                ServiceProviderMode = (ServiceProviderMode)Enum.Parse(typeof(ServiceProviderMode), value);
-            }
-        }
+        // internal ServiceProviderMode ServiceProviderMode { get; private set; }
```
JunTaoLuo, line 15: Hey @halter73 @davidfowl do you know what the history here is? Specifically, are these benchmarks still relevant? Given I don't have much context here, I'm inclined to keep the benchmarks for each of the DI modes, but if these are no longer useful in any way, we can remove them.

halter73: >  Given I don't have much context here, I'm inclined to keep the benchmarks for each of the DI modes, but if these are no longer useful in any way, we can remove them.

I'm inclined to keep them too. Even if we aren't doing active perf work on DI now, it could be useful later. Is there some reason we cannot make Microsoft.Extensions.DependencyInjection's internals visable to this project?

JunTaoLuo: The problem is that the benchmarks app isn't strong named so IVT doesn't work. But I think what we should do here is turn strong name on, instead of removing the benchmarks. @maryamariyan is looking into do that now.

maryamariyan:
closed in favor of https://github.com/dotnet/performance/pull/1362

(it's odd, push force didnt update the PR so I tried close and re-opening the PR but now I cant reopen the PR for whatever reason).

src/benchmarks/micro/libraries/Microsoft.Extensions.DependencyInjection/ServiceProviderEngineBenchmark.cs
maryamariyan: moved over the PR to https://github.com/dotnet/performance/pull/1362/commits/48ba3c46ecd4e62b5ab27fd012fdd0880a0797e0#r441228858

PR #1363, opened 2020-06-17
specify output folder for crossgen tests

ooooolivia:
Crossgen2 and Crossgen tests share the same payload so we shouldn't output artifacts to the payload folder directly. Specifically, crossgen2 composite scenario uses ```*.dll``` to include assemblies in Core_Root to compile and ```*.ni.dll``` from crossgen should be excluded from the list. Changing crossgen's output to a subfolder here to avoid this problem.


ooooolivia:
validation build in runtime repo: https://dev.azure.com/dnceng/public/_build/results?buildId=691985&view=results

billwert:
There's no reason the output folder needs to be in `core_root`. You could put it under the work item instead and then it gets cleaned up?

PR #1366, opened 2020-06-18
Complete the phase implementation

PeterSolMS:
This completes the phase implementation - changes are:
- parser changes (fix asserts, accept whitespace at the beginning of a line so phases can be indented, etc.)
- implement GotoNextPhase()
- some init code needed to move from the MemoryAlloc constructor to GotoNextPhase()

Phases are only supported when parsing options from a file instead of from the command line.

Here is a sample file with phases:

threadCount 1
[phase]
  totalAllocGB 10
  totalLiveGB 1
  [bucket]
    survInterval 30
    pinInterval 0
[phase]
  totalAllocGB 10
  totalLiveGB 1
  [bucket]
    survInterval 300
    pinInterval 0

This specifies two phases with equal allocation and live data, but different SOH survival intervals.


ivdiazsa:
Just tested this and it seems to work fine. The code looks well too, but I have a question. Could we omit all the messages it prints? In the beginning, it shows each phase's stats such as _time_, _tlgb_, _tagb_, and so on. Then, it prints lots of messages with the format "Thread # stopping phase after xMB". What would you say about keeping this under an `#if DEBUG` or something like that?

Other than the aforementioned observation, it looks ready. What do you think @Maoni0 ?
Once Maoni approves, we can merge this. Thanks a lot for doing this Peter!

ivdiazsa:
@billwert This PR is ready to merge.

PR #1373, opened 2020-06-23
add --disasm-diff argument

adamsitnik:
@DrewScoggins I've added a new `--disasm-diff` command line argument and extended the `PerfLabExporter` to build a string that contains the formatted disassembly. I had to use some reflection because some of the BDN types were internal. I'll later expose them in BDN so they become public in the next release.


adamsitnik:
Closing in favor of #1379

PR #1376, opened 2020-06-25
remove jobs on perfsnakes

ooooolivia:
Removing jobs running on PerfSnakes under "scheduled private job" because 
1) we are retiring perfsnakes 
2) we are running these tests under "private job" with a more frequent schedule


ooooolivia:
I think we are good to remove the same set of tests on Windows for the same reason? https://github.com/dotnet/performance/blob/ff5d8c3e9832f7960165aa7c50449e51513de4fe/azure-pipelines.yml#L413-L471

PR #1378, opened 2020-06-25
Add category for not running tests on interpreter

DrewScoggins:


DrewScoggins:
FYI @adamsitnik

billwert:
after I clicked approve it occurred to me that we have a spot for this string [here](https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/Categories.cs). Maybe put it there?

DrewScoggins:
Yes, I meant to do that, but forgot. Will do.

tannergooding:
What's the reason for not wanting to run certain tests under the interpreter?

DrewScoggins:
All of these tests take over twenty minutes to complete with some taking over two hours. As a result I have already given this list to the mono folks so they can investigate, but in the meantime to increase throughput and allow us to have sane timeouts I want to take these tests out of rotation.

DrewScoggins:
FYI @SamMonoRT

PR #1381, opened 2020-06-26
Add benchmarks for Utf8JsonReader.Get*

Jozkee:
Given that https://github.com/dotnet/runtime/pull/38056 touches several `Utf8JsonReader.Get*` methods, I wanted to make sure that performance was not being affected and wrote these benchmarks.

The results of these benchmarks can be used as a reference for future changes.

cc @layomia @steveharter

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
+++ b/src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
@@ -10,20 +10,55 @@ namespace System.Text.Json.Tests
     [BenchmarkCategory(Categories.Libraries, Categories.JSON)]
     public class Perf_Get
     {
-        private static readonly byte[] _jsonSByteBytes = Encoding.UTF8.GetBytes(sbyte.MaxValue.ToString());
-        private static readonly byte[] _jsonInt16Bytes = Encoding.UTF8.GetBytes(short.MaxValue.ToString());
-        private static readonly byte[] _jsonInt32Bytes = Encoding.UTF8.GetBytes(int.MaxValue.ToString());
-        private static readonly byte[] _jsonInt64Bytes = Encoding.UTF8.GetBytes(long.MaxValue.ToString());
+        private byte[] _jsonSByteBytes;
+        private byte[] _jsonInt16Bytes;
+        private byte[] _jsonInt32Bytes;
+        private byte[] _jsonInt64Bytes;
 
+        private byte[] _jsonSingleBytes;
+        private byte[] _jsonDoubleBytes;
+        private byte[] _jsonDecimalBytes;
 
-        private static readonly byte[] _jsonSingleBytes = Encoding.UTF8.GetBytes(float.MaxValue.ToString());
-        private static readonly byte[] _jsonDoubleBytes = Encoding.UTF8.GetBytes(double.MaxValue.ToString());
-        private static readonly byte[] _jsonDecimalBytes = Encoding.UTF8.GetBytes(decimal.MaxValue.ToString());
+        private byte[] _jsonStringBytes;
+        private byte[] _jsonGuidBytes;
+        private byte[] _jsonDateTimeBytes;
+        private byte[] _jsonDateTimeOffsetBytes;
 
-        private static readonly byte[] _jsonGuidBytes = Encoding.UTF8.GetBytes($"\"{Guid.Empty}\"");
-        private static readonly byte[] _jsonDateTimeBytes = Encoding.UTF8.GetBytes($"\"{DateTime.MaxValue:O}\"");
-        private static readonly byte[] _jsonDateTimeOffsetBytes = Encoding.UTF8.GetBytes($"\"{DateTimeOffset.MaxValue:O}\"");
-        private static readonly byte[] _jsonStringBytes = Encoding.UTF8.GetBytes("\"The quick brown fox jumps over the lazy dog.\"");
+        [Params(true, false)]
+        public bool UseLargeValue;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            if (UseLargeValue)
+            {
+                _jsonSByteBytes = Encoding.UTF8.GetBytes(sbyte.MaxValue.ToString());
+                _jsonInt16Bytes = Encoding.UTF8.GetBytes(short.MaxValue.ToString());
+                _jsonInt32Bytes = Encoding.UTF8.GetBytes(int.MaxValue.ToString());
+                _jsonInt64Bytes = Encoding.UTF8.GetBytes(long.MaxValue.ToString());
+
+                _jsonSingleBytes = Encoding.UTF8.GetBytes(float.MaxValue.ToString("F"));
+                _jsonDoubleBytes = Encoding.UTF8.GetBytes(double.MaxValue.ToString("F"));
+                _jsonDecimalBytes = Encoding.UTF8.GetBytes(decimal.MaxValue.ToString("F"));
+                _jsonStringBytes = Encoding.UTF8.GetBytes($"\"{new string('a', 1024 * 2)}\"");
+            }
+            else
+            {
+                _jsonSByteBytes = Encoding.UTF8.GetBytes(default(sbyte).ToString());
+                _jsonInt16Bytes = Encoding.UTF8.GetBytes(default(short).ToString());
+                _jsonInt32Bytes = Encoding.UTF8.GetBytes(default(int).ToString());
+                _jsonInt64Bytes = Encoding.UTF8.GetBytes(default(long).ToString());
+
+                _jsonSingleBytes = Encoding.UTF8.GetBytes(default(float).ToString());
+                _jsonDoubleBytes = Encoding.UTF8.GetBytes(default(double).ToString());
+                _jsonDecimalBytes = Encoding.UTF8.GetBytes(default(decimal).ToString());
+                _jsonStringBytes = Encoding.UTF8.GetBytes("\"The quick brown fox jumps over the lazy dog.\"");
+            }
+
+            _jsonGuidBytes = Encoding.UTF8.GetBytes($"\"{Guid.Empty}\"");
+            _jsonDateTimeBytes = Encoding.UTF8.GetBytes($"\"{DateTime.MaxValue:O}\"");
+            _jsonDateTimeOffsetBytes = Encoding.UTF8.GetBytes($"\"{DateTimeOffset.MaxValue:O}\"");
+        }
 
         [Benchmark]
         public byte GetByte()
--- a/src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
+++ b/src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
@@ -16,14 +16,14 @@ namespace System.Text.Json.Tests
         private static readonly byte[] _jsonInt64Bytes = Encoding.UTF8.GetBytes(long.MaxValue.ToString());
 
 
-        private static readonly byte[] _jsonSingleBytes = Encoding.UTF8.GetBytes(float.MaxValue.ToString("F"));
-        private static readonly byte[] _jsonDoubleBytes = Encoding.UTF8.GetBytes(double.MaxValue.ToString("F"));
-        private static readonly byte[] _jsonDecimalBytes = Encoding.UTF8.GetBytes(decimal.MaxValue.ToString("F"));
+        private static readonly byte[] _jsonSingleBytes = Encoding.UTF8.GetBytes(float.MaxValue.ToString());
+        private static readonly byte[] _jsonDoubleBytes = Encoding.UTF8.GetBytes(double.MaxValue.ToString());
+        private static readonly byte[] _jsonDecimalBytes = Encoding.UTF8.GetBytes(decimal.MaxValue.ToString());
 
         private static readonly byte[] _jsonGuidBytes = Encoding.UTF8.GetBytes($"\"{Guid.Empty}\"");
         private static readonly byte[] _jsonDateTimeBytes = Encoding.UTF8.GetBytes($"\"{DateTime.MaxValue:O}\"");
         private static readonly byte[] _jsonDateTimeOffsetBytes = Encoding.UTF8.GetBytes($"\"{DateTimeOffset.MaxValue:O}\"");
-        private static readonly byte[] _jsonStringBytes = Encoding.UTF8.GetBytes($"\"{new string('a', 1024*2)}\"");
+        private static readonly byte[] _jsonStringBytes = Encoding.UTF8.GetBytes("\"The quick brown fox jumps over the lazy dog.\"");
 
         [Benchmark]
         public byte GetByte()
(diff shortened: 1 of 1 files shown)


Jozkee:
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18362.900 (1903/May2019Update/19H1)
Intel Core i7-9750H CPU 2.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.100-preview.6.20314.3
  [Host]     : .NET Core 5.0.0 (CoreCLR 5.0.20.30506, CoreFX 5.0.20.30506), X64 RyuJIT
  Job-QHAWRL : .NET Core 5.0 (CoreCLR 42.42.42.42424, CoreFX 42.42.42.42424), X64 RyuJIT

PowerPlanMode=00000000-0000-0000-0000-000000000000  Arguments=/p:DebugType=portable  Toolchain=CoreRun  
IterationTime=250.0000 ms  MaxIterationCount=20  MinIterationCount=15  
WarmupCount=1  

```
|            Method |        Mean |     Error |    StdDev |      Median |         Min |         Max |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------------ |------------:|----------:|----------:|------------:|------------:|------------:|-------:|-------:|------:|----------:|
|           GetByte |    54.46 ns |  0.578 ns |  0.512 ns |    54.29 ns |    53.83 ns |    55.56 ns |      - |      - |     - |         - |
|          GetSByte |    54.99 ns |  0.394 ns |  0.349 ns |    55.00 ns |    54.34 ns |    55.58 ns |      - |      - |     - |         - |
|          GetInt16 |    58.55 ns |  0.565 ns |  0.528 ns |    58.54 ns |    57.77 ns |    59.48 ns |      - |      - |     - |         - |
|          GetInt32 |    62.45 ns |  0.365 ns |  0.324 ns |    62.43 ns |    61.95 ns |    63.13 ns |      - |      - |     - |         - |
|          GetInt64 |    82.53 ns |  0.933 ns |  0.873 ns |    82.56 ns |    81.05 ns |    84.13 ns |      - |      - |     - |         - |
|         GetUInt16 |    57.21 ns |  0.584 ns |  0.547 ns |    57.17 ns |    56.09 ns |    58.25 ns |      - |      - |     - |         - |
|         GetUInt32 |    61.11 ns |  0.492 ns |  0.436 ns |    61.00 ns |    60.52 ns |    61.97 ns |      - |      - |     - |         - |
|         GetUInt64 |    78.46 ns |  1.049 ns |  0.930 ns |    78.22 ns |    77.30 ns |    79.98 ns |      - |      - |     - |         - |
|         GetSingle |   325.17 ns |  3.718 ns |  3.478 ns |   324.31 ns |   320.09 ns |   331.08 ns |      - |      - |     - |         - |
|         GetDouble | 2,402.32 ns | 20.865 ns | 19.517 ns | 2,396.84 ns | 2,368.23 ns | 2,434.65 ns |      - |      - |     - |         - |
|        GetDecimal |   190.52 ns |  2.658 ns |  2.486 ns |   190.72 ns |   187.13 ns |   196.45 ns |      - |      - |     - |         - |
|       GetDateTime |    92.93 ns |  1.211 ns |  1.074 ns |    92.51 ns |    91.80 ns |    95.24 ns |      - |      - |     - |         - |
| GetDateTimeOffset |   119.01 ns |  1.944 ns |  1.723 ns |   118.39 ns |   117.04 ns |   122.62 ns |      - |      - |     - |         - |
|           GetGuid |   119.35 ns |  1.567 ns |  1.466 ns |   119.42 ns |   117.06 ns |   122.31 ns |      - |      - |     - |         - |
|         GetString |   459.04 ns |  5.955 ns |  5.570 ns |   457.71 ns |   452.04 ns |   467.47 ns | 0.6551 | 0.0092 |     - |    4120 B |

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -0,0 +1,148 @@
...
+    public class Perf_Get
+    {
+        private static readonly byte[] _jsonSByteBytes = Encoding.UTF8.GetBytes(sbyte.MaxValue.ToString());
+        private static readonly byte[] _jsonInt16Bytes = Encoding.UTF8.GetBytes(short.MaxValue.ToString());
+        private static readonly byte[] _jsonInt32Bytes = Encoding.UTF8.GetBytes(int.MaxValue.ToString());
+        private static readonly byte[] _jsonInt64Bytes = Encoding.UTF8.GetBytes(long.MaxValue.ToString());
+
+
+        private static readonly byte[] _jsonSingleBytes = Encoding.UTF8.GetBytes(float.MaxValue.ToString("F"));
+        private static readonly byte[] _jsonDoubleBytes = Encoding.UTF8.GetBytes(double.MaxValue.ToString("F"));
```
layomia, line 20:
```suggestion
        private static readonly byte[] _jsonDoubleBytes = Encoding.UTF8.GetBytes(double.MaxValue.ToString());
```

This "F" format representation for this value is quite long, and it shows in the results for `double` compared to the other number types.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -0,0 +1,148 @@
...
+        private static readonly byte[] _jsonDecimalBytes = Encoding.UTF8.GetBytes(decimal.MaxValue.ToString("F"));
+
+        private static readonly byte[] _jsonGuidBytes = Encoding.UTF8.GetBytes($"\"{Guid.Empty}\"");
+        private static readonly byte[] _jsonDateTimeBytes = Encoding.UTF8.GetBytes($"\"{DateTime.MaxValue:O}\"");
+        private static readonly byte[] _jsonDateTimeOffsetBytes = Encoding.UTF8.GetBytes($"\"{DateTimeOffset.MaxValue:O}\"");
+        private static readonly byte[] _jsonStringBytes = Encoding.UTF8.GetBytes($"\"{new string('a', 1024*2)}\"");
```
layomia, line 26: I think this string is too long. I think a string like the "The quick brown fox jumps over the lazy dog" pangram would do here, and be closer in length to strings founds in typical payloads.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
Jozkee: My point was to get the perf. result in the worst case scenario, why we should not aim to that?

layomia: If we had to pick one, I think it's more valuable to benchmark the most common scenarios. A 312 digit number would be a rare occurrence in a JSON payload.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
DrewScoggins: Hmm, I think it might be useful to have both here. A longer string is going to give us protection against any bad scaling that might be introduced that we could miss with a shorter string.

Jozkee: We could add
```cs
[Params(true, false)]
public bool UseLargeValue;
```
So we can test both scenarios, on `true` we use the `MaxValue` and on `false` we can use `default`.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -0,0 +1,183 @@
...
+            _jsonDateTimeOffsetBytes = Encoding.UTF8.GetBytes($"\"{DateTimeOffset.MaxValue:O}\"");
+        }
+
+        [Benchmark]
+        public byte GetByte()
+        {
+            var reader = new Utf8JsonReader(_jsonSByteBytes);
+            reader.Read();
+            return reader.GetByte();
+        }
```
adamsitnik, line 69: @Jozkee do you know what part of the total time is spent for the creation of `Utf8JsonReader`, calling `Read` and calling `GetByte()`?  What is the typical usage of `reader.GetX()` methods? 

I am asking because the creation of the reader and call to `reader.Read` might be more expensive than the call to `GetX` which we are trying to measure here. If this is true, we should reconsider rewriting the benchmark to something like:

```cs
public sbyte ReadAndGetSBytes()
{
    var reader = new Utf8JsonReader(_jsonSByteBytes); // _jsonSByteBytes should contain more than 1 SByte (100 should be enough)
    sbyte result = default;
    while (reader.Read())
    {
        result += reader.GetSByte();
    }
    return result;
}
```

this should allow for removal of `UseLargeValue` and hence reducing the number of new benchmarks and the time it takes to run them by half. Please see https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#Test-Cases for more

Jozkee: Good point, for Int32, ~80% time is spent in the first two lines:
```cs
[Benchmark]
public void CreateAndReadInt32()
{
    var reader = new Utf8JsonReader(_jsonInt32Bytes);
    reader.Read();
}
```

|             Method | UseLargeValue |     Mean |    Error |   StdDev |   Median |      Min |      Max | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------- |-------------- |---------:|---------:|---------:|---------:|---------:|---------:|------:|------:|------:|----------:|
| **CreateAndReadInt32** |         **False** | **39.95 ns** | **0.548 ns** | **0.512 ns** | **39.81 ns** | **39.42 ns** | **41.11 ns** |     **-** |     **-** |     **-** |         **-** |
|           GetInt32 |         False | 46.41 ns | 0.302 ns | 0.283 ns | 46.47 ns | 45.90 ns | 46.81 ns |     - |     - |     - |         - |
| **CreateAndReadInt32** |          **True** | **49.73 ns** | **0.211 ns** | **0.187 ns** | **49.71 ns** | **49.48 ns** | **50.02 ns** |     **-** |     **-** |     **-** |         **-** |
|           GetInt32 |          True | 62.31 ns | 0.647 ns | 0.573 ns | 62.42 ns | 60.74 ns | 63.15 ns |     - |     - |     - |         - |

Jozkee: Good point, for `GetInt32`, ~80% time is spent in the first two lines:
```cs
[Benchmark]
public void CreateAndReadInt32()
{
    var reader = new Utf8JsonReader(_jsonInt32Bytes);
    reader.Read();
}
```

|             Method | UseLargeValue |     Mean |    Error |   StdDev |   Median |      Min |      Max | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------- |-------------- |---------:|---------:|---------:|---------:|---------:|---------:|------:|------:|------:|----------:|
| **CreateAndReadInt32** |         **False** | **39.95 ns** | **0.548 ns** | **0.512 ns** | **39.81 ns** | **39.42 ns** | **41.11 ns** |     **-** |     **-** |     **-** |         **-** |
|           GetInt32 |         False | 46.41 ns | 0.302 ns | 0.283 ns | 46.47 ns | 45.90 ns | 46.81 ns |     - |     - |     - |         - |
| **CreateAndReadInt32** |          **True** | **49.73 ns** | **0.211 ns** | **0.187 ns** | **49.71 ns** | **49.48 ns** | **50.02 ns** |     **-** |     **-** |     **-** |         **-** |
|           GetInt32 |          True | 62.31 ns | 0.647 ns | 0.573 ns | 62.42 ns | 60.74 ns | 63.15 ns |     - |     - |     - |         - |

Jozkee: FWIW I added `ReadToEnd*` benchmarks in order to measure how much time is spent in traversing the 100 elements of the json.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -0,0 +1,265 @@
...
+
+        [Benchmark]
+        public float GetSingle()
+        {
+            float result = 0;
+            var reader = new Utf8JsonReader(_jsonNumberBytes);
+            reader.Read();
+            while (reader.Read() && reader.TokenType == JsonTokenType.Number)
+            {
+                result += reader.GetSingle();
```
adamsitnik, line 182: the input for `GetSingle` and `GetDouble()` should be sth more than `0.0` or `123.0`. Perhaps `123.456` would be a better number? Just to make sure it performs work that is closer to real-life scenarios

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -0,0 +1,265 @@
...
+            while (reader.Read() && reader.TokenType == JsonTokenType.String) ;
+        }
+
+        [Benchmark]
+        public void ReadToEndDateTimeOffset()
+        {
+            var reader = new Utf8JsonReader(_jsonDateTimeOffsetBytes);
+            reader.Read();
+            while (reader.Read() && reader.TokenType == JsonTokenType.String) ;
+        }
```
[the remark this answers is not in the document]
Jozkee: The JSON strings will be of different size depending on the type and its respective `_json*Bytes`.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -0,0 +1,265 @@
...
+            sb.Append(']');
+
+            return Encoding.UTF8.GetBytes(sb.ToString());
+        }
+
+        [Benchmark]
+        public void ReadToEndNumber()
+        {
+            var reader = new Utf8JsonReader(_jsonNumberBytes);
+            reader.Read();
```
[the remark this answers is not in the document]
Jozkee: Yes.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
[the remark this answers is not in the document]
adamsitnik: If the only difference is the string size, please keep just one of them. We want to have 1 benchmark per code path https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#Code-Paths

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
Jozkee: Another approach that we could take to avoid having the `ReadtoEnd*` methods could be just call `reader.GetInt3()2` over the same token a hundred times, that could remove the overhead of `reader.Read()`.

PR #1382, opened 2020-06-26
Fixing GCPerfSim bugs

cshung:
- When we call ` GC.AllocateArray<byte>`, make sure we set `isPinned` to `true` to make the allocation in the `POH`.
- Set the default `POH` allocation rate to 0.


cshung:
@brianrob

Maoni0:
actually I have permission to merge now so I will do it (this was before I got permission).

PR #1388, opened 2020-07-02
Added full suites documentation to GC Infra

ivdiazsa:
This PR adds missing documentation regarding the suite commands of GC Benchmarking Infra:

- suite-run
- suite-run-command

This is done in two sections:

1. Added how to run a full suite in the main README.md, as this is a fairly common use case. Linked to the new file containing the full doc.
2. Added a new doc detailing with examples how to use `suite-run` and `suite-run-command`, as well as how `suite.yaml` should be written, and the implications.


ivdiazsa:
@Maoni0 PTAL

src/benchmarks/gc/README.md
```diff
@@ -187,6 +192,34 @@ Each trace file will be named `{coreclr_name}__{config_name}__{benchmark_name}__
 
 _ARM NOTE_: Container tests are not supported on ARM/ARM64.
 
+### Running the Entire Suite
```
Maoni0, line 195: I would definitely put this as the first thing in the "Running" section and mentioned that this is what you should always run to make sure you don't regress perf if you are making a perf change in GC.

src/benchmarks/gc/README.md
```diff
@@ -150,7 +150,12 @@ The configuration values that can be used in the test `.yaml` file are described
 
 ## Running
 
-Running the full suite would take a while, so let's just run one:
+The benchmark tests created in the previous step can be run as a whole bundle, or individually.
+This is explained in the following sections.
+
+### Running a Single Test
```
Maoni0, line 156: I probably wouldn't call this a single *test* since each .yaml could include more than one test. perhaps call it a "scenario"? eg, this is the low memory container scenario and we run multiple tests for this scenario

src/benchmarks/gc/docs/suites.md
```diff
@@ -0,0 +1,127 @@
...
+- <Benchmark Name> -
+<Iteration Number>
+<Error Message>
+
+<Stack Trace>
+
+<Repeat with other errors that might have occurred>
+```
+
+## Suite Run Command
```
Maoni0, line 85: interesting, I had thought that we just ran the command_groups after running each scenario 'cause otherwise it seems a little odd to have it in the same .yaml if it's a complete different thing. is there any relation between command_groups and the rest in suite.yaml?

src/benchmarks/gc/README.md
ivdiazsa: Sounds good. I also forgot to add that to the `Contributing` doc. Will do that as well.

src/benchmarks/gc/docs/suites.md
ivdiazsa: I thought the `command_groups` were part of the `suite-run` as such, but after looking into the source code and running various experiments, I learned they are completely different. Maybe they were added here as to bundle all suite-like stuff in one file?

Maoni0: that seems like a weird design... I don't have a better explanation than what you have.

src/benchmarks/gc/README.md
ivdiazsa: This will require more rephrasing of other parts, especially in this section. Let me draft it and submit it later.

ivdiazsa:
@Maoni0 PTAL at the reordered and slightly reworded doc.

PR #1389, opened 2020-07-03
Fix CoreRun.exe path

felipepessoto:


adamsitnik:
Hi @felipepessoto 

Big thanks for improving our docs!

Could you please update the https://github.com/dotnet/performance/blob/master/docs/profiling-workflow-dotnet-runtime.md doc as well?

Thank you,
Adam

felipepessoto:
Sure. It is done

PR #1390, opened 2020-07-08
Enable dynamic event tracing and Crossgen2 parser on Linux

ooooolivia:


src/tools/ScenarioMeasurement/Startup/Crossgen2Parser.cs
```diff
@@ -60,60 +67,109 @@ public IEnumerable<Counter> Parse(string mergeTraceFile, string processName, ILi
     public sealed class EventParser
     {
         public string EventName { get; private set; }
+        public (int StartEventID, int StopEventID) EventID { get; private set; } // (StartEventID, StopEventID)
```
billwert, line 70: I think this is an odd spot to use a tuple. I'd probably make this two separate fields.

src/tools/ScenarioMeasurement/Startup/Crossgen2Parser.cs
```diff
@@ -20,29 +21,35 @@ public void EnableUserProviders(ITraceSession user)
...
-            var loadingParser = new EventParser("Loading");
-            var emittingParser = new EventParser("Emitting");
-            var jittingParser = new EventParser("Jit");
-            var compilationParser = new EventParser("Compilation");
-
+            var loadingParser = new EventParser("Loading", (1, 2));
+            var emittingParser = new EventParser("Emitting", (3, 4));
+            var compilationParser = new EventParser("Compilation", (5, 6));
+            var jittingParser = new EventParser("Jit", (7, 8));
+
```
billwert, line 28: Since this is the crossgen2 parser, can we have an enum with symbolic names for these numbers? It can match whatever is used in the EventSource is in crossgen2.

billwert:
(Also please check on the CI failures, I didn't look at them.)

PR #1393, opened 2020-07-11
Modify linker parameter

ooooolivia:
Linker has updated its dump file property in msbuild since late .NET 5 preview 7
Update it to unblock maestro update https://github.com/dotnet/performance/pull/1385


ooooolivia:
Merge to unblock maestro

PR #1396, opened 2020-07-14
Add microbenchmarks for System.Text.Encodings.Web

eiriktsarpalis:
Adds a rudimentary set of microbenchmarks for System.Text.Encodings.Web, to better track AdvSimd performance improvements. Cf. https://github.com/dotnet/runtime/pull/38707


src/benchmarks/micro/libraries/System.Text.Encodings.Web/Perf.DefaultJavaScriptEncoder.cs
```diff
@@ -0,0 +1,59 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+#if !NETFRAMEWORK
```
tannergooding, line 4: Would a conditional include at the project file level be better?

eiriktsarpalis: Done

src/benchmarks/micro/libraries/System.Text.Encodings.Web/Perf.UrlEncoder.cs
```diff
@@ -0,0 +1,58 @@
...
+using MicroBenchmarks;
+using System.Text.Encodings.Web;
+
+namespace System.Text.Tests
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public class Perf_UrlEncoder
+    {
+        [Params("&lorem ipsum=dolor sit amet")]
+        public string sourceString;
```
adamsitnik, line 15: why is the `sourceString` different for all 3 types? if we want to have apples-to-apples comparison, it should be the same

eiriktsarpalis: They are different encoders escaping different characters, so the variation in inputs exists merely to induce some escaping in all of them. The three types have very different implementations, so this is not (and should not be) attempting to compare them to each other.

src/benchmarks/micro/libraries/System.Text.Encodings.Web/Perf.DefaultJavaScriptEncoder.cs
```diff
@@ -0,0 +1,58 @@
...
+        public void SetupGetBytes()
+        {
+            _encoder = JavaScriptEncoder.Default;
+            _sourceBufferUtf16 = BuildSourceString();
+            _destinationBufferUtf16 = new char[paddingSize + 10 * sourceString.Length];
+
+            _sourceBufferUtf8 = Encoding.UTF8.GetBytes(_sourceBufferUtf16);
+            _destinationBufferUtf8 = new byte[paddingSize + 10 * sourceString.Length];
+            
+            string BuildSourceString()
```
kunalspathak, line 38: >BuildSourceString [](start = 19, length = 17)

can we have a common method somewhere that takes parameters from each of the encoders test?

src/benchmarks/micro/libraries/System.Text.Encodings.Web/Perf.UrlEncoder.cs
```diff
@@ -0,0 +1,58 @@
...
+            _destinationBufferUtf8 = new byte[paddingSize + 10 * sourceString.Length];
+
+            string BuildSourceString()
+            {
+                var sb = new StringBuilder();
+                // pad the string with `paddingSize` non-escapable ascii characters
+                var random = new Random(42);
+                for (int i = 0; i < paddingSize; i++)
+                {
+                    sb.Append((char)random.Next('a', 'z' + 1));
```
kunalspathak, line 45: There is code path to find non-ascii characters in the string. I didn't see anywhere that you are creating a string with non-ascii characters?

eiriktsarpalis: Addressed

src/benchmarks/micro/libraries/System.Text.Encodings.Web/Perf.DefaultJavaScriptEncoder.cs
eiriktsarpalis: I resisted extracting shared components into a helper class, since it seems that benchmarks are pretty self-contained in this project. @adamsitnik might be able to comment about DRY concerns here.

kunalspathak:
Any other update needed or is it ready to merge?

kunalspathak:
Ping @adamsitnik .

PR #1400, opened 2020-07-20
Refactor System.Collections/Sort and add Span test cases

nietras:
As part of the https://github.com/dotnet/runtime/pull/39543 we need this to fully cover and detail sorting code paths and scenarios. Includes floating point benchmarks for completeness. cc: @jkotas @stephentoub 

This also replaces the same invocations per iteration count for all scenarios with a per test case scenario that meant some of the benchmarks were unreliable or took "too long".

It doesn't appear `NETCOREAPP5_0` is defined, how can we specify this so Span tests only included for that?

### DIFF
--- a/src/benchmarks/micro/libraries/System.Collections/Sort.cs
+++ b/src/benchmarks/micro/libraries/System.Collections/Sort.cs
@@ -37,7 +37,12 @@ namespace System.Collections
         public void CleanupIteration() => _iterationIndex = 0; // after every iteration end we set the index to 0
 
         [IterationSetup(Targets = new []{ nameof(Array), nameof(Array_ComparerClass),
-            nameof(Array_ComparerStruct), nameof(Array_Comparison) })]
+            nameof(Array_ComparerStruct), nameof(Array_Comparison),
+//#if NETCOREAPP5_0
+            nameof(Span), nameof(Span_ComparerClass),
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.Collections/Sort.cs
```diff
@@ -108,6 +345,7 @@ private sealed class ComparableComparerClass : IComparer<T>
 
         private readonly struct ComparableComparerStruct : IComparer<T>
         {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
jkotas, line 348: Why can't JIT inline it? This is tiny method, it does not sound right to require `AggressiveInlining` on it for good performance.

jkotas: (multiple places)

nietras: will remove. old habits 🙄

danmoseley:
@nietras do you want to continue with this PR?

someone from @dotnet/area-system-collections  can review it.

eiriktsarpalis (changes requested):
We generally don't want to change existing benchmark suites as that tends to mess our performance reporting history cc @adamsitnik 

Also, Span benchmarks arguably belong in the `System.Memory` suites and not `System.Collections`. Have you checked if they define tests for sorting?

PR #1404, opened 2020-07-23
Add first features of CPU Samples Analysis to GC Benchmarking Infra

ivdiazsa:
This PR adds the first features of the awaited CPU Samples Analysis functionality to GC Benchmarking Infrastructure, as well as all the necessary documentation on how to run them. These features can be summarized in the following points:

* Chart CPU Samples metrics of one or more functions, for one or more GC's, given one or more traces. A list of supported metrics is described later on.
* Show various CPU Samples metrics, such as inclusive and exclusive sample count, for a function within a given time range.

Thi functionality relies on the use of _GCPerf_, which calls several components from _TraceEvent_.

Currently, this is only supported through the use of Jupyter Notebook by calling their respective functions, as shown in the examples down below.

### Chart CPU Samples

To chart samples metrics per GC's, let's take the following example. Here, we have will chart the number of inclusive CPU Samples of all Gen1 GC's for the functions `gc_heap::plan_phase` and `gc_heap::mark_phase`.

```python
chart_cpu_samples_per_gcs(
    ptraces=(_MY_TRACE,),
    symbol_path=Path("Path/To/PDB/Directory"),
    functions_to_chart=("gc_heap::plan_phase", "gc_heap::mark_phase",),
    x_property_name="gc_index",
    y_property_names=("inclusive_count",),
    gc_filter=lambda gc: gc.Generation == Gens.Gen1,
)
```

This function call results in the following chart as output:

SamplesChart

### Show Samples Metrics

Let's explain this with the following example. We want to know the main samples metrics for the function `gc_heap::plan_phase` between the 1-second mark and the 5-seconds mark of the trace.

```python
show_cpu_samples_metrics(
    ptrace=_MY_TRACE,
    symbol_path=Path("Path/To/PDB/Directory"),
    function="gc_heap::plan_phase",
    start_time_msec=1000.0,
    end_time_msec=5000.0,
)
```

This function call will display the following information:

SamplesMetrics


ivdiazsa:
@Maoni0 @cshung @PeterSolMS @mangod9

Maoni0:
seems unusual to have symbol_path be a parameter to a chart function, does this mean each time you chart, it'll have to go open the trace and decode the symbols again?

ivdiazsa:
> seems unusual to have symbol_path be a parameter to a chart function, does this mean each time you chart, it'll have to go open the trace and decode the symbols again? for example, with GCs, we can get back a list of gcs from _MY_TRACE, then we can get info from these GCs. this is to facilitate the Jupyter Notebook usage. is there an equivalent for the CPU samples?

Not right now. Although analyzing the code yesterday after submitting the PR, I was actually thinking about creating a new object to hold on to the TraceLog, the SymbolReader, and the full StackSource objects. That way, we can just pass that around instead for each chart, instead of reading the trace every time.

ivdiazsa:
Updated the way of running CPU Samples Analysis. Instead of reading and resolving the traces every time a chart or metrics information are called, we now resolve it in a separate object, which then is passed around to do the filtering as necessary for all the charts the user might want to do.

This slightly changes the steps to run the analysis, as explained below. The documentation was also updated accordingly to match this new behavior.

First, read the trace to get all the GC's information.

```python
_MY_TRACE = get_trace_with_everything("Path/To/Test/Status/File.yaml")
```

Next, create the utilities object, which parses and resolves the trace automatically.

```python
_MY_TRACE_ALL_DATA = TraceReadAndParseUtils(
    ptrace=_MY_TRACE,
    symbol_path=Path("Path/To/PDB/Directory"),
)
```

Once this is done, one can now run the two currently supported features of CPU Samples Analysis.

### Chart CPU Samples

To chart samples metrics per GC's, let's take the following example. Here, we have will chart the number of inclusive CPU Samples of all Gen1 GC's for the functions `gc_heap::plan_phase` and `gc_heap::mark_phase`.

```python
chart_cpu_samples_per_gcs(
    ptraces=(_MY_TRACE,),
    symbol_path=Path("Path/To/PDB/Directory"),
    functions_to_chart=("gc_heap::plan_phase", "gc_heap::mark_phase",),
    x_property_name="gc_index",
    y_property_names=("inclusive_count",),
    gc_filter=lambda gc: gc.Generation == Gens.Gen1,
)
```

This function call results in the following chart as output:

SamplesChart

### Show Samples Metrics

Let's explain this with the following example. We want to know the main samples metrics for the function `gc_heap::plan_phase` between the 1-second mark and the 5-seconds mark of the trace.

```python
show_cpu_samples_metrics(
    ptrace=_MY_TRACE,
    symbol_path=Path("Path/To/PDB/Directory"),
    function="gc_heap::plan_phase",
    start_time_msec=1000.0,
    end_time_msec=5000.0,
)
```

This function call will display the following information:

SamplesMetrics

Maoni0:
this looks great! I will give it a try later today. instead of having to copy&paste code from the doc for jupyter notebook, could you check in your modified version of jupyter_notebook.py?

danmosemsft:
@ivdiazsa just curious, did you consider using .NET in your Jupyter notebook? https://github.com/dotnet/interactive

Maoni0:
@danmosemsft using C# notebooks is actually a work item on Ivan's list. the charting functionality we needed just recently became available in VS code. obviously we'd love to get rid of the python part if we can because the interop between C# and Python has caused us problems before.

danmosemsft:
Cool just curious. Whatever gets the job done!

ivdiazsa:
CPU Samples Analysis example has been added to Jupyter Notebook! Also, the linter updated the `dead_code.py` file, as some pieces of code it complained about not being used, are either essential for debugging, or actually used behind the scenes, so removing them would be inconvenient and/or even potentially dangerous. Some reformatting with the linter was also done.

Maoni0:
this will need more work to actually optimize for the Jupyter Notebook scenario (which is the main use of this feature) and @ivdiazsa will do that in a separate PR.

PR #1406, opened 2020-07-23
add retry loop around downloading dotnet-install

billwert:


billwert:
It turns out that 404s from this endpoint redirect to the homepage of https://dot.net, so instead of relying on HTTP status codes I'm going to assume that if we get HTML back we didn't get what we want.

DrewScoggins:
LGTM

PR #1413, opened 2020-07-27
add version number as additional data

ooooolivia:
- add ```AdditionalData``` field to Build class
- add dotnet version number as ```AdditionalData```

### DIFF
--- a/src/tools/Reporting/Reporting/Build.cs
+++ b/src/tools/Reporting/Reporting/Build.cs
@@ -27,10 +27,5 @@ namespace Reporting
         public DateTime TimeStamp { get; set; }
 
         public Dictionary<string, string> AdditionalData { get; set; } = new Dictionary<string, string>();
-
-        public void AddData(string key, string payload)
-        {
-            AdditionalData.Add(key, payload);
-        }
     }
 }
--- a/src/tools/Reporting/Reporting/Build.cs
+++ b/src/tools/Reporting/Reporting/Build.cs
@@ -25,5 +25,12 @@ namespace Reporting
         public string BuildName { get; set; }
 
         public DateTime TimeStamp { get; set; }
+
+        public Dictionary<string, string> AdditionalData { get; set; } = new Dictionary<string, string>();
+
+        public void AddData(string key, string payload)
+        {
+            AdditionalData.Add(key, payload);
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/tools/Reporting/Reporting/Build.cs
```diff
@@ -25,5 +25,12 @@ public sealed class Build
         public string BuildName { get; set; }
 
         public DateTime TimeStamp { get; set; }
+
+        public Dictionary<string, string> AdditionalData { get; set; } = new Dictionary<string, string>();
+
+        public void AddData(string key, string payload)
+        {
+            AdditionalData.Add(key, payload);
```
billwert, line 33: Should this have an add/update/etc pattern? What happens if I need to change some set of data?

May be better to just let the callsite manipulate the dictionary.

src/tools/Reporting/Reporting/Build.cs
```diff
@@ -25,5 +25,12 @@ public sealed class Build
         public string BuildName { get; set; }
 
         public DateTime TimeStamp { get; set; }
+
+        public Dictionary<string, string> AdditionalData { get; set; } = new Dictionary<string, string>();
```
billwert, line 29: If you decide you should have the method for interacting with this the dictionary should be private.

src/tools/Reporting/Reporting/Build.cs
ooooolivia: FYI it's following the format of ```Test.cs``` https://github.com/dotnet/performance/blob/602872a6c8da1369c3d522676a50f053af3da4ac/src/tools/Reporting/Reporting/Test.cs#L43-L46
Unless we have a centralized place for methods manipulating the dictionary somewhere, I'd prefer callsite just accessing the dictionary, otherwise we might end up duplicating these methods for other library classes.  @DrewScoggins

ooooolivia:
PR failure caused by unrelated connection issue https://helix.dot.net/api/2019-06-17/jobs/6984fbff-b6fc-462c-98e8-495f4eed7c49/workitems/Partition0/console

src/tools/Reporting/Reporting/Build.cs
DrewScoggins: I did not put much thought into accessing this when I first added this in Test.cs. I think just moving to a place where we manipulate the dictionary directly is perfectly fine here.

DrewScoggins:
I assume you have run this locally and seen that the outputted json has the additional data field on it, correct?

ooooolivia:
> I assume you have run this locally and seen that the outputted json has the additional data field on it, correct?

Our PR can reflect this change. Link to the json report:
https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-1413-merge-34194f6734f4433aa4/SOD%20-%20New%20Blazor%20Template%20-%20Publish/traces/perf-lab-report.json?sv=2019-02-02&se=2020-08-17T00%3A52%3A13Z&sr=c&sp=rl&sig=5yFy9%2BgbhTQR8JPa07Fn5OeXUWRSb6EgHz5W38%2BgBp8%3D

src/tools/Reporting/Reporting/Build.cs
ooooolivia: This will stay public since we now switched to manipulating the dictionary from callsite.

PR #1424, opened 2020-07-31
Add WASM support for benchmarks_ci.py

DrewScoggins:
Add link to nightly BDN

Move to latest BDN

Fix obsolete property

### DIFF
--- a/scripts/micro_benchmarks.py
+++ b/scripts/micro_benchmarks.py
@@ -128,6 +128,15 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         return []
 
     parser.add_argument(
+        '--wasm',
+        dest='wasm',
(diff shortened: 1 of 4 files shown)


scripts/micro_benchmarks.py
```diff
@@ -127,6 +127,15 @@ def __get_bdn_arguments(user_input: str) -> list:
             return args
         return []
 
+    parser.add_argument(
+        '--wasm',
+        dest='wasm',
+        required=False,
+        default=False,
+        action='store_true',
+        help='''Tests should be run with the wasm runtime'''
```
ooooolivia, line 136: nit: extra quotes

DrewScoggins: That is the python feature for a string literal.

PR #1437, opened 2020-08-06
Add JIT label and label smaller subsets of tests

DrewScoggins:


src/benchmarks/micro/Categories.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Data;
```
billwert, line 5: why?

DrewScoggins: I saw that as well. I have no idea how that got there.

src/benchmarks/micro/runtime/Bytemark/ByteMark.cs
```diff
@@ -381,6 +382,7 @@ public void BenchNumericSortJagged()
     const int NumericSortRectangularIterations = 5;
 
     [Benchmark]
+    [BenchmarkCategory(Categories.JIT, Categories.ByteMark)]
```
kunalspathak, line 385: >BenchmarkCategory [](start = 5, length = 17)

In other places, I have seen this attribute at class level. Can we do the same for this file as well? Or do they have to be at method level?

DrewScoggins: As long as you can have class level annotations and also method level annotations and they are additive I think that moving it up the class level is fine. @adamsitnik

src/benchmarks/micro/libraries/System.Numerics.BitOperations/Perf_BitOperations.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace System.Numerics.Tests
 {
-    [BenchmarkCategory(Categories.Libraries, Categories.SIMD)]
+    [BenchmarkCategory(Categories.Libraries, Categories.SIMD, Categories.JIT)]
     public class Perf_BitOperations
```
kunalspathak, line 14: can you also tag Roslyn\CscBench.cs with "JIT"?

DrewScoggins: Good catch, I forgot to save this file when I pushed the other changes.

src/benchmarks/micro/Categories.cs
```diff
@@ -21,15 +21,20 @@ public static class Categories
...
         /// </summary>
         public const string Runtime = "Runtime";
-            public const string BenchmarksGame = "BenchmarksGame";
-            public const string Benchstones = "Benchstones";
-                public const string BenchF = "BenchF";
-                public const string BenchI = "BenchI";
-            public const string Inlining = "Inlining";
-            public const string V8 = "V8";
-            public const string Perflab = "Perflab";
-            public const string Virtual = "Virtual";
```
adamsitnik, line 31: the goal of the spacing was to show category structure, but I can live without it ;)

src/benchmarks/micro/runtime/Bytemark/ByteMark.cs
```diff
@@ -471,7 +471,6 @@ public void BenchAssignJagged()
     const int AssignRectangularIterations = 5;
 
     [Benchmark]
-    [BenchmarkCategory("TestTestTest")]
```
adamsitnik, line 474: great catch, I have no idea how it got into master branch

src/benchmarks/micro/Categories.cs
```diff
@@ -21,15 +21,20 @@ public static class Categories
...
+        public const string BenchmarksGame = "BenchmarksGame";
+        public const string Benchstones = "Benchstones";
+        public const string BenchF = "BenchF";
+        public const string BenchI = "BenchI";
+        public const string Inlining = "Inlining";
+        public const string V8 = "V8";
+        public const string Perflab = "Perflab";
+        public const string Virtual = "Virtual";
+        public const string ByteMark = "ByteMark";
```
adamsitnik, line 32: the `ByteMark` category is used only by a single type. What is the purpose of adding it?

PR #1439, opened 2020-08-07
Load all GC's Sample Data at first to optimize charting in Jupyter Notebook

ivdiazsa:
This PR made a few changes to the CPU Samples Analysis charting functionality, which was initially added in PR #1404.

Now, the `TraceReadAndParseUtils` reads the trace, and takes a list of functions of which to initialize the CPU Samples data. This was necessary because when working with bigger traces, wait times for a chart could take even a few minutes. With this change, the setup will take a while with these traces due to the enormous amount of data to process, but charts are displayed in seconds, which is optimal when doing experimenting and looking at different pieces of data for performance analysis.

If one forgets, or simply wants to then process samples of other functions, one can call the initializing function with the new functions and the `TraceReadAndParseUtils` object will be updated, so no need to recreate it or restart the Jupyter Notebook.

The documentation was also updated accordingly and some broken links were fixed. Also, I renamed a couple files to match the naming convention used in this repo.


ivdiazsa:
@dotnet/gc

src/benchmarks/gc/README.md
```diff
@@ -20,11 +20,14 @@ NOTE: If running under ARM/ARM64, the program's functionalities are limited to r
...
-This should be fixed in 3.8.1.
+
+WARNING: Python 3.8.0+ is known to cause issues while setting up, so make sure
+you use a 3.7 flavor.
 
 On Windows, just go to https://www.python.org/downloads/ and run the installer.
-It's recommended to install a 64-bit version if possible, but not required.
+For CPU Samples Analysis, it is required to have the 64-bit version. If you're
+not planning to use this functionality, any version works but 64-bit is still
+highly recommended.
```
Maoni0, line 30: might wanna put a simple explanation here (you'd easily get OOM).

ivdiazsa:
@dotnet/gc This PR is now ready for review again.

PR #1445, opened 2020-08-10
Benchmarks for System.Utf8String.Experimental

pgovind:
Putting it for initial review. The `IsNormalized_GetIndexOfFirstNonAsciiChar` and `IsNormalized_WidenAsciiToUtf16` have been setup to really accelerate the perf since those benchmarks don't contain any non-ascii data. A more comprehensive benchmark would need some non-ascii cases as well. 

@kunalspathak @carlossanlop


GrabYourPitchforks:
When I was benchmarking the UTF-8 code I used [these texts](https://github.com/GrabYourPitchforks/fast-utf8/tree/master/FastUtf8Tester/SampleTexts) from Project Gutenberg. These texts were chosen because they represent a nice variety of languages and non-ASCII character distributions.

* 11.txt - English, all-ASCII, should stay entirely within fast paths
* 11-0.txt - English, _mostly_ ASCII with some rare non-ASCII chars, exercises that the occasional non-ASCII char doesn't kill our fast paths
* 25249-0.txt - Chinese, exercises 3-byte scalar processing paths typical of East Asian languages
* 30774-0.txt - Cyrillic, exercises a combination of ASCII and 2-byte scalar processing paths
* 39251-0.txt - Greek, similar to the Cyrillic case but with a different distribution of ASCII and non-ASCII chars

The most important benchmark I ran over these tests was to load them into a ROS<byte> (remember strip the UTF-8 BOM if it exists!) and to run the "is this valid UTF-8?" method over them.

GrabYourPitchforks:
What I did _not_ benchmark: emoji and other 4-byte sequences, as they don't occur commonly enough to merit the optimization work. But we did unit test and fuzz the hell out of it so we have high confidence that the code works. :)

kunalspathak:
> What I did _not_ benchmark: emoji and other 4-byte sequences, as they don't occur commonly enough to merit the optimization work. But we did unit test and fuzz the hell out of it so we have high confidence that the code works. :)

Thanks for the pointers @GrabYourPitchforks . We should do similar thing for `AdvSimd` as well. More important to be functionally correct than performant.

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            expected.IsNormalized();
+            expected.IsNormalized();
+        }
+
+        [Benchmark]
+        [ArgumentsSource(nameof(AsciiData))]
+        public void IsNormalized_WidenAsciiToUtf16(string expected)
+        {
+            Utf8Span span = new Utf8Span(new Utf8String(expected));
```
kunalspathak, line 64: hhm, the return value is not used anywhere. Did you check what happens to the actual calls to `ToCharArray()`? Is your code even getting called?

pgovind: It is. I did explicitly check for that before running the benchmark :) Others may have the same question though, so I'll add some lines here to consume the result of `ToCharArray` in some way

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
```diff
@@ -0,0 +1,86 @@
...
+        public static IEnumerable<string> AsciiData()
+        {
+            yield return ascii_11;
+        }
+
+        [Benchmark]
+        [ArgumentsSource(nameof(NonAsciiData))]
+        public int ToUtf16(string expected)
+        {
+            Utf8Span span = new Utf8Span(new Utf8String(expected));
```
GrabYourPitchforks, line 63: Don't call `new Utf8String(string)` in these benchmarks, as it performs a considerable amount of work and will throw off your numbers. Instead, have a benchmark dedicated just to calling this method. For all other benchmarks, use the `[GlobalSetup]` routine to pre-populate an instance field of type `Utf8String`.

In your `[GlobalSetup]` routine, you should check whether the first char of each string is `'\uFEFF'`, and if so then strip that character off the front of the string (or just throw an exception from within your setup routine if you detect this). That will work around any stray BOM issues that might throw off the benchmark.

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
```diff
@@ -0,0 +1,86 @@
...
+            return span.IsAscii();
+        }
+
+        [Benchmark]
+        [ArgumentsSource(nameof(AsciiData))]
+        public bool IsNormalized_GetIndexOfFirstNonAsciiChar(string expected)
+        {
+            bool b1 = expected.IsNormalized();
```
GrabYourPitchforks, line 81: This isn't testing any `Utf8String` related code paths.

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
pgovind: That's the plan with `GlobalSetup`. This is still WIP. I'm hitting an exception that throws "System.ComponentModel.Win32Exception: 'The filename or extension is too long'" though, so I'm still trying to debug that. 

Yup, I'll strip out the BOM

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
pgovind: Wait are you sure? I see this path: `string.IsNormalized -> string.IsNormalized(NormalizationForm) -> string.IsAscii() -> ASCIIUtility.GetIndexOfFirstNonAsciiChar`

GrabYourPitchforks: Ah, I got confused by the name then. Carry on! :)

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
```diff
@@ -0,0 +1,109 @@
...
+            nonascii_chinese = File.ReadAllText(Path.Combine(path, "25249-0.txt"));
+            nonascii_chinese_ustring = new Utf8String(nonascii_chinese);
+            nonascii_cyrillic = File.ReadAllText(Path.Combine(path, "30774-0.txt"));
+            nonascii_cyrillic_ustring = new Utf8String(nonascii_cyrillic);
+            nonascii_greek = File.ReadAllText(Path.Combine(path, "39251-0.txt"));
+            nonascii_greek_ustring = new Utf8String(nonascii_greek);
+        }
+
+        [Benchmark]
+        public int ToUtf16_nonascii_110()
```
kunalspathak, line 43: If possible, consider having a name of method that you are testing in this benchmark?

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            string str = nonascii_110;
+            Utf8Span span = new Utf8Span(nonascii_110_ustring);
```
kunalspathak, line 46: Can we just put everything except last line in `[GlobalSetup]` so we want be allocating / triggering GC in every run and the measurements will be accurate?

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
```diff
@@ -0,0 +1,109 @@
...
+            Utf8Span span = new Utf8Span(nonascii_greek_ustring);
+            Memory<char> memory = new char[str.Length];
+            Span<char> destination = memory.Span;
+            return span.ToChars(destination);
+        }
+
+        [Benchmark]
+        public bool IsAscii_GetIndexOfFirstNonAsciiByte()
+        {
+            Utf8Span ascii_11_span = new Utf8Span(ascii_11_ustring);
```
kunalspathak, line 85: Same here. Move the allocation out of benchmark code.

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        [Benchmark]
+        public void IsNormalized_WidenAsciiToUtf16()
+        {
+            Utf8Span ascii_11_span = new Utf8Span(ascii_11_ustring);
+            char[] returned = ascii_11_span.ToCharArray();
+            char[] second = ascii_11_span.ToCharArray();
+            if (returned[0] != second[0])
+            {
+                Console.WriteLine("Just a line to consume returned and second");
```
kunalspathak, line 105: Instead of `Console.WriteLine()` , just do `return returned[0] == second[0];`

src/benchmarks/micro/libraries/System.Utf8String.Experimental/Perf.Utf8String.cs
pgovind: That's the first thing I tried :) The problem is that Utf8Span is a ref struct. I did check that the constructor for `Utf8Span` is cheap though. I don't see another way to initialize a `Span` here. @GrabYourPitchforks, do you think the `Utf8Span` constructor will skew our numbers here? From what I can tell it just get a reference to the first byte in the string and it's all on the stack, so I would expect it to be incredibly cheap. For reference, I'm trying to benchmark the `Utf8Utility.TranscodeToUtf16` method here (which internally calls both `ASCIIUtility.WidenAsciiToUtf16` and `ASCIIUtility.WidenFourAsciiBytesToUtf16AndWriteToBuffer` )

kunalspathak: I don't think it should matter. `ToChars()` just calls static `Utf8.Utf16()` method and return. So you can pass the same `utf8Span()` again and again. 
Ref: https://source.dot.net/#System.Private.CoreLib/Utf8Span.Conversion.cs,103

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -104,4 +105,37 @@
...
+    </None>
+    <None Update="25249-0.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="30774-0.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="39251-0.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>-->
```
carlossanlop, line 124: Remove if unused.
```suggestion
```

pgovind:
@carlossanlop I merged Adam's changes. You can re-apply your commits now if you want :)

src/benchmarks/micro/libraries/System.Utf8String/Perf.Utf8String.cs
```diff
@@ -0,0 +1,50 @@
...
+        [GlobalSetup]
+        public void Setup()
+        {
+            string unicode = File.ReadAllText(Path.Combine(Environment.CurrentDirectory, "libraries", "System.Utf8String", $"{Input}.txt"));
+            _utf8 = new Utf8String(unicode);
+            _destination = new char[unicode.Length];
+        }
+
+        [Benchmark]
+        public int ToChars() => new Utf8Span(_utf8).ToChars(_destination.Span);
```
kunalspathak, line 39: @adamsitnik  - Should we avoid creating `Utf8Span()` inside this method but instead have it created globally?

adamsitnik: @kunalspathak this is what we do by default, but `Utf8Span` is a stack-only struct so we can't store it in the class field and BDN does not support `ref structs` as type that define benchmarks

https://github.com/dotnet/runtime/blob/b5705587347d29d79cec830dc22b389e1ad9a9e0/src/libraries/System.Private.CoreLib/src/System/Text/Utf8Span.cs#L20

adamsitnik:
I have no idea why GH shows 61 modified files and changes that have already been merged to master a while ago. We need to squash this PR.

pgovind:
Local WSL2 numbers on my machine using the latest RC1 runtime:

With and without AdvSimd:
```
| Faster                                                             | base/diff | Base Median (ns) | Diff Median (ns) | Modality|
| ------------------------------------------------------------------ | ---------:| ----------------:| ----------------:| -------- |
| System.Text.Perf_Utf8String.IsAscii(Input: Greek)                  |      3.23 |           249.47 |            77.29 |         |
| System.Text.Perf_Utf8String.IsAscii(Input: Chinese)                |      3.05 |           156.38 |            51.23 |         |
| System.Text.Perf_Utf8String.IsAscii(Input: EnglishAllAscii)        |      2.89 |         40824.14 |         14134.85 |         |
| System.Text.Perf_Utf8String.ToChars(Input: EnglishAllAscii)        |      2.73 |        163875.12 |         60134.38 |         |
| System.Text.Perf_Utf8String.IsNormalized(Input: Cyrillic)          |      2.24 |       1281092.41 |        573072.40 | several?|
| System.Text.Perf_Utf8String.IsAscii(Input: Cyrillic)               |      1.93 |           236.73 |           122.49 | several?|
| System.Text.Perf_Utf8String.ToChars(Input: Cyrillic)               |      1.50 |        214020.16 |        142740.49 |         |
| System.Text.Perf_Utf8String.IsNormalized(Input: EnglishAllAscii)   |      1.40 |        858855.47 |        615500.23 | several?|
| System.Text.Perf_Utf8String.ToCharArray(Input: EnglishMostlyAscii) |      1.35 |        584513.22 |        434575.87 |         |
| System.Text.Perf_Utf8String.IsAscii(Input: EnglishMostlyAscii)     |      1.18 |            10.84 |             9.16 |         |
| System.Text.Perf_Utf8String.IsNormalized(Input: Chinese)           |      1.17 |       1572500.21 |       1345411.33 | several?|
| System.Text.Perf_Utf8String.ToChars(Input: Chinese)                |      1.05 |        187545.45 |        179036.58 |         |

No file given

```

Technically we should be getting 20 results, but I get only 12 here. Probably because I see some output on my machine about some iterations being less than 100ms.

adamsitnik:
> Technically we should be getting 20 results, but I get only 12 here.

The ResultsComparer shows only improvements and regressions. Here it means that for the other 8 benchmarks there was no difference for the threshold that you have provided

PR #1447, opened 2020-08-11
Eliminate usage of volatile variables used to prevent dead code elimination 

### LINKED ISSUE #1444: Eliminate usage of volatile variables used to prevent dead code elimination
Today, there are [various uses](https://github.com/dotnet/performance/search?p=2&q=volatile&unscoped_q=volatile) of `volatile` variable to make sure the result (and hence the operation itself) doesn't get eliminated through dead code. There is guidance [here](https://github.com/dotnet/performance/blob/6ce5b33651f6fe4ea27befec585e7d68d1f022db/docs/microbenchmark-design-guidelines.md#dead-code-elimination) that describes it. However `volatile` variables [are expensive on ARM64](https://github.com/dotnet/runtime/issues/34198) and should be avoided whenever possible. With `volatile` variables in place, the measurement could get hidden by just the cost of memory barriers needed to access `volatile` variable. We should consider eliminating this pattern (specially from BenchStone benchmarks) and instead either 
- accumulate result in a private variable which gets returned
- accumulate the result in an array that gets returned
- pass the result to a method that is marked with `[NoInling]`. 

I can do this for some benchmarks when I get time. But just want to listen to what others think about it.

adamsitnik:
fixes #1444

The reported timings can only decrease, so this change should not trigger a regression in the auto-filing issue

### DIFF
--- a/src/benchmarks/micro/libraries/System.Runtime/Perf.HashCode.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime/Perf.HashCode.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Runtime.CompilerServices;
 using BenchmarkDotNet.Attributes;
 using MicroBenchmarks;
 
@@ -11,23 +10,6 @@ namespace System.Tests
     [BenchmarkCategory(Categories.Libraries)]
     public class Perf_HashCode
     {
-        private static volatile int _valueStorage;
-
-        // Prevents the jitter from eliminating code that
-        // we want to test.
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private static void DontDiscard(int value)
-        {
-            _valueStorage = value;
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private static int DontFold(int value)
-        {
-            return value + _valueStorage;
-        }
-
         [Benchmark]
         public int Add()
         { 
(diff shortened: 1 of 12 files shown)


src/benchmarks/micro/runtime/Linq/Linq.cs
```diff
@@ -327,7 +320,6 @@ public bool Order00LinqQueryX()
             var productsInPriceOrder = from prod in products orderby prod.UnitPrice descending select prod;
             int count = productsInPriceOrder.Count();
             medianPricedProduct = productsInPriceOrder.ElementAt<Product>(count / 2);
-            Escape(medianPricedProduct);
```
kunalspathak, line 330: Any reason why this has been removed?

src/benchmarks/micro/runtime/Benchstones/BenchF/Regula.cs
```diff
@@ -15,13 +15,8 @@ public class Regula
...
     public const int Iterations = 4000000;
 
-    public static volatile object VolatileObject;
-
     [MethodImpl(MethodImplOptions.NoInlining)]
-    private static void Escape(object obj)
-    {
-        VolatileObject = obj;
-    }
+    private static void Escape(object _) { }
```
kunalspathak, line 19: should we have a common `Escape()` method that can be used everywhere?

src/benchmarks/micro/libraries/System.Runtime/Perf.HashCode.cs
```diff
@@ -41,114 +23,128 @@ public int Add()
...
             for (int i = 0; i < 10000; i++)
             {
-                DontDiscard(HashCode.Combine(
-                    DontFold(i),
-                    DontFold(i),
-                    DontFold(i),
-                    DontFold(i),
-                    DontFold(i),
-                    DontFold(i)));
+                result = HashCode.Combine(
```
kunalspathak, line 102: Just to be sure, did you confirm that this doesn't have any adverse effect on the intention of the benchmark?

src/benchmarks/micro/runtime/Benchstones/BenchF/Regula.cs
kunalspathak: Also, just to confirm, have you verified the disasm to make sure this does what we intent to do?

kunalspathak:
Could you please also update [this portion](https://github.com/dotnet/performance/blob/6ce5b33651f6fe4ea27befec585e7d68d1f022db/docs/microbenchmark-design-guidelines.md#dead-code-elimination) of "Microbenchmark Design Guidelines" to reflect the guidance.

src/benchmarks/micro/runtime/Linq/Linq.cs
adamsitnik: it's simply not needed, as the variable is used to calculate the result of the method, so it won't get dead code eliminated:

https://github.com/dotnet/performance/blob/413d1aa6aed0d05e69bff2de8ff9b59513779679/src/benchmarks/micro/runtime/Linq/Linq.cs#L325

kunalspathak: Ah, true. In this loop, we don't have any dependency on `i` so everything can be hoisted out. But I believe we don't do it because it has function calls. Anyways, nothing actionable here.

src/benchmarks/micro/libraries/System.Runtime/Perf.HashCode.cs
adamsitnik: I have not compared the produced disassembly but verified that the benchmarks are not optimized to a single call to `HashCode.Combine`. I hope that this is enough (we want to benchmark the particular methods using input that changes with every loop iteration)

BEFORE

|    Method |       Mean |     Error |    StdDev |     Median |        Min |        Max | Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------- |-----------:|----------:|----------:|-----------:|-----------:|-----------:|------:|------:|------:|----------:|
|       Add |   3.688 us | 0.0277 us | 0.0231 us |   3.702 us |   3.659 us |   3.721 us |     - |     - |     - |         - |
| Combine_1 |  68.680 us | 0.3437 us | 0.3215 us |  68.876 us |  68.058 us |  68.981 us |     - |     - |     - |         - |
| Combine_2 |  82.845 us | 0.5915 us | 0.5533 us |  83.093 us |  81.930 us |  83.595 us |     - |     - |     - |         - |
| Combine_3 |  95.124 us | 0.6227 us | 0.5520 us |  95.384 us |  94.127 us |  95.819 us |     - |     - |     - |         - |
| Combine_4 |  99.920 us | 0.7247 us | 0.6779 us | 100.305 us |  98.696 us | 100.681 us |     - |     - |     - |         - |
| Combine_5 | 117.198 us | 0.9800 us | 0.9167 us | 116.894 us | 115.617 us | 118.414 us |     - |     - |     - |         - |
| Combine_6 | 128.352 us | 1.0133 us | 0.8983 us | 128.091 us | 127.247 us | 129.667 us |     - |     - |     - |         - |
| Combine_7 | 142.305 us | 0.8045 us | 0.7525 us | 142.210 us | 140.931 us | 143.547 us |     - |     - |     - |         - |
| Combine_8 | 154.183 us | 1.5598 us | 1.4591 us | 154.196 us | 152.158 us | 156.671 us |     - |     - |     - |         - |

AFTER

|    Method |       Mean |     Error |    StdDev |     Median |        Min |        Max | Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------- |-----------:|----------:|----------:|-----------:|-----------:|-----------:|------:|------:|------:|----------:|
|       Add |   3.698 us | 0.0229 us | 0.0214 us |   3.691 us |   3.672 us |   3.748 us |     - |     - |     - |         - |
| Combine_1 |  55.353 us | 0.3684 us | 0.3446 us |  55.593 us |  54.724 us |  55.711 us |     - |     - |     - |         - |
| Combine_2 |  68.739 us | 0.4063 us | 0.3800 us |  68.869 us |  67.948 us |  69.091 us |     - |     - |     - |         - |
| Combine_3 |  82.176 us | 0.2895 us | 0.2708 us |  82.197 us |  81.258 us |  82.403 us |     - |     - |     - |         - |
| Combine_4 |  76.823 us | 0.5927 us | 0.5544 us |  76.916 us |  75.832 us |  77.733 us |     - |     - |     - |         - |
| Combine_5 |  88.316 us | 0.4261 us | 0.3777 us |  88.414 us |  87.312 us |  88.668 us |     - |     - |     - |         - |
| Combine_6 |  99.235 us | 0.7510 us | 0.7025 us |  98.875 us |  98.552 us | 100.343 us |     - |     - |     - |         - |
| Combine_7 | 114.906 us | 0.8534 us | 0.7982 us | 114.485 us | 114.092 us | 115.988 us |     - |     - |     - |         - |
| Combine_8 | 100.023 us | 0.3219 us | 0.2513 us | 100.094 us |  99.413 us | 100.279 us |     - |     - |     - |         - |

src/benchmarks/micro/runtime/Benchstones/BenchF/Regula.cs
adamsitnik: > Also, just to confirm, have you verified the disasm to make sure this does what we intent to do?

Passing an object to a method that can't be inlined is an old trick used even by [GC.KeepAlive](https://github.com/dotnet/runtime/blob/71a216819169a7ce609326e9404eacf8fb267637/src/coreclr/src/System.Private.CoreLib/src/System/GC.cs#L273)

src/benchmarks/micro/libraries/System.Runtime/Perf.HashCode.cs
kunalspathak: Excellent. Thanks for confirming.

src/benchmarks/micro/runtime/Benchstones/BenchF/Regula.cs
adamsitnik: > should we have a common Escape() method that can be used everywhere?

I would prefer to stick with the current guidelines that basically say that all you need to do to prevent from dead code elimination is to return the result from the benchmark and BDN is going to consume it for you

https://github.com/dotnet/performance/blob/6ce5b33651f6fe4ea27befec585e7d68d1f022db/docs/microbenchmark-design-guidelines.md#dead-code-elimination

when I was porting the benchmarks long time ago I was asked to not rerwrite them, this is the only reason why this old benchmark is still using this old pattern.

PR #1451, opened 2020-08-13
Add NoWASM label

DrewScoggins:


src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -66,6 +66,7 @@ public int Read()
         }
         
         [Benchmark]
+        [BenchmarkCategory(Categories.NoWASM)]
```
adamsitnik, line 69: are file operations not supported in general on WASM or just the async ones?

src/benchmarks/micro/libraries/System.Threading/Perf.CancellationToken.cs
```diff
@@ -17,6 +17,7 @@ public class Perf_CancellationToken
         public void RegisterAndUnregister_Serial() => _token.Register(() => { }).Dispose();
 
         [Benchmark(OperationsPerInvoke = 1_000_000)]
+        [BenchmarkCategory(Categories.NoWASM)]
```
adamsitnik, line 20: For the multi-threaded tests, we could introduce a `multithreaded` category and use it to exclude the benchmarks when affinitizing to a single core. Sth like:

```cmd
--filter * --category-exclusion-filter multithreaded --affinity X
```

And the full WASM filter would be:

```cmd
--filter * --category-exclusion-filter nowasm multithreaded
```

adamsitnik:
btw @buyaa-n is currently working on an analyzer that is going to print warnings for os|platform specific APIs. When it's ready we could use it here to verify if our "nowasm" category is complete

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
DrewScoggins: File operations do not seem to be allowed on WASM. From what I was able to glean from reading you need to hit a web endpoint if you want to access files.

DrewScoggins: So maybe that would be something we could do long term, but for now we can just turn them off.

PR #1456, opened 2020-08-14
Add support for blazor tests in runtime

ooooolivia:
- add ```--binlog``` flag to ```precommands.py``` so we can verify the runtime version used in publish; also useful for other diagnosis purpose
- add targets file to blazor project so it uses just-built runtime in the payload on CI

### DIFF
--- a/src/scenarios/blazor/Directory.Build.targets
+++ b/src/scenarios/blazor/Directory.Build.targets
@@ -1,7 +1,7 @@
 <Project>
  <Target Name="UpdateTargetingAndRuntimePack"
          AfterTargets="ResolveFrameworkReferences"
-         Condition="'$(HELIX_CORRELATION_PAYLOAD)' != ''">
+         Condition="Exists('$(HELIX_CORRELATION_PAYLOAD)/microsoft.netcore.app.ref')">
     <PropertyGroup>
         <MicrosoftNetCoreAppRefPackDir>$(HELIX_CORRELATION_PAYLOAD)/microsoft.netcore.app.ref</MicrosoftNetCoreAppRefPackDir>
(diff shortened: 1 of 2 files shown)


ooooolivia:
Here's a sample binlog for publishing blazor project with custom runtime in runtime CI https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-40567-merge-593a4bd95ed049b7bd/SOD%20-%20New%20Blazor%20Template%20-%20Publish/traces/blazor_publish.binlog?sv=2019-02-02&se=2020-09-03T19%3A45%3A58Z&sr=c&sp=rl&sig=yVDZsjKhA8gDXq2KYRSkW1GQHSiuXol95bNDXD0fQJY%3D Note the ```UpdateTargetingAndRuntimePack``` is called to use the just-built runtime packs.

billwert:
Hmm. So all blazor scenarios will need to go under the blazor directory so we use the same `Directory.Build.Targets`, right? What do you think about moving the existing thing in the blazor directory to a sub directory so we can have other scenarios share this?

ooooolivia:
> Hmm. So all blazor scenarios will need to go under the blazor directory so we use the same `Directory.Build.Targets`, right? What do you think about moving the existing thing in the blazor directory to a sub directory so we can have other scenarios share this?

The idea that sharing the custom runtime configuration is great! However my concern is that the content of this file is not generic enough to be shared across different scenarios, because it is for updating the wasm runtime specifically. I would take effort into making this generic when we have specific scenarios ready to add.

ooooolivia:
From offline discussion --- consider moving the targets file for updating runtime packs somewhere that could be shared by other scenarios (blazor or non-blazor) when necessary

PR #1457, opened 2020-08-15
System.Diagnostics.Activity enumeration benchmarks

CodeBlanch:
Adding the benchmarks used on https://github.com/dotnet/runtime/pull/40544

/cc @noahfalk @tarekgh

### DIFF
--- a/src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
+++ b/src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
@@ -14,27 +14,27 @@ namespace System.Diagnostics
 {
     [MemoryDiagnoser]
     [BenchmarkCategory(Categories.Libraries)]
-    public class Perf_Activity : IDisposable
+    public class Perf_Activity
     {
-        private readonly ActivitySource _ActivitySource;
-        private readonly ActivityListener _ActivityListener;
-        private readonly Activity _Activity;
-        private readonly ActivityLink _ActivityLink;
+        private static readonly ActivitySource s_ActivitySource;
+        private static readonly ActivityListener s_ActivityListener;
+        private static readonly Activity s_Activity;
+        private static readonly ActivityLink s_ActivityLink;
 
-        public Perf_Activity()
+        static Perf_Activity()
         {
-            _ActivitySource = new ActivitySource("TestActivitySource");
+            s_ActivitySource = new ActivitySource("TestActivitySource");
 
-            _ActivityListener = new ActivityListener
+            s_ActivityListener = new ActivityListener
             {
                 ShouldListenTo = s => s.Name == "TestActivitySource",
 
                 GetRequestedDataUsingContext = (ref ActivityCreationOptions<ActivityContext> o) => ActivityDataRequest.AllDataAndRecorded
             };
 
-            ActivitySource.AddActivityListener(_ActivityListener);
+            ActivitySource.AddActivityListener(s_ActivityListener);
 
-            _Activity = _ActivitySource.StartActivity(
+            s_Activity = s_ActivitySource.StartActivity(
                 "TestActivity",
                 ActivityKind.Internal,
                 parentContext: default,
@@ -53,14 +53,14 @@ namespace System.Diagnostics
                     new ActivityLink(default),
                 });
 
-            _Activity.AddEvent(new ActivityEvent("TestEvent1"));
-            _Activity.AddEvent(new ActivityEvent("TestEvent2"));
-            _Activity.AddEvent(new ActivityEvent("TestEvent3"));
-            _Activity.AddEvent(new ActivityEvent("TestEvent4"));
+            s_Activity.AddEvent(new ActivityEvent("TestEvent1"));
+            s_Activity.AddEvent(new ActivityEvent("TestEvent2"));
+            s_Activity.AddEvent(new ActivityEvent("TestEvent3"));
+            s_Activity.AddEvent(new ActivityEvent("TestEvent4"));
 
-            _Activity.Stop();
+            s_Activity.Stop();
 
-            _ActivityLink = new ActivityLink(
+            s_ActivityLink = new ActivityLink(
                 default,
                 new ActivityTagsCollection(
                     new Dictionary<string, object>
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
```diff
@@ -0,0 +1,373 @@
...
+
+        [Params(5000)]
+        public int NumberOfActivities { get; set; }
+
+        [Benchmark]
+        public void EnumerateActivityTags()
+        {
+            int total = 0;
+
+            for (int i = 0; i < NumberOfActivities; i++)
```
tarekgh, line 83: >for [](start = 12, length = 3)

do we really need to loop? Benchmark already repeat the method execution anyway and we can get he exact number for the enumeration. 

I would suggest too to create multiple static activities with different tags count. we can have the perf numbers for short, medium and long tags count.

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
```diff
@@ -0,0 +1,373 @@
...
+            {
+                foreach (var tagObject in s_Activity.TagObjects)
+                {
+                    total++;
+                }
+            }
+        }
+
+        [Benchmark]
+        public void OTelHelperActivityTagObjects()
```
tarekgh, line 107: >OTelHelperActivityTagObjects [](start = 20, length = 28)

could you please use the full word OpenTelemetry instead?

tarekgh: Also, are we using OpenTelemetry  because this how it call the Activity for enumeration?

---
In reply to: [473440210](https://github.com/dotnet/performance/pull/1457#discussion_r473440210) [](ancestors = 473440210)

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
```diff
@@ -0,0 +1,373 @@
...
+                = Enumerator<IEnumerable<KeyValuePair<string, object>>, KeyValuePair<string, object>, int>.FindAllocationFreeForEach(s_ActivityLink.Tags);
+
+            for (int i = 0; i < NumberOfActivities; i++)
+            {
+                foreachDelegate(s_ActivityLink.Tags, ref total, ForeachTagObjectRef);
+            }
+        }
+
+        // A helper class for enumerating over IEnumerable<TItem> without allocation if a struct enumerator is available.
+        private class Enumerator<TEnumerable, TItem, TState>
```
tarekgh, line 229: >Enumerator [](start = 22, length = 10)

I am not really sure if it is good to test the internals with the reflection and code emit. what is really the goal here?

tarekgh:
CC @adamsitnik if he has any comment.

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
CodeBlanch: These benchmarks were written for the other PR, really. I needed to make sure the struct enumerators worked for completely removing the allocation with the reflection mechanism, and also show there was an allocation reduction for the standard enumeration too. I opened this PR because I think @noahfalk said it would be useful. So my goal now is just to provide usefulness to future users of Activity 😄 I can remove the OpenTelemetry stuff, or I can close the PR completely if we don't think it will be that useful. LMK

tarekgh: Adding perf tests for Activity is useful. I suggest removing all OTel stuff and keep the simple test cases for the enumerating tags, links...etc. Thanks a lot for all your provided help.

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
CodeBlanch: Removed the looping. Added small/large tests.

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
CodeBlanch: Removed all the OpenTelemetry stuff.

CodeBlanch:
@tarekgh Ya it's running fine. Here's the output:

|                           Method |         Mean |        Error |       StdDev |       Median |          Min |          Max |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------- |-------------:|-------------:|-------------:|-------------:|-------------:|-------------:|-------:|------:|------:|----------:|
|       EnumerateActivityTagsSmall |     64.20 ns |     2.596 ns |     2.990 ns |     64.97 ns |     59.36 ns |     70.10 ns | 0.0065 |     - |     - |      56 B |
|       EnumerateActivityTagsLarge | 10,649.76 ns | 1,301.307 ns | 1,498.587 ns |  9,959.51 ns |  9,232.34 ns | 13,946.43 ns |      - |     - |     - |      56 B |
| EnumerateActivityTagObjectsSmall |     89.92 ns |     3.850 ns |     4.434 ns |     88.89 ns |     83.85 ns |     98.86 ns | 0.0044 |     - |     - |      40 B |
| EnumerateActivityTagObjectsLarge | 16,737.40 ns |   808.777 ns |   794.327 ns | 16,513.94 ns | 15,501.91 ns | 18,173.94 ns |      - |     - |     - |      40 B |
|      EnumerateActivityLinksSmall |    116.74 ns |     0.812 ns |     0.634 ns |    116.70 ns |    115.73 ns |    117.67 ns | 0.0072 |     - |     - |      64 B |
|      EnumerateActivityLinksLarge | 22,298.54 ns |   409.880 ns |   383.402 ns | 22,225.34 ns | 21,698.86 ns | 22,929.51 ns |      - |     - |     - |      64 B |
|     EnumerateActivityEventsSmall |     88.37 ns |     1.154 ns |     1.080 ns |     88.34 ns |     86.60 ns |     90.12 ns | 0.0065 |     - |     - |      56 B |
|     EnumerateActivityEventsLarge | 18,992.88 ns |   290.204 ns |   271.457 ns | 19,048.33 ns | 18,594.74 ns | 19,382.68 ns |      - |     - |     - |      56 B |
|   EnumerateActivityLinkTagsSmall |     89.10 ns |     9.694 ns |    10.775 ns |     85.42 ns |     74.65 ns |    114.33 ns | 0.0057 |     - |     - |      48 B |
|   EnumerateActivityLinkTagsLarge | 14,855.37 ns | 1,645.641 ns | 1,760.818 ns | 14,261.70 ns | 13,465.29 ns | 19,684.47 ns |      - |     - |     - |      48 B |

Issue #1432, opened 2020-08-05
benchmarks_ci.py does not work on Ubuntu 20.04


danmosemsft:
Ubuntu 20.04 has Python 3.8, unless you build from source. Then benchmarks_ci.py fails with:
```
dan@dan:~/git/performance$ python scripts/benchmarks_ci.py -f netcoreapp3.1 netcoreapp5.0 --filter 'System.Collections.IterateForEach<Int32>.ConcurrentBag*'
The `linux_distribution` method is missing from the `platform` module, which is used to find out information about the OS flavor/version we are using.
The Python Docs state that `platform.linux_distribution` is "Deprecated since version 3.5, will be removed in version 3.8: See alternative like the distro package.
"Most systems in the lab have Python versions 3.5 and 3.6 installed, so we are good at the moment.
If we are hitting this issue, then it might be time to look into using the `distro` module, and possibly packaing as part of the dependencies of these scripts/repo.
dan@dan-Virtual-Machine:~/git/performance$ 
```
Maybe I'm the only one who uses this script on a dev machine.

danmosemsft:
Not sure how @alexcovington got past this https://github.com/dotnet/performance/issues/1412

alexcovington:
I don't use `benchmarks_ci.py`, I get the same output when I use it. I usually do something like this:

```
cd /path/to/performance
cd src/benchmarks/micro
dotnet run -c Release -f netcoreapp5.0 --filter $SOME_FILTER
```

billwert:
Fixed by #1459

PR #1459, opened 2020-08-18
Remove warning about missing method

billwert:
As far as I can tell the method we are complaining about is never called.


billwert:
Fixes #1432

danmosemsft:
thanks

PR #1460, opened 2020-08-18
update the libs that are not part of SDK

adamsitnik:
I need to update these libs to make sure we don't have any regressions in 5.0


adamsitnik:
The CI failure is not related to this PR, merging

PR #1467, opened 2020-08-19
Disable Perf_Utf8String

billwert:


GrabYourPitchforks:
Might it be a little less churn to surround this with `#if FALSE` or similar? (I don't know what pattern you typically follow for this.)

pgovind:
Tagging other people for visibility. For some reason, these tests produce the following error on Mono:

```
[2020/08/19 16:09:45][INFO] System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
[2020/08/19 16:09:45][INFO]  ---> System.MissingMethodException: Method not found: void System.Utf8String..ctor(string)
[2020/08/19 16:09:45][INFO]    at BenchmarkDotNet.Engines.EngineFactory.CreateReadyToRun(EngineParameters engineParameters)
[2020/08/19 16:09:45][INFO]    at BenchmarkDotNet.Autogenerated.Runnable_19.Run(IHost host, String benchmarkName) in /home/billwert/correlation/performance/artifacts/bin/MicroBenchmarks/Release/netcoreapp5.0/Job-JXYTLT/Job-JXYTLT.notcs:line 14174
[2020/08/19 16:09:45][INFO]    at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture) in /_/src/mono/netcore/System.Private.CoreLib/src/System/Reflection/RuntimeMethodInfo.cs:line 384
[2020/08/19 16:09:45][INFO]    --- End of inner exception stack trace ---
[2020/08/19 16:09:45][INFO]    at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture) in /_/src/mono/netcore/System.Private.CoreLib/src/System/Reflection/RuntimeMethodInfo.cs:line 396
[2020/08/19 16:09:45][INFO]    at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters) in /_/src/libraries/System.Private.CoreLib/src/System/Reflection/MethodBase.cs:line 49
[2020/08/19 16:09:45][INFO]    at BenchmarkDotNet.Autogenerated.UniqueProgramName.AfterAssemblyLoadingAttached(String[] args) in /home/billwert/correlation/performance/artifacts/bin/MicroBenchmarks/Release/netcoreapp5.0/Job-JXYTLT/Job-JXYTLT.notcs:line 70
```

cc @carlossanlop @kunalspathak @adamsitnik @safern (see anything obviously wrong here?) @GrabYourPitchforks

billwert:
Ah, github ate the commit message out of the description. Annoying. Thanks for pasting the error and some context.

I'm reasonably sure this is going to come down to something to do with how we build the mono package that we use to actually run the benchmarks (through corerun.) I validated that `System.Utf8String.Experimental.dll` is showing up in the publish dir of the test, but for some reason it isn't getting loaded. I'll debug that in the morning.



> Might it be a little less churn to surround this with `#if FALSE` or similar? (I don't know what pattern you typically follow for this.)

We don't usually do this so we don't have a pattern. :) I'm not particularly sensitive to the churn here but perhaps I'm missing something.

billwert:
The one failure is unrelated.

adamsitnik:
@billwert @DrewScoggins Would adding the `NoInterpreter` category fix it? https://github.com/dotnet/performance/pull/1378

billwert:
No. Failing on all mono runs. I will figure it out in the morning.

Thanks,
-Bill

pgovind:
@billwert pointed out that the mono version of System.Private.CoreLib.dll does not have the same Utf8String as the coreclr library does. Indeed, `Utf8String.ctor(string value)` is defined in `Utf8String.CoreCLR.cs` which is in S.P.CoreLib (**NOT** S.P.CoreLib.Shared). Also, `Utf8String.CoreCLR.cs` is only compiled when `FeatureUtf8String` is `true` and `FeatureUtf8String` is defined like this:

```
    <!-- Features we're currently flighting, but don't intend to ship in officially supported releases -->
    <PropertyGroup Condition="'$(IsPrerelease)' == 'true'">
        <FeatureUtf8String>true</FeatureUtf8String>
        <!-- FeatureXXX>true</FeatureXXX -->
    </PropertyGroup>
```

There is some manual hand editing in the Utf8String.Experimental.csproj later, but nothing that affects `Utf8String.CoreCLR.cs`.

1. Given this, it makes sense that mono is throwing an error. But shouldn't an official release branch also throw this same error on non-mono runs?
2. To fix the benchmarks, it looks like `Utf8String` doesn't expose any constructors in mono at all? Is there no way to construct a Utf8String in mono?

@GrabYourPitchforks

adamsitnik:
Would the proper fix then be to add a `ctor` to Mono implementation? cc @eerhardt 

@billwert @pgovind thanks for a detailed investigation!

eerhardt:
> Would the proper fix then be to add a ctor to Mono implementation?

It depends on your definition of "proper" 😄.

To fix this error in the product, yes Mono would need to implement that `ctor`, which it doesn't appear to today:

https://github.com/dotnet/runtime/blob/master/src/mono/netcore/System.Private.CoreLib/src/System/Utf8String.cs

But if you look at the rest of that file - all those methods throw PNSE. So even if we got past this error, we'd get a runtime exception.

Now if your definition of "proper" means that Utf8String shouldn't be in the official repo until it is ready to be shipped as a feature, then the proper fix would be to remove Utf8Sting from dotnet/runtime and put it in https://github.com/dotnet/runtimelab/tree/Utf8String

pgovind:
I'll spend a little time later today to see if I can benchmark the methods I really want without going through `Utf8String`. That should fix all our issues. And if `Utf8` is the ONLY path to get to some methods, I guess I can argue that it's not ready to be benchmarked yet?

adamsitnik:
>  So even if we got past this error, we'd get a runtime exception.

I did not know that they just throw. You are right that adding a throwing `ctor` would not help at all.

> Now if your definition of "proper" means that Utf8String shouldn't be in the official repo until it is ready to be shipped as a feature, then the proper fix would be to remove Utf8Sting from dotnet/runtime and put it in 

No, this would delay the release of `Utf8String` and make the Chicken and Egg problem even worse.

I think that the proper fix is to filter the `Utf8String` benchmarks and simply don't run them for Mono. We already support such a scenario for Mono Interpreter and Mono Wasm scenarios. I am going to send PRs

PR #1478, opened 2020-08-23
allow a filter for displaying condemned reasons

Maoni0:
the original max_gcs arg was not very useful. better to allow to specify a filter like other functions.


Maoni0:
@dotnet/gc PTAL

PR #1479, opened 2020-08-24
Fix System.IO.Tests.Perf_Copy.RecursiveCreateDeleteDirectory test on Alpine by removing excessively deep tree

danmosemsft:
* Remove the scenario that deletes a directory tree that's over 1000 folders deep. On my box, we cannot successfully delete it. For full gory details see https://github.com/dotnet/runtime/issues/41195. 

I believe this scenario is contrived, and the 100-deep scenario is enough performance coverage of this codepath, so I suggest to remove the 1000 folder scenario. Also the comment says the limit is 1024 UTF8 characters - it's actually creating a tree that's over 2000 characters deep, so perhaps it wasn't what was intended anyway. 

* Before I figured it out, I found a couple lines we were cleaning up without being robust to any files within them. They're reasonable changes, even though turned out to be not necessary.

### DIFF
--- a/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.Directory.cs
+++ b/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.Directory.cs
@@ -59,7 +59,7 @@ namespace System.IO.Tests
         public void CleanupExists() => Directory.Delete(_testFile, recursive: true);
 
         public IEnumerable<object> RecursiveDepthData()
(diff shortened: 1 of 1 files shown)


danmosemsft:
Note that `uname -n 4096` will work around this problem, so we could update all our images, but I believe it's a synthetic scenario so better to simply remove it.

adamsitnik (approved):
LGTM, big thanks for improving the benchmarks @danmosemsft !

PR #1482, opened 2020-08-25
Add NoWASM flag to new tests

DrewScoggins:


adamsitnik (approved):
LGTM, I should have remembered that reading a file in setup won't work for Wasm ;)

kg:
It seems to be possible to remove this filter now for some of these tests, if I remove it and run the Utf8 encoding benchmarks locally in wasm they seem to work.

PR #1484, opened 2020-08-26
Branch out DotNet 5

ooooolivia:


azure-pipelines.yml
```diff
@@ -462,7 +487,8 @@ jobs:
       csproj: src/benchmarks/micro/MicroBenchmarks.csproj
       runCategories: 'runtime libraries'
       channels: # for private jobs we want to benchmark .NET Core 5.0 and 3.1
-        - master
+        #- master
+        - release/5.0.1xx
```
adamsitnik, line 491: big thanks for adding `release/5.0.1xx`. I am not sure about commenting out `master` (6.0 early bits). It would be nice if we could have at least 1 windows and 1 Linux CI leg that runs it (just to make sure we can still run the benchmarks for it)

scripts/channel_map.py
```diff
@@ -6,6 +6,14 @@ class ChannelMap():
...
         },
+        'release/5.0.1xx':{
+            'tfm': 'netcoreapp5.0',
+            'branch': 'release/5.0.1xx'
+        },
+        '5.0':{
+            'tfm': 'netcoreapp5.0',
+            'branch': 'release/5.0'
+        },
         'release/3.1.3xx':{
```
adamsitnik, line 17: not related to the goal of this PR, but should we add `release/3.1.4xx` to the list as well? (we are currently using `release/3.1.3xx`)

scripts/dotnet.py
```diff
@@ -67,7 +67,7 @@ def get_target_framework_moniker(framework: str) -> str:
         To run CoreRT benchmarks we need to run the host BDN process as latest
         .NET Core the host process will build and run CoreRT benchmarks
         '''
-        return 'netcoreapp5.0' if framework == 'corert' else framework
+        return ChannelMap.get_target_framework_moniker("master") if framework == 'corert' else framework
```
adamsitnik, line 70: nice, definitely more futureproof than what we had so far! 👍

adamsitnik (commented):
Overall LGTM, but I think that it would be great to have at least two CI legs that test 6.0 bits (master channel)

I've reported https://github.com/dotnet/runtime/issues/41521 to unblock this PR.

@ooooolivia thank you for working on this!

azure-pipelines.yml
ooooolivia: Yes after talking with Drew offline we've decided to keep the master branch by just updating the sdk version.

scripts/channel_map.py
ooooolivia: Thanks for pointing it out! Yes I think we should replace ```release/3.1.3xx``` with ```release/3.1.4xx``` now as there's no active development on the former.. will include the change in this PR.

adamsitnik:
@ooooolivia the feedback I got in https://github.com/dotnet/runtime/issues/41521 tells me that `Utf8String` won't be included in 5.0 RTM. I think that for now to unblock this important PR you can conditionally exclude https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/libraries/System.Utf8String/Perf.Utf8String.cs for `netcoreapp5.0` and later we can find a proper solution

azure-pipelines.yml
adamsitnik: > by just updating the sdk version

do you mean the `Microsoft.Dotnet.Sdk.Internal` version? does it mean that Arcade is going to download 6.0 SDK but then the python script is going to download `release/5.0.1xx` and verify the benchmarks against `release/5.0.1xx` SDK?

ooooolivia: If my thought is right, for master jobs, our ```ci_setup.py``` script is going to download the ```Microsoft.Dotnet.Sdk.Internal``` version 6.0 SDK (after this PR got checked in to unblock maestro update). The script will download from channel ```release/5.0.1xx``` for ```release/5.0.1xx``` jobs. Will uncomment the master jobs in yml.

scripts/channel_map.py
adamsitnik: @ooooolivia FWIW if `release/3.1.4xx` is problematic we can just keep using the `release/3.1.3xx` channel. To me, the biggest value of this PR is that it allows me to benchmark 5.0 bits produced from the RC branch, not the 6.0 stuff from master branch.

billwert: agreed. let's do as much as we can and sort the 3.1.4 thing out later.

PR #1486, opened 2020-08-27
Add exclusoin filter for Fannkuck_9 test on ARM64

DrewScoggins:


adamsitnik (changes requested):
I think that it would be better to fix the benchmark (logic bug) rather than exclude it. I am going to send a PR with a fix today

adamsitnik:
Closing in favor of #1488

PR #1488, opened 2020-08-27
don't check the check sum because it can be different for different number of processors

adamsitnik:
I've compared `FannkuchRedux_5` vs `FannkuchRedux_9` and:

* `FannkuchRedux_5` returns `maxFlips`, `FannkuchRedux_9`  returns `checkSum`: https://github.com/dotnet/performance/blob/ce1417e53e9af9cf3a158f327608d20a05594c67/src/benchmarks/micro/runtime/BenchmarksGame/fannkuch-redux-5.cs#L136
https://github.com/dotnet/performance/blob/ce1417e53e9af9cf3a158f327608d20a05594c67/src/benchmarks/micro/runtime/BenchmarksGame/fannkuch-redux-9.cs#L182
* the `checkSum` calculated by both FannkuchRedux_5 vs FannkuchRedux_9 is different depending on `Environment.ProcessorCount`, we can not rely on it

In my opinion, the best solution for `FannkuchRedux_9` is just to not assert the returned value, similarly to what `FannkuchRedux_5 ` already does:

https://github.com/dotnet/performance/blob/ce1417e53e9af9cf3a158f327608d20a05594c67/src/benchmarks/micro/runtime/BenchmarksGame/fannkuch-redux-5.cs#L106-L107

I've kept the unused `expectedSum` argument to keep the benchmark ID unchanged.

@DrewScoggins PTAL

### DIFF
--- a/src/benchmarks/micro/runtime/BenchmarksGame/fannkuch-redux-9.cs
+++ b/src/benchmarks/micro/runtime/BenchmarksGame/fannkuch-redux-9.cs
@@ -142,13 +142,7 @@ namespace BenchmarksGame
         // Official runs use [Arguments(12, 3968050)] which takes ~4.2 sec vs ~330ms for 11
         [Benchmark(Description = nameof(FannkuchRedux_9))]
         [Arguments(11, 556355)]
-        public int RunBench(int n, int expectedSum)
-        {
-            int chkSum = Bench(n, false);
-            if (chkSum != expectedSum)
-                    throw new Exception($"Expected {expectedSum} actual {chkSum}");
-            return chkSum;
-        }
+        public int RunBench(int n, int expectedSum) => Bench(n, false);
 
         public static int Bench(int n, bool verbose)
         {


DrewScoggins:
LGTM.

adamsitnik:
the CI failure is unrelated (scenarios), I am merging

kunalspathak:
I am wondering if we have considered eliminating `Environment.ProcessorCount` and instead having a fixed constant like "10" or something. When we compare performance of x64 vs. arm64 that run on different h/w and cores, the gap might be different because of no. of cores?

AndyAyersMS:
I think the checksum should be consistent?

My impression was that the work division algorithm was flawed and that's the part that needed a fix, basically the number of tasks needed to be an even divisor of the range of work, and it isn't for the 46 core machine.

> having a fixed constant

I would expect typical customer algorithms would adapt to use more cores when available, so perhaps we should continue to mimic that here in some fashion -- in particular oversubscribing can lead to bad perf effects (as in dotnet/runtime#13629).

adamsitnik:
> I am wondering if we have considered eliminating Environment.ProcessorCount and instead having a fixed constant like "10" or something. When we compare performance of x64 vs. arm64 that run on different h/w and cores, the gap might be different because of no. of cores?

adamsitnik:
> I am wondering if we have considered eliminating Environment.ProcessorCount and instead having a fixed constant like "10" or something. When we compare performance of x64 vs. arm64 that run on different h/w and cores, the gap might be different because of no. of cores?

I was thinking about something similar a while ago: setting it to two cores as every machine that we run benchmarks should have two cores. @kunalspathak if you believe that this would be beneficial from ARM64 vs x64 comparison please create a new issue and we can discuss it with others

adamsitnik:
> I think the checksum should be consistent?

That was my assumption as well. However, I've verified that the other multithreaded implementation of this benchmark (`FannkuchRedux_5`) is having exact same checksums for the same numbers of threads.

danmosemsft:
If we have to pick a value, maybe it would be good to pick the one that is effective on the official benchmarkgame measurement machine.

PR #1490, opened 2020-08-28
remove PAShort benchmark that uses 1 byte long input and add a "Medim" that consists of 6 bytes

adamsitnik:
As agreed in https://github.com/dotnet/runtime/issues/39720 measuring the performance of `.ToString` operation of a single byte address does not add any value

As suggested in https://github.com/dotnet/runtime/issues/39720#issuecomment-678320654 I have added a six bytes benchmark. I hope it's OK.


DrewScoggins:
LGTM

PR #1491, opened 2020-08-28
Remove Utf8String benchmarks, add more UTF8Encoding benchmarks

adamsitnik:
To unblock https://github.com/dotnet/performance/pull/1484, please see https://github.com/dotnet/runtime/issues/41521 for full details


adamsitnik:
Since more and more people start using RC1 SDK and get the wonderful runtime error about missing the `Utf8String` type I am merging it now.

/cc @AndyAyersMS

PR #1498, opened 2020-09-03
fix version numbers for dotnet.js files

billwert:


ooooolivia:
It might not be important but it would help if there could be a short description for the fix, just for review and future references ( I admit sometimes I slack on writing PR description as well...)

DrewScoggins:
This a good point. I will make sure to have a description of changes in PRs going forward.

PR #1536, opened 2020-09-25
make sure that benchmarks with explicit InvocationCount are promoted to Tier 1 BEFORE Actual Workloads

adamsitnik:
when InvocationCount is set, BDN does not run Pilot Stage, so to get the code promoted to Tier 1 before Actual Workload, we enforce more Warmups

Before:

obraz

obraz

After:

obraz

obraz


adamsitnik:
The CI legs failed with:

```log
Azure DevOps services are currently unavailable. Try again later
```

AndyAyersMS (approved):
Thanks.

Do you think this problem exists elsewhere?

adamsitnik:
> Do you think this problem exists elsewhere?

I've checked that this attribute is used only in two places in the perf repo and fixed both of them.

kunalspathak (approved):

billwert (approved):

PR #1546, opened 2020-10-08
Added basic FSharp benchmarks

TIHan:
Adds FSharp benchmarks to the repo. Contains a simple 'hello world' compilation benchmark.


TIHan:
@billwert I need to fix a merge conflict, but is there anything else that I need to do here?

eng/performance/pr.yml
```diff
@@ -116,4 +116,35 @@ jobs:
       csproj: src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
       frameworks: # for Roslyn jobs we want to check .NET Core 3.0 only
         - netcoreapp3.0
-        
\ No newline at end of file
+
```
billwert, line 119: I don't think this file is actually used - @DrewScoggins can you confirm?

billwert (changes requested):
Couple things: 
1) The pr.yml change may not be necessary. I pinged @DrewScoggins to see what he thinks.
2) Can you explain what you're trying to accomplish with the 3.1 reference check? That will be problematic in our lab for two reasons. One, as you can see from the failures in CI it's going to fail in the release/5.0.1xx branch due to missing nuget paths (which is something that's not really fixable.) Second, we don't install a 3.1 .NET Core unless we're trying to run 3.1, so even if it had restored it would fail to execute. Depending on what you're trying to do we maybe can fix some of this, but it's better if we can just.. not. :)

TIHan:
@billwert I need to get a list of .NET Core DLL references in order to compile the benchmark test. I'm also fine just with testing .NET 5 and not .NET Core 3.1.

TIHan:
Closing this as the scenarios trying to be addressed here require testing F# compiler APIs directly. Instead, we will just add F# scenarios in the "scenarios" folder of this repo.

PR #1547, opened 2020-10-09
Add base template files for GC Infra CI tests in Azure Pipelines

ivdiazsa:
This PR adds a couple template `yml` files to describe Azure Pipelines jobs for GC Benchmarking Infra, as well as a quick infra setup Python script.


ivdiazsa:
Here's the PR with the initial templates @DrewScoggins

azure-pipelines-gc.yml
```diff
@@ -0,0 +1,27 @@
...
+  containers:
+    - container: ubuntu_x64_build_container
+      image: microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-c103199-20180628134544
+
+# Trigger builds for PR's targeting master
+pr:
+  branches:
+    include:
+    - master
+  paths:
```
DrewScoggins, line 11: Should probably add some include paths here for just the GC area

ivdiazsa: Done! Would this be enough? I'm setting it to trigger anytime the main script, or any source files are modified.

DrewScoggins (approved):
LGTM

PR #1548, opened 2020-10-14
Initial attempt at expanding Crossgen2 perf testing

trylek:
Based on offline Bill's advice I'm trying to expand on two aspects
of Crossgen2 testing:

1) I have expanded the list of test assemblies to 12 by hand-picking
assemblies covering a broad range of sizes (10K, 20K, 50K, 100K, 200K,
500K, 1M, 2M, 3M, 4M*2, 10M).

2) I have added a new runner argument for crossgen2 "--singlethreaded"
that activates the "--parallelism 1" option for crossgen2.

In my local testing I see the "--singlethreaded" flag propagated
to the Crossgen2 command-line. I'm definitely no Python expert so
please let me know if my change can be done in a more elegant manner.

Thanks

Tomas

### DIFF
--- a/eng/common/performance/crossgen_perf.proj
+++ b/eng/common/performance/crossgen_perf.proj
@@ -27,11 +27,18 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <SingleAssembly Include="System.Private.Xml.dll"/>
-    <SingleAssembly Include="System.Linq.Expressions.dll"/>
-    <SingleAssembly Include="Microsoft.CodeAnalysis.VisualBasic.dll"/>
-    <SingleAssembly Include="Microsoft.CodeAnalysis.CSharp.dll"/>
-    <SingleAssembly Include="System.Private.CoreLib.dll"/>
+    <SingleAssembly Include="System.Net.WebProxy.dll"/>                 <!-- Approx. 10 KB -->
+    <SingleAssembly Include="System.Net.Http.Json.dll"/>                <!-- Approx. 20 KB -->
+    <SingleAssembly Include="System.Drawing.Primitives.dll"/>           <!-- Approx. 50 KB -->
+    <SingleAssembly Include="System.ServiceModel.Syndication.dll"/>     <!-- Approx. 100 KB -->
+    <SingleAssembly Include="System.Net.Sockets.dll"/>                  <!-- Approx. 200 KB -->
+    <SingleAssembly Include="System.Linq.Expressions.dll"/>             <!-- Approx. 500 KB -->
+    <SingleAssembly Include="System.Data.Common.dll"/>                  <!-- Approx. 1 MB -->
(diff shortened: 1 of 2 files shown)


trylek:
@billwert - is there a formal way to add you as a reviewer in the GitHub UI akin to the runtime repo?
@nattress 

/cc @dotnet/crossgen-contrib

billwert (changes requested):
Hello @trylek!

Unfortunately I had one more detail wrong. The change to crossgen_perf.proj needs to happen in Arcade, not here. https://github.com/dotnet/arcade/blob/master/eng/common/performance/crossgen_perf.proj is the correct location. (In the future I would like this to just live in the runtime repo, but as of today that's not how the system works.)

Otherwise this looks great! The CI failures appear to be unrelated. I'll take a look at them separately.

trylek:
Thanks @billwert for confirming - it has already dawned on me that <code>eng/common</code> in <code>dotnet/runtime</code> gets synchronized with Arcade as the master and I speculated that the perf repo might be in the same situation. Frankly speaking, however, I'm somewhat concerned about the incurred overhead - perhaps not in this particular case but in general, it simply doesn't make sense to me why such technical detail specific to the perf / runtime repo (construction of the Crossgen2 command line) should be encoded in the Arcade repo - actually it was just the other day I tried to push a YAML cleanup to Arcade and @riarenas pushed back against putting repo-specific logic into Arcade scripts. I don't want to wreak havoc in your test harness but couldn't we somehow extract the command-line construction into a bit residing in the perf repo, e.g. somewhere around src/scenarios/crossgen[2]?

Thanks for patience

Tomas

riarenas:
@trylek, The issue with the change that had to be reverted in Arcade was that one of the YAML files that is used by all repos was changed to reference a file that only existed in the runtime repo, so it would've broken everyone else if we had delivered that version of arcade to anyone else.

We know that the perf projs and templates live in Arcade and we won't push back against adding more related work there.

trylek:
Thanks @riarenas for clarifying - am I then right to understand that, if I port the "Arcade" part of this PR (the change to crossgen_perf.proj) to a PR targeting the Arcade repo, there's a reasonable chance it's going to be reviewed and ultimately approved after review feedback has been addressed?

riarenas:
@trylek Yep, submit away!

billwert:
> I'm somewhat concerned about the incurred overhead - perhaps not in this particular case but in general, it simply doesn't make sense to me why such technical detail specific to the perf / runtime repo (construction of the Crossgen2 command line) should be encoded in the Arcade repo

Great question. In general we have this arrangement so that we can run perf tests in more repos. In practice that design hasn't been necessary (we thought it would be at the start of .NET Core 3.0, which is when we made these choices.) It's something we're taking a look at changing for .NET 6. In the mean time, as @riarenas says we have the ability to change any of the perf files in Arcade as often as we want, so it's inconvenient but not a blocker. I appreciate you bearing with it.

billwert (approved):

PR #1549, opened 2020-10-14
Added Numeric Analysis feature to GC Benchmarking Infra

ivdiazsa:
Numeric Analysis is a feature that allows you to use the `pandas` library to analyze GC metrics from various runs of a *GCPerfSim* test. It reads all the metrics numbers from all the iterations for the test run, and builds a list with them, ready for `pandas` to consume and interpret it. This feature is only available in Jupyter Notebook.

First, you have to run the test you want to analyze multiple times. The simplest way to do this is to set the field `default_iteration_count` in the test's YAML file to a higher number. Once tests have run, you run the following code snippets in Jupyter Notebook.

```python3
_BENCH = Path("bench")
_SUITE = Path("bench") / "suite"
_TRACE_PATH = _SUITE / "normal_server.yaml"

metrics_data = get_gc_metrics_numbers_for_jupyter(
    traces=ALL_TRACES,
    bench_file_path=_TRACE_PATH,
    run_metrics=parse_run_metrics_arg(("important",)),
    machines=None,
)

data_frame = pandas.DataFrame.from_dict(metrics_data)
```

`ALL_TRACES` is a constant already defined in the Jupyter Notebook provided in this repo. For the `machines` argument you can provide a list with the names of the machines where tests where run, but this is not necessary to get the numeric and statistical analysis of this feature.

The first function run here is `get_gc_metrics_numbers_for_jupyter()`. This function reads the test's `yaml` file (by default located in `bench/suite`), which then uses to lookup the resulting traces in the corresponding *out* folder. Then, it reads the traces of each time the test was run, fetching the numbers data, processing it, and returning the dictionary with the lists of values of each metric.

Once this data has been acquired, it is given to `pandas` to create the `DataFrame` object it requires. The dictionary is composed as follows:

* **Keys**: Metric Name
* **Values**: List with said metric's numbers from each iteration of the test run.

NOTE: Aside from the metric's names, this dictionary also holds two additional keys:

* **Config_Name**: Name of the configuration run (e.g. _only\_config_).
* **Benchmark_Name**: Name of the benchmark run (e.g. _2gb_)

These are used to group values by configuration and/or benchmark for more specific analysis.

With that `DataFrame`, you can use all of `pandas` functionalities for analysis, charting, etc. The documentation can be found [here](https://pandas.pydata.org/docs/).

Let's see two examples.

First, the `describe()` method will show the main statistics values of your data.

```python3
data_frame.describe()
```

PandasDescribe

Let's see an example a bit more complicated now. Let's suppose we want to see how the heap sizes before and after garbage collection took place behaved through the runs. For this, we first create another data frame with this subset of data, and then chart it.

```python3
heap_sizes = data_frame[["HeapSizeBeforeMB_Mean", "HeapSizeAfterMB_Mean"]]
heap_sizes.plot()
```

PandasPlot

You can also get statistical values grouped by test parameters (e.g. configuration, benchmark). Let's build another example to illustrate this. In this example, we repeated the entire setup but this time we ran 4 different benchmarks of a test:

* _1gb_ and _2gb_ without concurrent GC's.
* _1gb_ and _2gb_ with concurrent GC's.

We want to see the mean value for each metric from each flavor's iterations. The `groupby()` code snippet separates the values as we require them, and then `pandas' mean()` method calculates for each flavor, resulting in the table shown in the picture after the code snippet below.

```python3
data_frame.groupby(["config_name", "benchmark_name"]).mean()
```

PandasMeanGroupBy

Many more things can be done for analyzing this data using `pandas`. The main purpose of this feature is to get all the information from the traces to have it ready for use with `pandas`.


ivdiazsa:
The numeric and statistical analysis feature is done. Please take a look when you're able to. Thanks! @dotnet/gc

src/benchmarks/gc/docs/bench_file.md
```diff
@@ -7,8 +7,7 @@ Each is a map with keys being arbitrary names.
 Here's an example benchfile:
 
 ```yaml
-test_executables:
-  defgcperfsim: /performance/artifacts/bin/GCPerfSim/release/netcoreapp5.0/GCPerfSim.dll
+test_executables: {}
```
Maoni0, line 10: what does {} mean here?

ivdiazsa: Linter screwed up some things again when adding Andrew's encapsulation changes. I've converted the PR to draft to fix this and other things. I'll let you know once it's ready for review again.

ivdiazsa:
I've now fixed what was needed and updated the PR description. It's now ready for review again @Maoni0 @cshung

src/benchmarks/gc/src/mypy.ini
```diff
@@ -0,0 +1,5 @@
+[mypy]
+python_version = 3.7
```
Maoni0, line 2: so why are we limiting this to 3.7? does this mean it would complain if I'm running on 3.8?

ivdiazsa: No, it means to use at least 3.7 version. This is in accordance with the original docs and tests made by Andy and specified in the README. This file is only to add `mypy` exceptions (this should be kept to a minimum), as `mypy` sometimes fails in linting. This is equivalent to the pylintrc file Andy added for some pylint exceptions.

Maoni0: I see; when I looked up the doc for this it said this is the version, not the oldest version. that's fine then.

PR #1551, opened 2020-10-20
Fix broken links in performance docs

nattress:
Add `.md` to a few places where it was omitted so links are navigable.


nattress:
@billwert PTAL at this doc fix spam PR

adamsitnik (approved):
LGTM, thank you @nattress 

@billwert I've clicked every link and tested that they now work as expected

PR #1554, opened 2020-10-22
Basic Activity allocation benchmarks

CodeBlanch:
Adding benchmarks used (more or less) for https://github.com/dotnet/runtime/pull/43710 in case anyone might find them useful in the future.

/cc @tarekgh


tarekgh:
CC @adamsitnik @DrewScoggins

tarekgh (approved):

billwert:
Hello @CodeBlanch  (and @tarekgh)! 

What is the part of the test that allocates? Should we break this apart into separate benchmarks for the constructor, `SetIdFormat`, `Start`, and `Stop`? 

Also, you could parameterize the `ActivityIdFormat` to reduce duplication. [This](https://benchmarkdotnet.org/articles/features/parameterization.html) is a good reference for the support available in Benchmark.NET.

tarekgh:
@billwert 

> What is the part of the test that allocates? 

The issue https://github.com/dotnet/runtime/issues/43704 description has the details and why adding this test now. 

> Should we break this apart into separate benchmarks for the constructor, SetIdFormat, Start, and Stop?

I don't think we need to do that. This test is testing specific scenario which mainly concerned with the `Stop` while the Activity Id is not allocated yet.

> Also, you could parameterize the ActivityIdFormat to reduce duplication. 

This is a good idea.

billwert:
Got it. I would parameterize this but otherwise I think it's fine.

CodeBlanch:
@billwert I moved the enumeration benchmarks to their own class/file & updated the allocation benchmarks to use params.

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity_Enumerations.cs
```diff
@@ -0,0 +1,186 @@
...
+
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+using System.Collections.Generic;
+
+namespace System.Diagnostics
+{
+    [MemoryDiagnoser]
+    [BenchmarkCategory(Categories.Libraries)]
+    public class Perf_Activity_Enumerations
```
adamsitnik, line 13: by moving the existing benchmarks to a new type with a new name you are changing the benchmark ID, which is used by our Reporting System for tracking performance over time. Please revert this change.

https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
```diff
@@ -4,183 +4,23 @@
...
-
-        [Benchmark]
-        public void EnumerateActivityLinkTagsSmall()
-        {
-            foreach (var _ in s_ActivityLinkSmall.Tags)
-            {
-            }
-        }
+        [Params(ActivityIdFormat.Hierarchical, ActivityIdFormat.W3C)]
+        public ActivityIdFormat IdFormat { get; set; }
```
adamsitnik, line 15: `[Params]` are applied to all the benchmarks in the given type. Since we would like to keep the old type name (and not move all the benchmarks to a different type), you can use `[Arguments]` instead of `[Params]` which would add a new benchmark with two test cases

https://benchmarkdotnet.org/articles/features/parameterization.html#sample-introarguments

it would look like this:

```cs
[Benchmark]
[Arguments(ActivityIdFormat.Hierarchical)]
[Arguments(ActivityIdFormat.W3C)]
public Activity ActivityAllocations(ActivityIdFormat idFormat)
{
    Activity activity = new Activity("TestActivity");
    activity.SetIdFormat(idFormat);
    activity.Start();
    activity.Stop();
    return activity;
}
```

An alternative would be to keep the existing type as is and introduce a new type with a new name and this new benchmark

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity_Enumerations.cs
```diff
@@ -0,0 +1,186 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+using System.Collections.Generic;
+
+namespace System.Diagnostics
+{
+    [MemoryDiagnoser]
```
adamsitnik, line 11: nit: `MemoryDiagnoser` is enabled by default in this repo, there is no need for the attribute  here

https://github.com/dotnet/performance/blob/7d9d860af8cfd8807975517015ca6d9f98985852/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L43

```suggestion
```

adamsitnik (changes requested):
Hi @CodeBlanch 

Big thanks for adding new benchmarks!

Could you please revert the changes that affect the benchmark ID and either add a new benchmark to the type using `[Arguments]` or introduce a new type with a new benchmark?

https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable

Thanks,
Adam

CodeBlanch:
@adamsitnik Sorry about that, updated. Thanks for the tip about `[Arguments]` haven't used that before, super useful.

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
```diff
@@ -8,7 +8,6 @@
 
 namespace System.Diagnostics
 {
-    [MemoryDiagnoser]
```
tarekgh, line 11: why removed the MemoryDiagnoser attribute?

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity_Enumerations.cs
CodeBlanch: /cc @tarekgh

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
CodeBlanch: @adamsitnik left a comment about that above, I tagged you on it.

adamsitnik (approved):
LGTM, thank you @CodeBlanch !

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Activity.cs
tarekgh: thanks, I missed this comment.

adamsitnik:
the build failures are unrelated, I am merging

PR #1556, opened 2020-10-22
Fix release branch

DrewScoggins:
In order to run the 5.0.1xx branch of installer we need to have darc update
    our Nuget.config to contain the paths to the binaries that match the version.
    As part of this we can no longer have a dependency on the master branch of
    installer. So this change moves all of the master runs over to a scheduled
    kickoff and leaves the CI runs to only by the release branches.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -163,7 +163,7 @@ jobs:
       csproj: src\benchmarks\micro\MicroBenchmarks.csproj
       runCategories: 'runtime libraries'
       channels: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
-        #- LTS # use LTS channel for net461 framework
+        - LTS # use LTS channel for net461 framework
 
   # Windows x86 micro benchmarks
   - template: /eng/performance/benchmark_jobs.yml
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,15 +9,10 @@ trigger:
     include:
     - master
   paths:
+    include:
+    - eng/Version.Details.xml
     exclude: # don't trigger the CI if only a doc file was changed
-    - docs/*
-    - eng/common/README.md
-    - src/benchmarks/micro/Serializers/README.md
-    - src/benchmarks/micro/README.md
-    - src/benchmarks/real-world/JitBench/README.md
-    - src/benchmarks/real-world/Microsoft.ML.Benchmarks/Input/README.md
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -9,15 +9,8 @@ trigger:
...
-    - docs/*
-    - eng/common/README.md
-    - src/benchmarks/micro/Serializers/README.md
-    - src/benchmarks/micro/README.md
-    - src/benchmarks/real-world/JitBench/README.md
-    - src/benchmarks/real-world/Microsoft.ML.Benchmarks/Input/README.md
-    - src/tools/ResultsComparer/README.md
-    - README.md
+    include:
+    - eng/Version.Details.xml
```
billwert, line 13: Won't this mean we only trigger on the schedule updates and not on other changes? What will that mean for CI runs on PRs etc?

eng/performance/benchmark_jobs.yml
```diff
@@ -99,9 +99,7 @@ jobs:
         steps:
         - checkout: self
           clean: true
-        - script: $(Python) scripts/parse_props.py --branch-name $(_Channel)
-          displayName: Parse Version.props
-        - script: $(Python) scripts/ci_setup.py $(DotnetVersion) --channel $(_Channel) --architecture ${{parameters.architecture}} --perf-hash $(Build.SourceVersion) --queue ${{parameters.queue}} --build-number $(Build.BuildNumber) --build-configs $(_Configs)
+        - script: $(Python) scripts/ci_setup.py --channel $(_Channel) --architecture ${{parameters.architecture}} --perf-hash $(Build.SourceVersion) --queue ${{parameters.queue}} --build-number $(Build.BuildNumber) --build-configs $(_Configs)
```
billwert, line 102: For non-`master` runs why wouldn't we keep parsing the props?

azure-pipelines.yml
DrewScoggins: This is only for the CI trigger. I did not change the PR trigger. So nothing will change for the PRs, but we will only do triggered runs when we update the version of the SDK for the release branch

eng/performance/benchmark_jobs.yml
DrewScoggins: If you look at parse_props.py we were never doing this for non-master branches any way. We only ever had the master version number ever checked in so there was nothing to really parse on non-master channels.

PR #1558, opened 2020-10-23
Fix Scenario Name

billwert:
The scenario name is computed in a way I didn't notice previously so our new single threaded mode is polluting the old data.

cc: @trylek @nattress


trylek (commented):
Ahh, great, thanks for fixing this!

PR #1563, opened 2020-10-27
Crossgen/Crossgen2 size-on-disk tests

nattress:
* Add support for size-on-disk assembly testing for crossgen/crossgen2
* This works by running crossgen in the precommand phase to generate a ready-to-run image for the assembly in the given test scenario. The test phase runs the size-on-disk tool to collect size information for the ready-to-run binary.
* Allow crossgen / crossgen2 execution to be hooked into pre.py scripts by factoring out the command line argument handling. crossgen.py can be used as a subparser during argument parsing (via `add_crossgen[2]_arguments`), when populating the evaluated args (`parse_crossgen[2]_args`), and finally returning the crossgen command line arguments to execute the scenario (via `get_crossgen[2]_command_line`).
* Align the crossgen and crossgen2 arguments. We were using `--test-name` and `--single` to represent assembly name. For now, accept both for crossgen until we stop using `--test-name` in Arcade. Later we can remove the deprecated option.
* Always emit crossgen images to `crossgen.out` and clean that up in `post.py`. We can add hierarchy there if we need, but we were starting to pollute the directory structure that the post.py commands take care to clean.
* Add documentation for the new scenario along with example pre, test, post commands

### DIFF
--- a/src/scenarios/shared/crossgen.py
+++ b/src/scenarios/shared/crossgen.py
@@ -1,0 +1,199 @@
+'''
+Module for running Crossgen
+'''
+
+import sys
+import os
+
+from logging import getLogger
+from argparse import ArgumentParser
+from shared import const
+
+class CrossgenArguments:
+    '''
+    Common helper for parsing crossgen args
+    This is intended for runtime tests where we must execute crossgen directly.
+    AOT code for SDK and higher level tests is generated during publish.
+    '''
+
+    def __init__(self):
+        self.coreroot = None
+        self.singlefile = None
+        self.compositefile = None
+        self.singlethreaded = None
+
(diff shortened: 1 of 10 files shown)


nattress:
@billwert PTAL

/cc @dotnet/crossgen-contrib

nattress:
This will be enabled in scheduled lab runs with the matching change in Arcade repo: https://github.com/dotnet/arcade/pull/6466

src/scenarios/shared/crossgen.py
```diff
@@ -0,0 +1,199 @@
+'''
+Module for running Crossgen
```
billwert, line 2: nit: this doesn't run Crossgen.
```suggestion
Module for parsing Crossgen args
```

PR #1568, opened 2020-10-28
Add logging time on thread

billwert:


DrewScoggins:
LGTM. It looks like we were returning before we could ever set the parallelism to one for single file. Is that what the deleted return is about?

billwert:
Yup exacty

nattress:
LGTM. Yep that early return was unintentional. If it's easy to wire up, I'd like to add one of each crossgen scenario to the CI so we can catch these sorts of bugs.

nattress (approved):

PR #1574, opened 2020-10-30
Add microbenchmarks for manual structure marshalling methods on System.Runtime.InteropServices.Marshal

jkoritzinsky:
Add microbenchmarks for Marshal.StructureToPtr, Marshal.PtrToStructure, and Marshal.DestroyStructure.

We accidentally regressed performance on these methods in .NET 5. Let's add some benchmarks so we can catch any regressions from this point onwards more easily.

cc: @AaronRobinsonMSFT  @elinor-fung


billwert (approved):
These look reasonable to me. Thanks @jkoritzinsky!

PR #1575, opened 2020-10-30
Update Core_Root building docs with the new changes in the runtime repo

ivdiazsa:
This PR updates `docs/building_coreroot.md` within GC Benchmarking Infra, as the way of generating the *Core_Root* from the *runtime* repo has changed. Previously, one would run `build-test` from `src/coreclr`, but that script is no longer located there. It is now called just `build` and was moved to `src/tests`. So, the command was changed:

From:

```powershell
cd \path\to\runtime\src\coreclr\
.\build-test.cmd Release generatelayoutonly
```

To:

```powershell
cd \path\to\runtime\src\tests\
.\build.cmd Release generatelayoutonly
```

The same applies to the Linux sh script.

This PR also changed the steps to build the runtime and libraries in the doc, to use `clr+libs` in one single command instead of building them separately. Notes on those flags were added as well.


ivdiazsa:
@dotnet/gc

ivdiazsa:
Also removed a piece of documentation pointing to non-existing functionality in `contributing.md`. This change was done due to it being unnecessary to open another PR for another small documentation change.

Maoni0 (approved):

PR #1578, opened 2020-11-03
Use Release Env since debug msvcrt is not distributed.

mangod9:
To be able to use the perf harness on a non-VS machine, need to use Release build of Env.


mangod9:
@cshung @Maoni0 @ivdiazsa please review, weird that I cant add reviewers (possibly permission issue)

ivdiazsa:
Looks good to me. I would just comment out `_EXEC_ENV_BUILD_DEBUG_PATH` if it's not used anywhere else. To avoid running into problems with the linter. Thanks for this contribution Manish!

mangod9:
@ivdiazsa @cshung do you have permissions to give me write access to this repo? I thought everyone in the dotnet org should have access, but doesnt seem like that.

ivdiazsa:
Unfortunately, the `performance` repo follows a much more restricted write policy. From the people I know, only Maoni, Bill, and Drew can merge PR's into the repo. @mangod9

PR #1582, opened 2020-11-05
Computing the allocation statistics

cshung:
In order to debug what is wrong with the POH allocation test case, I added this support in `GCPerfSim` to count how many bytes we allocated into the individual object heaps. 

To start with the review, read the changes in `GCPerfSim.cs`. 

The change accounts for the bytes allocated in individual object heaps in two ways. First, it is done in `GetObjectSpec()` and then it is done again in the `Statistics` class which is enabled only for debug, which allows me to be sure that the count is correct when I debug.

The change I did found a bug. In order for this accounting scheme to work, we need to make sure whenever `GetObjectSpec()` is called, there is exactly one allocation according to the spec happen downstream in `JustMakeObject()`. Just so it is clear, we need this constraint for another reason, if that is not true, then the allocation ratio arguments won't work. 

The various thing that the original code did for `ReferenceItemWithSize` does not comply with this constraint, and therefore I changed the logic there, it will still create a linked list, but it will allocate exactly once.

The python code is driven by the fact that the `GCPerfSim` output is actually parsed by the infrastructure, so we need to make sure the infrastructure is aware of these extra outputs, and then `py . lint` do all the other changes.


src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -669,7 +739,7 @@ public override string ToString()
     {
         Util.AlwaysAssert(totalAllocBytes != 0); // Must be set
 
-        this.testKind = testKind;;
+        this.testKind = testKind; ;
```
ivdiazsa, line 742: The second ';' was a typo right?

ivdiazsa:
Looks good! Just the typo I pointed and drop the linter's changes to `bench_file.md`. For some reason it messes with certain things like removing the example `test_executables`.

Maoni0:
looks fine to me. a note for future reference, if you could refrain from having a significant chunk of your changes be formatting changes only that'd be much appreciated. I'm all for have formatting changes to make the formatting more consistent but having them mixed with actual functional changes makes CRs more work than it should be. I'm not gonna ask you to separate them out for this PR but please keep it in mind for future PRs. I'm guessing these changes simply come from whatever editor you are using that does the formatting. you can just do that once and submit a PR to cover all formatting changes.

I see this is marked as Draft, please let me know when you think this is ready for checkin.

cshung:
**Status**
1. Once the [formatting change PR](https://github.com/dotnet/performance/pull/1638) is merged, changes to GCPerfSim.cs should be purely functional changes.
2. The Statistics functionality I introduced in this PR has some duplication with `allocatedBytesTotalSum` defined on the `Bucket`. Further investigation uncovered a few bugs with the accounting in correcting them is a work in progress. This PR will eventually turn into a fix for those accounting problems.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -343,13 +409,26 @@ private Item(uint size, bool isPinned, bool isWeakLong, bool isPoh)
         if (isPoh)
         {
 #if NET5_0
+#if STATISTICS
+            statistics.pohAllocatedBytes += size - SohSize;
```
Maoni0, line 413: nit - not duplicate (size - SohSize) 3 times?

Maoni0, line 413: >size - SohSize [](start = 44, length = 14)

this size doesn't seem to add up... let's talk in our 1:1.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -1725,11 +1845,13 @@ struct OldArr
 {
     public readonly ITypeWithPayload?[] items;
     public ulong TotalLiveBytes { get; private set; }
+    public uint NonEmptyLength;
 
     public OldArr(ulong numElements)
     {
         items = new ITypeWithPayload?[numElements];
         TotalLiveBytes = Util.ArraySize(items);
+        NonEmptyLength = (uint)numElements;
```
Maoni0, line 1854: >NonEmptyLength  [](start = 8, length = 15)

I don't understand why this is called NonEmptyLength

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -2193,129 +2271,54 @@ private void DoSurvive(ITypeWithPayload item, in ObjectSpec spec)
...
-                // 
-                // If the live data size is > what's set, we randomly choose a non null entry 
-                // and set it to null.
-
-                if (oldArr.TotalLiveBytes < curPhase.totalLiveBytes)
-                {
-                    MixItUp((ReferenceItemWithSize)item, spec);
-                }
-                else
+                // Step 1: Decide which list to pick a victim to free
```
Maoni0, line 2274: >               // Step 1: Decide which list to pick a victim to fr [](start = 1, length = 66)

this changes what it used to do.. let's talk about it in our meeting.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -2129,7 +2194,20 @@ bool GoToNextPhase()
 
     (ITypeWithPayload, ObjectSpec spec) JustMakeObject()
     {
-        ObjectSpec spec = bucketChooser.GetNextObjectSpec(rand);
+        uint overhead;
+        switch (curPhase.allocType)
+        {
+            case ItemType.SimpleItem:
+                overhead = Item.SohSize;
```
Maoni0, line 2201: read from the field also named Overhead?

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
Maoni0: could you please open an issue to add back the variability?

---
In reply to: [584232243](https://github.com/dotnet/performance/pull/1582#discussion_r584232243) [](ancestors = 584232243)

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
cshung: In the `allocType=reference` case, we might delete some nodes from a list so that the list becomes empty. To make it easy to find a victim node to delete, these empty lists are swapped on the right (i.e. larger index) so that we can have a single number to determine where do the lists end. Any lists with an index less than this number are non-empty, and any slot greater than or equal to this number is guaranteed to be `null`. This is why it is called `NonEmptyLength`.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
cshung: https://github.com/dotnet/performance/issues/1712

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -419,24 +503,43 @@ class SimpleRefPayLoad
...
 
     public SimpleRefPayLoad(uint size, bool isPinned, bool isPoh)
     {
-        uint sizePayload = size - Overhead;
+#if STATISTICS
+        Statistics statistics = Statistics.GetStatistics();
+        statistics.sohAllocatedBytes += ReferenceItemWithSize.SohOverhead;
+#endif
+        uint sizePayload = size - ArrayOverhead;
+        uint remainingSize = size + ReferenceItemWithSize.SimpleOverhead - ReferenceItemWithSize.SohOverhead;
```
Maoni0, line 519: these are still using data for ReferenceItemWithSize, not SimpleRefPayLoadSize

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -1616,6 +1740,20 @@ public ObjectSpec GetObjectSpec(Rand rand)
         bool shouldBePinned = shouldSurvive && Util.isNth(pinInterval, count / survInterval);
         bool shouldBeFinalizable = shouldSurvive && Util.isNth(finalizableInterval, count / survInterval);
 
+        if (isPoh)
+        {
+            sohAllocatedBytes += overhead;
+            pohAllocatedBytes += (size - overhead);
+        }
+        else if (size >= 85000)
```
Maoni0, line 1748: you might want to not hard code this -  you could be running with the LOH threshold changed with the config

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -1725,11 +1853,13 @@ struct OldArr
 {
     public readonly ITypeWithPayload?[] items;
     public ulong TotalLiveBytes { get; private set; }
+    public uint NonEmptyLength;
```
Maoni0, line 1856: why is the var that indicates the # of elements in the array is called NonEmptyLength?

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
Maoni0: makes sense.

weird, these comments simply don't show up in codeflow.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -419,24 +503,43 @@ class SimpleRefPayLoad
...
+
+    public static uint ArrayOverhead = SimpleRefPayLoadSize + Util.ARRAY_HEADER_SIZE;
+    public static uint SohOverhead = SimpleRefPayLoadSize;
 
     public SimpleRefPayLoad(uint size, bool isPinned, bool isPoh)
     {
-        uint sizePayload = size - Overhead;
+#if STATISTICS
+        Statistics statistics = Statistics.GetStatistics();
+        statistics.sohAllocatedBytes += ReferenceItemWithSize.SohOverhead;
```
Maoni0, line 516: this one is still not changed?

cshung: This is intentional.

For each allocation when `allocType=reference`, we create an instance of `ReferenceItemWithSize`, an instance of `SimpleRefPayLoad`, and a byte array. The number `ReferenceItemWithSize.SohOverhead` includes both the size of `ReferenceItemWithSize` and `SimpleRefPayLoad`, so it is the right number of bytes allocated on the SOH for these two objects.

I could have made it such that I account for the size of `ReferenceItemWithSize` in the constructor of `ReferenceItemWithSize` and then account for the size of `SimpleRefPayLoad` in the constructor of `SimpleRefPayLoad`, but then I will need to add one more thread static access and one more addition in the constructor `ReferenceItemWithSize`. This can be seen as an optimization so that I saved those operations.

cshung:
This PR is subsumed by https://github.com/dotnet/performance/pull/1704

PR #1584, opened 2020-11-05
Add support for POH allocations in GCPerfSim from GC Benchmarking Infra

ivdiazsa:
*GCPerfSim*, the tool that runs GC Infra's tests, has support for Pinned Object Heap allocations. However, the only way to trigger them was by calling GCPerfSim directly in the command line and passing the `-pohar` (Pinned Object Heap Allocation Ratio) flag there. This was because the infra's `yaml` benchmark files did not support this value.

This PR adds it to the `suite-create` command and also removed the `-pohpi` flag. This one stood for Pinned Object Heap Pinning Interval. However, all items by default are pinned when allocated in POH so it made no sense to have it. With these changes, the `benchmark` sections of the `yaml` files will look like this:

```yaml
benchmarks:
  2gb:
    arguments:
      tc: 6
      tagb: 300
      tlgb: 2
      lohar: 0
      pohar: 0  # This is the new flag.
      sohsi: 50
      lohsi: 0
      pohsi: 0
      sohpi: 0
      lohpi: 0
      sohfi: 0
      lohfi: 0
      pohfi: 0
      allocType: reference
      testKind: time
# "pohpi" was removed because it didn't provide any additional functionality.
```

### DIFF
--- a/src/benchmarks/gc/docs/bench_file.md
+++ b/src/benchmarks/gc/docs/bench_file.md
@@ -35,12 +35,12 @@ benchmarks:
       tagb: 500
       tlgb: 1
       lohar: 0
+      pohar: 0
       sohsi: 50
       lohsi: 0
       pohsi: 0
       sohpi: 0
       lohpi: 0
-      pohpi: 0
       sohfi: 0
       lohfi: 0
       pohfi: 0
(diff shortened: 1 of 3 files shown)


ivdiazsa:
@dotnet/gc

Maoni0 (approved):
LGTM! note that there's a problem with how the -*ohar args are handled including -pohar which @cshung has a fix for but that's a separated issue.

PR #1587, opened 2020-11-10
Refactor initialization logic to allow for enabling Memory Randomization

adamsitnik:
Edit: see https://github.com/dotnet/performance/pull/1587#issuecomment-768249711

~~I am currently working on an experiment with Memory Randomization in BDN (https://github.com/dotnet/BenchmarkDotNet/pull/1587)~~

~~The goal is to find benchmarks that are heavily dependent on memory alignment and make it hard for us to reliably detect regressions, discuss the solution, and then apply it.~~

~~Currently, I am opening this PR just so others can see the code changes that were needed to make it work and perhaps give it a try.~~

~~Next, I am going to create an issue with my findings, we are going to discuss it and see what we do next.~~


danmosemsft:
This will be interesting to @kunalspathak

adamsitnik:
I've changed this PR to not enable the Memory Randomization, but just refactor the initialization logic to allow for enabling Memory Randomization in the near future. So this PR should have no effect on the reported time as of now.

Memory Randomization will change the behavior of `[GlobalSetup]` methods, which will be invoked not just at the beginning of the benchmarking, but once before every benchmark iteration. The actual randomization is going to happen before it, so everything allocated in the setup will have random alignment and this will allow us to get truly random memory alignment for each iteration (so far everything was allocated just once and whatever alignment we got, it remained unchanged for all iterations).

@DrewScoggins @kunalspathak please take a look.

adamsitnik:
The CI failures are not related to microbenchmarks, so the PR should be ready to merge

kunalspathak:
Thanks @adamsitnik for putting this together. Are we also doing anything to avoid allocating inside the benchmark method itself to reduce the effect of GC? Several array benchmarks allocate an array inside the benchmark method.

kunalspathak:
@AndyAyersMS - FYI.

DrewScoggins:
I am working through this PR file by file. Should have comments later today.

kunalspathak:
I skimmed through the PR and looks ok. I will let @DrewScoggins do the sign off.

adamsitnik:
> Are we also doing anything to avoid allocating inside the benchmark method itself to reduce the effect of GC? Several array benchmarks allocate an array inside the benchmark method.

We recommend not to allocate any arrays in the new benchmarks (https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#setup). When it comes to the old ones they were left mostly unchanged when I was porting them to BenchmarkDotNet. One of the requirements was to use the existing Reporting System (called BenchView back then). This has enforced the rule of not changing the IDs (type, method, and argument names), and not modifying the benchmarks (unless they had bugs) to not introduce any changes in the reported time (to keep the trends as they were).

If these allocations are a problem, we should move them to setup methods. @kunalspathak would you like to send a PR based on your recent research?

kunalspathak:
> If these allocations are a problem, we should move them to setup methods. @kunalspathak would you like to send a PR based on your recent research?

Absolutely.

adamsitnik:
@DrewScoggins ping

kunalspathak:
I just realized that there are certain benchmarks which allocates data to `static` variables. 
https://github.com/dotnet/performance/blob/358051eb43feed5a036954fbc3b956f4774f270a/src/benchmarks/micro/libraries/System.Linq/LinqTestData.cs#L14-L22

Should this also need fixup?

src/benchmarks/micro/libraries/System.Text.Json/Document/EnumerateArray.cs
```diff
@@ -63,7 +63,7 @@ public void EnumerateUsingIndexer()
             int arrayLength = _element.GetArrayLength();
             for (int j = 0; j < arrayLength; j++)
             {
-                JsonElement dummy = _element[j];
```
DrewScoggins, line 66: Will this still work in 3.1 and 5.0? I don't remember what version of C# this is from.

src/benchmarks/micro/libraries/System.Runtime/Perf.Array.cs
```diff
@@ -273,43 +285,27 @@ public void ArrayResize()
...
-        public void CleanupClearUnaligned() => _clearableArrayHandle.Free();
+        [GlobalSetup(Target = nameof(Clear))]
+        public void SetupClear() => _clearableArray = new byte[8192];
  
-        private GCHandle _clearableArrayHandle;
-        private byte[] _clearableArray;
-
         [Benchmark]
-        public void ClearUnaligned() => Array.Clear(_clearableArray, 0, _clearableArray.Length);
+        public void Clear() => Array.Clear(_clearableArray, 0, _clearableArray.Length);
```
DrewScoggins, line 298: Are we getting rid of the ClearUnaligned test here?

DrewScoggins (approved):
LGTM

src/benchmarks/micro/libraries/System.Text.Json/Document/EnumerateArray.cs
adamsitnik: Yes, the discards were added in C# 7.0 ([doc](https://docs.microsoft.com/en-us/dotnet/csharp/discards)) and according to [this](https://stackoverflow.com/a/51729639/5852046) StackOverflow question, it's supported in .NET Core 1.0+. The CI leg for .NET 4.6.1 is also green, so it's 👍

src/benchmarks/micro/libraries/System.Runtime/Perf.Array.cs
adamsitnik: Yes, as we should rather force BDN to perform the memory alignment randomization instead of having a setup that tries to allocate an unaligned array ([source code](https://github.com/dotnet/performance/blob/4e14ecc54759224121b807c70e46fb1d9ee0e7d6/src/benchmarks/micro/libraries/System.Runtime/Perf.Array.cs#L281-L296)) (so we have always unaligned memory compared to having aligned and unaligned)

adamsitnik:
@kunalspathak @DrewScoggins thank you for your reviews!

PR #1590, opened 2020-11-10
Support the various size range parameters

cshung:


cshung:
@dotnet/gc

src/benchmarks/gc/docs/bench_file.md
```diff
@@ -58,7 +61,7 @@ benchmarks:
 comment: `str | None`
   (ignored)
 
-vary: `"machine" | "executable" | "coreclr" | "config" | "benchmark" | None`
+vary: `"machine" | "coreclr" | "config" | "benchmark" | "executable" | None`
```
Maoni0, line 64: not sure why the order was changed :P

src/benchmarks/gc/docs/bench_file.md
```diff
@@ -36,6 +36,9 @@ benchmarks:
       tlgb: 1
       lohar: 0
       pohar: 0
+      sohsr: 10-4000
```
Maoni0, line 39: do you want to also change the name of this profile from nosurvive to something like surv_2percent? 'cause it's clearly surviving... seems odd to call it nosurvive.

ivdiazsa:
Looks good to me aside from Maoni's observations to be fixed :)

src/benchmarks/gc/docs/bench_file.md
cshung: The file is automatically generated by `py . lint`. Presumably, it wants the options to be alphabetically ordered.

Maoni0: but it's not alphabetically ordered 😄... I can't tell what order this is. @ivdiazsa if you run the linter does it generate the same order for you? if not we may want to disable this from the linter (also I'm surprised since I thought you ran the linter on the whole code base).

ivdiazsa: Yes, it also happened to me. I don't know how the linter determines which order to use for these values. Since it looked odd to me, I manually undid this change in my PR's.

cshung: Let's keep it the way the linter likes so we don't have to revert it all the time? I don't know how the linter determines that order either.

Maoni0 (approved):

PR #1594, opened 2020-11-17
Added individual GC's statistics analysis feature to GC Benchmarking Infra

ivdiazsa:
This PR adds a new component to the currently existing `pandas` feature in GC Infra's Jupyter Notebook. This new component allows the user to perform statistics analysis on individual GC's metrics from a given trace.

First of all, you need the trace you want to analyze. For this component, it doesn't have to necessarily be a *GCPerfSim* trace. If got from elsewhere, you only need to write its own *test status yaml file* (More info on `docs/test_status_files.md`):

```yaml
success: true
trace_file_name: mytrace.etl
process_id: 123   # You can also give the process_name instead.
```

Once you have your trace ready to be processed by GC Infra, set everything up with the following code snippet in Jupyter Notebook:

```python3
_BENCH = Path("bench")
_SUITE = Path("bench") / "suite"
_TRACE_PATH = _SUITE / "normal_server.yaml.out"
_TRACE_DATA = get_trace_with_everything(_TRACE_PATH / "defgcperfsim__a__only_config__2gb__0.yaml")

gc_metrics_values = get_pergc_metrics_numbers_for_jupyter(_TRACE_DATA.gcs[0:10])

dframe = pandas.DataFrame.from_dict(gc_metrics_values)
```

In this example, we are using a trace generated by running *GCPerfSim's* `normal_server` test. You adjust the path variable to your own trace's location. The function in charge of retrieving the GC's data is `get_pergc_metrics_numbers_for_jupyter()`. It receives a list of GC's as argument, which you can get from the object returned by `get_trace_with_everything()`. This function in turn receives the path to your test status file. In this example, we are only analyzing the first 10 GC's for simplicity, but you can adjust this range or omit it entirely to process all the GC's from the trace.

We have all the data now, but it's not ready to be consumed by `pandas`. `Pandas` expects a dictionary which symbolizes the table you would usually build in statistics, and transforms it into a `DataFrame` of its own. The last line in the previous code snippet does this. To summarize, the dictionary is composed as follows:

* **Keys**: GC Metric Name
* **Values**: List with said metric's numbers from each processed GC.

With that `data frame`, you can now use all of `pandas` functionalities. Let's see three quick examples. Starting at the basic `describe()` use case, you can see overall statistics from the GC's:

```python
dframe.describe()
```

PandasGCDescribe

Now, you will most likely want to see each GC's metrics printed nicely on a table. This is easily done by asking `pandas` to group the data by GC Number. It needs a statistical method to display however. Since each GC only has one value per metric, there is no harm in calling the `mean()` method.

```python
dframe.groupby(["Number"]).mean()
```

PandasGCByNumber

Another common use case is to look only at certain metrics from the GC's as having all of them sometimes makes it hard to find the ones you are most interested in, at any certain point of your investigation. In this example, we want to check the allocation rate, as well as Large Object Heap sizes.

```python
dframe[["Number", "AllocRateMBSec", "LOHSizeAfterMB", "LOHSizeBeforeMB"]].groupby("Number").mean()
```

PandasGCGroupBy

It is worth noting that to have `pandas` display the information of the GC's ordered by number, you have to use the `groupby(["Number"])` function on the data frame, as shown in the last two examples.

Many more things can be done for analyzing this data using pandas. The main purpose of this feature is to get all the information from the GC's of a trace, in order to have it ready for use with pandas.


ivdiazsa:
@dotnet/gc

cshung:
I was thinking about this work Ivan and I were doing recently. I think we should keep it and get it merged, and here is why:

The key contribution to this PR is that we can drill into the measurement that we take per GC aggregated across multiple runs that are potentially running under different configurations. In a table format, we can plot the per GC measurement aggregates against the configurations, something that wasn't easy before.

Looking a bit further, having the data available in the `pandas` form allows us to use data science tools (e.g. `scikit-learn`). In particular, I am interested in doing some regression analysis that tries to predict key performance statistics when the scenario changes, which will give us a better understanding of how scenario variations impact performance.

PR #1603, opened 2020-11-18
extend the workflow doc with instructions for .dlls that are not part of the shared SDK

adamsitnik:


adamsitnik:
@eerhardt @maryamariyan I know that it's more or less a workaround, but before we find a clean solution I would be happy to have it documented so everyone can benchmark local builds of Microsoft.Extensions etc

maryamariyan (approved):
Thanks for the note. 

cc: @safern @tarekgh

PR #1606, opened 2020-12-01
add more MemoryCache benchmarks

adamsitnik:


src/benchmarks/micro/libraries/Microsoft.Extensions.Caching.Memory/MemoryCacheTests.cs
```diff
@@ -22,7 +26,7 @@ public void Setup()
             }
         }
 
-        [GlobalCleanup]
+        [GlobalCleanup(Targets = new[] { nameof(GetHit), nameof(TryGetValueHit), nameof(GetMiss), nameof(TryGetValueMiss), nameof(SetOverride) })]
         public void Cleanup() => _memCache.Dispose();
```
eerhardt, line 30:
```suggestion
        public void CleanupBasic() => _memCache.Dispose();
```

To match `SetupBasic`?

src/benchmarks/micro/libraries/Microsoft.Extensions.Caching.Memory/MemoryCacheTests.cs
```diff
@@ -39,5 +43,66 @@ public void Setup()
...
+                {
+                    cache.Set(item.key, item.value, relative);
+                }
+
+                foreach (var item in _items)
+                {
+                    cache.Remove(item.key);
+                }
+            }
+        }
```
eerhardt, line 106: Should we add a `SlidingExpiration` test as well? The above test is for `absoluteExpirationRelativeToNow` AFAICT.

eerhardt (approved):
LGTM

src/benchmarks/micro/libraries/Microsoft.Extensions.Caching.Memory/MemoryCacheTests.cs
adamsitnik: @eerhardt we should, thanks for pointing this out! I could not find an [extension method](https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs) that would allow for setting `SlidingExpiration` so I've changed all the benchmarks to use the `using` pattern for adding the entries (used by the extension methods)

adamsitnik:
the CI failures are not related to this change, I am merging

Issue #1616, opened 2020-12-11
GCPerfSim fails with an Overflow Arithmetic Exception


ivdiazsa:
As of recent, `GCPerfSim` has been no longer working. When running any benchmark (either standalone or through Python), it fails with an `overflow Arithmetic Exception`. Following is the command line and the error message.

```
C:\performance\artifacts\bin\GCPerfSim\release\netcoreapp5.0\GCPerfSim.exe -tc 1 -tagb 150.0 -tlgb 2.0 -lohar 0 -pohar 0 -sohsi 50 -lohsi 0 -pohsi 0 -sohsr 10-4000 -lohsr 102400-204800 -pohsr 100-204800 -sohpi 0 -lohpi 0 -sohfi 0 -lohfi 0 -pohfi 0 -allocType reference -testKind time
GCPerfSim C:\performance\artifacts\bin\GCPerfSim\release\netcoreapp5.0\GCPerfSim.dll -tc 1 -tagb 150.0 -tlgb 2.0 -lohar 0 -pohar 0 -sohsi 50 -lohsi 0 -pohsi 0 -sohsr 10-4000 -lohsr 102400-204800 -pohsr 100-204800 -sohpi 0 -lohpi 0 -sohfi 0 -lohfi 0 -pohfi 0 -allocType reference -testKind time
allocating 161,061,273,600 per thread
Running 64-bit? True
PID: 13084
Running 1 threads.
time, ReferenceItem, tlgb 2, tagb 150, totalMins 0, buckets:
    10-4000; surv every 50; pin every 0; weight 1000; isPoh False
Arithmetic operation resulted in an overflow.
   at SimpleRefPayLoad..ctor(UInt32 size, Boolean isPinned, Boolean isPoh) in C:\performance\src\benchmarks\gc\src\exec\GCPerfSim\GCPerfSim.cs:line 429
   at ReferenceItemWithSize..ctor(UInt32 size, Boolean isPinned, Boolean isPoh) in C:\performance\src\benchmarks\gc\src\exec\GCPerfSim\GCPerfSim.cs:line 527
   at ReferenceItemWithSize.ReferenceItemWithSizeNonFinalizable..ctor(UInt32 size, Boolean isPinned, Boolean isPoh) in C:\performance\src\benchmarks\gc\src\exec\GCPerfSim\GCPerfSim.cs:line 586
[...]

PR #1621, opened 2020-12-12
Fix wrong SOH size range in GCPerfSim

ivdiazsa:
This PR fixes issue #1616.

The test bench files were being created with a size range of *10-4000* for SOH objects, when running `suite-create`. Having a low boundary this small was causing `GCPerfSim` to fail with an `Arithmetic Exception` when attempting to create `SimpleRefPayload` objects. This is the code snippet that failed:

```csharp
    public SimpleRefPayLoad(uint size, bool isPinned, bool isPoh)
    {
        uint sizePayload = size - Overhead;
        if (isPoh)
        {
#if NET5_0
            payload = GC.AllocateArray<byte>((int)sizePayload, pinned: true);
#else
            throw new Exception("UNREACHABLE: POH allocations require netcoreapp5.0 or higher");
#endif
        }
        else
        {
            payload = new byte[sizePayload];
        }
        (...)
    }
```

Having a boundary as low as 10 caused `sizePayload` to be negative in certain iterations, thus throwing the exception. This makes sense because this value is then used to create the `payload` array.

The fix was to set the lower boundary to 100 in the Python suite create, in order to match GCPerfSim's default value.

### DIFF
--- a/src/benchmarks/gc/src/commonlib/bench_file.py
+++ b/src/benchmarks/gc/src/commonlib/bench_file.py
@@ -606,7 +606,7 @@ class GCPerfSimArgs:
     totalMins: Optional[float] = None
     lohar: int = 0
     pohar: int = 0
-    sohsr: str = "10-4000"
+    sohsr: str = "100-4000"
     lohsr: str = "102400-204800"
     pohsr: str = "100-204800"
     sohsi: int = 0


ivdiazsa:
@dotnet/gc

cshung:
Thanks for the fix, that was a typo.

Maoni0:
ideally we'd like to be able to allocate smaller objects - for the sizes that are too small for what SimpleRefPayload can handle we should just create another type of objects. I'll merge this for now but we should make that a workitem since the average size of objects are somewhere around 60 or 70 bytes. @ivdiazsa could you please do that so we don't forget?

ivdiazsa:
Hi @Maoni0 Sure thing! I will open an issue here in the `performance` repo to keep track of that. The purpose of its PR when it's worked on will be to add a new type of objects to GCPerfSim so we are able to give smaller size ranges without it failing correct?

cshung:
@Maoni0, shall we actually merge this?

Maoni0:
yes, I'll merge. thanks for the reminder!

PR #1622, opened 2020-12-12
Remove Extension and Activity tests

DrewScoggins:
Just removed tests that seemed to be 5.0 and above. I have run these locally and they seem to work.


billwert (commented):
LGTM.

adamsitnik:
> Just removed tests that seemed to be 5.0 and above

@DrewScoggins I am not sure what you are working on, but if you are working on generating some 3.1 baselines it would be great to have the 3.1 numbers for  `Microsoft.Extensions.*`  benchmarks. Have you considered downgrading `Microsoft.Extensions.*` libraries to their older versions? Like https://www.nuget.org/packages/Microsoft.Extensions.Caching.Memory/3.1.10 ?

PR #1633, opened 2021-01-18
FileStream benchmarks

adamsitnik:
Main changes:

- add benchmark that copies one file into another (existing benchmark was copying from file to `Stream.Null`)
- add Open&Close benchmark
- add Lock&Unlcok benchmark
- add Seek (forward and backward) benchmarks
- give better names to differentiate user buffer (passed to stream) and internal stream buffer (passed by Stream to OS)
- add more file sizes (1kB, 1MB, 100MB) where it makes sense

cc @stephentoub

### DIFF
--- a/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
+++ b/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
@@ -74,7 +74,7 @@ namespace System.IO.Tests
             
             using (FileStream reader = new FileStream(_filePath, FileMode.Open, FileAccess.Read, FileShare.Read, DefaultBufferSize, FileOptions.None))
             {
-                for (int i = 0; i < TotalSize / BufferSize; i++)
+                while (bytesRead < TotalSize)
                 {
                     bytesRead += reader.Read(buffer, 0, buffer.Length);
                 }
@@ -92,7 +92,7 @@ namespace System.IO.Tests
             
             using (FileStream reader = new FileStream(_filePath, FileMode.Open, FileAccess.Read, FileShare.Read, DefaultBufferSize, FileOptions.Asynchronous))
             {
-                for (int i = 0; i < TotalSize / BufferSize; i++)
+                while (bytesRead < TotalSize)
                 {
                     bytesRead += await reader.ReadAsync(buffer, 0, buffer.Length);
                 }
--- a/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
+++ b/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
@@ -30,7 +30,7 @@ namespace System.IO.Tests
             _userBuffers = new Dictionary<int, byte[]>()
             {
                 { HalfKiloByte, ValuesGenerator.Array<byte>(HalfKiloByte) },
-                { FourKiloBytes, ValuesGenerator.Array<byte>(HalfKiloByte) },
+                { FourKiloBytes, ValuesGenerator.Array<byte>(FourKiloBytes) },
             };
             _sourceFilePaths = fileSizes.ToDictionary(size => size, size => CreateFileWithRandomContent(size));
             _destinationFilePaths = fileSizes.ToDictionary(size => size, size => CreateFileWithRandomContent(size));
--- a/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
+++ b/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
@@ -142,7 +142,7 @@ namespace System.IO.Tests
             }
         }
 
-        [GlobalSetup(Targets = new[] { nameof(Read), nameof(ReadAsyncArray), "ReadAsyncMemory", nameof(CopyToFile), nameof(CopyToFileAsync), nameof(Write), nameof(WriteAsyncArray), "WriteAsyncMemory" })]
+        [GlobalSetup(Targets = new[] { nameof(Read), "ReadAsync", nameof(Write), "WriteAsync", nameof(CopyToFile), nameof(CopyToFileAsync) })]
         public void SetupBigFileBenchmarks() => Setup(OneKibibyte , OneMibibyte, HundredMibibytes);
 
         [Benchmark]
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -2,152 +2,299 @@
...
 namespace System.IO.Tests
 {
     [BenchmarkCategory(Categories.Libraries)]
     public class Perf_FileStream
     {
-        private const int DefaultBufferSize = 4096;
-        
-        [Params(DefaultBufferSize / 8, 200000)]
-        public int BufferSize;
+        private const int OneKiloByte = 1_000;
```
danmosemsft, line 18: Utter nit, I believe we write as "Kilobyte" -- it's one word. Eg., "Kilometer", "Kilogram" etc

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -2,152 +2,299 @@
...
+        private const int HalfKiloByte = OneKiloByte / 2;
+        private const int FourKiloBytes = OneKiloByte * 4; // default Stream buffer size
+        private const int OneMegaByte = OneKiloByte * 1_000;
+        private const int HundredMegaBytes = 1_000_000 * 100;
```
danmosemsft, line 22: should these all be powers of two? Ie., what you call Kibibytes, Mibibytes etc. 

Since the default FileStream buffer is a Kibibyte = 1024, not 1000.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
danmosemsft, line 18: May be interesting to review "real world" values:

https://grep.app/search?q=new%20FileStream.%2A0x&regexp=true&filter[lang][0]=C%23
https://grep.app/search?q=new%20FileStream.%2A%5Cd%7B4%2C%7D&regexp=true&filter[lang][0]=C%23
https://grep.app/search?q=new%20FileStream.%2Asize%3A%5Cs%3F.%5B%5E4%5D%5Cd%7B3%2C%7D&regexp=true&filter[lang][0]=C%23

0x1000 / 4096 dominate (redundantly)
I also see 0x1_000 and )x16, 2000, 1024, 512 << 10, 64 * 1024

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -2,152 +2,299 @@
...
+        [Arguments(HundredMegaBytes, FourKiloBytes, FileOptions.None)]
+        public long Read(long fileSize, int userBufferSize, FileOptions options)
         {
-            byte[] buffer = _buffer;
-            int bytesRead = 0;
-            
-            using (FileStream reader = new FileStream(_filePath, FileMode.Open, FileAccess.Read, FileShare.Read, DefaultBufferSize, FileOptions.Asynchronous))
+            byte[] userBuffer = _userBuffers[userBufferSize];
+            long bytesRead = 0;
+            using (FileStream fileStream = new FileStream(_sourceFilePaths[fileSize], FileMode.Open, FileAccess.Read, FileShare.Read, FourKiloBytes, options))
```
stephentoub, line 159: We also end up measuring the open/close time.  That's ok?

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -2,152 +2,299 @@
...
         }
 
         [Benchmark]
-        public void WriteByte()
+        [Arguments(OneKiloByte, HalfKiloByte, FileOptions.Asynchronous)]
+        [Arguments(OneKiloByte, FourKiloBytes, FileOptions.Asynchronous)]
+        [Arguments(OneMegaByte, HalfKiloByte, FileOptions.Asynchronous)]
+        [Arguments(OneMegaByte, FourKiloBytes, FileOptions.Asynchronous)]
+        [Arguments(HundredMegaBytes, HalfKiloByte, FileOptions.Asynchronous)]
+        [Arguments(HundredMegaBytes, FourKiloBytes, FileOptions.Asynchronous)]
```
stephentoub, line 195: Only FileOptions.Asynchronous?

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -2,152 +2,299 @@
...
-        
+
         [Benchmark]
-        public void Write()
+        [Arguments(OneKiloByte, HalfKiloByte, FileOptions.Asynchronous)]
+        [Arguments(OneKiloByte, FourKiloBytes, FileOptions.Asynchronous)]
+        [Arguments(OneMegaByte, HalfKiloByte, FileOptions.Asynchronous)]
+        [Arguments(OneMegaByte, FourKiloBytes, FileOptions.Asynchronous)]
+        [Arguments(HundredMegaBytes, HalfKiloByte, FileOptions.Asynchronous)]
+        [Arguments(HundredMegaBytes, FourKiloBytes, FileOptions.Asynchronous)]
```
stephentoub, line 218: Same question.  I'd have thought we'd want FileOptions.None as well.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -2,152 +2,299 @@
...
-            using (FileStream writer = new FileStream(_filePath, FileMode.Create, FileAccess.Write, FileShare.Read, DefaultBufferSize, FileOptions.None))
+            byte[] userBuffer = _userBuffers[userBufferSize];
+            long bytesRead = 0;
+            using (FileStream fileStream = new FileStream(_sourceFilePaths[fileSize], FileMode.Open, FileAccess.Read, FileShare.Read, FourKiloBytes, options))
             {
-                for (int i = 0; i < TotalSize; i++)
+                while (bytesRead < fileSize)
                 {
-                    writer.WriteByte(default);
+                    bytesRead += await fileStream.ReadAsync(userBuffer, 0, userBuffer.Length);
```
stephentoub, line 205: If we're only going to choose one overload, I'd prefer it be the one that takes a Memory.  Same for Write.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -2,152 +2,299 @@
...
+            using (FileStream destination = new FileStream(_destinationFilePaths[fileSize], FileMode.Create, FileAccess.Write, FileShare.Read, FourKiloBytes, options))
+            {
+                source.CopyTo(destination);
+            }
+        }
+
+        [Benchmark]
+        [Arguments(OneKiloByte, FileOptions.None)]
+        [Arguments(OneMegaByte, FileOptions.None)]
+        [Arguments(HundredMegaBytes, FileOptions.None)]
```
stephentoub, line 248: Only FileOptions.None and not FileOptions.Asynchronous?

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
adamsitnik: @stephentoub Did you mean to use other flags like `FileOptions.RandomAccess`, `.FileOptions.SequentialScan` or `FileOptions.None`?

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
adamsitnik: I would like this benchmark to answer the question "How long does it take to read X bytes from a file?". Since opening (and closing) the file is mandatory to read it, I think that it's OK to include it here.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
adamsitnik: Perhaps we should have both of them? The array overload allows for apples to apples comparison with Full Framework, which sometimes is nice to have. And it's also most probably the most frequently used overload?

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
adamsitnik: I thought that `FileOptions.None` combined with async file operations is more like an anti-pattern. Is it not?

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
stephentoub: None

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
stephentoub: > is more like an anti-pattern

It's not as simple as that.  It's very common to have a file opened for non-async I/O and then use the async methods with it, often because of separation of concerns or composition.  For example, some component opens a file and hands the stream to you, and then you use it as part of an operation like `networkStream.CopyToAsync(destination)`.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
stephentoub: > The array overload allows for apples to apples comparison with Full Framework

Does that matter?

> And it's also most probably the most frequently used overload?

It might be today.  It shouldn't be in the future.  For example, we've changed almost all use of Stream.ReadAsync(byte[], ...) throughout the core libraries where we're consuming some stream to be Stream.ReadAsync(Memory, ...), and [CA1835](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1835) will recommend to developers that they switch overloads and provides an auto-fixer to do so.

If someone cares about performance, they should be using these newer overloads.  Today ReadAsync(Memory, ...) enables zero-allocation for synchronously completing operations, and I'm hopeful that with the .NET 6 FileStream work that will be extended to both ReadAsync and WriteAsync for asynchronously completion as well (ammortized zero-allocation).  They enable pre-pinned memory to be used and avoid the GCHandle costs that are otherwise required.  And they enable reading/writing with non-array-based sources, potentially saving on copying costs to get things into arrays.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
adamsitnik: @stephentoub now I get it. I've added `FileOptions.None` to all async benchmarks. Thanks!

adamsitnik:
@Jozkee @carlossanlop the CI failures are unrelated, PTAL

Jozkee (approved):
I wonder if we should measure `Flush` as well?
That aside, looks great to me!

adamsitnik:
> I wonder if we should measure Flush as well?

Great point, I've added benchmark and verified with profiler that most of the time (85%) is spent in Flushing

obraz

PR #1634, opened 2021-01-19
Test System.Threading.Channels with non-default ExecutionContext

KorneiDontsov:
The added test shows the difference between usages of channels with and without ExecutionContext. It's good to catch ExecutionContext-related regressions in future.
Also it's used to test https://github.com/dotnet/runtime/pull/46458.

### DIFF
--- a/src/benchmarks/micro/libraries/System.Threading.Channels/Perf.Channel.cs
+++ b/src/benchmarks/micro/libraries/System.Threading.Channels/Perf.Channel.cs
@@ -26,9 +26,12 @@ namespace System.Threading.Channels.Tests
     [BenchmarkCategory(Categories.Libraries)]
     public abstract class PerfTests
     {
+        private const int PingPongCount = 10_000;
+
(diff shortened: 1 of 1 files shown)


KorneiDontsov:
Didn't get why some checks failed. Need help

danmoseley:
@dotnet/area-system-threading-channels could one of you please review? I'm going through old PRs. is this a change we want?

stephentoub:
> https://github.com/orgs/dotnet/teams/area-system-threading-channels could one of you please review? I'm going through old PRs. is this a change we want?

If we don't have perf tests around ExecutionContext capture and flow in general, or in the context of async/await, that would be fine to add.  But I don't think there's anything really specific to channels here.  It does have its own implementation of IValueTaskSource, but that's just using ExecutionContext.Capture and ExecutionContext.Run as would any IValueTaskSource.

danmoseley:
I'm going to close this as it's not clear that it is the right way to fill that gap. Thanks.

PR #1637, opened 2021-01-26
Add tests for new Random API and for the distinct unseeded path

danmosemsft:
Tests for changes in https://github.com/dotnet/runtime/pull/47390

cc @colgreen

### DIFF
--- a/src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Random.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Random.cs
@@ -32,7 +32,7 @@ namespace System.Tests
 
         [Benchmark]
         public int Next_int_int_unseeded() => _randomUnseeded.Next(100, 10000);
-
+#if NET6_0 // New API in .NET 6.0
         [Benchmark]
         public long Next_long() => _random.NextInt64(2^20);
 
@@ -44,23 +44,25 @@ namespace System.Tests
 
         [Benchmark]
         public long Next_long_long_unseeded() => _randomUnseeded.NextInt64(100, 10000);
-
+#endif // NET6_0
         [Benchmark]
         public void NextBytes() => _random.NextBytes(_bytes);
 
         [Benchmark]
         public void NextBytes_unseeded() => _randomUnseeded.NextBytes(_bytes);
 
+#if NET6_0 // New API in .NET 6.0
         [Benchmark]
-        public single NextSingle() => _random.NextSingle();
+        public float NextSingle() => _random.NextSingle();
 
         [Benchmark]
-        public single NextSingle_unseeded() => _randomUnseeded.NextSingle();
-
+        public float NextSingle_unseeded() => _randomUnseeded.NextSingle();
+#endif // NET6_0
         [Benchmark]
         public double NextDouble() => _random.NextDouble();
 
         [Benchmark]
         public double NextDouble_unseeded() => _randomUnseeded.NextDouble();
+
     }
 }
--- a/src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Random.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Random.cs
@@ -32,7 +32,28 @@ namespace System.Tests
 
         [Benchmark]
         public int Next_int_int_unseeded() => _randomUnseeded.Next(100, 10000);
-#if NET6_0 // New API in .NET 6.0
+
+        [Benchmark]
+        public void NextBytes() => _random.NextBytes(_bytes);
+
+        [Benchmark]
+        public void NextBytes_unseeded() => _randomUnseeded.NextBytes(_bytes);
+
+        [Benchmark]
+        public double NextDouble() => _random.NextDouble();
+
+        [Benchmark]
+        public double NextDouble_unseeded() => _randomUnseeded.NextDouble();
+
+#if !NETFRAMEWORK // New API in .NET Core 2.1
+        [Benchmark]
+        public void NextBytes_span() => _random.NextBytes(_bytes.AsSpan());
+
+        [Benchmark]
+        public void NextBytes_span_unseeded() => _randomUnseeded.NextBytes(_bytes.AsSpan());
+#endif
+
+#if !NETFRAMEWORK && !NETCOREAPP2_1 && !NETCOREAPP3_1 && !NET5_0 // New API in .NET 6.0
         [Benchmark]
         public long Next_long() => _random.NextInt64(2^20);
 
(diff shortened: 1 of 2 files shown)


danmosemsft:
BenchmarkDotNet=v0.12.1.1466-nightly, OS=Windows 10.0.19042
Intel Core i7-10510U CPU 1.80GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.100-preview.1.21075.9
  [Host]     : .NET 6.0.0 (6.0.21.7403), X64 RyuJIT
  Job-CFOLDV : .NET 6.0.0 (6.0.21.7403), X64 RyuJIT

PowerPlanMode=00000000-0000-0000-0000-000000000000  Arguments=/p:DebugType=portable  IterationTime=250.0000 ms
MaxIterationCount=20  MinIterationCount=15  WarmupCount=1

|                  Method |         Mean |       Error |      StdDev |       Median |          Min |           Max |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------ |-------------:|------------:|------------:|-------------:|-------------:|--------------:|-------:|------:|------:|----------:|
|                    ctor |   266.672 ns |  34.0699 ns |  39.2349 ns |   269.363 ns |   203.979 ns |    351.375 ns | 0.0163 |     - |     - |      72 B |
|             ctor_seeded | 1,180.033 ns | 202.4371 ns | 233.1268 ns | 1,171.310 ns |   806.414 ns |  1,609.254 ns | 0.0727 |     - |     - |     312 B |
|                Next_int |    31.807 ns |   1.5263 ns |   1.6965 ns |    31.997 ns |    29.147 ns |     35.128 ns |      - |     - |     - |         - |
|       Next_int_unseeded |    12.383 ns |   0.7647 ns |   0.7853 ns |    12.044 ns |    11.532 ns |     14.451 ns |      - |     - |     - |         - |
|            Next_int_int |    15.260 ns |   0.8519 ns |   0.9115 ns |    15.455 ns |    13.209 ns |     16.281 ns |      - |     - |     - |         - |
|   Next_int_int_unseeded |    13.295 ns |   0.5646 ns |   0.6275 ns |    12.951 ns |    12.849 ns |     15.017 ns |      - |     - |     - |         - |
|               Next_long |   148.044 ns |   2.2936 ns |   2.1455 ns |   147.844 ns |   145.846 ns |    151.795 ns |      - |     - |     - |         - |
|      Next_long_unseeded |     8.163 ns |   0.0936 ns |   0.0782 ns |     8.126 ns |     8.096 ns |      8.374 ns |      - |     - |     - |         - |
|          Next_long_long |   227.569 ns |  53.0941 ns |  61.1432 ns |   193.907 ns |   177.535 ns |    342.688 ns |      - |     - |     - |         - |
| Next_long_long_unseeded |    16.942 ns |   0.6433 ns |   0.7151 ns |    16.899 ns |    15.933 ns |     18.531 ns |      - |     - |     - |         - |
|               NextBytes | 9,092.806 ns | 697.9947 ns | 746.8465 ns | 8,789.110 ns | 8,379.180 ns | 11,109.769 ns |      - |     - |     - |         - |
|      NextBytes_unseeded |   353.442 ns |   6.5098 ns |   6.3935 ns |   350.555 ns |   347.158 ns |    370.248 ns |      - |     - |     - |         - |
|              NextDouble |    14.138 ns |   0.2560 ns |   0.2138 ns |    14.213 ns |    13.725 ns |     14.430 ns |      - |     - |     - |         - |
|     NextDouble_unseeded |     3.877 ns |   0.1161 ns |   0.1140 ns |     3.890 ns |     3.720 ns |      4.119 ns |      - |     - |     - |         - |

colgreen:
Couple of nits:
- You flipped from ctor_seeded to Next_int_ **unseeded**  :)  
- There is a NextSingle() method now too.

danmosemsft:
> You flipped from ctor_seeded to Next_int_ unseeded :)

Yes, this is because I want to avoid changing benchmark names, so we don't break history. And unfortunately the ctor was the exception in that it measured unseeded. (@adamsitnik I assume there isn't an attribute that would allow me to have consistent method names while still being recorded under the old test name...)

> There is a NextSingle() method now too.

I'll add.

src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Random.cs
```diff
@@ -10,22 +10,50 @@ namespace System.Tests
...
+        public long Next_long_long() => _random.NextInt64(100, 10000);
+
+        [Benchmark]
+        public long Next_long_long_unseeded() => _randomUnseeded.NextInt64(100, 10000);
+
         [Benchmark]
         public void NextBytes() => _random.NextBytes(_bytes);
 
+        [Benchmark]
+        public void NextBytes_unseeded() => _randomUnseeded.NextBytes(_bytes);
```
stephentoub, line 51: If you're going for API completeness, NextBytes(Span) and NextBytes(byte[]) have different perf characteristics, at least for seeded.

danmosemsft:
OK, should be done now.

I added a bunch of #if's for new API since it seemed more readable in this case than separate files which is what we mostly seem to do in this repo.

billwert:
It looks like the actual failure was that installation of the SDK was not reliable this afternoon. [(For example)](https://dev.azure.com/dnceng/public/_build/results?buildId=966893&view=logs&j=bdf114a7-fb55-5dd7-7bb6-a93c797038fb&t=d2c22e55-057a-5805-846b-9ec89feb82fb&l=23). These are probably fine to ignore. I've got a work item to fix this up.

danmosemsft:
@stephentoub any more feedback ? As noted above, I can change to 3 partial classes, but I think this is one of those cases where #if actually makes it much clearer what's going on.

danmosemsft:
Thanks @BillWert.. would retrying help or not?

stephentoub:
> As noted above, I can change to 3 partial classes, but I think this is one of those cases where #if actually makes it much clearer what's going on.

I'll defer to @adamsitnik.  I prefer the partial files approach, but if the preferred scheme for this repo is ifdefs, that's fine.

src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Random.cs
```diff
@@ -10,22 +10,67 @@ namespace System.Tests
     [BenchmarkCategory(Categories.Libraries)]
     public class Perf_Random
     {
+        Random _randomUnseeded = new Random();
```
stephentoub, line 13: Nit: these fields should be `private`

stephentoub (approved):

danmosemsft:
Merging with CI failures due to missing SDK on some machines, noted by @billwert 

@adamsitnik if you believe partial classes would be clearer, LMK and I'll put up a new PR.

adamsitnik:
> I assume there isn't an attribute that would allow me to have consistent method names while still being recorded under the old test name.

You are right. Perhaps we should create such an attribute?

> I added a bunch of #if's for new API since it seemed more readable in this case than separate files which is what we mostly seem to do in this repo.

Great!

@danmosemsft big thanks for adding the benchmarks!

@colgreen very nice improvements!

PR #1639, opened 2021-01-27
Add BinaryWriter perf tests

GrabYourPitchforks:
Adds benchmarks for various `BinaryWriter.Write` APIs.

### DIFF
--- a/src/benchmarks/micro/libraries/System.IO/BinaryWriterExtendedTests.cs
+++ b/src/benchmarks/micro/libraries/System.IO/BinaryWriterExtendedTests.cs
@@ -1,0 +1,41 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.IO.Tests
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public class BinaryWriterExtendedTests
+    {
+        private string _input;
+        private char[] _inputAsChars;
+        private BinaryWriter _bw;
+
+        [Params(4, 16, 512, 10_000, 100_000, 500_000, 2_000_000)]
+        public int StringLengthInChars;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            _bw = new BinaryWriter(new NullWriteStream());
+
+            _input = new string('x', StringLengthInChars);
+            _inputAsChars = _input.ToCharArray();
+        }
+
+        [Benchmark]
+        public void WriteAsciiCharArray()
+        {
+            _bw.Write(_inputAsChars);
+        }
+
+        [Benchmark]
+        public void WriteAsciiString()
+        {
+            _bw.Write(_input);
+        }
+    }
+}
--- a/src/benchmarks/micro/libraries/System.IO/BinaryWriterExtendedTests.cs
+++ b/src/benchmarks/micro/libraries/System.IO/BinaryWriterExtendedTests.cs
@@ -14,7 +14,7 @@ namespace System.IO.Tests
         private char[] _inputAsChars;
         private BinaryWriter _bw;
 
-        [Params(4, 16, 512, 10_000, 100_000, 500_000, 2_000_000)]
+        [Params(32, 8_000, 2_000_000)]
         public int StringLengthInChars;
 
         [GlobalSetup]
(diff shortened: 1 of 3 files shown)


src/benchmarks/micro/libraries/System.IO/NullWriteStream.cs
```diff
@@ -0,0 +1,62 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.IO.Tests
+{
+    /// <summary>
+    /// A <see cref="Stream"/> that acts as a null sink for data. Overrides members that
+    /// <see cref="Stream.Null"/> does not. Used for benchmarking wrappers around Stream
```
adamsitnik, line 12: thank you for adding a clear explanation why `Stream.Null` is not enugh 👍

src/benchmarks/micro/libraries/System.IO/NullWriteStream.cs
```diff
@@ -0,0 +1,62 @@
...
+        public override void Write(byte[] buffer, int offset, int count) { }
+
+        public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            return Task.CompletedTask;
+        }
+
+        public override void WriteByte(byte value) { }
+
+#if NETCOREAPP2_1_OR_GREATER // these virtual methods only exist in .NET Core 2.1+
```
adamsitnik, line 53: this is something new to me. Are you sure that it works outside of `dotnet/runtime`? cc @ViktorHofer

src/benchmarks/micro/libraries/System.IO/NullWriteStream.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public override long Seek(long offset, SeekOrigin origin)
+        {
+            throw new NotSupportedException();
+        }
+
+        public override void SetLength(long value)
+        {
+            throw new NotSupportedException();
+        }
```
adamsitnik, line 42: minor nit: it would be nice to be consistent with other methods and properties in this type and use method expressions everywhere

```suggestion
        public override int Read(byte[] buffer, int offset, int count) => throw new NotSupportedException();

        public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();

        public override void SetLength(long value) => throw new NotSupportedException();
```

src/benchmarks/micro/libraries/System.IO/BinaryWriterExtendedTests.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace System.IO.Tests
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public class BinaryWriterExtendedTests
+    {
+        private string _input;
+        private char[] _inputAsChars;
+        private BinaryWriter _bw;
+
+        [Params(4, 16, 512, 10_000, 100_000, 500_000, 2_000_000)]
```
adamsitnik, line 17: can we reduce the number of test cases to the number of different code paths? https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#Code-Paths

adamsitnik (approved):
LGTM, thank you @GrabYourPitchforks !

src/benchmarks/micro/libraries/System.IO/NullWriteStream.cs
GrabYourPitchforks: @terrajobst Aren't these defines part of the latest compilers? In theory they should work in any project type in any repo, even non-MSFT code.

src/benchmarks/micro/libraries/System.IO/BinaryWriterExtendedTests.cs
GrabYourPitchforks: Sure, there should be 3 different code paths. How about 32 chars, 8k chars, and 2M chars?

GrabYourPitchforks:
BTW I'm not quite sure how to debug the CI failures. Looks like a .nupkg dependency is missing in one of the legs?

billwert:
CI is a real mess right now for a bunch of reasons, not the least of which today is that there's been Nuget outages. Don't worry about it.

src/benchmarks/micro/libraries/System.IO/BinaryWriterExtendedTests.cs
adamsitnik: >How about 32 chars, 8k chars, and 2M chars?

👍

PR #1644, opened 2021-02-01
Add inner loop tests

DrewScoggins:


src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
```diff
@@ -135,7 +135,8 @@ public ProcessHelper(Logger logger)
                     if (!exited)
                     {
                         process.Kill();
-                        return (Result.CloseFailed, pid);
+                        Thread.Sleep(3000);
+                        return (Result.Success, pid);
```
billwert, line 139: Why is this a success case now?

eng/performance/sdk_scenarios.proj
```diff
@@ -103,5 +103,15 @@
       <Timeout>4:00</Timeout>
     </HelixWorkItem>
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetsWindows)' == 'true'">
+    <HelixWorkItem Include="SDK Inner Loop Console">
```
billwert, line 108: nit: "SDK" sort of implies ownership in the SDK team. Is that appropriate here?

src/scenarios/emptyconsoletemplateinnerloop/test.py
```diff
@@ -0,0 +1,14 @@
...
+'''
+from shared.runner import TestTraits, Runner
+
+EXENAME = 'emptyconsoletemplateinnerloop'
+
+if __name__ == "__main__":
+    traits = TestTraits(exename=EXENAME, 
+                        startupmetric='InnerLoop',
+                        guiapp='false',
+                        innerloopcommandargs='-c "from shutil import copyfile; copyfile(\'src/Program.cs\', \'app/Program.cs\')"'
```
billwert, line 12: should this use pathsep?

src/scenarios/shared/util.py
```diff
@@ -34,6 +34,10 @@ def builtexe(exename: str):
     'gets binary path'
     return os.path.join(const.BINDIR, '%s%s' % (exename, extension()))
 
+def appfolder(projname: str):
+    'gets path for callign dotnet run'
```
billwert, line 38:
```suggestion
    'gets path for calling dotnet run'
```

src/scenarios/shared/util.py
```diff
@@ -34,6 +34,10 @@ def builtexe(exename: str):
...
+    return os.path.join(const.APPDIR, '%s.csproj' % projname)
```
billwert, line 39: should this include the project extension? We have some VB floating around etc.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -161,13 +183,48 @@ static void checkArg(string arg, string name)
...
+
+                            logger.LogIterationHeader($"Iteration {i} - Diff");
+                            iterationResult = RunIteration(null, TestProcess, cleanupProcHelper, logger);
+                            if (!iterationResult.Success)
+                            {
+                                failed = true;
+                                break;
+                            }
+                            pids.Add(iterationResult.Pid);
+                            InnerLoopMarkerEventSource.Log.EndIteration();
```
billwert, line 225: could this be just after `RunIteration`?

billwert: oh, I guess it doesn't matter since these are just fences and not measurement spots.

billwert (changes requested):
Some questions, but otherwise looks great!

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
DrewScoggins: This is dead code that I need to rip out. Good catch!

src/tools/ScenarioMeasurement/Startup/Startup.cs
DrewScoggins: Yeah, I just use them as fences to let me know which side of the test we are in.

eng/performance/sdk_scenarios.proj
DrewScoggins: Yeah, this was just a copy paste. I will change this.

src/scenarios/emptyconsoletemplateinnerloop/test.py
DrewScoggins: Yeah, I should probably use os.join to create these paths.

src/scenarios/shared/util.py
DrewScoggins: Done, had to solve this by adding a new test trait, but it is pretty clean.

billwert (approved):
LGTM!

PR #1652, opened 2021-02-09
Enable AMD64 Owl machines

billwert:


adamsitnik:
@billwert does it mean that we are now going to have regular perf runs on AMD? That would be great!

cc @tannergooding

billwert:
That's the plan! Once we get the bugs worked out. ;) Need to figure out the plan for dotnet/runtime. We have far fewer of these (10:1) than we do Intel machines, so we won't be able to run them as often as we run the other queue. We'll probably setup a timed pipeline for the runtime repo to get a run or two a day like we do with the installer repo.

PR #1660, opened 2021-02-09
Add AzDo feed for BDN

billwert:


billwert:
@adamsitnik the consequence of this nuget feed change is that we will need to push new versions of BDN if we want a new nightly to it. @DrewScoggins and I can do that. (We do this rarely, but I wanted to FYI you.)

adamsitnik:
@billwert thanks for letting me know 👍

PR #1665, opened 2021-02-15
Add PriorityQueue benchmarks

eiriktsarpalis:
Draft PR, requires an SDK build containing PriorityQueue.

Cf. https://github.com/dotnet/runtime/pull/46009

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -109,6 +109,10 @@
     <Compile Remove="libraries\System.Formats.Cbor\*.cs" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0')">
+    <Compile Remove="libraries\System.Collections\PriorityQueue\Perf_PriorityQueue.cs" />
+  </ItemGroup>
+
   <ItemGroup>
     <!-- Workaround https://github.com/dotnet/project-system/issues/935 -->
     <None Include="**/*.cs" />
--- a/src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
+++ b/src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Linq;
 using System.Collections.Generic;
 using BenchmarkDotNet.Attributes;
 using MicroBenchmarks;
@@ -15,7 +16,9 @@ namespace System.Collections.Tests
         public int Size;
 
         private int[] _priorities;
+        private (int Element, int Priority)[] _elements;
         private PriorityQueue<int, int> _priorityQueue;
+        private PriorityQueue<int, int> _populatedPriorityQueue;
 
         [GlobalSetup]
         public void Setup()
@@ -27,20 +30,55 @@ namespace System.Collections.Tests
                 _priorities[i] = random.Next();
             }
 
+            _elements = _priorities.Select((i, x) => (i, x)).ToArray();
             _priorityQueue = new PriorityQueue<int, int>(initialCapacity: Size);
+            _populatedPriorityQueue = new PriorityQueue<int, int>(_elements);
         }
 
         [Benchmark]
         public void HeapSort()
         {
             var queue = _priorityQueue;
+            queue.EnqueueRange(_elements);
+
+            while (queue.Count > 0)
+            {
+                queue.Dequeue();
+            }
+        }
+
+        [Benchmark]
+        public void Enumerate()
+        {
+            foreach (var _ in _populatedPriorityQueue.UnorderedItems)
+            {
+
+            }
+        }
+
+        [Benchmark]
+        public void Dequeue_And_Enqueue()
+        {
+            // benchmarks dequeue and enqueue operations
+            // for heaps of fixed size.
+
+            var queue = _priorityQueue;
             var priorities = _priorities;
 
+            // populate the heap: incorporated in the 
+            // benchmark to achieve determinism.
             for (int i = 0; i < Size; i++)
             {
                 queue.Enqueue(i, priorities[i]);
             }
 
+            for (int i = 0; i < Size; i++)
+            {
+                queue.Dequeue();
+                queue.Enqueue(i, priorities[i]);
+            }
+
+            // Drain the heap
             while (queue.Count > 0)
             {
                 queue.Dequeue();
--- a/src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
+++ b/src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
@@ -10,36 +10,38 @@ using MicroBenchmarks;
 namespace System.Collections.Tests
 {
     [BenchmarkCategory(Categories.Libraries, Categories.Collections, Categories.GenericCollections)]
-    public class Perf_PriorityQueue
+    public abstract class Perf_PriorityQueue<TElement, TPriority>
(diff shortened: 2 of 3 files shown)


src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
```diff
@@ -0,0 +1,73 @@
...
+                queue.EnqueueDequeue(i, _priorities[i]);
+            }
+
+            for (int i = 0; i < k; i++)
+            {
+                queue.Dequeue();
+            }
+        }
+    }
+}
```
stephentoub, line 73: Worth adding tests for EnqueueRange, enumeration, or anything else?

eiriktsarpalis:
Looks like the latest nightly SDK includes PriorityQueue, I presume this is safe to merge now.

src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.Collections.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.Collections, Categories.GenericCollections)]
+    public abstract class Perf_PriorityQueue<TElement, TPriority>
+    {
+        [Params(10, 100, 1000, 10_000, 100_000)]
```
adamsitnik, line 15: does change the Size affects the code path that is being executed? If not, we should limit the number of test cases (1-2) https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#code-paths

edit: I can see that it's being passed as `initialCapacity` so it does. We should then most probably go with the default value and some typical size value.

src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
```diff
@@ -0,0 +1,142 @@
...
+            for (int i = 0; i < k; i++)
+            {
+                queue.Dequeue();
+            }
+        }
+    }
+
+    public class Perf_PriorityQueue_String_String : Perf_PriorityQueue<string, string>
+    {
+        public override IEnumerable<(string Element, string Priority)> GenerateItems(int count)
```
adamsitnik, line 110: Please use [ValuesGenerator](https://github.com/dotnet/performance/blob/master/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs) instead of generating the test data on your own (it supports both arrays of strings and integers).

https://github.com/dotnet/performance/blob/master/docs/microbenchmark-design-guidelines.md#always-the-same-input-data

src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
```diff
@@ -0,0 +1,142 @@
...
+
+using System.Linq;
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.Collections.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.Collections, Categories.GenericCollections)]
+    public abstract class Perf_PriorityQueue<TElement, TPriority>
```
adamsitnik, line 13: there is no need to introduce derived types, you can use the `GenericTypeArguments` attribute and BDN is going to instantiate the right thing for you.

```suggestion
    [GenericTypeArguments(typeof(int), typeof(int))]
    [GenericTypeArguments(typeof(string), typeof(string))]
    public class Perf_PriorityQueue<TElement, TPriority>
```

adamsitnik (commented):
Overall looks good to me, but we can use some more of the existing BDN features to simplify the benchmarks. Thank you for adding new benchmarks as soon as it was possible @eiriktsarpalis !

src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
eiriktsarpalis: Thanks, I wasn't aware of ValuesGenerator which is why I resorted to inheritance instead.

src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
```diff
@@ -0,0 +1,110 @@
...
+        public int Size;
+
+        private (TElement Element, TPriority Priority)[] _items;
+        private PriorityQueue<TElement, TPriority> _priorityQueue;
+        private PriorityQueue<TElement, TPriority> _prePopulatedPriorityQueue;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            _items = ValuesGenerator.Array<TElement>(Size).Zip(ValuesGenerator.Array<TPriority>(Size)).ToArray();
```
adamsitnik, line 30: `ValuesGenerator` always generates the same data and here `TElement` is always the same as `TPriority` so for each `(TElement Element, TPriority Priority)` `Element` and `Priority` are going to be equal. Is that OK?

If not we could do sth like this:

```suggestion
            var elements = ValuesGenerator.Array<TElement>(Size * 2);
            var priorities = ValuesGenerator.Array<TPriority>(Size * 2);
            _items = elements.Take(Size).Zip(priorities.Skip(Size)).ToArray();
```

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -135,5 +137,12 @@ private static string GenerateRandomString(Random random, int minLength, int max
 
             return builder.ToString();
         }
+
+        private static Guid GenerateRandomGuid(Random random)
+        {
+            byte[] bytes = new byte[16];
+            random.NextBytes(bytes);
```
adamsitnik, line 144: thank you for adding `Guid` support and ensuring that we are always going to generate the same data by reusing the `random` instance! 👍

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -109,6 +109,10 @@
     <Compile Remove="libraries\System.Formats.Cbor\*.cs" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0')">
```
adamsitnik, line 112: I wonder whether it would be possible to simplify the condition to just `<= 6.0` check:

```suggestion
  <ItemGroup Condition=" '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0' ">
```

src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
eiriktsarpalis: It shouldn't matter in this case, since `priorities` is what really drives the heap formation. Alternatively I was thinking we could update `ValuesGenerator` to also include support for tuple generation.

src/benchmarks/micro/MicroBenchmarks.csproj
eiriktsarpalis: I just commited it, let's see if CI works.

src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
stephentoub: Is performance going to be impacted measurably based on whether all priorities are different or many priorities are the same?

eiriktsarpalis: If many priorities are the same then I expect that the average depth of a siftup/siftdown operation would be shortened. If all priorities were the same then enqueue and dequeue becomes _O(1)_ (since the heap property is always satisfied).

adamsitnik (approved):
:shipit:

PR #1691, opened 2021-02-18
Delete RegisterAndUnregister_Parallel cancellation test

stephentoub:
We've decided this isn't a relevant scenario to track.  The type was originally optimized for this scenario at the expense of serial usage, but since then the majority use case is the serial one.

cc: @adamsitnik


adamsitnik (approved):
LGTM, thank you @stephentoub

PR #1693, opened 2021-02-19
Fixed GCPerHeapHistories in Workstation GC's in GC Infra

ivdiazsa:
This PR removes some code that was preventing GC Infra from fetching `GCPerHeapHistories` data when the trace in question came from an app running Workstation GC. The GC properties that were detected that required this information are the following:

* `Gen2FreeListSpaceBeforeMB`
* `Gen2FreeListSpaceAfterMB`
* `LOHFreeListSpaceBeforeMB`
* `LOHFreeListSpaceAfterMB`

When one tried to access these properties in the scenario described above, they would be greeted with the following error:

`Exception: Workstation GC has no AbstractGCPerHeapHistories.`

This behavior is wrong as if we check *GCStats* directly in PerfView, we can observe that regardless of GC Type, they do have *GCPerHeapHistories*.

With this fix, the infra now reads this information properly and can be used for charting, calculations, and any other supported functionality.

### DIFF
--- a/src/benchmarks/gc/src/analysis/process_trace.py
+++ b/src/benchmarks/gc/src/analysis/process_trace.py
@@ -236,18 +236,15 @@ def _get_processed_gc(
 
 
 def _get_per_heap_histories(gc: AbstractTraceGC) -> Sequence[Result[str, AbstractGCPerHeapHistory]]:
-    if gc.HeapCount == 1:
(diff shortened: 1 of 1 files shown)


ivdiazsa:
@dotnet/gc

Maoni0 (approved):

PR #1694, opened 2021-02-19
Add Version perf tests

stephentoub:


danmoseley (approved):
Is the codegen different when the out parameter of format is a discard?

danmoseley:
@billwert  could you please look at the build failure? It is not related to the change.

stephentoub:
> Is the codegen different when the out parameter of format is a discard?

Only if it could be inlined (it can't, and I don't see it ever being inlineable by default).

src/benchmarks/micro/libraries/System.Runtime/Perf.Version.cs
```diff
@@ -0,0 +1,52 @@
...
+        private Version _v2 = new Version(1, 2);
+        private Version _v3 = new Version(1, 2, 3);
+        private Version _v4 = new Version(1, 2, 3, 4);
+        private Version _vL = new Version(int.MaxValue, int.MaxValue, int.MaxValue, int.MaxValue);
+
+        [Benchmark]
+        public Version Ctor4() => new Version(1, 2, 3, 4);
+
+        [Benchmark]
+        public Version Parse4() => Version.Parse("1.2.3.4");
```
adamsitnik, line 22: it would be great to have some `TryParse` benchmarks as well:

```suggestion
        [Benchmark]
        public Version Parse4() => Version.Parse("1.2.3.4");
        
        [Benchmark]
        public bool TryParse4() => Version.TryParse("1.2.3.4", out _);
```

src/benchmarks/micro/libraries/System.Runtime/Perf.Version.cs
```diff
@@ -0,0 +1,52 @@
...
+    [BenchmarkCategory(Categories.Libraries)]
+    public class Perf_Version
+    {
+        private Version _v2 = new Version(1, 2);
+        private Version _v3 = new Version(1, 2, 3);
+        private Version _v4 = new Version(1, 2, 3, 4);
+        private Version _vL = new Version(int.MaxValue, int.MaxValue, int.MaxValue, int.MaxValue);
+
+        [Benchmark]
+        public Version Ctor4() => new Version(1, 2, 3, 4);
```
adamsitnik, line 19: is there any particular reason why we have the `2`, `3` and `4` variants for `ToString` and `TryFormat` but not for `ctor` and `Parse`?

stephentoub:
Thanks, @adamsitnik.  I added the tests you requested.

adamsitnik (approved):
Perfect! thank you @stephentoub :shipit:

PR #1703, opened 2021-03-01
Upload Failed Test Runs

Tohron:
This PR is to update the python scripts so that failed runs upload JSONs and use a different blob container, called failed_results. I haven't created the container yet, since I want other people to look over things first.

### DIFF
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -105,6 +105,15 @@ jobs:
         - ${{ if eq(parameters.osName, 'windows') }}:
           - script: (robocopy $(Build.SourcesDirectory) $(Build.SourcesDirectory)\notLocked /E /XD $(Build.SourcesDirectory)\notLocked $(Build.SourcesDirectory)\artifacts $(Build.SourcesDirectory)\.git) ^& IF %ERRORLEVEL% LEQ 1 exit 0
             displayName: Workaround to avoid locked file # https://github.com/dotnet/arcade/issues/2125
+          - script: $(CorrelationStaging)dotnet\dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r win-${{parameters.architecture}} $(Build.SourcesDirectory)\src\tools\Reporting\FailureReporting\FailureReporting.csproj
+            displayName: Build FailureReporter tool
+            env:
+              PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
+        - ${{ if eq(parameters.osName, 'linux') }}:
+          - script: $(CorrelationStaging)dotnet/dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r linux-${{parameters.architecture}} $(Build.SourcesDirectory)/src/tools/Reporting/FailureReporting/FailureReporting.csproj
+            displayName: Build FailureReporter tool
+            env:
+              PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
         - template: /eng/performance/send-to-helix.yml
           parameters:
             HelixSource: '$(HelixSourcePrefix)/dotnet/performance/$(Build.SourceBranch)' # sources must start with pr/, official/, prodcon/, or agent/
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -105,12 +105,12 @@ jobs:
         - ${{ if eq(parameters.osName, 'windows') }}:
           - script: (robocopy $(Build.SourcesDirectory) $(Build.SourcesDirectory)\notLocked /E /XD $(Build.SourcesDirectory)\notLocked $(Build.SourcesDirectory)\artifacts $(Build.SourcesDirectory)\.git) ^& IF %ERRORLEVEL% LEQ 1 exit 0
             displayName: Workaround to avoid locked file # https://github.com/dotnet/arcade/issues/2125
-          - script: $(CorrelationStaging)dotnet\dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r win-${{parameters.architecture}} $(Build.SourcesDirectory)\src\tools\Reporting\FailureReporting\FailureReporting.csproj
+          - script: $(Build.SourcesDirectory)\notLocked\tools\dotnet\${{parameters.architecture}}\dotnet publish -c Release -o $(Build.SourcesDirectory)\notLocked\FailureReporter -f $(_Framework) -r win-${{parameters.architecture}} $(Build.SourcesDirectory)\src\tools\Reporting\FailureReporting\FailureReporting.csproj
             displayName: Build FailureReporter tool
             env:
               PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
         - ${{ if eq(parameters.osName, 'linux') }}:
-          - script: $(CorrelationStaging)dotnet/dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r linux-${{parameters.architecture}} $(Build.SourcesDirectory)/src/tools/Reporting/FailureReporting/FailureReporting.csproj
+          - script: $(Build.SourcesDirectory)\tools\dotnet\${{parameters.architecture}}\dotnet publish -c Release -o $(Build.SourcesDirectory)\FailureReporter -f $(_Framework) -r linux-${{parameters.architecture}} $(Build.SourcesDirectory)/src/tools/Reporting/FailureReporting/FailureReporting.csproj
             displayName: Build FailureReporter tool
             env:
               PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
--- a/eng/performance/scenarios.yml
+++ b/eng/performance/scenarios.yml
@@ -93,6 +93,10 @@ jobs:
             displayName: Build SOD tool
             env:
               PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
+          - script: $(CorrelationStaging)dotnet\dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r win-${{parameters.architecture}} $(Build.SourcesDirectory)\src\tools\Reporting\FailureReporting\FailureReporting.csproj
+            displayName: Build FailureReporter tool
+            env:
+              PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
         - ${{ if ne(parameters.osName, 'windows') }}:
           - script: cp ./NuGet.config $(CorrelationStaging);cp -r ./scripts $(CorrelationStaging)scripts;cp -r ./src/scenarios/shared $(CorrelationStaging)shared;cp -r ./src/scenarios/staticdeps $(CorrelationStaging)staticdeps
             displayName: Copy python libraries and NuGet.config
--- a/eng/performance/scenarios.yml
+++ b/eng/performance/scenarios.yml
@@ -108,6 +108,10 @@ jobs:
             displayName: Build SOD tool
             env:
               PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
+          - script: $(CorrelationStaging)dotnet\dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r linux-${{parameters.architecture}} $(Build.SourcesDirectory)\src\tools\Reporting\FailureReporting\FailureReporting.csproj
+            displayName: Build FailureReporter tool
+            env:
+              PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
         - template: /eng/performance/send-to-helix.yml
           parameters:
             HelixSource: '$(HelixSourcePrefix)/dotnet/performance/$(Build.SourceBranch)' # sources must start with pr/, official/, prodcon/, or agent/
--- a/eng/performance/scenarios.yml
+++ b/eng/performance/scenarios.yml
@@ -108,7 +108,7 @@ jobs:
             displayName: Build SOD tool
             env:
               PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
-          - script: $(CorrelationStaging)dotnet\dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r linux-${{parameters.architecture}} $(Build.SourcesDirectory)\src\tools\Reporting\FailureReporting\FailureReporting.csproj
+          - script: $(CorrelationStaging)dotnet/dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r linux-${{parameters.architecture}} $(Build.SourcesDirectory)/src/tools/Reporting/FailureReporting/FailureReporting.csproj
             displayName: Build FailureReporter tool
             env:
               PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -221,7 +221,7 @@ def __main(args: list) -> int:
 
     # Run micro-benchmarks
     if not args.build_only:
-        upload_container = UPLOAD_CONTAINER # Unsure if correct approach
+        upload_container = UPLOAD_CONTAINER
         try:
             for framework in args.frameworks:
                 micro_benchmarks.run(
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -232,7 +232,7 @@ def __main(args: list) -> int:
                     args
                 )
         except CalledProcessError:
-            upload_container = 'failed_results'
+            upload_container = 'failedresults'
             
         dotnet.shutdown_server(verbose)
 
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -221,6 +221,7 @@ def __main(args: list) -> int:
 
     # Run micro-benchmarks
     if not args.build_only:
+        upload_container = UPLOAD_CONTAINER # Unsure if correct approach
         for framework in args.frameworks:
             micro_benchmarks.run(
                 BENCHMARKS_CSPROJ,
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -31,6 +31,7 @@ from performance.common import validate_supported_runtime, get_artifacts_directo
 from performance.logger import setup_loggers
 from performance.constants import UPLOAD_CONTAINER, UPLOAD_STORAGE_URI, UPLOAD_TOKEN_VAR, UPLOAD_QUEUE
 from channel_map import ChannelMap
+from subprocess import Popen
 
 import dotnet
 import micro_benchmarks
@@ -222,6 +223,7 @@ def __main(args: list) -> int:
     # Run micro-benchmarks
     if not args.build_only:
         upload_container = UPLOAD_CONTAINER
+        globpath
         try:
             for framework in args.frameworks:
                 micro_benchmarks.run(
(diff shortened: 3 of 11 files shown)


scripts/benchmarks_ci.py
```diff
@@ -239,7 +243,7 @@ def __main(args: list) -> int:
                 '**',
                 '*perf-lab-report.json')
 
-            upload.upload(globpath, UPLOAD_CONTAINER, UPLOAD_QUEUE, UPLOAD_TOKEN_VAR, UPLOAD_STORAGE_URI)
+            upload.upload(globpath, upload_container, UPLOAD_QUEUE, UPLOAD_TOKEN_VAR, UPLOAD_STORAGE_URI)
```
DrewScoggins, line 246: So currently, this will follow the same path as when we succeed. Which is to grovel the artifact directory for all of the report JSONs, and then upload them one by one. We need to change this to instead, when we have a failure, create a result JSON without any test results, but with the run information, and upload just that one file.

DrewScoggins (changes requested):

scripts/benchmarks_ci.py
Tohron: How would I create a JSON in Python and where is the run information?

src/tools/Reporting/FailureReporting/FailureReporter.cs
```diff
@@ -0,0 +1,47 @@
...
+    public class FailureReporter
+    {
+        public static void CreateFailureReport(string reportJsonPath, Logger logger)
+        {
+            var reporter = Reporter.CreateReporter();
+            if (reporter.InLab && !String.IsNullOrEmpty(reportJsonPath))
+            {
+                File.WriteAllText(reportJsonPath, reporter.GetJson());
+            }
+            logger.Log(reporter.WriteResultTable());
```
DrewScoggins, line 17: I think we should remove this, because we will never have any results to write to a table. This also means that you can get rid of the Logger class.

src/tools/Reporting/FailureReporting/FailureReporting.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
DrewScoggins, line 4: I believe that you will need to target netcoreapp3.1 here, because by default when you use net standard you get a dll, and not a standalone executable. cc @billwert

scripts/benchmarks_ci.py
```diff
@@ -221,25 +222,31 @@ def __main(args: list) -> int:
...
+                    args
+                )
             globpath = os.path.join(
                 get_artifacts_directory() if not args.bdn_artifacts else args.bdn_artifacts,
                 '**',
                 '*perf-lab-report.json')
+        except CalledProcessError:
+            upload_container = 'failedresults'
+            globpath = os.path.join(get_artifacts_directory(), "failure-report.json")
+            Popen(["/bin", "FailureReporter.exe",  globpath])
```
DrewScoggins, line 243: Have you tested that this is path is where the FailureReporter executable ends up as part of the build?

DrewScoggins (changes requested):
This looks like a good start!

src/tools/Reporting/FailureReporting/FailureReporter.cs
billwert: agreed

src/tools/Reporting/FailureReporting/FailureReporting.csproj
billwert: Correct. This is a library project, not a console exe. You need to add `<OutputType>Exe</OutputType>` as well, and for completeness so it doesn't cause problems in the future I would use our standard version setting technique, seen here:

https://github.com/dotnet/performance/blob/master/src/tools/ScenarioMeasurement/SizeOnDisk/SizeOnDisk.csproj#L5-L7

Tohron: Do I just change those lines, or is there more I need to do?

src/tools/Reporting/FailureReporting/FailureReporter.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.IO;
+
+namespace FailureReporting
+{
+    public class FailureReporter
+    {
+        public static void CreateFailureReport(string reportJsonPath, Logger logger)
+        {
+            var reporter = Reporter.CreateReporter();
+            if (reporter.InLab && !String.IsNullOrEmpty(reportJsonPath))
```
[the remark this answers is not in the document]
Tohron: Okay, so seems that's correct then.

src/tools/Reporting/FailureReporting/FailureReporting.csproj
```diff
@@ -1,7 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+	<OutputType>Exe</OutputType>
```
billwert, line 4: nit: indention

src/tools/Reporting/FailureReporting/FailureReporting.csproj
```diff
@@ -1,7 +1,10 @@
...
+    <TargetFramework>$(PERFLAB_TARGET_FRAMEWORKS)</TargetFramework> 
+    <!-- Supported target frameworks --> 
+    <TargetFramework Condition="'$(TargetFramework)' == ''">netcoreapp3.0</TargetFramework>
```
billwert, line 7:
```suggestion
    <TargetFramework Condition="'$(TargetFramework)' == ''">netcoreapp3.1</TargetFramework> 
```

eng/performance/scenarios.yml
```diff
@@ -93,6 +93,10 @@ jobs:
             displayName: Build SOD tool
             env:
               PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
+          - script: $(CorrelationStaging)dotnet\dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r win-${{parameters.architecture}} $(Build.SourcesDirectory)\src\tools\Reporting\FailureReporting\FailureReporting.csproj
+            displayName: Build FailureReporter tool
+            env:
+              PERFLAB_TARGET_FRAMEWORKS: $(_Framework)
```
billwert, line 99: This also needs to be built on Linux (below.)

scripts/benchmarks_ci.py
```diff
@@ -221,25 +222,37 @@ def __main(args: list) -> int:
...
+            upload_container = 'failedresults'
+            globpath = os.path.join(
+                get_artifacts_directory(), 
+                'FailureReporter', 
+                'failure-report.json')
+            cmdline = [
+                'FailureReporting.exe', globpath
+            ]
+            RunCommand(cmdline, verbose=verbose).run(
+                os.path.join(get_artifacts_directory()))
```
billwert, line 249: Is this going to work properly in the context of the lab machines? In the other examples the build location is `$(CorrelationStaging)`. In local testing `get_artifacts_directory()` will work out, because the build winds up putting the tool in (mostly) the right place. (I'm actually a bit surprised this didn't need to be `get_artifacts_directory()\FailureReporting`.)

src/scenarios/init.ps1
```diff
@@ -27,7 +27,7 @@ function Setup-Env($directory){
 function Download-Dotnet($channel){
     Write-Host "Downloading dotnet from channel $channel"
     $dotnetScript= Join-Path "$scripts" 'dotnet.py' -Resolve
-    python $dotnetScript install --channels $channel -v
+    py -3 $dotnetScript install --channels $channel -v
```
billwert, line 30: Why is this necessary?

src/tools/Reporting/FailureReporting/FailureReporter.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using System.IO;
+
+namespace FailureReporting
+{
+    public class FailureReporter
+    {
+        public static void Main(string[] args)
+        {
+            CreateFailureReport(args[0]);
```
billwert, line 11: Can you add error handling and help if this is called without a parameter? Or called with a json that doesn't exist?

billwert (changes requested):

eng/performance/scenarios.yml
Tohron: How are the Linux build instructions distinct from the Windows ones?

billwert: The `yml` files have branches everywhere for Linux and Windows to work around path construction issues. This change is in the Windows branch, so it needs to be mirrored in the Linux branch below (but with the correct path separators.)

src/scenarios/init.ps1
Tohron: The original command wasn't running properly on my machine, and Drew suggested doing this general change to make things work reliably.

scripts/benchmarks_ci.py
Tohron: Do you mean get_artifacts_directory()\FailureReporter (based on the yaml)?  Is there a way to confirm that that directory would work?

src/tools/Reporting/FailureReporting/FailureReporter.cs
Tohron: What would be the desired outcome if the executable is called without a parameter (or with an invalid path)?

scripts/benchmarks_ci.py
billwert: If I check out your branch and build FailureReporter, it shows up here:
```
 D:\source\performance\src\tools\Reporting\FailureReporting   upload_failed_runs ≢ +1 ~0 -0 !                                                                                                                                                                                                       [12:11]
❯ dir D:\source\performance\artifacts\bin\FailureReporting\Debug\netcoreapp3.1\

    Directory: D:\source\performance\artifacts\bin\FailureReporting\Debug\netcoreapp3.1

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---           3/24/2021 12:11 PM          17036 FailureReporting.deps.json
-a---           3/24/2021 12:11 PM          11264 FailureReporting.dll
-a---           3/24/2021 12:11 PM         175104 FailureReporting.exe
-a---           3/24/2021 12:11 PM            180 FailureReporting.runtimeconfig.dev.json
-a---           3/24/2021 12:11 PM            154 FailureReporting.runtimeconfig.json
-a---           5/15/2018  9:01 PM          24024 Microsoft.DotNet.PlatformAbstractions.dll
-a---           4/22/2019  1:06 AM         669608 Newtonsoft.Json.dll
-a---           3/24/2021 12:11 PM          23040 Reporting.dll
```

`get_artifacts_directory()` simply returns `performance\artifacts`:
https://github.com/dotnet/performance/blob/5927768185cd876206f4001c5077964887ee6ce7/scripts/performance/common.py#L88

So I'm surprised your existing code worked, as `FailureReporting.exe` should not exist in the root of the `artifacts` directory.

However, this is a slightly different case than normal. `benchmarks_ci.py` tries hard to work in both local user contexts as well as lab contexts. Some pieces only need to work in the lab, however. This piece of it only ever has to work in the lab, and should never work in local user contexts. The important place to know this landed is somewhere in the correlation payload. By building to `$(CorrelationStaging)`, you ensure that the binary will get picked up and put in the right place by Helix. To find that on the other end, you can use a method like we have in the scenarios code:
https://github.com/dotnet/performance/blob/09356faeb90bbfefa0a9f128b63208ad5f95735f/src/scenarios/shared/util.py#L11-L15

I would not reference that file (`util.py`) directly from here, so perhaps some refactoring is useful to move that method to `common.py` or something (and then fix up existing call sites.)

src/tools/Reporting/FailureReporting/FailureReporter.cs
billwert: I'd probably print an error to the console (so we can see it in the logs) and exit.

src/tools/Reporting/FailureReporting/FailureReporter.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            if (args.Length == 0)
+            {
+                throw new ArgumentException("FailureReporting.exe was called without the required JSON path argument");
+            }
+            else if (args.Length > 1)
+            {
+                throw new ArgumentException("FailureReporting.exe was called with more than one argument");
+            }
+            else if (!Directory.Exists(Path.GetDirectoryName(args[0]))) {
```
billwert, line 19: nit: newline.
```suggestion
            else if (!Directory.Exists(Path.GetDirectoryName(args[0]))) 
            {
```

billwert, line 19: You can just use `File.Exists` here instead of `Directory.Exists`

Tohron: The argument is a directory name, since I need to know whether the directory exists, and the json file hasn't been created yet.  Are you sure File.Exists is correct for this?

scripts/benchmarks_ci.py
```diff
@@ -221,25 +222,37 @@ def __main(args: list) -> int:
...
+                    args
+                )
             globpath = os.path.join(
                 get_artifacts_directory() if not args.bdn_artifacts else args.bdn_artifacts,
                 '**',
                 '*perf-lab-report.json')
+        except CalledProcessError:
+            upload_container = 'failedresults'
+            globpath = os.path.join(
+                get_artifacts_directory(),
```
billwert, line 242: this use of `get_artifacts_directory()` is also not correct for the lab.

Tohron: Should I replace it with get_artifacts_directory() if not args.bdn_artifacts else args.bdn_artifacts as in the above case?

src/tools/Reporting/FailureReporting/FailureReporter.cs
billwert: Ah. This bit of  your change made me think you were passing in the location of an existing json file and not making a new one. My mistake.

https://github.com/dotnet/performance/pull/1703/files#diff-55f4d5691c43ac42f361039296fae993f269730c60f83fbd7c5aa103d3892879R241

In that case the directory passed in to create the file should be the helix upload directory. Look in `util.py` in the scenarios folder for a pointer there, and again you may want to refactor that code into `util.py`.

scripts/benchmarks_ci.py
DrewScoggins: Yes, this will ensure that the failure json will get written into the correct artifacts location.

scripts/benchmarks_ci.py
```diff
@@ -221,25 +222,37 @@ def __main(args: list) -> int:
...
+        except CalledProcessError:
+            upload_container = 'failedresults'
+            globpath = os.path.join(
+                get_artifacts_directory(), 
+                'FailureReporter', 
+                'failure-report.json')
+            cmdline = [
+                'FailureReporting.exe', globpath
+            ]
+            RunCommand(cmdline, verbose=verbose).run(
```
DrewScoggins, line 248: This path is not correct, you need to look back to the yml changes you made. We are saving the build executable to `$(CorrelationStaging)FailureReporter` which means that when we want to access it on the build machines we are going to need to look in the correlation payload directory. This is why Bill had mentioned the need to port over the helixpayload function from util.py to common.py. You will want to use that to get the proper directory that the FailureReporting.exe will be in. Also since you specify an out folder it will not use the `bin\FailureReporting\Debug\netcoreapp3.11, but will instead just be sitting in the folder you specified.

Tohron: Already fixed locally

eng/Version.Details.xml
```diff
@@ -19,13 +19,13 @@
     </Dependency>
   </ProductDependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="6.0.0-beta.21167.3">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="6.0.0-beta.21175.1">
```
DrewScoggins, line 22: Why is this in your PR now?

eng/common/generate-locproject.ps1
```diff
@@ -79,7 +79,7 @@ $locJson = @{
 }
 
 $json = ConvertTo-Json $locJson -Depth 5
-Write-Host "(NETCORE_ENGINEERING_TELEMETRY=Build) LocProject.json generated:`n`n$json`n`n"
+Write-Host "LocProject.json generated:`n`n$json`n`n"
```
DrewScoggins, line 82: Why is this in your PR now?

DrewScoggins:
It seems like this PR now has changes from other places. This includes the updates to Version.xml and changes in the performance setup scripts. These should be taken out for clarity.

eng/Version.Details.xml
Tohron: I think I mistakenly rebased off of darc-main-0fade34f-2916-414b-9f7c-5b5d8ffdef60.  How do I un-rebase off that commit?

src/scenarios/shared/util.py
```diff
@@ -8,12 +8,6 @@
 from shared import const
 from performance.constants import UPLOAD_TOKEN_VAR
 
-def helixpayload():
```
DrewScoggins, line 11: If you are going to remove this function from util.py then you will need to go through all of the places where we reference it from here and update them to use common.py instead. I would however recommend that you just instead keep the function definition here.

DrewScoggins:
Besides the comments I have it looks good so far.

The thing we now need to do is get the tool building and packaged as part of the payloads we send down to the Helix machines for the microbenchmarks. That will involve making changes in `benchmark_jobs.yml`. To add the building of the tool and making sure that it is included in what we send down to Helix.

The other piece of this that remains is adding the error checking and uploading for the scenario runs. The code we modified in `benchmark_ci.py` only affects the micro benchmarks. We will also need to put the same logic around the running an uploading that happens in `src\scenarios\shared\startup.py`

eng/performance/benchmark_jobs.yml
```diff
@@ -105,6 +105,15 @@ jobs:
         - ${{ if eq(parameters.osName, 'windows') }}:
           - script: (robocopy $(Build.SourcesDirectory) $(Build.SourcesDirectory)\notLocked /E /XD $(Build.SourcesDirectory)\notLocked $(Build.SourcesDirectory)\artifacts $(Build.SourcesDirectory)\.git) ^& IF %ERRORLEVEL% LEQ 1 exit 0
             displayName: Workaround to avoid locked file # https://github.com/dotnet/arcade/issues/2125
+          - script: $(CorrelationStaging)dotnet\dotnet publish -c Release -o $(CorrelationStaging)FailureReporter -f $(_Framework) -r win-${{parameters.architecture}} $(Build.SourcesDirectory)\src\tools\Reporting\FailureReporting\FailureReporting.csproj
```
DrewScoggins, line 108: The `$(CorrelationStaging)` variable does not exist here. Instead we either use the `$(Build.SourcesDirectory)` variable directly (this is on Ubuntu) or `$(Build.SourcesDirectory)\notLocked` to work around a locked file issue in AzDO. This means that you will need to adjust where we are outputting the FailureReporter binary as well as change where we are calling dotnet from. If you look at the `install` method in `dotnet.py` you can see the algorithm we use for installing dotnet. This algorithm is used by the call to `ci_setup.py` and how we download dotnet. We do not specify a location when doing micro benchmark runs so we install in the default location, but that means we need to use that location instead of the ones we have here.

billwert (approved):
:shipit:

scripts/benchmarks_ci.py
```diff
@@ -221,25 +222,37 @@ def __main(args: list) -> int:
...
+        except CalledProcessError:
+            upload_container = 'failedresults'
+            globpath = os.path.join(
+                get_artifacts_directory() if not args.bdn_artifacts else args.bdn_artifacts,
+                'FailureReporter', 
+                'failure-report.json')
+            cmdline = [
+                'FailureReporting.exe', globpath
+            ]
+            RunCommand(cmdline, verbose=verbose).run(
```
DrewScoggins, line 248: After we create the path to the executable, we should check to make sure it exists, and if it doesn't we should throw here so we don't upload something that we do not intend. This should also happen in the startup.py file.

src/scenarios/shared/startup.py
```diff
@@ -101,11 +102,24 @@ def runtests(self, traits: TestTraits):
...
+            upload_container = 'failedresults'
+            reportjson = os.path.join(
+                TRACEDIR,
+                'FailureReporter', 
+                'failure-report.json')
+            cmdline = [
+                'FailureReporting.exe', reportjson
+            ]
+            RunCommand(cmdline, verbose=True).run(
+                os.path.join(helixpayload(), 'FailureReporter'))
```
DrewScoggins, line 119: After we create the path to the executable, we should check to make sure it exists, and if it doesn't we should throw here so we don't upload something that we do not intend.

DrewScoggins (approved):
LGTM

PR #1724, opened 2021-03-12
[FileStream] No buffering benchmarks

adamsitnik:
This PR adds benchmarks for a scenario, where the user has disabled the `FileStream` buffering by specifying `1` as `bufferSize` in `FileStream` `ctor`:

```cs
new FileStream($path, FileMode.Create, FileAccess.Write, FileShare.Read, 1 /* disable the buffering */ , options))
```


adamsitnik:
Since it's important to get some historical data before we merge more `FileStream` changes, I am going to merge it now.

PR #1728, opened 2021-03-16
Update microbenchmarks readme

danmoseley:
Update to be net6.0 by default.
Move into docs as I think it's important enough to be there. Most people will come to this repo to run microbenchmarks.

### DIFF
--- a/src/benchmarks/micro/README.md
+++ b/src/benchmarks/micro/README.md
@@ -12,58 +12,58 @@ To learn more about designing benchmarks, please read [Microbenchmark Design Gui
 
 ## Quick Start
 
-The first thing that you need to choose is the Target Framework. Available options are: `netcoreapp2.1|netcoreapp3.1|net5.0|net461`. You can specify the target framework using `-f|--framework` argument. For the sake of simplicity, all examples below use `net5.0` as the target framework.
+The first thing that you need to choose is the Target Framework. Available options are: `netcoreapp2.1|netcoreapp3.1|net5.0|net6.0|net461`. You can specify the target framework using `-f|--framework` argument. For the sake of simplicity, all examples below use `net6.0` as the target framework.
 
 The following commands are run from the `src/benchmarks/micro` directory.
 
 To run the benchmarks in Interactive Mode, where you will be asked which benchmark(s) to run:
 
 ```cmd
-dotnet run -c Release -f net5.0
+dotnet run -c Release -f net6.0
 ```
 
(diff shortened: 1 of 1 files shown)


docs/microbenchmarks_workflow.md
```diff
@@ -12,58 +12,58 @@ To learn more about designing benchmarks, please read [Microbenchmark Design Gui
...
 
 ```cmd
-dotnet run -c Release -f netcoreapp3.1 --filter * --runtimes netcoreapp3.1 net5.0 corert
+dotnet run -c Release -f net5.0 --filter * --runtimes net5.0 net6.0
 ```
 
 ## Private Runtime Builds
 
-If you contribute to [dotnet/runtime](https://github.com/dotnet/runtime) and want to benchmark **local builds of .NET Core** you need to build [dotnet/runtime](https://github.com/dotnet/runtime) in Release (including tests) and then provide the path(s) to CoreRun(s). Provided CoreRun(s) will be used to execute every benchmark in a dedicated process:
+If you contribute to [dotnet/runtime](https://github.com/dotnet/runtime) and want to benchmark **local builds of .NET Core** you need to build [dotnet/runtime](https://github.com/dotnet/runtime) in Release (including tests - so a command similar to `build clr+lib+libs.tests -rc release -lc release`) and then provide the path(s) to CoreRun(s). Provided CoreRun(s) will be used to execute every benchmark in a dedicated process:
```
adamsitnik, line 51: Should it be clr+lib**s** ?

```suggestion
If you contribute to [dotnet/runtime](https://github.com/dotnet/runtime) and want to benchmark **local builds of .NET Core** you need to build [dotnet/runtime](https://github.com/dotnet/runtime) in Release (including tests - so a command similar to `build clr+libs+libs.tests -rc release -lc release`) and then provide the path(s) to CoreRun(s). Provided CoreRun(s) will be used to execute every benchmark in a dedicated process:
```

adamsitnik (changes requested):
@danmoseley thank you for keeping our docs up to date!

> Most people will come to this repo to run microbenchmarks.

If they come to this repo to run microbenchmarks, they already have the link to the right doc at the top of the docs list. If they go directly to the `src/benchmarks/micro`, the `README.me` is nicely printed:

obraz

With your change we lose the latter:

obraz

To be honest with you I don't like it and I would prefer to keep the old setting.

danmoseley:
You are right. I removed the move. How does it look now?

adamsitnik (approved):
LGTM, thank you @danmoseley !

PR #1734, opened 2021-03-17
Add NoAOT category

DrewScoggins:
Adding the NoAOT label to the JSON tests which cause a crash when run.


adamsitnik (approved):
LGTM, thank you @DrewScoggins

PR #1735, opened 2021-03-17
Update BDN version and fix OS type ambiguity

DrewScoggins:


src/benchmarks/micro/libraries/System.Drawing/Perf_Graphics_Transforms.cs
```diff
@@ -30,7 +30,7 @@ public void Setup()
         }
 
         [Benchmark]
-        [AllowedOperatingSystems("Graphics.TransformPoints is not implemented in libgdiplus yet. See dotnet/corefx 20884", OS.Windows)]
+        [AllowedOperatingSystems("Graphics.TransformPoints is not implemented in libgdiplus yet. See dotnet/corefx 20884", BenchmarkDotNet.Extensions.OS.Windows)]
```
adamsitnik, line 33: the type name conflict is caused by the fact that BDN has recently introduced an attribute that allows for exactly the same thing `AllowedOperatingSystems` does.

If you want you can remove [our implementation](https://github.com/dotnet/performance/blob/master/src/harness/BenchmarkDotNet.Extensions/OperatingSystems.cs) and use the new attribute provided by BDN:

https://github.com/dotnet/BenchmarkDotNet/blob/314a27fc26cbbc9aa15d7250ad12578aab632641/src/BenchmarkDotNet/Attributes/Filters/OperatingSystemsFilterAttribute.cs#L21

DrewScoggins: Is it OK, if we leave this for now, and I fix this in another PR? I am trying to get the mono AOT runs going ASAP, and when I tried to switch over I was getting some errors that I didn't want to deal with right then.

billwert (approved):
I think it's fine to clean up the `AllowedOperatingSystems` thing in another PR - please open an item to track doing so.

src/benchmarks/micro/libraries/System.Drawing/Perf_Graphics_Transforms.cs
adamsitnik: > Is it OK, if we leave this for now, and I fix this in another PR?

Sure!

DrewScoggins:
https://github.com/dotnet/performance/issues/1737

PR #1739, opened 2021-03-18
Move from master -> main

DrewScoggins:


scripts/dotnet.py
```diff
@@ -854,7 +854,7 @@ def __process_arguments(args: list):
         dest='channels',
         required=False,
         nargs='+',
-        default=['master'],
+        default=['main'],
```
billwert, line 857: This is the channel for the `dotnet-install` script, yes? Did they also change that?

DrewScoggins: I added a new channel mapping that goes from main -> tfm: net6.0, channel: master

adamsitnik (approved):
The microbenchmarks part LGTM, thank you for working on this @DrewScoggins !

billwert (approved):

billwert:
@DrewScoggins is this mergable?

DrewScoggins:
I need to rip out the changes to the bootstrap code. I will do that and then we can merge.

PR #1743, opened 2021-03-22
Read console output as utf8

### LINKED ISSUE #1742: Benchmarks failed with UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 223: character maps to <undefined>
I tried running perf benchmarks and ended up with this error: 

```
  File "D:\src\repos\performance\scripts\benchmarks_ci.py", line 247, in <module>
    __main(sys.argv[1:])
  File "D:\src\repos\performance\scripts\benchmarks_ci.py", line 225, in __main
    micro_benchmarks.run(
  File "D:\src\repos\performance\scripts\micro_benchmarks.py", line 309, in run
    BENCHMARKS_CSPROJ.run(
  File "D:\src\repos\performance\scripts\dotnet.py", line 457, in run
    RunCommand(cmdline, verbose=verbose).run(
  File "D:\src\repos\performance\scripts\performance\common.py", line 187, in run
    (returncode, quoted_cmdline) = self.__runinternal(working_directory)
  File "D:\src\repos\performance\scripts\performance\common.py", line 176, in __runinternal
  File "C:\Python39\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 223: character maps to <undefined>
```

@danmoseley had sent out an (internal) email suggesting those with a access to non-standard machines give this a try to help out .NET 6.0 perf efforts and I have a homebrew Ryzen 9 3900X box with lots of RAM which I thought would be a good machine to run this on. Too bad this failed  partway through (it didn't produce any net6.0 logs at all). If someone could help puzzle out what went wrong and fix this, I'll be happy to try again in a week. 

Are there additional logs I can share here (or offline) that could be of more help, for e.g.? 

/cc @adamsitnik

danmoseley:
Fix #1742

Python is apparently using the OEM codepage to decode the bytes it reads from console output pipe of the spawned dotnet process. If the dotnet process emits a sequence of bytes that is not represented in the OEM codepage, it will cause this error. System.Console.Write() will emit strings in the encoding returned by GetConsoleOutputCP(). I am guessing in this repro the console output encoding is set to UTF-8, and something in the output ends up encoding to 0x81.

E.g, if I do `chcp 65001` to change console codepage to UTF-8 and replace the top of microbenchmark.csproj with:
```xml
<Project Sdk="Microsoft.NET.Sdk" InitialTargets="echo">
<Target Name="echo">
<Warning Text="&#x81;"/>
</Target>
```
then run the benchmark script, it will fail in this way when trying to build that project. In this case the MSBuild console logger is using Console.Write(). I guess that dotnet.exe itself does so as well.

I am not sure why Python is decoding using the OEM codepage in this case. Grepping the Python code, I see a call GetConsoleOutputCP() but I didn't debug to see what is going on.

The fix is to add `errors='ignore'` to tell Python to ignore decoding errors. I considered also adding `encoding='utf-8'` but this could be equally wrong if the spawned process isn't using that encoding. If we see garbled text when folks have their console CP set to UTF-8 we would need to investigate further to figure out how to have Python properly read the console CP.

### DIFF
--- a/scripts/performance/common.py
+++ b/scripts/performance/common.py
@@ -170,7 +170,6 @@ class RunCommand:
                     stdout=PIPE if should_pipe else DEVNULL,
                     stderr=STDOUT,
                     universal_newlines=True,
-                    encoding='utf-8',
                     errors='ignore'
             ) as proc:
                 if proc.stdout is not None:
--- a/scripts/performance/common.py
+++ b/scripts/performance/common.py
@@ -170,6 +170,8 @@ class RunCommand:
                     stdout=PIPE if should_pipe else DEVNULL,
                     stderr=STDOUT,
                     universal_newlines=True,
+                    encoding='utf-8',
+                    errors='ignore'
             ) as proc:
                 if proc.stdout is not None:
                     with proc.stdout:


danmoseley:
@vatsan-madhavan are you using regular cmd.exe or the newer Windows Terminal? what does `chcp` return in this console?

danmoseley:
Also @vatsan-madhavan could you please confirm this change avoids the failure you saw?

vatsan-madhavan:
Windows Terminal + Powershell Core 7.1.3.

vatsan-madhavan:
```
chcp
Active code page: 437
```

danmoseley:
Interesting, maybe Powershell writes UTF-8 encoded anyway? My knowledge of how all this works is limited. I'm fairly confident this will fix the specific problem though.

vatsan-madhavan:
I'm kicking off another run off of your topic branch now - hoping I'll have something useful to report in the AM.

danmoseley:
CI errors suggest build machines are using an older version of Python 3.

Man pages say "New in version 3.6: encoding and errors were added."

vatsan-madhavan:
I'm on python 3.9.2. The net5.0 portion seems to have completed around 6am today and net6.0 started successfully around that time - so looks like the fix worked at least for my use-case. I'll be using the machine during the day for RDP + Teams alongside these runs - if that messes up the perf runs I'll do it again next weekend when I can leave it the box untouched for ~20 hours.

danmoseley:
@DrewScoggins do you control the Python installs on these machines? Some of them seem to have a version earlier than 3.6, which [apparently](https://www.python.org/downloads/) is the oldest version of 3.x that's in support.

eg https://dev.azure.com/dnceng/public/_build/results?buildId=1050209&view=logs&j=da90480d-12b5-5a42-6484-03e90d2bb558&t=c66bac4a-d2fb-5afc-1cc4-f8624f8f9de3
`Agent machine name: 'fv-az50-457'`

DrewScoggins:
Yes, and no. We just use a container and take what is in the container. However, we have a very old container, so I am reaching out to dnceng to figure out how to get a newer container for this job.

I am still curious why we do not see this behavior in the internal performance lab. I wonder if it has something to do with running in cmd instead of powershell.

Finally, is there a reason you chose ignore over the other options listed here, https://docs.python.org/3/library/io.html#io.TextIOWrapper? I was looking and was thinking it might be nice to have the offending character escaped to the console instead of just ignored.

danmoseley:
> However, we have a very old container, so I am reaching out to dnceng to figure out how to get a newer container for this job.

I think that is a good idea as we should not be using unsupported infrastructure.

> I am still curious why we do not see this behavior in the internal performance lab. I wonder if it has something to do with running in cmd instead of powershell.

I don't think so (although given that I don't fully understand it). If you run 'chcp' in powershell, you get 437 as well -- even in the new Windows Terminal. So either something else is involved, or the repro involved output from a tool that was ignoring that codepage setting.

> Finally, is there a reason you chose ignore over the other options listed here, https://docs.python.org/3/library/io.html#io.TextIOWrapper? I was looking and was thinking it might be nice to have the offending character escaped to the console instead of just ignored.

I am a Python beginner and I chose the smallest change that worked. Are you comfortable writing it in terms of TextIOWrapper, or can point me at a snippet and I can try that?

DrewScoggins:
Here is the PR that makes the container update, https://github.com/dotnet/performance/pull/1744.

I have tried to reproduce this locally on my machine and have not been able to do it. When I call the code as is I get the following output

```
[2021/03/22 13:47:25][INFO] |               Method |                                input |   value |             comparisonType |      Mean | Error |    Median |       Min |       Max | Gen 0 | Gen 1 | Gen 2 | Allocated |
[2021/03/22 13:47:25][INFO] |--------------------- |------------------------------------- |-------- |--------------------------- |----------:|------:|----------:|----------:|----------:|------:|------:|------:|----------:|
[2021/03/22 13:47:25][INFO] |         StringAsSpan |                                    ? |       ? |                          ? | 782.20 us |    NA | 782.20 us | 782.20 us | 782.20 us |     - |     - |     - |      48 B |
[2021/03/22 13:47:25][INFO] | GetPinnableReference |                                    ? |       ? |                          ? |  31.76 us |    NA |  31.76 us |  31.76 us |  31.76 us |     - |     - |     - |       2 B |
[2021/03/22 13:47:25][INFO] |                 Trim |                                      |       ? |                          ? | 786.00 us |    NA | 786.00 us | 786.00 us | 786.00 us |     - |     - |     - |      48 B |
[2021/03/22 13:47:25][INFO] |                 Trim |                             abcdefg  |       ? |                          ? | 753.30 us |    NA | 753.30 us | 753.30 us | 753.30 us |     - |     - |     - |      48 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |                           AAAAA5AAAA |       5 |           InvariantCulture |  62.86 us |    NA |  62.86 us |  62.86 us |  62.86 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString | AAAAAAAAAAAA(...)AAAAAAAAAAAA [1000] |       X |                    Ordinal |  39.52 us |    NA |  39.52 us |  39.52 us |  39.52 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |  AAAAAAAAAAAA(...)AAAAAAAAAAAA [100] |       x | InvariantCultureIgnoreCase |  54.10 us |    NA |  54.10 us |  54.10 us |  54.10 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |  AAAAAAAAAAAA(...)AAAAAAAAAAAA [100] |       x |          OrdinalIgnoreCase |  52.33 us |    NA |  52.33 us |  52.33 us |  52.33 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |                                ABCDE |       c | InvariantCultureIgnoreCase |  62.98 us |    NA |  62.98 us |  62.98 us |  62.98 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |  Hello Worldb(...)allylong!xyz [186] |       w |          OrdinalIgnoreCase |  64.99 us |    NA |  64.99 us |  64.99 us |  64.99 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |  Hello Worldb(...)allylong!xyz [187] |       ~ |                    Ordinal |  56.01 us |    NA |  56.01 us |  56.01 us |  56.01 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString | Hello Worldbb(...)bbbbbbbbbbba! [47] |       y |                    Ordinal |  48.44 us |    NA |  48.44 us |  48.44 us |  48.44 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |                          More Test's |   Tests |          OrdinalIgnoreCase |  76.19 us |    NA |  76.19 us |  76.19 us |  76.19 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |                               StrIng |  string |          OrdinalIgnoreCase |  73.43 us |    NA |  73.43 us |  73.43 us |  73.43 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |                 Trim |                              abcdefg |       ? |                          ? | 530.40 us |    NA | 530.40 us | 530.40 us | 530.40 us |     - |     - |     - |      48 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |                         foobardzsdzs |   rddzs |           InvariantCulture |  70.89 us |    NA |  70.89 us |  70.89 us |  70.89 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |                              string1 | string2 |           InvariantCulture |  61.26 us |    NA |  61.26 us |  61.26 us |  61.26 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |                                    ? |       ? |           InvariantCulture | 399.68 us |    NA | 399.68 us | 399.68 us | 399.68 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString |  ????????????(...)???????????? [100] |       ? |                    Ordinal |  48.84 us |    NA |  48.84 us |  48.84 us |  48.84 us |     - |     - |     - |       3 B |
[2021/03/22 13:47:25][INFO] |        IndexOfString | ????????????(...)???????????? [1000] |       x |                    Ordinal |  56.86 us |    NA |  56.86 us |  56.86 us |  56.86 us |     - |     - |     - |       3 B |
```

This is with Powershell 7.1 and Windows Terminal, as well as just straight cmd.

This is the repro command I used `py -3 .\scripts\benchmarks_ci.py -f net6.0 --filter *ReadOnlySpan* --bdn-arguments="--iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart --stopOnFirstError true"`

danmoseley:
@DrewScoggins did you try the steps in the topmost comment? If you're trying with the regular codepage, I agree, I couldn't repro it either -- but does it matter? The code should not fail no matter what bytes come out of stdout. Either by suppressing errors, try/catch, or whatever.

billwert:
Wasn't @vatsan-madhavan also using the default code page? You're right that this fix is probably reasonable, but I am kind of interested in knowing what the actual problem is..

billwert:
Particularly since this is the only report we've got of this issue after running on dozens and dozens of machines.

DrewScoggins:
When I tried your repro steps on my computer I could not repro the issue.  Instead I get this output `[2021/03/22 14:23:55][INFO] F:\git\performance\src\benchmarks\micro\MicroBenchmarks.csproj(3,1): warning : ?`, and then it just runs normally after that.

danmoseley:
And if you type `chcp` it confirms it's 65001? Odd.

DrewScoggins:
image

vatsan-madhavan:
Forgot to mention something that might be important (or not 😉) : I launch my pwsh under [gsudo](https://github.com/gerardog/gsudo). 

```
Windows Terminal (unelevated)
|
-- gsudo
     |
      -- pwsh (elevated)
```

vatsan-madhavan:
image

danmoseley:
I wonder whether @GrabYourPitchforks has thoughts to share.

From the console output in the zip, it may have been a line containing these characters that it failed on.
https://github.com/danmoseley/performance/blob/be991a99b1b6ee379638fb4b366fd68773bf734f/src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs#L71

billwert:
Given that this isn't 100% blocking (except for @vatsan-madhavan maybe?) I'd like to not merge this until we have a strong understanding of what's going on for real.

@vatsan-madhavan  can you try without gsudo? Their readme.md says this:
```
Supports being used on scripts:
Outputs of the elevated commands can be interpreted: E.g. StdOut/StdErr can be piped or captured (e.g. gsudo dir | findstr /c:"bytes free" > FreeSpace.txt) and exit codes too (%errorlevel%). If gsudo fails to elevate, the exit code will be 999.
```

Given that it claims to already be doing things with the streams it would not surprise me if it is somehow implicated.

vatsan-madhavan:
Not blocking for me anymore - I was able to complete the perf runs and share logs offline via email using @danmoseley's topic branch to run the experiment.

vatsan-madhavan:
>can you try without gsudo

i'll try.

GrabYourPitchforks:
[This comment](https://github.com/dotnet/performance/pull/1743#issuecomment-803775345) mentioned code page 437. In that code page, the byte `0x81` corresponds to the code point __ü__ (_U+00FC LATIN SMALL LETTER U WITH DIAERESIS_). But the byte 0x81 doesn't have any particular meaning in CP1252 (the normal U.S. English system code page, used by most apps by default). And I can't find any interesting terminal escape sequence `\x81` that would explain it. So... unless __ü__ (or `\u00fc`) appears in a unit test somewhere, and unless the code page got horked while running the test, I'm stumped.

vatsan-madhavan:
By the time the perf scripts are launched, isn't the launching process pwsh (and gsudo merely an ancestor process) ? Should we really expect gsudo to have as much (or any) impact here on the python process?

billwert:
> and gsudo merely an ancestor process

I've read less than @danmoseley has about consoles, but I know that stdin/stdout/stderr are often in some way inherited from the child processes, and the examples on the `gsudo` page suggest that ultimately `gsudo` is doing something to be able to capture the output. I bet there's something to it.

@GrabYourPitchforks I think we all latched onto 437 since that's what `chcp` says in a new `cmd.exe` or `pwsh` window.

vatsan-madhavan:
I ran it w/o gsudo from the tip of main overnight and the whole thing worked cleanly. So gsudo's stdout/stderr redirection seems to be the cause of the original hiccup I encountered!

billwert:
Thanks for doing the experiment @vatsan-madhavan!

In light of `gsudo` being the culprit what shall we do here? I'm inclined to just close this PR as in general I am in favor of not ignoring such errors, but can be swayed.

danmoseley:
I have no opinion. Happy to close and we can revisit if needed.

PR #1746, opened 2021-03-23
Add MSBuild timing Dev Inner Loop Tests

DrewScoggins:


DrewScoggins:
PTAL @arkalyanms @ladipro

src/scenarios/emptyconsoletemplateinnerloopmsbuild/src/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
billwert, line 1: jamming this comment here for lack of a  better place - I don't think this should be `emptyconsoletempalteinnerloopmsbuild` if it is static code. I think generally we use `empty` to mean "having run `dotnet new`". Contrast `staticconsoletemplate` for example. Maybe this convention is already broken and I hadn't noticed before.

billwert (approved):
Going to approve this but please look at the one comment.

src/scenarios/emptyconsoletemplateinnerloopmsbuild/src/Program.cs
DrewScoggins: This does run dotnet new, yeah? The last five lines of the pre.py makes it seem like that is what we are doing.

billwert: Oh this is the 'changed' source, I understand now. Forgot what context we were in.

PR #1759, opened 2021-04-05
Add dotnet watch with hot reload scenario

DrewScoggins:
This change adds dotnet watch with hot reload as a parsable scenario for startup. I
have added new capabilties to startup in order to allow this. I have refactored the
call to `RunIteration` to allow for all of the steps that can happen in any startup
test that we have. I have also added new steps for the dotnet watch, console
listeners, that can be used to understand when a process had made certain progress.
I have also added the ability to run a process such that we return from our call,
and the underlying process can still be executing.

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -349,7 +349,7 @@ class CSharpProject:
 
                 if args:
                     cmdline = cmdline + list(args)
-                
+                print(cmdline)
                 RunCommand(cmdline, verbose=verbose).run(
                     self.working_directory)
     @staticmethod
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -349,7 +349,7 @@ class CSharpProject:
 
                 if args:
                     cmdline = cmdline + list(args)
-                print(cmdline)
+                
                 RunCommand(cmdline, verbose=verbose).run(
                     self.working_directory)
     @staticmethod
--- a/src/scenarios/blazorwasmdotnetwatch/pre.py
+++ b/src/scenarios/blazorwasmdotnetwatch/pre.py
@@ -1,0 +1,20 @@
+'''
+pre-command
+'''
+import sys
+import shutil
+import os.path
+from performance.logger import setup_loggers
+from shared import const
+from shared.precommands import PreCommands
+from test import EXENAME
+
+setup_loggers(True)
+precommands = PreCommands()
+precommands.new(template='blazorwasm',
+                output_dir=const.APPDIR,
+                bin_dir=const.BINDIR,
+                exename=EXENAME,
+                working_directory=sys.path[0])
+precommands.execute()
+#shutil.copy(os.path.join('src', 'Program.cs'), os.path.join(const.APPDIR, 'Program.cs'))
--- a/src/scenarios/blazorwasmdotnetwatch/pre.py
+++ b/src/scenarios/blazorwasmdotnetwatch/pre.py
@@ -17,4 +17,3 @@ precommands.new(template='blazorwasm',
                 exename=EXENAME,
                 working_directory=sys.path[0])
 precommands.execute()
-#shutil.copy(os.path.join('src', 'Program.cs'), os.path.join(const.APPDIR, 'Program.cs'))
--- a/src/tools/ScenarioMeasurement/Startup/Startup.cs
+++ b/src/tools/ScenarioMeasurement/Startup/Startup.cs
@@ -155,8 +155,17 @@ namespace ScenarioMeasurement
                         Console.WriteLine(e.Data);
                     }
                 };
+                DataReceivedEventHandler stdErrProcessor = (s, e) =>
+                {
+                    if (!String.IsNullOrEmpty(e.Data))
+                    {
+                        Console.WriteLine(e.Data);
+                    }
+                };
                 Proc.OutputDataReceived += stdOutProcessor;
+                Proc.ErrorDataReceived += stdErrProcessor;
                 Proc.BeginOutputReadLine();
+                Proc.BeginErrorReadLine();
                 bool isSteadyState = false;
                 int timeoutCount = 0;
                 while (!isSteadyState && timeoutCount < timeout)
@@ -172,7 +181,9 @@ namespace ScenarioMeasurement
                     timeoutCount++;
                     Thread.Sleep(1000);
                 }
+                Proc.CancelErrorRead();
                 Proc.CancelOutputRead();
+                Proc.ErrorDataReceived -= stdErrProcessor;
                 Proc.OutputDataReceived -= stdOutProcessor;
                 return true && timeoutCount < timeout;
             }
--- a/src/tools/ScenarioMeasurement/Startup/Startup.cs
+++ b/src/tools/ScenarioMeasurement/Startup/Startup.cs
@@ -1,8 +1,11 @@
 ﻿using Reporting;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Tracing;
 using System.IO;
+using System.Text;
+using System.Threading;
 
 namespace ScenarioMeasurement
 {
@@ -14,7 +17,8 @@ namespace ScenarioMeasurement
         WPF,
         Crossgen2,
         InnerLoop,
-        InnerLoopMsBuild
+        InnerLoopMsBuild,
+        DotnetWatch
     }
 
     public class InnerLoopMarkerEventSource : EventSource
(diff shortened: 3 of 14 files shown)


src/scenarios/blazorwasmdotnetwatch/pre.py
```diff
@@ -0,0 +1,20 @@
...
+
+setup_loggers(True)
+precommands = PreCommands()
+precommands.new(template='blazorwasm',
+                output_dir=const.APPDIR,
+                bin_dir=const.BINDIR,
+                exename=EXENAME,
+                working_directory=sys.path[0])
+precommands.execute()
+#shutil.copy(os.path.join('src', 'Program.cs'), os.path.join(const.APPDIR, 'Program.cs'))
```
billwert, line 20: should this be commented?

scripts/dotnet.py
```diff
@@ -349,7 +349,7 @@ def build(self,
 
                 if args:
                     cmdline = cmdline + list(args)
-                
+                print(cmdline)
```
billwert, line 352: remove?

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -114,28 +119,102 @@ static void checkArg(string arg, string name)
                 Arguments = appArgs,
                 WorkingDirectory = workingDir,
                 GuiApp = guiApp,
-                EnvironmentVariables = envVariables
+                EnvironmentVariables = envVariables,
+                IsRunWithExit = metricType == MetricType.DotnetWatch ? false : true
```
billwert, line 123: I wonder if this should not key off of `DotnetWatch` and instead just be a command line argument. You could imagine other things that might want this behavior.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -114,28 +119,102 @@ static void checkArg(string arg, string name)
...
+                        }
+                    }
+                    timeoutCount++;
+                    Thread.Sleep(1000);
+                }
+                Proc.CancelOutputRead();
+                Proc.OutputDataReceived -= stdOutProcessor;
+                return true && timeoutCount < timeout;
+            }
+            : null;
```
billwert, line 196: These look to be basically the same method, can we refactor them so they take the string as input also?

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -114,28 +119,102 @@ static void checkArg(string arg, string name)
...
+                DataReceivedEventHandler stdOutProcessor = (s, e) =>
+                {
+                    if (!String.IsNullOrEmpty(e.Data))
+                    {
+                        output.AppendLine(e.Data);
+                        Console.WriteLine(e.Data);
+                    }
+                };
+                Proc.OutputDataReceived += stdOutProcessor;
+                Proc.BeginOutputReadLine();
```
billwert, line 141: Should this also read stderr?

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
```diff
@@ -36,19 +36,89 @@ public class ProcessHelper
...
+        }
+
+        /// <summary>
+        /// Runs the specified process and does not wait for it to exit.
+        /// </summary>
+        /// <param name="appExe">Full path to executable</param>
+        /// <param name="appArgs">Optional arguments</param>
+        /// <param name="workingDirectory">Optional working directory (defaults to current directory)</param>
+        /// <returns></returns>
+        public Process RunWithNoExit()
```
billwert, line 70: I understand you moved the handling of stderr/stdout out of this method so you can search it instead, but it feels like a weird break of the contract for `ProcessHelper` to not handle those in the case you want to let the process run. I'd like to see there be either a default handling here that can be overridden, or a rework of the handling generally in this type that lets you insert text processing into the chain.

billwert (changes requested):

src/scenarios/blazorwasmdotnetwatch/pre.py
DrewScoggins: That needs to be taken out. I will just remove it.

scripts/dotnet.py
DrewScoggins: Yes

src/tools/ScenarioMeasurement/Startup/Startup.cs
DrewScoggins: That is probably the right approach here. We can just include it as a test trait that is set, and also add it as a command line argument to startup.

src/tools/ScenarioMeasurement/Startup/Startup.cs
DrewScoggins: I was already thinking about doing this after I realized how close they are. Given that other tests would almost certainly want to do this same thing, we can start by this taking a string. If we want to eventually do more complex checking we can look into doing this as some sort of verification function that we pass in.

src/tools/ScenarioMeasurement/Startup/Startup.cs
DrewScoggins: I am not sure on this. Maybe we could use output on stderr to indicate that we have reached a failure point, but I am not sure that that would always hold. I am having a hard time imagining any scenario where we would want to inspect and know the contents of stderr as part of a correctly functioning test, but maybe there is one? Maybe it is just good to have from a logging perspective.

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
DrewScoggins: So I can imagine how we could do it either way, but I have a question. The way I had imagined this method being used, basically required the use of some kind of text processing on the process to understand the current state. In all other places we are either using a GUI app, and allow for a timer to elapse, or some process that we expect to shutdown of it's own volition. But here I would posit that if you are just relying on the default output processing that you will be holding it wrong, and almost certainly not doing what you want to do. Mainly because `RunWithoutExit` gives up on handling control flow for you, where `RunWithExit` does. Maybe the right thing do here is to change the ProcessHelper type to an interface, and have a "managed" and "unmanaged" process execution. That way you at least know what you are signing up for when you use the `RunWithoutExit` mode.

src/tools/ScenarioMeasurement/Startup/Startup.cs
DrewScoggins: I just have it read it and write it out to the console for now.

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
```diff
@@ -6,7 +6,27 @@
...
 {
-    public class ProcessHelper
+    public enum Result
+        {
+            Success,
+            TimeoutExceeded,
+            ExitedEarly,
+            CloseFailed,
+            ExitedWithError
+        }
```
billwert, line 16:
```suggestion
    public enum Result
    {
        Success,
        TimeoutExceeded,
        ExitedEarly,
        CloseFailed,
        ExitedWithError
    }
```

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -105,37 +112,105 @@ static void checkArg(string arg, string name)
...
+                }
+                Proc.CancelErrorRead();
+                Proc.CancelOutputRead();
+                Proc.ErrorDataReceived -= stdErrProcessor;
+                Proc.OutputDataReceived -= stdOutProcessor;
+                return true && timeoutCount < timeout;
+            }
+            : null;
+
+            Func<Process, string, bool> waitForRecompile = waitForSteadyState;
```
billwert, line 192: nit: you don't really need a second name here right?

billwert (approved):
LGTM!

src/tools/ScenarioMeasurement/Startup/Startup.cs
DrewScoggins: I think I want to keep it because I could imagine wanting to do different things between these two steps. Right now we don't but I would like to keep that option.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -105,37 +112,105 @@ static void checkArg(string arg, string name)
...
+                        }
+                    }
+                    timeoutCount++;
+                    Thread.Sleep(1000);
+                }
+                Proc.CancelErrorRead();
+                Proc.CancelOutputRead();
+                Proc.ErrorDataReceived -= stdErrProcessor;
+                Proc.OutputDataReceived -= stdOutProcessor;
+                return true && timeoutCount < timeout;
```
danmoseley, line 188: nit `true &&` redundant

PR #1761, opened 2021-04-06
Fix Failure Reporting

Tohron:
This is a PR for resolving issues with the job-running code.  It currently fixes an issue by setting verbose=True rather than setting it to a nonexistent variable.


DrewScoggins:
This is for fixing the Failure reporting stuff, correct?

DrewScoggins:
LGTM

PR #1769, opened 2021-04-13
Add Blazor Pizza scenario for SOD

DrewScoggins:


billwert (approved):
Looks good. One question: should we use `PERFLAB_TARGET_FRAMEWORK` in the app source so we can roll it forward later, or do we expect we'd have to pick up a whole new version anyway?

PR #1771, opened 2021-04-13
Failure reporting doesn't run outside of lab

Tohron:
This PR is to fix an issue where failure reporting tries to run outside of the lab and causes unintended startup test failure.


scripts/benchmarks_ci.py
```diff
@@ -31,6 +31,7 @@
 from performance.logger import setup_loggers
 from performance.constants import UPLOAD_CONTAINER, UPLOAD_STORAGE_URI, UPLOAD_TOKEN_VAR, UPLOAD_QUEUE
 from channel_map import ChannelMap
+from shared.util import runninginlab
```
billwert, line 34: This won't work as `shared` won't be in the scope for this script. You either need to move this to `common.py` (and fixup uses of it in other places) or something similar.

billwert (changes requested):

billwert (approved):
LGTM

PR #1772, opened 2021-04-14
Add CI (cold) performance runs

DrewScoggins:
This PR adds first run performance tests for dotnet.


billwert (dismissed):

danmoseley:
Looks like the perf tests took only 5 mins each but then hung for over 4 hours and were killed.

danmoseley:
I'm going to be bold here and close this.

PR #1779, opened 2021-04-19
Add NoneType check when obtaining helixpayload

L2:
* Currently when the return value of helixpayload() is None,
  it's still passed along to os.path.join which raises a TypeError.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -131,7 +131,7 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         required=False,
         help="Causes results files to be uploaded to perf container",
         action='store_true'
-    )   
+    )
 
     # Generic arguments.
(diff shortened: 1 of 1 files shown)


billwert:
Hi @L2! This code path should only execute inside our perf lab (where `helixpayload()` should return a value.) Can you help us understand how you hit this?

L2:
Looks like it's been fixed with 0c988339004e2

I was investigating a crash result I had prior to this so I thought this issue was still around. Although if inside the lab you may still encounter the os.path.join call expecting something other than NoneType if that env var has not been set (which this PR handles). If you don't expect this case to happen, please feel free to close this PR.

Sorry for the trouble, thanks.

billwert:
Hey, no trouble at all. I wanted to make sure we weren't missing something. You're correct about the fact that there are many things in the lab that we guard with this one variable.

PR #1787, opened 2021-04-26
Remove custom OS attribute

### LINKED ISSUE #1737: Change from our custom OS exclusion type to the built in one in BDN
the type name conflict is caused by the fact that BDN has recently introduced an attribute that allows for exactly the same thing `AllowedOperatingSystems` does.

If you want you can remove [our implementation](https://github.com/dotnet/performance/blob/master/src/harness/BenchmarkDotNet.Extensions/OperatingSystems.cs) and use the new attribute provided by BDN:

https://github.com/dotnet/BenchmarkDotNet/blob/314a27fc26cbbc9aa15d7250ad12578aab632641/src/BenchmarkDotNet/Attributes/Filters/OperatingSystemsFilterAttribute.cs#L21

adamsitnik:
fixes #1737


src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -94,15 +92,12 @@
     <Compile Remove="runtime\Math\Functions\Single\ILogB.cs" />
     <Compile Remove="runtime\Math\Functions\Single\Log2.cs" />
     <Compile Remove="runtime\Math\Functions\Single\ScaleB.cs" />
+    <Compile Remove="runtime\PacketTracer\**\*.cs" />
```
DrewScoggins, line 95: Has this functionality been ported over from ARM?

DrewScoggins:
LGTM

DrewScoggins (approved):

src/benchmarks/micro/MicroBenchmarks.csproj
adamsitnik: No, but with https://github.com/dotnet/performance/blob/58e6c75de7a4ceb2f84af30d17d839ec7ffbd300/src/benchmarks/micro/runtime/PacketTracer/Render.cs#L15 we are prohibiting it to be executed on ARM, as only x86 and x64 are allowed

PR #1802, opened 2021-05-10
Resolve Ubuntu Scenario Test Failures

Tohron:
This PR is to resolve scenario test failures in pipeline runs such as
https://dev.azure.com/dnceng/internal/_build/results?buildId=1101991&view=logs&j=9d3763e3-464f-5ce4-a590-b3a4c86731a8&t=0357c9b3-0714-5dd0-78d8-213761c0279d
The New Console Template, New VB Console Template, and Static Console Template scenario tests were failing on machines where the lttng kernel modules had not been installed because the check for whether lttng was installed was returning false positives, causing the test to proceed when the necessary modules were not present..  The  modified check has been confirmed to return false if lttng is not installed, and true once it has been installed.

### DIFF
--- a/src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
+++ b/src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
@@ -52,7 +52,7 @@ namespace ScenarioMeasurement
                 RootAccess = true
             };
 
-            if (Install(logger) != Result.Success)
+            if (Install() != Result.Success)
             {
                 throw new Exception("Lttng installation failed. Please try manual install.");
             }
@@ -105,9 +105,9 @@ namespace ScenarioMeasurement
             return result;
         }
 
-        public Result Install(Logger logger)
+        public Result Install()
         {
-            if (LttngInstalled(logger))
+            if (LttngInstalled())
             {
                 Console.WriteLine("Lttng is already installed.");
                 return Result.Success;
@@ -115,10 +115,10 @@ namespace ScenarioMeasurement
             perfCollectProcess.Arguments = "install -force";
             perfCollectProcess.Run();
 
-            int retry = 1;
+            int retry = 10;
             for(int i=0; i<retry; i++)
             {
-                if (!LttngInstalled(logger))
+                if (!LttngInstalled())
                 {
                     Console.WriteLine($"Lttng not installed. Retry {i}...");
                     perfCollectProcess.Run();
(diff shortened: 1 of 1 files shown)


src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
```diff
@@ -147,7 +148,14 @@ public void AddKernelKeyword(KernelKeyword keyword)
...
         {
-            return File.Exists("//usr/bin/lttng");
+            ProcessStartInfo procStartInfo = new ProcessStartInfo("modinfo", "lttng_probe_writeback");
+            Process proc = new Process() { StartInfo = procStartInfo, };
+            proc.StartInfo.RedirectStandardOutput = true;
+            proc.Start();
+            proc.WaitForExit();
+            string result = proc.StandardOutput.ReadToEnd();
+            // If the lttng_probe_writeback module is installed, the modinfo output will include the filename field
+            return File.Exists("//usr/bin/lttng") && result != null && result.Contains("filename:");
```
billwert, line 158:
```suggestion
            return result.Contains("filename:");
```

As we know that the existence of `//usr/bin/lttng` is poor indicator, we don't need to bother checking for it. Additionally, [`StreamReader.ReadToEnd()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.readtoend?view=net-5.0#System_IO_StreamReader_ReadToEnd) returns an empty string in the case that there was nothing there, not `null`. Thus the null check is unnecessary.

src/tools/ScenarioMeasurement/Startup/PerfCollect.cs
```diff
@@ -147,7 +148,14 @@ public void AddKernelKeyword(KernelKeyword keyword)
 
         private bool LttngInstalled()
         {
-            return File.Exists("//usr/bin/lttng");
+            ProcessStartInfo procStartInfo = new ProcessStartInfo("modinfo", "lttng_probe_writeback");
+            Process proc = new Process() { StartInfo = procStartInfo, };
+            proc.StartInfo.RedirectStandardOutput = true;
+            proc.Start();
+            proc.WaitForExit();
+            string result = proc.StandardOutput.ReadToEnd();
```
billwert, line 156: This pattern is prone to deadlocking if the process writes enough output to the buffer. Consider using a pattern as we do in `ProcessHelper`:
https://github.com/dotnet/performance/blob/90b537906d686846ec7c8293cefd268e0143452c/src/tools/ScenarioMeasurement/Util/ProcessHelper.cs#L222-L243

if you're confident the output of `modinfo` will not get very long this is fine. It's something to be aware of.

Tohron: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=net-5.0
seems to indicate that moving string result = proc.StandardOutput.ReadToEnd(); before proc.WaitForExit(); is sufficient to avoid the deadlock problem.  I've tested it and it seems to run fine both for code tests and in startup tests.

Tohron: Since the volume of output returned by modinfo is pretty consistent, I don't think any further measures are necessary.

billwert (approved):

PR #1808, opened 2021-05-13
Make path prepend instead of append and print dotnet location

DrewScoggins:


scripts/ci_setup.py
```diff
@@ -209,7 +209,8 @@ def __main(args: list) -> int:
     repo_url = None if args.repository is None else args.repository.replace('-','/')
 
     variable_format = 'set %s=%s\n' if sys.platform == 'win32' else 'export %s=%s\n'
-    path_variable = 'set PATH=%%PATH%%;%s\n' if sys.platform == 'win32' else 'export PATH=$PATH:%s\n'
+    path_variable = 'set PATH=%s;%%PATH%%\n' if sys.platform == 'win32' else 'export PATH=%s:$PATH\n'
+    which = 'where dotnet\n' if sys.platform == 'win32' else 'which dotnet\n'
```
billwert, line 213: We run in powershell right? This needs to be `where.exe`, not `where`.
```suggestion
    which = 'where.exe dotnet\n' if sys.platform == 'win32' else 'which dotnet\n'
```

billwert (approved):

PR #1809, opened 2021-05-13
Add no pgo config to setup_ci

DrewScoggins:


scripts/ci_setup.py
```diff
@@ -242,6 +272,7 @@ def __main(args: list) -> int:
 
         with open(args.output_file, 'w') as out_file:
             out_file.write(which)
+            out_file.write(pgo_config)
```
billwert, line 275: this needs to use the variable format, right?

DrewScoggins: Yes, good catch.

PR #1818, opened 2021-05-24
Make BTL traces compatible with GC Infra's CPU Samples Analysis

ivdiazsa:
When working with the CPU Samples Analysis feature of GC Infra, the API used to open trace files only supported _ETL_ and _ETLX_ traces. However, _BTL_ traces are also a common occurrence, so this PR changes to another API that supports all trace file formats.


ivdiazsa:
@dotnet/gc

Maoni0 (approved):

PR #1847, opened 2021-06-11
Add Information To JSON Upload For Use In Improved Alerting

Tohron:
This PR adds WorkItemName to the Run section of the JSON so that the logger in the TableUploader will have enough information to generate a link to the console logs for a run.  It also adds MachineName to the OS section for general use.

Used by https://nettel.visualstudio.com/perflab/_git/perflab/pullrequest/284


src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -72,6 +73,7 @@ private void Init()
             os = new Os()
             {
                 Name = $"{RuntimeEnvironment.OperatingSystem} {RuntimeEnvironment.OperatingSystemVersion}",
+                MachineName = environment.GetEnvironmentVariable("COMPUTERNAME"),
```
billwert, line 76: This won't work on non-Windows OSes. Just use `Environment.MachineName`.

billwert (approved):

PR #1872, opened 2021-06-28
Add NoneType check before iterating args.bdn_arguments

L2:
- Fixes a crash with running benchmarks_ci.py due to args.bdn_arguments being iterated over even when it's of NoneType value (the bdn_arguments flag was not given, so the variable is left uninitialized).


billwert:
Apologies. Thanks for the fix! Appreciate you helping out with the codepaths that don't run in our automation. :)

PR #1877, opened 2021-07-01
Fix SSL issues when downloading dotnet-install.sh

premun:


billwert:
@premun where is this failure showing up? I don't think we've seen it in any of our CI runs..

premun:
@billwert I saw it locally when running this while downloading .NET this way works on the machine for other repos

billwert:
cool thanks!

PR #1880, opened 2021-07-02
Rethrow the caught exception so that it is bobbled up correctly

Lxiamail:
This is for fixing a regression introduced by PR https://github.com/dotnet/performance/pull/1703.  CalledProcessError exception was caught to do failure report handling, however, it was not rethrew. Because of that, the job reports pass even when  micro_benchmarks.run() failed with exception.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -260,6 +260,8 @@ def __main(args: list) -> int:
                     RunCommand(cmdline, verbose=True).run(reporterpath)
             else:
                 args.upload_to_perflab_container = False
+            # rethrow the caught CalledProcessError exception so that the exception being bobbled up correctly.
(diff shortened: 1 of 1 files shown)


scripts/benchmarks_ci.py
```diff
@@ -260,6 +260,8 @@ def __main(args: list) -> int:
                     RunCommand(cmdline, verbose=True).run(reporterpath)
             else:
                 args.upload_to_perflab_container = False
+            # rethrow the caught CalledProcessError exception so that the exception being bobbled up correctly.
```
billwert, line 263:
```suggestion
            # rethrow the caught CalledProcessError exception so that the exception being bubbled up correctly.
```

billwert (approved):

PR #1884, opened 2021-07-09
Add workload uninstall

DrewScoggins:


Lxiamail (changes requested):
I think that microsoft-net-sdk-blazorwebassembly-aot workload had been renamed into "wasm-tools"

src/scenarios/blazor/pre.py
```diff
@@ -1,14 +1,15 @@
...
-import sys, os
+import sys, os, subprocess
 from performance.logger import setup_loggers
 from shared import const
 from shared.precommands import PreCommands
 from test import EXENAME
 
 # For blazor3.2 the linker argument '--dump-dependencies' should be added statically to enable linker dump
 # For blazor5.0 the linker argument can be added to the command line as an msbuild property
+subprocess.run(["dotnet", "workload", "uninstall", "microsoft-net-sdk-blazorwebassembly-aot"])
```
Lxiamail, line 12: Why do we need to uninstall the workload before installing wasm-tools workload later?

Lxiamail (commented):
I think that microsoft-net-sdk-blazorwebassembly-aot workload had been renamed into "wasm-tools"

src/scenarios/blazor/pre.py
DrewScoggins: For the non-AOT runs I want this toolset removed because it seems to be causing weird behavior, where the runs will try and AOT the builds even though that is not specified in the csproj. It also just seems like good hygiene to make sure that we have these thing uninstalled after we are done using them.

DrewScoggins: And this way we can make sure that all the machines have them cleaned off before we ever do any runs. Once we have the machines back in a good state we can take these out and only run the post step.

Lxiamail: Sounds good to me

Lxiamail (approved):

src/scenarios/blazor/pre.py
```diff
@@ -9,7 +9,9 @@
 
 # For blazor3.2 the linker argument '--dump-dependencies' should be added statically to enable linker dump
 # For blazor5.0 the linker argument can be added to the command line as an msbuild property
+
 subprocess.run(["dotnet", "workload", "uninstall", "microsoft-net-sdk-blazorwebassembly-aot"])
```
Lxiamail, line 13: Should "microsoft-net-sdk-blazorwebassembly-aot" uninstall be removed.

DrewScoggins: I want to keep it around to make sure that package gets removed from all of our machines. We can remove it once it has run everywhere.

PR #1890, opened 2021-07-16
Add Pipeline For Arm64 Runs in 3.1

Tohron:
This PR adds a pipeline to run microbenchmark tests for our release/3.1 branch on Arm64 machines.


Lxiamail (commented):
@Tohron , was we discussed with @DrewScoggins, since we only run tests two times a day, ARM64 job won't slow down the run, so it's not naccessory to create a separated pipeline for ARM64 only.

azure-pipelines-slow.yml
```diff
@@ -0,0 +1,67 @@
...
+        - main
+        - release/3.1.3xx
+
+  # Windows Arm x64 micro benchmarks
+  - template: /eng/performance/benchmark_jobs.yml
+    parameters:
+      osName: windows
+      osVersion: 10.19041
+      kind: micro
+      architecture: x64
```
Lxiamail, line 58: Should it be arm64?

Tohron:
Alright, before I make that change, does anyone spot any problems with the formulations for the two jobs?  Am I missing any jobs?  @DrewScoggins ?

Lxiamail (changes requested):
@DrewScoggins can you help CR the PR for double checking.

azure-pipelines-slow.yml
```diff
@@ -0,0 +1,67 @@
...
+      kind: micro
+      architecture: arm64
+      pool: Hosted Ubuntu 1604
+      machinePool: Open
+      queue: Ubuntu.1804.Arm64.perf
+      csproj: src\benchmarks\micro\MicroBenchmarks.csproj
+      runCategories: 'runtime libraries' 
+      channels:
+        - main
+        - release/3.1.3xx
```
DrewScoggins, line 50: This should be release/3.1.4xx

azure-pipelines-slow.yml
```diff
@@ -0,0 +1,67 @@
...
+      kind: micro
+      architecture: arm64
+      pool: Hosted VS2017
+      machinePool: Open
+      queue: Windows.10.Arm64.perf.surf
+      csproj: src\benchmarks\micro\MicroBenchmarks.csproj
+      runCategories: 'runtime libraries' 
+      channels:
+        - main
+        - release/3.1.3xx
```
DrewScoggins, line 66: Same as above

DrewScoggins:
I believe that you will also need to make sure that this change makes its way into the release/3.1.4xx branch, or it will not work.

azure-pipelines-slow.yml
```diff
@@ -0,0 +1,67 @@
...
+    - src/benchmarks/real-world/Microsoft.ML.Benchmarks/Input/README.md
+    - src/tools/ResultsComparer/README.md
+    - README.md
+
+schedules:
+- cron: "0 */12 * * *"
+  displayName: Every 12 hours build
+  branches:
+    include:
+    - main
```
DrewScoggins, line 32: We should remove this from triggering on main. We have coverage of Arm64 on the runtime side, and given how little Arm64 capacity we have we should only run what we have to here, the 3.1 runs.

PR #1897, opened 2021-07-27
Revise the PipeStream perf tests

stephentoub:
There's no need to run the same tests with different pipe directions.

We were missing read/write async tests.

And we were only running tests for named pipe with FileOptions.Asynchronous; we should also have None.


adamsitnik (approved):
LGTM, thank you @stephentoub !

PR #1898, opened 2021-07-28
Remove scenarios broken by top level C#

billwert:
Recent changes in the SDK templates mean that building them with previous TFMs fails. This scenario works if the older SDKs are present, but that's not the situation in our test runs.

This also contains a fix to startup.py to re-throw the previously caught exception so that CI status is correct.

### DIFF
--- a/eng/performance/sdk_scenarios.proj
+++ b/eng/performance/sdk_scenarios.proj
@@ -39,7 +39,7 @@
   </ItemDefinitionGroup>
   
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


billwert:
The remaining failed checks are other issues that have been masked for a while. This PR is doing the thing I intend to do - we need to clean those other things up also.

DrewScoggins:
LGTM

DrewScoggins (approved):

PR #1901, opened 2021-07-29
Add property with path to wasm runtime config

DrewScoggins:


billwert (approved):

DrewScoggins:
Had to update this PR to include actually parsing the the values passed to --msbuild, as before we said you could provide a semicolon deliminated list, but we did not parse the input correctly to actually handle that.

src/scenarios/shared/precommands.py
```diff
@@ -71,6 +71,7 @@ def __init__(self):
         self.framework = args.framework
         self.runtime_identifier = args.runtime
         self.msbuild = args.msbuild
+        print(self.msbuild)
```
billwert, line 74: Should this be deleted?

billwert (approved):
looks fine with one question

PR #1915, opened 2021-08-10
Fix TimeToFirstService tests to dispose the ServiceProvider correctly

eerhardt:
Contributes to https://github.com/dotnet/runtime/issues/57146

cc @DrewScoggins @maryamariyan @davidfowl 

(I don't have write permissions, so I can't request reviews)

### DIFF
--- a/src/benchmarks/micro/libraries/Microsoft.Extensions.DependencyInjection/TimeToFirstServiceBenchmark.cs
+++ b/src/benchmarks/micro/libraries/Microsoft.Extensions.DependencyInjection/TimeToFirstServiceBenchmark.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Extensions.DependencyInjection
             temp.Foo();
         }
 
-        [GlobalCleanup(Targets = new[] { nameof(BuildProvider), nameof(Transient)})]
+        [IterationCleanup(Targets = new[] { nameof(BuildProvider), nameof(Transient)})]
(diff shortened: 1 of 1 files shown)


DrewScoggins:
LGTM

adamsitnik (changes requested):
Changing `[GlobalCleanup]` to `[IterationCleanup]` means "run this cleanup after each benchmark run", the problem is that it also means "run the benchmark once per iteration" and for short-running benchmarks it usually affects the accuracy in a bad way (stopowatch resolution).

Please move the disposal of the scope to the `[Benchmark]` methods and remove the cleanups.

eerhardt:
> Please move the disposal of the scope to the [Benchmark] methods and remove the cleanups.

The intention of the benchmark, AFAICT, is that it wants to measure how long it takes to "get the first service". This doesn't include disposing the ServiceProvider. So I kind of think this suggestion is contradicting the original intention of the test.

>  the problem is that it also means "run the benchmark once per iteration" and for short-running benchmarks it usually affects the accuracy in a bad way (stopowatch resolution).

This sounds like a bug/problem with BenchmarkDotnet. Have we considered fixing the issue there? It also sounds like `IterationCleanup` should never be used. So why do we have it?

eerhardt:
Is the CI currently broken in this repo? I don't see my change causing any of the errors I'm seeing on CI.

billwert:
Yes we're working on it. :( Sorry. This looks good. I'll merge.

PR #1920, opened 2021-08-13
Add JSON benchmarks for nullable types

eiriktsarpalis:
Adding a benchmark measuring a codegen-related performance regression that we've had to deal with a few times. See also https://github.com/dotnet/runtime/pull/46788, https://github.com/dotnet/runtime/pull/57327 and https://github.com/dotnet/runtime/issues/50915

cc @steveharter @EgorBo


steveharter:
Agree; we having a single test for a nullable will help future cases.

Note that `MyEventsListerItemTask` has some `DateTimeOffet?` nullables as well although as mentioned that takes some debugging to figure out why it regressed.

adamsitnik (approved):
LGTM, thank you @eiriktsarpalis !

PR #1921, opened 2021-08-13
Fix: MS Build Console Pipeline Failure

LoopedBard3:
Updated the event being looked for to the new updated ETW name. (EvaluateStop/Stop -> Evaluate/Stop)
PR changing name: [https://github.com/dotnet/msbuild/pull/6725](https://github.com/dotnet/msbuild/pull/6725)


LoopedBard3:
@DrewScoggins Looks like this fixed the MSBuild issue. Should be good to merge.

PR #1924, opened 2021-08-16
Disable ResultTable for Scenario Testing

Tohron:
This change skips generation of the result table if the current issues with the 3.1 runtime lead to necessary events not occurring.

### DIFF
--- a/eng/performance/scenarios.proj
+++ b/eng/performance/scenarios.proj
@@ -44,14 +44,6 @@
     </Scenario>
   </ItemGroup>
 
-  <!-- Startup FDD publish -->
-  <ItemGroup>
-    <HelixWorkItem Include="@(Scenario -> 'Startup - %(Identity) - FDD Publish')">
-      <PreCommands>$(Python) pre.py publish -f $(_Framework) -c Release</PreCommands>
-      <Command>$(Python) test.py startup --scenario-name &quot;%(Identity)&quot;</Command>
-    </HelixWorkItem>
-  </ItemGroup>
-
   <!-- SOD FDD publish -->
   <ItemGroup>
     <HelixWorkItem Include="@(Scenario -> 'SOD - %(Identity) - FDD Publish')">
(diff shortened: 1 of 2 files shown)


DrewScoggins:
I think that a better approach here would be to disable the execution of the underlying scenarios in the yml file, rather than trying do this on the reporting side.

DrewScoggins (changes requested):
See my other comment

src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -119,12 +119,12 @@ public string WriteResultTable()
...
                 ret.AppendLine(test.Name);
-                ret.AppendLine($"{LeftJustify("Metric", counterWidth)}|{LeftJustify("Average",resultWidth)}|{LeftJustify("Min", resultWidth)}|{LeftJustify("Max",resultWidth)}");
+                ret.AppendLine($"{LeftJustify("Metric", counterWidth)}|{LeftJustify("Average", resultWidth)}|{LeftJustify("Min", resultWidth)}|{LeftJustify("Max", resultWidth)}");
                 ret.AppendLine($"{new String('-', counterWidth)}|{new String('-', resultWidth)}|{new String('-', resultWidth)}|{new String('-', resultWidth)}");
 
-           
+
                 ret.AppendLine(Print(defaultCounter, counterWidth, resultWidth));
-                foreach(var counter in topCounters)
+                foreach (var counter in topCounters)
```
billwert, line 127: Please undo all these changes as they are white-space only.

eng/performance/scenarios.proj
```diff
@@ -44,14 +44,6 @@
...
   </ItemGroup>
 
-  <!-- Startup FDD publish -->
-  <ItemGroup>
-    <HelixWorkItem Include="@(Scenario -> 'Startup - %(Identity) - FDD Publish')">
-      <PreCommands>$(Python) pre.py publish -f $(_Framework) -c Release</PreCommands>
-      <Command>$(Python) test.py startup --scenario-name &quot;%(Identity)&quot;</Command>
-    </HelixWorkItem>
-  </ItemGroup>
-
```
billwert, line 54: I would comment this out instead of delete it, as we will want to turn it back on at some point.

billwert (changes requested):

billwert (approved):

DrewScoggins (approved):
LGTM

PR #1930, opened 2021-08-18
Limit Drawing tests to Windows

DrewScoggins:


billwert (approved):

DrewScoggins:
Failures are unrelated.

adamsitnik:
@DrewScoggins @billwert I assume there was some breaking change introduced recently and it stopped working? Please let me know if there are any other `net7.0` "surprises" I could help you with (in theory BDN should work fine, but historically we always needed some fixes)

DrewScoggins:
Yeah, this is because libgdiplus was no longer being maintained, and since that is how System.Drawing worked on Linux, we are just giving up on this support. I am not sure of anything else, but we will keep you in the loop when we learn stuff :)

PR #1936, opened 2021-08-20
Disable AzurePipelinesReporter

billwert:
The reporter runs python in the context of our automation, and we break it. It turns out we don't need it so we disable it.

### DIFF
--- a/eng/performance/blazor_scenarios.proj
+++ b/eng/performance/blazor_scenarios.proj
@@ -6,6 +6,10 @@
     </HelixCorrelationPayload>
   </ItemGroup>
 
+  <PropertyGroup>
+    <EnableAzurePipelinesReporter>false</EnableAzurePipelinesReporter>
+  </PropertyGroup>
+
   <PropertyGroup Condition="'$(TargetsWindows)' == 'true'">
     <ScenariosDir>$(WorkItemDirectory)\src\scenarios\</ScenariosDir>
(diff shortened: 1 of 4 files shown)


Lxiamail (approved):

Lxiamail:
Do we currently have active test failures, or just for future failure prevention?

billwert:
I noticed it causing failures in our pipeline while looking at something else with Parker.

LoopedBard3:
Please remind me what the goal of this fix was? It appears that the change fixed many of the build problems but now the tests that did fail are not showing up in a tests tab (The tests tab is not appearing next to summary)

billwert:
This fix fixes instances where we get logs that end like this:
```
  File "C:\h\w\A69D0947\p\reporter\run.py", line 13, in <module>
    from test_results_reader import read_results
  File "C:\h\w\A69D0947\p\reporter\test_results_reader\__init__.py", line 3, in <module>
    from helix.public import TestResult, TestResultAttachment
ModuleNotFoundError: No module named 'helix'
```

Not all of our pipelines currently do this. I'm proactively fixing it everywhere so we don't have a problem in the future. 

I'm not sure why the test failures aren't showing up, but it isn't related to this change. (Test failures were showing up previously and this has been broken for many days now.)

billwert:
Other failures unrelated.

PR #1938, opened 2021-08-20
Another fix for retry logic

billwert:
The previous fix (and revert) for this broke lots of things. This is the correct fix.


billwert:
@DrewScoggins you might find merging this fixes some issues. I think it is in a good state.

PR #1939, opened 2021-08-20
Preserve PYTHONPATH

billwert:
We need to preserve PYTHONPATH to enable the AzurePipelinesReporter task.


DrewScoggins:
LGTM, if this runs at the right time.

Lxiamail (approved):
LGTM

billwert:
This works now. We need a similar fix in the runtime repo now.

PR #1944, opened 2021-08-23
Add no https option for generation

LoopedBard3:
Found that some templates have the option to have no https enabled when they are generated so I added a precommands arg and piped it to where we generate the projects enabling this flag to be set.


LoopedBard3:
Don't forget!!! @LoopedBard3 reset the yaml, the separate commit for it got squashed when you were fixing your fork.

DrewScoggins:
Hmm, I am not sure if we want to actually test on just http. https is the common scenario, so I think I would want to do that.

LoopedBard3:
The main snag with testing with https is that the dev-cert needs to be installed. I didn't see anything having to do with it finishing the install and, at least locally, there is a gui popup that needs to be confirmed for the cert to install when using dotnet dev-cert --trust. I agree that if we figure out the blazor wasm hot reload and this is not necessary, then it should not be merged.

billwert:
@sebastienros Do you have an easy way to automate dev cert install for SSL testing?

sebastienros:
> there is a gui popup that needs to be confirmed for the cert to install

Mac ?

PR #1952, opened 2021-08-24
Removed non-main pipeline jobs from pipeline

LoopedBard3:
Removed non-main branch pipeline jobs from YML file. This is a part of the larger segmenting of the different release pipelines so that they exist individuallyenabling the status badge scenario and making the release that individual pipelines ran for more clear. This also allows each release to have its own jobs run and cuts down on the time it takes for an individual run for a release takes.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,7 +35,6 @@ schedules:
   branches:
     include:
     - main
-    - release/3.1.4xx
   always: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


LoopedBard3:
I was not quite sure what to do with the jobs that had main already commented out and the 5.0 jobs, so I just commented out the complete job for ones with main commented out and left the 5.0 jobs. If we don't need either of these I will clean them up, I just need to know.

PR #1953, opened 2021-08-24
Removed non-5.0.1xx jobs from pipeline.

LoopedBard3:
Segmented 5.0.1xx channel from the other channels so only release/5.0.1xx will run in this repo. This is a part of a trifecta of PRs on this branch, 3.1.4, and the main branch.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,7 +7,7 @@ resources:
 trigger:
   branches:
     include:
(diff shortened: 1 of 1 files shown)


LoopedBard3:
This should be pretty much ready to use, the only thing that may be desirable is having the cron job build enabled. I disabled it for the time being.

PR #1986, opened 2021-09-03
Fix blaz hr again

LoopedBard3:


danmoseley:
This seems dead. I will close.

PR #1987, opened 2021-09-03
Disabling blazor hot reload

LoopedBard3:
Disabling blazor hot reload as we fix its running errors, especially in prep for the long weekend. An exact copy of the last time we disabled blazor hot reload.


DrewScoggins (approved):
LGTM

LoopedBard3:
Errors are unrelated.

PR #1999, opened 2021-09-10
Add more `StringBuilder` benchmarks.

teo-tsirpanis:
This PR adds additional benchmarks for the `StringBuilder` type, specifically measuring appending read-only spans (on non-.NET Framework only), appending value types using string interpolation (taking advantage of the .NET 6 optimizations), and inserting primitives.

related to dotnet/runtime#58907

c.c. @danmoseley

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text/Perf.StringBuilder.cs
+++ b/src/benchmarks/micro/libraries/System.Text/Perf.StringBuilder.cs
@@ -178,6 +178,26 @@ namespace System.Text.Tests
             return builder;
         }
 
+        // on .NET 6+, interpolated string handlers make appending more efficient by avoiding boxing and using ISpanFormattable.
+        [Benchmark]
+        public StringBuilder Append_ValueTypes_Interpolated()
+        {
+            var builder = new StringBuilder();
+
+            var dateTime = new DateTime(2018, 12, 14);
+            var timeSpan = new TimeSpan(1, 2, 0);
+            var dateTimeOffset = new DateTimeOffset(dateTime, timeSpan);
+            var guid = new Guid(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11);
+
+            // to amortize the cost of creating the value types we add them few times
+            for (int i = 1; i <= 4; i++)
+            {
+                builder.Append($"{guid} {dateTime} {timeSpan} {dateTimeOffset}");
+            }
+
+            return builder;
+        }
+
         [Benchmark]
         public StringBuilder Append_Memory()
         {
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.Text/Perf.StringBuilder.cs
```diff
@@ -178,6 +178,26 @@ public StringBuilder Append_ValueTypes()
...
+            var dateTime = new DateTime(2018, 12, 14);
+            var timeSpan = new TimeSpan(1, 2, 0);
+            var dateTimeOffset = new DateTimeOffset(dateTime, timeSpan);
+            var guid = new Guid(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11);
+
+            // to amortize the cost of creating the value types we add them few times
+            builder.Append($"{guid}"); builder.Append($"{dateTime}"); builder.Append($"{timeSpan}"); builder.Append($"{dateTimeOffset}");
+            builder.Append($"{guid}"); builder.Append($"{dateTime}"); builder.Append($"{timeSpan}"); builder.Append($"{dateTimeOffset}");
+            builder.Append($"{guid}"); builder.Append($"{dateTime}"); builder.Append($"{timeSpan}"); builder.Append($"{dateTimeOffset}");
+            builder.Append($"{guid}"); builder.Append($"{dateTime}"); builder.Append($"{timeSpan}"); builder.Append($"{dateTimeOffset}");
```
adamsitnik, line 196: Is C# compiler smart enough to detect that the value types don't change their values and the each string needs to be interpolated just once? If so, this benchmark is measuring `builder.Append(string)` performance which is already covered here and we need to modify the values. 

```cs
var dateTime = new DateTime(2018, 12, 14);
var timeSpan = new TimeSpan(1, 2, 0);
var dateTimeOffset = new DateTimeOffset(dateTime, timeSpan);
var guid = new Guid(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11);

for (int i = 1; i <= 10; i++)
{
    builder.Append($"{guid} {dateTime} {timeSpan} {dateTimeOffset}");
    dateTime = date.AddSeconds(i);
}
```

teo-tsirpanis: I don't think the compiler is _that_ smart to recognize such pattern, but I still changed the benchmark to use a loop and a single interpolated string with many holes (a more real-word case).

adamsitnik (approved):
LGTM, thank you @teo-tsirpanis !

danmoseley:
@teo-tsirpanis if you are comfortable connecting on email, would you reach out to me at danmose - at - microsoft.com ? I have a question!

Issue #2009, opened 2021-09-13
Add support for net7.0 to dotnet.py script


kunalspathak:
Currently, when we do `dotnet.py install --architecture x64 --install-dir D:\workspace\_work\1\s\workitem\performance\tools\dotnet\x64 --verbose` , it installs net6.0 but it should start installing net7.0

billwert:
Good call. We need to spin up a release/6.0.1xx branch in the perf repo before we can change this. Is this blocking you?

kunalspathak:
Yes...superpmi collection is failing with following errors:

```
Running: C:\h\w\B38009EC\w\ABE2096B\u\performance\tools\dotnet\x64\dotnet build C:\h\w\B38009EC\w\ABE2096B\u\performance\src\benchmarks\micro\MicroBenchmarks.csproj --configuration Release --framework net7.0 --no-restore /p:NuGetPackageRoot=C:\h\w\B38009EC\w\ABE2096B\u\performance\artifacts\packages -o C:\h\w\B38009EC\w\ABE2096B\u\performance\artifacts
Microsoft (R) Build Engine version 17.0.0-preview-21460-01+8f208e609 for .NET

Copyright (C) Microsoft Corporation. All rights reserved.



C:\h\w\B38009EC\w\ABE2096B\u\performance\tools\dotnet\x64\sdk\6.0.100-rc.2.21460.37\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(144,5): error NETSDK1045: The current .NET SDK does not support targeting .NET 7.0.  Either target .NET 6.0 or lower, or use a version of the .NET SDK that supports .NET 7.0. [C:\h\w\B38009EC\w\ABE2096B\u\performance\src\benchmarks\micro\MicroBenchmarks.csproj]



Build FAILED.



C:\h\w\B38009EC\w\ABE2096B\u\performance\tools\dotnet\x64\sdk\6.0.100-rc.2.21460.37\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(144,5): error NETSDK1045: The current .NET SDK does not support targeting .NET 7.0.  Either target .NET 6.0 or lower, or use a version of the .NET SDK that supports .NET 7.0. [C:\h\w\B38009EC\w\ABE2096B\u\performance\src\benchmarks\micro\MicroBenchmarks.csproj]

    0 Warning(s)

    1 Error(s)

```

[...]

PR #2012, opened 2021-09-14
Update channel map for net7.0 and servicing runtime branches

DrewScoggins:
These are the changes that go along with https://github.com/dotnet/runtime/pull/59070, in order to keep both main and the release branches working. Also should fix https://github.com/dotnet/performance/issues/2009.

### DIFF
--- a/scripts/channel_map.py
+++ b/scripts/channel_map.py
@@ -2,14 +2,19 @@ from argparse import ArgumentParser
 
 class ChannelMap():
     channel_map = {
+        'release/6.0': {
+            'tfm': 'net6.0',
+            'branch': '6.0.1xx',
+            'quality': 'daily'
+        },
         '6.0': {
             'tfm': 'net6.0',
-            'branch': 'main',
+            'branch': '6.0.1xx',
             'quality': 'daily'
         },
         'main': {
-            'tfm': 'net6.0',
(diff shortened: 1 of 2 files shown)


scripts/channel_map.py
```diff
@@ -2,14 +2,19 @@
...
             'tfm': 'net6.0',
-            'branch': 'main',
+            'branch': '6.0.1xx',
             'quality': 'daily'
         },
         'main': {
-            'tfm': 'net6.0',
-            'branch': 'main',
+            'tfm': 'net7.0',
+            'branch': '7.0.1xx',
```
billwert, line 17: I'm surprised this isn't `main` now?

billwert (changes requested):
see comment

scripts/channel_map.py
DrewScoggins: From talking with Matt Mitchell they are moving away from branches that imply source, and instead are moving to channel names that imply intent. So there will be no main channel going forward.

billwert (approved):

LoopedBard3 (approved):

DrewScoggins:
Did the change for the release branch go in @LoopedBard3? If so we can finally merge this.

Issue #2005, opened 2021-09-10
dotnet-performance pipeline release/5.0.1xx all Linux jobs fail


Lxiamail:
pipeline: dotnet-performance 
job: [all Linux jobs](https://dev.azure.com/dnceng/internal/_build/results?buildId=1352171&view=results)
Failure ([full failure log](https://helix.dot.net/api/2019-06-17/jobs/05aeb171-cc51-4c2b-8ab0-bff88370a89b/workitems/Partition0/console?access_token=)): 
+ pip3 install azure.storage.blob==12.0.0 --force-reinstall
Collecting azure.storage.blob==12.0.0
  Using cached https://files.pythonhosted.org/packages/1b/15/d3d5d1cd178433982346b87317c25be78c1bcdcce8404bf7bbed05ac3a15/azure_storage_blob-12.0.0-py2.py3-none-any.whl
Collecting azure-core<2.0.0,>=1.0.0 (from azure.storage.blob==12.0.0)
  Using cached https://files.pythonhosted.org/packages/d5/7c/d6f248b79bfc9f276987d0e9546525deacfca0df47447f341d87069cbf73/azure_core-1.18.0-py2.py3-none-any.whl
Collecting cryptography>=2.1.4 (from azure.storage.blob==12.0.0)
  Using cached https://files.pythonhosted.org/packages/cc/98/8a258ab4787e6f835d350639792527d2eb7946ff9fc0caca9c3f4cf5dcfe/cryptography-3.4.8.tar.gz
    Complete output from command python setup.py egg_info:
    
            =============================DEBUG ASSISTANCE==========================
            If you are seeing an error here please try the following to
            successfully install cryptography:
    
            Upgrade to the latest pip and try again. This will fix errors for most
            users. See: https://pip.pypa.io/en/stable/installing/#upgrading-pip
            =============================DEBUG ASSISTANCE==========================
    
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/home/helixbot/work/B41E09C9/t/pip-build-4lclz2nt/cryptography/setup.py", line 14, in <module>
        from setuptools_rust import RustExtension
    ModuleNotFoundError: No module named 'setuptools_rust'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /home/helixbot/work/B41E09C9/t/pip-build-4lclz2nt/cryptography/

LoopedBard3:
[...]

PR #2018, opened 2021-09-15
Add-pip-update-to-5.0

billwert:
We're not properly updating pip in release branches which leads to #2005.

### DIFF
--- a/eng/performance/scenarios.yml
+++ b/eng/performance/scenarios.yml
@@ -61,7 +61,7 @@ jobs:
               value: 'py -3 -m venv %HELIX_WORKITEM_PAYLOAD%\.venv;call %HELIX_WORKITEM_PAYLOAD%\.venv\Scripts\activate.bat;set PYTHONPATH=;py -3 -m pip install azure.storage.blob==12.0.0 --force-reinstall;py -3 -m pip install azure.storage.queue==12.0.0 --force-reinstall;set "PERFLAB_UPLOAD_TOKEN=$(PerfCommandUploadToken)";$(AdditionalHelixPreCommands)'
           - ${{ if ne(parameters.osName, 'windows') }}:
             - name: HelixPreCommand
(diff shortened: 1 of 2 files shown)


eng/performance/scenarios.yml
```diff
@@ -61,7 +61,7 @@ jobs:
...
           - ${{ if ne(parameters.osName, 'windows') }}:
             - name: HelixPreCommand
-              value: 'sudo apt-get -y install python3-venv;python3 -m venv $HELIX_WORKITEM_PAYLOAD/.venv;. $HELIX_WORKITEM_PAYLOAD/.venv/bin/activate;export PYTHONPATH=;pip3 install azure.storage.blob==12.0.0 --force-reinstall;pip3 install azure.storage.queue==12.0.0 --force-reinstall;export PERFLAB_UPLOAD_TOKEN="$(PerfCommandUploadTokenLinux)";$(AdditionalHelixPreCommands)'
+              value: 'sudo apt-get -y install python3-venv;python3 -m venv $HELIX_WORKITEM_PAYLOAD/.venv;. $HELIX_WORKITEM_PAYLOAD/.venv/bin/activate;export PYTHONPATH=;python3 -m pip install -U pip;pip3 install azure.storage.blob==12.0.0;pip3 install azure.storage.queue==12.0.0;export PERFLAB_UPLOAD_TOKEN="$(PerfCommandUploadTokenLinux)";$(AdditionalHelixPreCommands)'
```
Lxiamail, line 64: Why "--force-reinstall" are removed here? Don't we still want reinstall?

billwert: I think that's a vestige of before we were using a venv. We don't force reinstall in our `main` branch:

https://github.com/dotnet/performance/blob/ef497aa104ae7abe709c71fbb137230bf5be25e9/eng/performance/scenarios.yml#L73

Lxiamail: > @Lxiamail @DrewScoggins thoughts? should we merge?

LGTM

PR #2047, opened 2021-10-05
Add the release 6.0 mapping to the channel mapping list

LoopedBard3:
Add the release 6.0 mapping to the channel mapping list to fix release 6.0 runtime perf issue.


LoopedBard3:
Covered in #2012

PR #2054, opened 2021-10-08
Add C# multi-dimension array versions of the Benchstones microbenchmarks

BruceForstall:
The current versions of the Benchstones microbenchmarks use C#
"jagged" arrays for multi-dimension arrays. This adds a set of all the
BenchF/BenchI benchmarks that use multi-dimensions arrays, using C#
true multi-dimension array access. The code is exactly the same as the
non-multi-dimension versions except for array creation and indexing
expressions.

This will help track and compare multi-dimension versus jagged array
performance.


BruceForstall:
@billwert

Issue #1822, opened 2021-05-25
Enable markdown lint


billwert:
As suggested in #1821.

danmoseley:
@bouchraRekhadda any interest in picking this up? Even if you're not familiar with yaml and Github actions, it should be fairly easy to just duplicate what was done in the dotnet/runtime repo: just copy https://github.com/dotnet/runtime/pull/40887

bouchraRekhadda:
>@bouchraRekhadda any interest in picking this up?

More than happy to. I'll give it a try

danmoseley:
Thanks, assigned. It will likely need some tuning to exclude low-value rules (eg perhaps trailing spaces) and retain useful ones (eg broken links)

billwert:
@bouchraRekhadda any luck? :)

bouchraRekhadda:
Thanks for the reminder @billwert 😅; I was busy with work in the past few months and it totally slipped my mind.
I'll take a look today

PR #2060, opened 2021-10-11
Add markdownlint

bouchraRekhadda:
Config was copied from [dotnet/runtime/.github/workflows/markdownlint.yml](https://github.com/dotnet/runtime/blob/5d5c3e7a589e21e43ad6464b6ed6732e98e7fc8a/.github/workflows/markdownlint.yml).
This closes https://github.com/dotnet/performance/issues/1822

### DIFF
--- a/docs/benchmarking-workflow-dotnet-runtime.md
+++ b/docs/benchmarking-workflow-dotnet-runtime.md
@@ -199,7 +199,7 @@ The next step is to send a PR to this repository with the aforementioned benchma
 
 ### Profiling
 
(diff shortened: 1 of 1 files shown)


billwert:
> I am a bit surprised to see all `performance-ci` workflows launched on my PR that doesn't modify any sources.

The exclusions for what gets launched are [path based](https://github.com/dotnet/performance/blob/1a7d35d3b1749e505f367d954a979853818e63e1/azure-pipelines.yml#L22-L30) so it makes sense the whole thing ran.

Can you modify a .md file in your branch so we can see the workflow work?

PR #2077, opened 2021-10-14
Fix-markdownlint-errors

billwert:
Cleaning up a bunch of errors reported by the new action so the pipeline stays green.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,6 +1,6 @@
 {
+    "MD004": false,
     "MD013": false,
     "MD024": false,
-    "MD004": false,
     "MD034": false
 }


billwert:
@cshung @Maoni0 I cleaned up some things in the GC docs with this change. If any of them are a problem please let us know.

.markdownlint.json
```diff
@@ -1,6 +1,6 @@
 {
-    "default": false,
-    "MD009": {
-        "br_spaces": 0
-    }
+    "MD013": false,
+    "MD024": false,
+    "MD004": false,
+    "MD034": false
```
LoopedBard3, line 5: Nit, MD004 should come before the other MDs, also want to make sure that this is the correct rule since it is between 24 and 34.

docs/profiling-workflow-dotnet-runtime.md
```diff
@@ -2,41 +2,42 @@
...
+  - [Identifying Regressions](#identifying-regressions)
+  - [VTune](#vtune)
+  - [When to use](#when-to-use)
+  - [Identifying Hotspots](#identifying-hotspots)
+  - [Troubleshooting](#troubleshooting)
+  - [Code](#code)
+  - [Skids](#skids)
+  - [Linux](#linux)
+  - [PerfCollect](#perfcollect)
+  - [Preparing Your Machine](#preparing-your-machine)
```
LoopedBard3, line 31: Why did we change headers from being sub-headings?

billwert: Ah. An extension I had installed automatically fixes up the TOC to be in-line with the headers below. I think fixing the headings is a better idea anyway. I'll do that.

PR #2078, opened 2021-10-14
Add net7.0 to the include list for Cbor library

DrewScoggins:


LoopedBard3:
Merging to get the change in, especially given the low impact it should have aside from the fix.

PR #2082, opened 2021-10-15
Add net.70 TFMto Microbenchmarks.csproj and net7 Nuget feed

DrewScoggins:


LoopedBard3:
@DrewScoggins We will have to check out the scenario errors, but it otherwise LGTM

PR #2090, opened 2021-10-21
Temporarily Allow Use of OpenSSL 1.0.2 In Pipeline

Tohron:
This is a temporary fix to deal with OpenSSL 1.0.2 cryptography still being present in some parts of our pipeline until the underlying issue can be addressed.

### DIFF
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -47,7 +47,7 @@ jobs:
           - name: PreservePythonPath
             value: 'export ORIGPYPATH=$PYTHONPATH'
           - name: SetAllowOpenSsl102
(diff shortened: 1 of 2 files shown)


eng/performance/benchmark_jobs.yml
```diff
@@ -46,6 +46,8 @@ jobs:
             value: 'chmod +x ./tools/machine-setup.sh;. ./tools/machine-setup.sh'
           - name: PreservePythonPath
             value: 'export ORIGPYPATH=$PYTHONPATH'
+          - name: SetAllowOpenSsl102
+            value: 'export CRYPTOGRAPHY_ALLOW_OPENSSL_102 =true'
```
billwert, line 50: This is not sufficient. Notice how we then use variables like `PreservePythonPath` when building the command line down on line 92. You need to include it there as well.

billwert (approved):

Issue #1602, opened 2020-11-18
Memory Randomization


adamsitnik:
## The Problem

The problem that we are facing with our current setup are benchmarks that:

- have a flat distribution when run just once locally because the memory is allocated just once and BenchmarkDotNet uses the same input for all the iterations in a given run.
- have a multimodal distribution when considering historical data from multiple runs because the memory alignment changes over time and we don't control it. Example: today we are sorting an aligned array, tomorrow an unaligned one, and the reported time is different.

## Goal

The goal of the study was to see how randomizing memory alignment could improve long-term benchmark stability. What might seem to be surprising is that we don't want to get a flat distribution when running locally, but get a full picture for all possible scenarios (aligned and not aligned).

The idea comes from [stabilizer](https://github.com/ccurtsinger/stabilizer/blob/master/README.md) project, which was suggested as a possible solution by @AndyAyersMS a few years ago when we started discussing this problem.

## Implementation

The "randomization" has been implemented as a new, optional feature in BenchmarkDotNet in the following way:

- allocate a random-size stack memory at the beginning of every iteration (idea comes from @AndyAyersMS, [code](https://github.com/dotnet/BenchmarkDotNet/blob/7caa3c500ba7ae225d82da5baf1fc07d27e917c6/src/BenchmarkDotNet/Engines/Engine.cs#L167)) and keep it alive for the iteration period
- between every iteration:
    - call the `[GlobalCleanup]` method that should be disposing of all resources ([code](https://github.com/dotnet/BenchmarkDotNet/blob/7caa3c500ba7ae225d82da5baf1fc07d27e917c6/src/BenchmarkDotNet/Engines/Engine.cs#L225))
    - allocate a random-size small byte array (Gen 0 object) (idea comes from @jkotas, [code](https://github.com/dotnet/BenchmarkDotNet/blob/7caa3c500ba7ae225d82da5baf1fc07d27e917c6/src/BenchmarkDotNet/Engines/Engine.cs#L227))
    - allocate a random-size large byte array (LOH object) (idea comes from @AndyAyersMS, [code](https://github.com/dotnet/BenchmarkDotNet/blob/7caa3c500ba7ae225d82da5baf1fc07d27e917c6/src/BenchmarkDotNet/Engines/Engine.cs#L228))
[...]

PR #2111, opened 2021-11-01
Add initial memory randomization

DrewScoggins:
Added the memory randomization as laid out by @adamsitnik in https://github.com/dotnet/performance/issues/1602

### DIFF
--- a/src/benchmarks/micro/libraries/System.Collections/Contains/ContainsFalse.cs
+++ b/src/benchmarks/micro/libraries/System.Collections/Contains/ContainsFalse.cs
@@ -37,6 +37,9 @@ namespace System.Collections
         [GlobalSetup(Targets = new[] { nameof(Array), "Span" })]
         public void SetupArray() => _array = Setup();
 
+        [MemoryRandomization(true)]
+        [IterationCount(50)]
+        [Outliers(Perfolizer.Mathematics.OutlierDetection.OutlierMode.DontRemove)]
         [Benchmark]
         public bool Array()
         {
@@ -49,6 +52,9 @@ namespace System.Collections
         }
 
 #if !NETFRAMEWORK
+        [MemoryRandomization(true)]
+        [IterationCount(50)]
+        [Outliers(Perfolizer.Mathematics.OutlierDetection.OutlierMode.DontRemove)]
         [BenchmarkCategory(Categories.Span)]
         [Benchmark]
         public bool Span()
@@ -64,6 +70,9 @@ namespace System.Collections
         [GlobalSetup(Targets = new[] { nameof(List), nameof(ICollection) })]
         public void SetupList() => _list = new List<T>(Setup());
 
+        [MemoryRandomization(true)]
+        [IterationCount(50)]
+        [Outliers(Perfolizer.Mathematics.OutlierDetection.OutlierMode.DontRemove)]
         [Benchmark]
         public bool List()
         {
--- a/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
+++ b/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
@@ -135,6 +135,9 @@ namespace System.IO.Tests
             }
         }
 
+        [MemoryRandomization(true)]
+        [IterationCount(50)]
+        [Outliers(Perfolizer.Mathematics.OutlierDetection.OutlierMode.DontRemove)]
         [Benchmark]
         [Arguments(OneKibibyte, FileOptions.None)]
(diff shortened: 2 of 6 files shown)


src/benchmarks/micro/libraries/System.Memory/Span.cs
```diff
@@ -71,12 +77,18 @@ public void SetupIndexOf()
...
         public int IndexOfAnyTwoValues() => new System.Span<T>(_emptyWithSingleValue).IndexOfAny(_notDefaultValue, _notDefaultValue);
 
+        [MemoryRandomization(true)]
+        [IterationCount(50)]
+        [Outliers(Perfolizer.Mathematics.OutlierDetection.OutlierMode.DontRemove)]
         [Benchmark]
         public int IndexOfAnyThreeValues() => new System.Span<T>(_emptyWithSingleValue).IndexOfAny(_notDefaultValue, _notDefaultValue, _notDefaultValue);
 
         [Benchmark]
         public int IndexOfAnyFourValues() => new System.Span<T>(_emptyWithSingleValue).IndexOfAny(new ReadOnlySpan<T>(_fourValues));
```
LoopedBard3, line 87: Is there a specific reason not to do the randomization for both?

LoopedBard3 (approved):
LGTM once the CI finishes.

kunalspathak:
How are we  planning to represent these numbers in the test history graph? What happens to the existing test history?

cc: @AndyAyersMS

src/benchmarks/micro/libraries/System.Memory/Span.cs
DrewScoggins: I just went with the initial set that Adam had laid out. We might expand these going forward however.

DrewScoggins:
Right now the plan is still use the average, with error bars, to display these tests. As for the existing test history, I didn't really believe that these tests were of much value, given how they behave with respect to memory alignment So I was not too worried about the trend changing. We could give the test a new name, or duplicate the tests and have one run with alignment and one without if we prefer.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.FileStream.cs
```diff
@@ -135,6 +135,9 @@ public void SeekBackward(long fileSize, FileOptions options)
             }
         }
 
+        [MemoryRandomization(true)]
+        [IterationCount(50)]
+        [Outliers(Perfolizer.Mathematics.OutlierDetection.OutlierMode.DontRemove)]
```
adamsitnik, line 140: After my recent work on `FileStream` rewrite I doubt that any of the File IO benchmarks are affected by alignment issues.

@DrewScoggins have you tried comparing the results for MemoryRandomization enabled vs disabled for this benchmark? If we can't observe any gains, we should not enable it.

```suggestion
```

adamsitnik:
> Right now the plan is still use the average, with error bars, to display these tests.

Do you have some sample screen shots?

I was wondering if we could display every histogram bucket on the chart with a size equal to the number of iterations belonging to it.

We could use: https://plotly.com/javascript/bubble-charts/#bubble-size-scaling-on-charts. Example:

```ini
-------------------- Histogram --------------------
[ 79.009 ns ; 183.239 ns) | @@@@@@@@@@
[183.239 ns ; 287.047 ns) | @
[287.047 ns ; 391.277 ns) | @@@@@@@@
[391.277 ns ; 501.821 ns) | @
---------------------------------------------------
```

Would give us 4 points:
* for each of them, x would be the same
* y would be equal to median of values in given histogram
* size would be equal the number of iterations belonging to given bucket

We could also try box plots, but most likely we have too many data points to visualize

src/benchmarks/micro/libraries/System.Collections/Contains/ContainsFalse.cs
```diff
@@ -37,6 +37,9 @@ public class ContainsFalse<T> where T : IEquatable<T>
         [GlobalSetup(Targets = new[] { nameof(Array), "Span" })]
         public void SetupArray() => _array = Setup();
 
+        [MemoryRandomization(true)]
+        [IterationCount(50)]
+        [Outliers(Perfolizer.Mathematics.OutlierDetection.OutlierMode.DontRemove)]
```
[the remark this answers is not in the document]
DrewScoggins: OK, I was just going off of the way that it was called in your PR. I will take this out.

danmoseley:
@DrewScoggins just curious whether you plan to proceed with this one, it sounds intriguingly useful..

DrewScoggins:
I have been fully focused on other work, that should be finishing up in a week or two. My hope was to get back this along with other requests around test execution and visualization after.

cincuranet:
Closing in favor of https://github.com/dotnet/runtime/pull/95291.

PR #2112, opened 2021-11-01
Add Maui iOS and MacCatalyst Performance SOD tests

LoopedBard3:
Add Maui iOS and MacCatalyst Performance SOD test folders to the performance repo so they can be activated in the runtime repo.


DrewScoggins (approved):
LGTM

Issue #2089, opened 2021-10-20
Upload failures don't fail pipeline


LoopedBard3:
Upload failures in the performance and runtime perf pipelines did not cause the pipelines to register as failed. These failures should register a failure so we don't miss uploading data without knowing it isn't being uploaded.

Tohron:
@LoopedBard3 could you link to some of the pipeline runs where this occurred?

LoopedBard3:
@Tohron DM'd you some specific instances, but any runs around October 17-19th that upload data should demonstrate the same problem, at least in the runtime perf.

Tohron:
https://github.com/dotnet/performance/pull/2117 merged

PR #2117, opened 2021-11-03
Perf Scripts Fail if Results Upload Fails

Tohron:
PR to address https://github.com/dotnet/performance/issues/2089
by having perf scripts return a failure code if the results upload returns that failure code.
Failure when upload fields are invalid: https://dev.azure.com/dnceng/internal/_build/results?buildId=1478721&view=results
Pass when upload fields are valid: https://dev.azure.com/dnceng/internal/_build/results?buildId=1478740&view=results


LoopedBard3:
@Tohron Do we have any runs that show that this properly errors out the pipeline?

scripts/benchmarks_ci.py
```diff
@@ -243,7 +243,8 @@ def __main(args: list) -> int:
 
         if args.upload_to_perflab_container:
             import upload
-            upload.upload(globpath, upload_container, UPLOAD_QUEUE, UPLOAD_TOKEN_VAR, UPLOAD_STORAGE_URI)
+            upload_code = upload.upload(globpath, upload_container, UPLOAD_QUEUE, UPLOAD_TOKEN_VAR, UPLOAD_STORAGE_URI)
```
LoopedBard3, line 246: It may be better to only return on failure incase things get added that run after upload in the future.

LoopedBard3 (approved):
Once we have verified that this causes the failure as expected either locally or in the pipeline, it should be good to merge. Although only returning on failure would be a nice implement in this case if possible.

LoopedBard3 (approved):
LGTM for covering the main cases. We should make an issue to track testing and fixing other issue types that may not be caught.

PR #2122, opened 2021-11-05
Diasble Cbor tests until crypto change makes its way to SDK

DrewScoggins:


DrewScoggins:
https://github.com/dotnet/performance/issues/2123

LoopedBard3 (approved):
Generally LGTM.

DrewScoggins:
@naricc I just added on the BDN update to this PR. Once tests pass should be good to go.

adamsitnik:
@DrewScoggins @LoopedBard3 By commenting out entire file you have to modify multiple files and later become an owner of the file from the "git blame" perspective.

In case you need to disable some benchmarks in the future, the easiest way of doing that is by disabling the compilation for given file in the [Microbenchmarks.csproj](https://github.com/dotnet/performance/blob/main/src/benchmarks/micro/MicroBenchmarks.csproj)

Example:

```xml
```

PR #2132, opened 2021-11-15
add some new string(char, int) benchmarks

adamsitnik:
Needed for https://github.com/dotnet/runtime/pull/60269

### DIFF
--- a/src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
@@ -21,6 +21,12 @@ namespace System.Tests
         }
 
         [Benchmark]
+        [Arguments(1)]
+        [Arguments(10)]
+        [Arguments(100)]
+        public string CtorCharCount(int size) => new string('a', size);
+
+        [Benchmark]
         [ArgumentsSource(nameof(TestStringSizes))]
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
```diff
@@ -20,6 +20,12 @@ public static IEnumerable<object> TestStringSizes()
             yield return new StringArguments(1000);
         }
 
+        [Benchmark]
+        [Arguments(1)]
+        [Arguments(10)]
+        [Arguments(100)]
```
danmoseley, line 26: would a large number be worth adding? there's certainly 10K strings eg.

danmoseley (approved):

src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
GrabYourPitchforks: Not sure if that would be worth adding. The typical use case for this ctor is to add whitespace padding to a line of text. Example: `string combined = new string(' ', indentationLevel * 4) + originalLine;` Realistically, this value should never get above a few dozen or so.

danmoseley: fair enough

PR #2139, opened 2021-11-18
Add Utf8JsonReader.GetBoolean() benchmark

eiriktsarpalis:
Add a `GetBoolean()` benchmark and simplify existing benchmarks in the same class. Used by https://github.com/dotnet/runtime/pull/61746.

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
+++ b/src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
@@ -4,14 +4,13 @@
 
 using BenchmarkDotNet.Attributes;
 using MicroBenchmarks;
+using System.Linq;
 
 namespace System.Text.Json.Tests
 {
     [BenchmarkCategory(Categories.Libraries, Categories.JSON)]
     public class Perf_Get
     {
-        private const int OperationsPerBenchmark = 100;
-
         private static readonly byte[] _jsonFalseBytes = GetJsonBytes("false");
         private static readonly byte[] _jsonIntegerNumberBytes = GetJsonBytes(123);
         private static readonly byte[] _jsonDecimalNumberBytes = GetJsonBytes(123.456f);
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -23,214 +25,189 @@ private static byte[] GetJsonBytes<T>(T elem)
             return Encoding.UTF8.GetBytes(elem.ToString());
         }
 
-        [Benchmark]
-        public byte GetByte()
+        [Benchmark(OperationsPerInvoke = OperationsPerBenchmark)]
```
layomia, line 28: What is the benefit of using `OperationsPerInvoke` here?

layomia (commented):
cc @steveharter @adamsitnik

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
adamsitnik: @layomia https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#operationsperinvoke

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -23,214 +25,189 @@ private static byte[] GetJsonBytes<T>(T elem)
...
-            byte result = 0;
-            var reader = new Utf8JsonReader(_jsonIntegerNumberBytes);
+            var reader = new Utf8JsonReader(_jsonFalseBytes);
             reader.Read();
 
-            for (int i = 0; i < 100; i++)
+            for (int i = 0; i < OperationsPerBenchmark; i++)
             {
-                result += reader.GetByte();
+                reader.GetBoolean();
```
adamsitnik, line 36: We need to consume the result to make sure that some smart compilers (like the Mono AOT compiler) are not going to treat it as dead code and remove it:

```cs
public bool GetBoolean()
{
    var reader = new Utf8JsonReader(_jsonFalseBytes);
    reader.Read();

    bool result = false;
    for (int i = 0; i < OperationsPerBenchmark; i++)
    {
        result ^= reader.GetBoolean();
    }
    return result;
}
```

https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#dead-code-elimination

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -23,214 +25,189 @@ private static byte[] GetJsonBytes<T>(T elem)
...
             {
-                result += reader.GetInt16();
+                reader.GetSByte();
             }
-            return result;
         }
 
-        [Benchmark]
-        public int GetInt32()
+        [Benchmark(OperationsPerInvoke = OperationsPerBenchmark)]
```
adamsitnik, line 64: Overall it's a good idea to use `OperationsPerInvoke` for such benchmarks, but in case of an existing benchmark we can't do it. It would suddenly change the time reported by BDN (`newTime = oldTime / OperationsPerBenchmark`) which is stored in the [ReportingSystem](https://pvscmdupload.blob.core.windows.net/reports/allTestHistory%2frefs%2fheads%2fmain_x64_Windows%2010.0.18362%2fSystem.Text.Json.Tests.Perf_Get.GetInt16.html) and would be recognized by the automation as a sudden perf improvement. 

We could do that if we were also about to change the benchmark ID (namespace.typeName.MethodName), but this would be done at a price of loosing the historical data in the Reporting System (it would be recognized as a new benchmark).

```suggestion
        [Benchmark]
```

That is why we hardly ever rename the benchmarks. For more detail please refer to:

https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
```diff
@@ -23,214 +25,189 @@ private static byte[] GetJsonBytes<T>(T elem)
...
-                result += reader.GetByte();
+                reader.GetBoolean();
             }
-            return result;
         }
 
-        [Benchmark]
-        public sbyte GetSByte()
+        [Benchmark(OperationsPerInvoke = OperationsPerBenchmark)]
+        public void GetByte()
```
adamsitnik, line 41: same as above, the result needs to be consumed

adamsitnik (changes requested):
@eiriktsarpalis thank you for your contribution! Please see my comments for full details.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
eiriktsarpalis: I'd be surprised if any compiler applied DCE to method calls, since it would need to be able to reason about mutations happening to the `reader` instance. But maybe some compilers are smarter than I would have thought 🤷 

In this case I think it makes sense to remove aggregation: these really are micro-level benchmarks, so I would anticipate that the added instructions contribute to noise.

src/benchmarks/micro/libraries/System.Text.Json/Utf8JsonReader/Perf.Get.cs
eiriktsarpalis: I imagined this might be a concern, but my assumption was that it wouldn't be an issue provided that it's an improvement versus a regression.

adamsitnik (approved):
LGTM, thank you @eiriktsarpalis !

PR #2147, opened 2021-11-20
Add performance portion of android startup scenario

LoopedBard3:
Add performance portion of android startup scenario. This consists of adding the Device Startup scenario to the perf runner, updating the android test files (pre.py) to account for both versions of the apk, and adding support for measuring from traces for startup.


scripts/performance/common.py
```diff
@@ -197,7 +202,9 @@ def __runinternal(self, working_directory: str = None) -> tuple:
             ) as proc:
                 if proc.stdout is not None:
                     with proc.stdout:
+                        self.__stdout = StringIO()
                         for line in iter(proc.stdout.readline, ''):
+                            self.__stdout.write(line)
```
DrewScoggins, line 207: Is this just so that we can log what is being output from the process we started?

src/scenarios/shared/startup.py
```diff
@@ -63,7 +94,7 @@ def runtests(self, traits: TestTraits):
         for key in ['apptorun', 'startupmetric', 'guiapp']:
             if not getattr(traits, key):
                 raise Exception('startup tests require %s' % key)
-        reportjson = os.path.join(TRACEDIR, 'perf-lab-report.json')
```
DrewScoggins, line 66: Why did we get rid of this line?

src/tools/ScenarioMeasurement/Startup/DeviceTimeToMain.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace ScenarioMeasurement
+{
+
+    /// <summary>
+    /// This is a custom parser that does not enable any profiling. Instead, it relies on being passed nettrace files collected
```
DrewScoggins, line 13: I am wondering if instead of using this we should pass a path to a folder, and a Regex of all of the files to match of that type. I think I would prefer that to having magic that looks for certain values.

src/scenarios/shared/runner.py
```diff
@@ -49,6 +54,16 @@ def parseargs(self):
...
         self.add_common_arguments(startupparser)
 
+        # parse only command
+        parseonlyparser = subparsers.add_parser(const.DEVICESTARTUP,
+                                              description='measure time to main for Android apps')
+        parseonlyparser.add_argument('--device-type', choices=['android','ios'],type=str.lower,help='Device type for testing', dest='devicetype')
+        parseonlyparser.add_argument('--package-path', help='Location of test application', dest='packagepath')
+        parseonlyparser.add_argument('--package-name', help='Classname of application', dest='packagename')
+        parseonlyparser.add_argument('--exit-code', help='Success exit code', dest='expectedexitcode')
+        parseonlyparser.add_argument('--startup-trials', help='Startups to run (1+)', type=int, default=5, dest='startuptrials')
```
DrewScoggins, line 64: nit: trials -> iterations

src/scenarios/shared/startup.py
LoopedBard3: It was moved out of this specific case so other portions also have access to the reportjson variable.

scripts/performance/common.py
LoopedBard3: This is more so we are able to get out the adb command location from the running of a specific command in runner.py, line 318 for example.

LoopedBard3: This is so that we can capture the adb location from the output of the xharness command.

src/tools/ScenarioMeasurement/Startup/DeviceTimeToMain.cs
LoopedBard3: Implemented a parse method specifically for parsing multiple files from a directory.

src/scenarios/helloandroid/pre.py
```diff
@@ -16,27 +16,29 @@
...
-else:
+if not os.path.exists(PUBDIR):
     os.mkdir(PUBDIR)
-    copyfile(apkname, os.path.join(PUBDIR, apkname))
+apknames = ['HelloAndroid.apk', 'HelloAndroidWithDiag.apk']
+for apkname in apknames:
+    apknamezip = '%s.zip' % (apkname)
+    if not os.path.exists(apkname):
+        getLogger().error('Cannot find %s' % (apkname))
+        exit(-1)
```
LoopedBard3, line 26: Do we want to fail if an APK is not found, or do we want to rely on the calling of the test method at test time to error if the apk is not found.

DrewScoggins: I think that we should fail at the first time that we discover that there is an error. So in that case, it should be when we realize that we do not have the APKs in the pre step.

LoopedBard3:
Errors are unrelated, merging with runtime portion once final checks finish.

LoopedBard3:
Successfully ran with corresponding runtime update and errors are unrelated.

Issue #2140, opened 2021-11-18
MicroBenchmark runs are failing when executing Perf_Regex_Industry_Leipzig tests


LoopedBard3:
MicroBenchmark runs are failing when executing Perf_Regex_Industry_Leipzig tests. The specific error is 
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xec in position 21: invalid continuation byte
```
As a part of 
```
[2021/11/18 12:22:02][INFO] // Found 5 benchmarks:
[2021/11/18 12:22:02][INFO] //   Perf_Regex_Industry_Leipzig.Count: Job-RCHHKQ(PowerPlanMode=00000000-0000-0000-0000-000000000000, Arguments=/p:DebugType=portable,-bl:benchmarkdotnet.binlog, InvocationCount=1, IterationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=0) [Pattern=(?i)Tom|Sawyer|Huckleberry|Finn, Options=Compiled]
[2021/11/18 12:22:02][INFO] //   Perf_Regex_Industry_Leipzig.Count: Job-RCHHKQ(PowerPlanMode=00000000-0000-0000-0000-000000000000, Arguments=/p:DebugType=portable,-bl:benchmarkdotnet.binlog, InvocationCount=1, IterationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=0) [Pattern=Huck[a-zA-Z]+|Saw[a-zA-Z]+, Options=None]
[2021/11/18 12:22:02][INFO] //   Perf_Regex_Industry_Leipzig.Count: Job-RCHHKQ(PowerPlanMode=00000000-0000-0000-0000-000000000000, Arguments=/p:DebugType=portable,-bl:benchmarkdotnet.binlog, InvocationCount=1, IterationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=0) [Pattern=Tom.{10,25}river|river.{10,25}Tom, Options=None]
[...]

PR #2150, opened 2021-11-20
Change reenable regex industry tests.

LoopedBard3:
Worked locally, but it worked locally before, look to the PR tests to see if the fix worked.


LoopedBard3:
This if for #2140.

Tohron (approved):

PR #2154, opened 2021-11-24
Add benchmarks for InitBlock/CopyBlock operations

echesakovMSFT:
This adds benchmark for `GT_STORE_BLK` node in the JIT that responsibly for two operations in .NET - InitBlock and CopyBlock.

### DIFF
--- a/src/benchmarks/micro/runtime/StoreBlock/StoreBlock.cs
+++ b/src/benchmarks/micro/runtime/StoreBlock/StoreBlock.cs
@@ -6,1239 +6,334 @@ using System;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Extensions;
 using MicroBenchmarks;
 
-[BenchmarkCategory(Categories.Runtime, Categories.JIT)]
-public class StoreBlock
+namespace StoreBlock
 {
-    const int OperationsPerInvoke = 1000;
-
-    byte[] _srcData = new byte[1024];
-    byte[] _dstData = new byte[1024];
-
-    [GlobalSetup]
-    public void Setup()
-    {
-        for (int i = 0; i < _srcData.Length; i++)
-            _srcData[i] = (byte)(i % 256);
-
-        for (int i = 0; i < _dstData.Length; i++)
-            _dstData[i] = (byte)(i % 256);
-    }
-
-    [StructLayout(LayoutKind.Explicit, Size=8)]
-    struct Struct8
-    {
-    }
-
-    Struct8 fld8;
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public void InitBlockAllZerosHeap8()
-    {
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.InitBlock(ref _dstData[0], 0, 8);
-        }
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public unsafe void InitBlockAllZerosLocalAddr8()
-    {
-        Struct8 dstLcl;
-
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.InitBlock(&dstLcl, 0, 8);
-        }
-
-        fld8 = dstLcl;
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public void InitBlockAllOnesHeap8()
-    {
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.InitBlock(ref _dstData[0], 255, 8);
-        }
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public unsafe void InitBlockAllOnesLocalAddr8()
-    {
-        Struct8 dstLcl;
-
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.InitBlock(&dstLcl, 1, 8);
-        }
-
-        fld8 = dstLcl;
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public void CopyBlockHeap8()
-    {
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.CopyBlock(ref _dstData[0], ref _srcData[0], 8);
-        }
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public unsafe void CopyBlockLocalAddr8()
-    {
-        Struct8 srcLcl;
-        Struct8 dstLcl;
-
-        srcLcl = fld8;
-
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.CopyBlock(&dstLcl, &srcLcl, 8);
-        }
-
-        fld8 = dstLcl;
-    }
-
-    [StructLayout(LayoutKind.Explicit, Size=16)]
-    struct Struct16
-    {
-    }
-
-    Struct16 fld16;
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public void InitBlockAllZerosHeap16()
-    {
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.InitBlock(ref _dstData[0], 0, 16);
-        }
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public unsafe void InitBlockAllZerosLocalAddr16()
-    {
-        Struct16 dstLcl;
-
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.InitBlock(&dstLcl, 0, 16);
-        }
-
-        fld16 = dstLcl;
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public void InitBlockAllOnesHeap16()
-    {
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.InitBlock(ref _dstData[0], 255, 16);
-        }
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public unsafe void InitBlockAllOnesLocalAddr16()
-    {
-        Struct16 dstLcl;
-
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.InitBlock(&dstLcl, 1, 16);
-        }
-
-        fld16 = dstLcl;
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public void CopyBlockHeap16()
-    {
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.CopyBlock(ref _dstData[0], ref _srcData[0], 16);
-        }
-    }
-
-    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
-    public unsafe void CopyBlockLocalAddr16()
-    {
-        Struct16 srcLcl;
-        Struct16 dstLcl;
-
-        srcLcl = fld16;
-
-        for (int i = 0; i < OperationsPerInvoke; i++)
-        {
-            Unsafe.CopyBlock(&dstLcl, &srcLcl, 16);
-        }
-
-        fld16 = dstLcl;
-    }
-
-    [StructLayout(LayoutKind.Explicit, Size=24)]
-    struct Struct24
-    {
-    }
-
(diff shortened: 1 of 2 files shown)


echesakovMSFT:
@DrewScoggins @adamsitnik PTAL
cc @dotnet/jit-contrib

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.cs
```diff
@@ -0,0 +1,1244 @@
...
+    }
+
+    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
+    public unsafe void InitBlockAllOnesLocalAddr8()
+    {
+        Struct8 dstLcl;
+
+        for (int i = 0; i < OperationsPerInvoke; i++)
+        {
+            Unsafe.InitBlock(&dstLcl, 1, 8);
```
adamsitnik, line 74: same as above

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.cs
```diff
@@ -0,0 +1,1244 @@
...
+    public void CopyBlockHeap8()
+    {
+        for (int i = 0; i < OperationsPerInvoke; i++)
+        {
+            Unsafe.CopyBlock(ref _dstData[0], ref _srcData[0], 8);
+        }
+    }
+
+    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
+    public unsafe void CopyBlockLocalAddr8()
```
adamsitnik, line 90: Are we executing a different `Unsafe.CopyBlock` code path compared to the `CopyBlockHeap8` above?

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
```diff
@@ -0,0 +1,114 @@
...
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
```
adamsitnik, line 12: nit: the old benchmarks from the xunit-performance era have no namespace for historical reasons and usually we add namespaces for new benchmarks

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
```diff
@@ -0,0 +1,114 @@
...
+{
+    const int OperationsPerInvoke = 1000;
+
+    byte[] _srcData = new byte[1024];
+    byte[] _dstData = new byte[1024];
+
+    [GlobalSetup]
+    public void Setup()
+    {
+        for (int i = 0; i < _srcData.Length; i++)
```
adamsitnik, line 24: These benchmarks might be sensitive to memory alignment so we should allocate the arrays in `[GlobalSetup]` so in the future we can enable memory randomization for it (https://github.com/dotnet/performance/pull/1587). It's relatively new feature that very few ppl are aware of.

```suggestion
    byte[] _srcData;
    byte[] _dstData;

    [GlobalSetup]
    public void Setup()
    {
        _srcData = new byte[1024];
        _dstData = new byte[1024];
        for (int i = 0; i < _srcData.Length; i++)
```

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
```diff
@@ -0,0 +1,114 @@
...
+    Struct<#= byteCount #> fld<#= byteCount #>;
+
+    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
+    public void InitBlockAllZerosHeap<#= byteCount #>()
+    {
+        for (int i = 0; i < OperationsPerInvoke; i++)
+        {
+            Unsafe.InitBlock(ref _dstData[0], 0, <#= byteCount #>);
+        }
+    }
```
adamsitnik, line 49: If we reuse the first `n` bytes over and over we might be testing the perfect scenario. Would it be more realistic to initialize entire array instead of the first `n` bytes?

```suggestion
    [Benchmark(OperationsPerInvoke = 1024 / <#= byteCount #>)]
    public void InitBlockAllZerosHeap<#= byteCount #>()
    {
        for (int i = 0; i < _dstData.Length / <#= byteCount #>; i++)
        {
            Unsafe.InitBlock(ref _dstData[i], 0, <#= byteCount #>);
        }
    }
```

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
```diff
@@ -0,0 +1,114 @@
...
+    {
+        Struct<#= byteCount #> dstLcl;
+
+        for (int i = 0; i < OperationsPerInvoke; i++)
+        {
+            Unsafe.InitBlock(&dstLcl, 0, <#= byteCount #>);
+        }
+
+        fld<#= byteCount #> = dstLcl;
+    }
```
adamsitnik, line 62: Are we executing a different `Unsafe.InitBlock` code path compared to the `InitBlockAllZerosHeap` above?

https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#test-cases

If not, we should have a single `Unsafe.InitBlock`  benchmark for given `byteCount`

```suggestion
```

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
```diff
@@ -0,0 +1,114 @@
...
+    {
+        Struct<#= byteCount #> dstLcl;
+
+        for (int i = 0; i < OperationsPerInvoke; i++)
+        {
+            Unsafe.InitBlock(&dstLcl, 1, <#= byteCount #>);
+        }
+
+        fld<#= byteCount #> = dstLcl;
+    }
```
adamsitnik, line 84: Are we executing a different code path for `value == 0` (benchmark `InitBlockAllZerosHeap` above) and `value == 255` and `value == 1`? If not, there is no need for more benchmarks.

```suggestion
```

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
```diff
@@ -0,0 +1,114 @@
...
+
+    [Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
+    public void CopyBlockHeap<#= byteCount #>()
+    {
+        for (int i = 0; i < OperationsPerInvoke; i++)
+        {
+            Unsafe.CopyBlock(ref _dstData[0], ref _srcData[0], <#= byteCount #>);
+        }
+    }
```
adamsitnik, line 93: Similarly to Init benchmarks, should we copy entire buffer instead of the first `n` bytes?

```suggestion
    [Benchmark(OperationsPerInvoke = 1024 / <#= byteCount #>)]
    public void CopyBlockHeap<#= byteCount #>()
    {
        for (int i = 0; i < _dstData.Length / <#= byteCount #>; i++)
        {
            Unsafe.CopyBlock(ref _dstData[i], ref _srcData[i], <#= byteCount #>);
        }
    }
```

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
```diff
@@ -0,0 +1,114 @@
...
+
+        srcLcl = fld<#= byteCount #>;
+
+        for (int i = 0; i < OperationsPerInvoke; i++)
+        {
+            Unsafe.CopyBlock(&dstLcl, &srcLcl, <#= byteCount #>);
+        }
+
+        fld<#= byteCount #> = dstLcl;
+    }
```
adamsitnik, line 109: Are we testing a different `Unsafe.CopyBlock` code path compared to `CopyBlockHeap`?

```suggestion
```

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
```diff
@@ -0,0 +1,114 @@
...
+<#
+    }
+#>
+}
+<#+ int[] byteCounts = new int[] { 7, 8, 15, 16, 31, 32, 48, 64, 96, 128 }; #>
```
adamsitnik, line 114: If possible please focus on the most common values: https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-not-unit-tests

I am guessing that it could be

```suggestion
<#+ int[] byteCounts = new int[] { 8, 32, 64, 128 }; #>
```

adamsitnik (commented):
Benchmarks overall look good, but I am afraid that they are mostly testing the same code path. We should reduce the numbers of benchmarks. PTAL at my comments and links that I've provided.

@echesakovMSFT thank you!

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
echesakovMSFT: > Are we executing a different Unsafe.InitBlock code path compared to the InitBlockAllZerosHeap above?

At the moment they are the same.

However, I have been working on a JIT optimization on Arm64 to make a use of SIMD instructions for `InitBlock`/`CopyBlock`- see https://github.com/dotnet/runtime/pull/61030.

If/when that PR is merged - the JIT can utilize 16-byte alignment of `sp` and knowledge about distance between `fp` and `sp` and use 16-byte store SIMD operations for `InitBlock*LocalAddr`/`CopyBlock*LocalAddr`. In that case, we will have two different code paths.

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
echesakovMSFT: At the moment, when `value` is `0` on Arm64 we don't require allocating a register for that value in the JIT and use `wzr`/`xzr`. We do need to allocate a register and fill it with non-zero `value`.

However, if the JIT starts using SIMD operations for `InitBlock` (e.g. as in https://github.com/dotnet/runtime/pull/61030), then we will always have to initialize a SIMD register by using `movi Vd.16B, #value`.

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
echesakovMSFT: Same answer as for `InitBlock`: not at the moment, but we might end up with a different code path.

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
echesakovMSFT: Agree, I can try to do this.

src/benchmarks/micro/runtime/StoreBlock/StoreBlock.tt
echesakovMSFT: Thank you for suggestion - I haven't heard about this feature!

echesakovMSFT (commented):
@adamsitnik Thank you for your review! As I mentioned in the replies, at the moment, there is no difference between heap and localAddr code paths for `InitBlock` and `CopyBlock`. However, I am trying to implement an optimization to see if we can utilize 16-byte alignment of `sp` (and known alignment of `fp`) in the JIT on Arm64 and make a use of appropriate store operations. 

For context: according to [Arm Cortex-A76 Software Optimization Guide
](https://developer.arm.com/documentation/swog307215/a/) (and similar guides for other microarchitectures) store operations that cross a 16-byte boundary can incur additional latency on Arm64. Hence, we should use these only when the JIT can proof that the store won't cross such boundary. The JIT won't be able to do this in a general case, but can do this for locals.

If you prefer, I can hold off on the benchmark PR until after https://github.com/dotnet/runtime/pull/61030 is merged.

adamsitnik:
> If you prefer, I can hold off on the benchmark PR until after dotnet/runtime#61030 is merged.

@echesakovMSFT thank you for detailed answers! In such case it would be better to merge the benchmarks before dotnet/runtime#61030 so our Reporting System can start gathering data and when the PR is merged we can see how it affects the performance for all configs.

DrewScoggins:
I am a little confused by the last line of the .tt file. We don't use it when generating the tests, we use the loop for that, so what is it there for? I assume the plan is to instead of walking by eight bytes in the for loop we instead iterate over the array and use the values there to decide on the different sizes for the copying. If not, we should reduce the number of testcases that we generate across the different sizes and only use sizes we believe will give us meaningful differences between them.

echesakovMSFT:
> I am a little confused by the last line of the .tt file. We don't use it when generating the tests, we use the loop for that, so what is it there for? I assume the plan is to instead of walking by eight bytes in the for loop we instead iterate over the array and use the values there to decide on the different sizes for the copying. If not, we should reduce the number of testcases that we generate across the different sizes and only use sizes we believe will give us meaningful differences between them.

You are right, I was using the array at the last line before - but then I wanted to do some extra testing locally and forgot to revert the change. Updated and reduced the number of tests.

echesakovMSFT:
@adamsitnik @DrewScoggins PTAL one more time - I believe I addressed all the suggestions.

adamsitnik (approved):
LGTM, thank you @echesakovMSFT !

PR #2169, opened 2021-12-07
add Process.ProcessName benchmark 

adamsitnik:
for https://github.com/dotnet/runtime/pull/59672

### DIFF
--- a/src/benchmarks/micro/libraries/System.Diagnostics/Perf_Process.cs
+++ b/src/benchmarks/micro/libraries/System.Diagnostics/Perf_Process.cs
@@ -15,7 +15,14 @@ namespace System.Diagnostics
         
         [Benchmark]
         public void GetCurrentProcess() => Process.GetCurrentProcess().Dispose();
-        
+
+        [Benchmark]
+        public string GetCurrentProcessName()
+        {
+            using var process = Process.GetCurrentProcess();
+            return process.ProcessName;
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.Diagnostics/Perf_Process.cs
```diff
@@ -15,7 +15,14 @@ public class Perf_Process
         
         [Benchmark]
         public void GetCurrentProcess() => Process.GetCurrentProcess().Dispose();
-        
+
+        [Benchmark]
+        public string GetCurrentProcessName()
+        {
+            using var process = Process.GetCurrentProcess();
```
danmoseley, line 22: should we be including the cost of Dispose()? it does do some work but perhaps it's not significant.

although I see the test above is doing that too.

danmoseley (approved):

src/benchmarks/micro/libraries/System.Diagnostics/Perf_Process.cs
adamsitnik: > should we be including the cost of Dispose()?

I am afraid that we don't have an alternative (due to BDN design). The good thing is that this benchmark is dominated by the sys-call performed by `Process.ProcessName` so the cost of `Dispose` should be negligible.

PR #2172, opened 2021-12-09
Rename some Perf_String arguments for better reporting

nietras:
cc: @adamsitnik


adamsitnik (changes requested):
I am sorry but I can't accept the arguments rename. The reason for that is that we are using the arg names in benchmark ID which is used by our Reporting System: https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable

If we change namespace, type|method|argument name or argument value we loose the track ;(

Fell free to create a new type with new `string` benchmarks. Once we run better benchmarks for a while and gather enough data in the Reporting System, we can remove the old benchmark(s)

nietras:
Alright.

PR #2217, opened 2022-01-20
Add Categories.Regex

stephentoub:
@DrewScoggins, per our conversation yesterday, is this the right way to do this?

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Industry.cs
+++ b/src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Industry.cs
@@ -40,7 +40,7 @@ namespace System.Text.RegularExpressions.Tests
     }
 
     /// <summary>Performance tests adapted from https://github.com/mariomka/regex-benchmark</summary>
-    [BenchmarkCategory(Categories.Libraries, Categories.NoWASM)]
+    [BenchmarkCategory(Categories.Libraries, Categories.Regex, Categories.NoWASM)]
     public class Perf_Regex_Industry_Mariomkas
     {
         [Params(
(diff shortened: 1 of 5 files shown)


src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Industry.cs
```diff
@@ -40,7 +40,7 @@ public static int Count(Regex r, string input)
     }
 
     /// <summary>Performance tests adapted from https://github.com/mariomka/regex-benchmark</summary>
-    [BenchmarkCategory(Categories.Libraries, Categories.NoWASM)]
+    [BenchmarkCategory(Categories.Libraries, Categories.Regex, Categories.NoWASM)]
```
DrewScoggins, line 43: You will need the Regex category for all of the test classes, unless you don't want them included

DrewScoggins:
Only the one comment, other than that LGTM.

src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Industry.cs
stephentoub: Oops, thanks.

stephentoub:
FYI, @danmoseley, @joperezr.

stephentoub:
@DrewScoggins, fixed, thanks.

danmoseley:
What cmd line do you use to run just these and how long does it take?

stephentoub:
> What cmd line do you use to run just these and how long does it take?

I'm not sure.  That wasn't actually my goal here.  Rather, I asked Drew how we could get a daily/weekly/etc. report generated that shows the history of all regex tests in an easily browsable manner, and this was his recommendation.

danmoseley (approved):

stephentoub:
I don't have permission to merge, if someone who does can do so.

DrewScoggins:
Done.

Also @danmoseley to answer your question you just need to use the `--anyCategories` command line argument and specify Regex as the category.

stephentoub:
Thanks, @DrewScoggins.

PR #2218, opened 2022-01-20
Remove the dmlib folder

LoopedBard3:
Removing the dmlib folder. Some configuration issues were found, but there is no obvious place that it is used and it seems stale, so we are going to remove it unless a reason to keep it pops up. The dotnet runtime perf pipeline ran as expected without it included so we should be clear to remove it is the perf run goes as expected.


LoopedBard3:
This will be good to merge once checks have passed (sdk_scenarios may not pass due to an ongoing issue).

DrewScoggins:
@adamsitnik who did the port of this and maybe knows what it is.

adamsitnik:
> @adamsitnik who did the port of this and maybe knows what it is.

To be honest I don't remember what is it and why I've ported it.

It's most probably some old app that we have used to profile a specific issue for one of the Azure teams (the code says it's "Azue Data Movement Library Benchmark")

Anyway deleting it is a good idea. 👍

LoopedBard3:
SDK_Scenario failures are not specific to this change and the net461 failures are due to an issue with dotnet-install, merging.

Issue #2075, opened 2021-10-14
Enable Maui Test Scenarios


LoopedBard3:
Enable Maui Android Size on Disk and Startup scenarios

- [x] Enable Size on Disk tests
- [x] Setup Size on Disk PowerBI
- [x] Enable Startup tests
- [x] Setup Startup PowerBI
- [x] Start testing iOS pipeline (need macos vm update)

PR #2219, opened 2022-01-21
Update Android Scenario Startup Flow

LoopedBard3:
Update the Android Scenario Startup Flow to use adb instead of test runners for more accurate startup times and the ability to test start up to first drawn frame on any apk. This effort is part of #2075 as it more easily enables the ability for Maui app testing.

### DIFF
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -318,22 +318,22 @@ ex: C:\repos\performance;C:\repos\runtime
             adb = RunCommand(cmdline, verbose=True)
             adb.run()
 
-            #cmdline = xharnesscommand() + [
-            #    self.devicetype,
-            #    'install',
-            #    '--app', self.packagepath,
-            #    '--package-name',
-            #    self.packagename,
-            #    '-o',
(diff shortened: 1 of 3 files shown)


src/scenarios/shared/runner.py
```diff
@@ -328,48 +338,129 @@ def run(self):
...
+                'am',
+                'start-activity',
+                '-W',
+                '-n',
+                activityname
+            ]
+            testRun = RunCommand(startAppCmd, verbose=True)
+            testRun.run()
+            testRunStats = re.findall(runSplitRegex, testRun.stdout) 
+            getLogger().info(f"Test run activity: {testRunStats[3]}")
```
DrewScoggins, line 397: What is testRunStats[3]? Can you break down what the different groups of this regex are.

LoopedBard3: Added a break down the data split at the top of this switch section (line 310)

DrewScoggins (approved):
LGTM

LoopedBard3:
Errors are unrelated, merging

PR #2223, opened 2022-01-26
Fix InnerLoopMsBuild Evaluate/Stop not found

LoopedBard3:
PR to fix InnerLoopMsBuild Evaluate/Stop not found issue. Initially just testing with public building.


LoopedBard3:
Looks like this is not an issue on our side, closing PR. https://github.com/dotnet/msbuild/pull/7277

PR #2233, opened 2022-02-03
Dump DOTNET_ env vars

danmoseley:

PR #2239, opened 2022-02-04
Hide env vars that are impl details

danmoseley:
Feedback on https://github.com/dotnet/performance/pull/2233


danmoseley:
performance-ci (windows RS4 x64 micro_net461 Open LTS)  results can be ignored. it's a dead queue

PR #2244, opened 2022-02-07
Fix Append_Span benchmark

adamsitnik:
* add missing setup method so the benchmark does not measure appending an empty string
* rename the benchmark so ReportingSystem does not detect the fix as a regression

related to https://github.com/dotnet/runtime/issues/64751

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text/Perf.StringBuilder.cs
+++ b/src/benchmarks/micro/libraries/System.Text/Perf.StringBuilder.cs
@@ -213,8 +213,11 @@ namespace System.Text.Tests
         }
 
 #if !NETFRAMEWORK
+        [GlobalSetup(Target = nameof(Append_NonEmptySpan))]
+        public void Setup_Append_NonEmptySpan() => _string100 = new string('a', 100);
+
         [Benchmark]
-        public StringBuilder Append_Span()
+        public StringBuilder Append_NonEmptySpan()
         {
             ReadOnlySpan<char> span = _string100.AsSpan();
             StringBuilder builder = new StringBuilder();


src/benchmarks/micro/libraries/System.Text/Perf.StringBuilder.cs
```diff
@@ -213,8 +213,11 @@ public StringBuilder Append_Memory()
         }
 
 #if !NETFRAMEWORK
+        [GlobalSetup(Target = nameof(Append_NonEmptySpan))]
+        public void Setup_Append_NonEmptySpan() => _string100 = new string('a', 100);
+
         [Benchmark]
-        public StringBuilder Append_Span()
+        public StringBuilder Append_NonEmptySpan()
```
stephentoub, line 220: Do we really need to rename it?  Every time there's a bug in a test we need to rename the test, and then we're left with some tests that just don't have any more data after their fixed data?

adamsitnik: It depends on how the reported time changes. In this case, if we keep the old name the automation is going to recognize it as a regression. We would need to have a possibility to hint the automation that "it's not a regression, it's a fix" but looking at how rarely it happens I don't think it's worth the investment.

https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable

cc @DrewScoggins

stephentoub: > We would need to have a possibility to hint the automation that "it's not a regression, it's a fix" but looking at how rarely it happens I don't think it's worth the investment.

Or just ignore the regression.  Just as we do when a regression issue is filed, we look at it, and we say "we don't care about that one".

kunalspathak: It is difficult for us (the triage team) to memorize all the issues that fall under this category and ignore the regression. In the past, we have suggested to @DrewScoggins to add an ability to annotate a benchmark from the historical data page or something. That way, we can see the annotation in future and take required actions.

stephentoub: Does it need to fall on you to remember and ignore it?  The issue can be filed if needed and then the team that goes to investigate can say "yeah, the benchmark changed".  Is that significantly different from a product change that's by design known by the team making the change to impact a benchmark?

kunalspathak: the team that goes to investigate - do you mean the triage team or the one that is investigating the regression?

stephentoub: > the team that goes to investigate - do you mean the triage team or the one that is investigating the regression?

The latter.

kunalspathak: > e.g. not you, me :)

in that case, I don't have any problem :)

DrewScoggins: I think in general I am less worried about us finding the regression once, and more about people looking at the graph, and seeing a big jump and not remembering why it happened. Like @kunalspathak said there is work on the backlog to add annotations for stuff like this, but we don't really have an ETA for when we would have it. 

All that to say, I don't think it will be that big a deal as we don't regularly look at every test result and it's full history.

PR #2248, opened 2022-02-08
TFM-specific fixes

adamsitnik:
So far, we were using hardcoded versions of all `Microsoft.Extension*` packages and few `System*` out-of-band packages that are not included in NetCoreApp.

It was compiling fine, but when running without CoreRun (using dotnet shipped with the SDK) we were using the same library for all TFMs.

For example the following two commands were benchmarking same library version using different runtimes:

```cmd
py .\scripts\benchmarks_ci.py -f netcoreapp3.1 --filter *XmlConfigurationProviderBenchmarks*
py .\scripts\benchmarks_ci.py -f net5.0 --filter *XmlConfigurationProviderBenchmarks*
```

In this PR I am proposing the following: use versions that were shipped for given TFMs. So when we are benchmarking `net5.0` vs `net6.0` using the SDK, we can see the actual difference.

It's not going to affect the dotnet/runtime runs that are using CoreRun as the BDN's CoreRun toolchain always prefers newer version of the files:

https://github.com/dotnet/BenchmarkDotNet/blob/569f4f5bb0395c0b730d88588e58172eacd4edf6/src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunPublisher.cs#L34-L37

I've also fixed few issues related to `System.Drawing` benchmarks caused by https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only

I've validated these changes locally and it works for `net461` (it took 2.5h just to run every benchmark once), `netcoreapp3.1`, `net5.0` and `net6.0`. The CI validated `net7.0`


src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -16,29 +16,50 @@
...
+    <PackageReference Include="Microsoft.Extensions.Logging.EventSource" Version="$(ExtensionsVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Primitives" Version="$(ExtensionsVersion)" />
     <PackageReference Include="protobuf-net" Version="3.0.73" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="5.0.0-rc.1.20417.14" />
-    <PackageReference Include="System.Drawing.Common" Version="5.0.0-rc.1.20417.14" />
-    <PackageReference Include="System.Formats.Cbor" Version="5.0.0-rc.1.20417.14" Condition="'$(TargetFramework)' == 'net5.0' or '$(TargetFramework)' == 'net6.0' or '$(TargetFramework)' == 'net7.0'" />
-    <PackageReference Include="System.IO.Pipelines" Version="5.0.0-rc.1.20417.14" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemVersion)" />
+    <PackageReference Include="System.Drawing.Common" Version="$(SystemVersion)" />
+    <PackageReference Include="System.Formats.Cbor" Version="$(SystemVersion)" Condition="'$(TargetFramework)' == 'net5.0' or '$(TargetFramework)' == 'net6.0' or '$(TargetFramework)' == 'net7.0'" />
```
danmoseley, line 56: These will need updates when we move to net8.0. Would it be better/possible to use this function instead: `Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net5.0'))">`

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -16,29 +16,50 @@
...
+        </PropertyGroup>
+    </When>
+    <When Condition="'$(TargetFramework)' == 'net5.0'">
+        <PropertyGroup>
+            <ExtensionsVersion>5.0.0</ExtensionsVersion>
+            <SystemVersion>5.0.0</SystemVersion>
+        </PropertyGroup>
+    </When>
+    <Otherwise>
+        <!-- when comparing against Full .NET Framework we are usually interested in CLR differences, so net461 belongs to this block -->
```
danmoseley, line 34: Would it be better to explicitly condition this block on TargetFramework=net461 (or whatever)?

If I understand right, these will be ignored for .net6.0 or later because they're in box at that point -- so that would be more explicit.
Or if they won't be ignored, then it would be also good to add the condition so things break when we test a future TFM that needs us to update this version number here?

src/benchmarks/micro/MicroBenchmarks.csproj
adamsitnik: @danmoseley Good idea, I was not aware of `IsTargetFrameworkCompatible` existence. Is it supported by older SDKs? (3.1 for example?)

src/benchmarks/micro/MicroBenchmarks.csproj
adamsitnik: > If I understand right, these will be ignored for .net6.0 or later because they're in box at that point

Do you mean that `Microsoft.Extensions.Caching.Memory` and others are now in-box? That is new to me. And it makes my life easier as it means that we won't ever need to update these versions?

src/benchmarks/micro/MicroBenchmarks.csproj
danmoseley: It looks like it would be 6.0 onwards , judging by the date :|
https://github.com/dotnet/msbuild/commit/51e8dd25010de68b07047ac49254e29d2af41241#

PR #1719, opened 2021-03-10
Add benchmarks for Microsoft.Extensions.Configuration.Xml

amoerie:
Hi

While creating [a pull request in the dotnet runtime repo](https://github.com/dotnet/runtime/pull/44608), I created some benchmarks to ensure my changes did not introduce a regression.

By request of @safern, I am adding these benchmarks here. 

I'm still struggling a bit with properly running them, and I'm short on time, so please give me a while to get this PR into a more polished state.

PR #2249, opened 2022-02-08
Use Regex.Count in regex perf tests

stephentoub:
cc: @joperezr, @danmoseley

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Industry.cs
+++ b/src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Industry.cs
@@ -26,6 +26,9 @@ namespace System.Text.RegularExpressions.Tests
 
         public static int Count(Regex r, string input)
         {
+#if NET7_0_OR_GREATER
+            return r.Count(input);
+#else
             int count = 0;
             
             Match m = r.Match(input);
@@ -36,6 +39,7 @@ namespace System.Text.RegularExpressions.Tests
             }
 
             return count;
+#endif
         }
     }
 


danmoseley (approved):
hopefully should get a nice twitch in the graph..

stephentoub:
> hopefully should get a nice twitch in the graph..

Yes.  I don't think we graph allocation, but it should drop allocation on many of these tests from "a lot" to "zero".

stephentoub:
@DrewScoggins, please review and merge when you have a moment (I don't have write access).

danmoseley:
@DrewScoggins any reason why we shouldn't give write access to all the folks that have write access to dotnet/runtime?
@terrajobst can help tell us the right way to do this.

adamsitnik:
> any reason why we shouldn't give write access to all the folks that have write access to dotnet/runtime?

When I was porting the benchmarks from coreclr and corefx repos, I've realized that they were more like unit tests rather than benchmarks. Example: focusing on edge cases, having hundreds of test cases etc ([more](https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#mindset)). 

To address these issues I've added few validators that simply produce errors ([example](https://github.com/dotnet/performance/blob/main/src/harness/BenchmarkDotNet.Extensions/TooManyTestCasesValidator.cs)) when things are obviously wrong and published the [Microbenchmark Design Guidelines docs](https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md) to share all of the know-how. 

Together with other members of the .NET Perf Team we restricted the write access only to Perf Team members and Perf Champions so we could ensure that all the benchmarks that we merge follow the best practices.

Are we ready to lift the restrictions? Many of the dotnet/performance contributors contribute great benchmarks and they definitely don't need our reviews anymore, however some still do.

Example: https://github.com/dotnet/performance/pull/1719 got three approvals despite violating multiple rules:

1. the build was red as one of the benchmarks was throwing
2. it was using `[IterationSetup]` (https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#iterationsetup)
3. the namespace was incorrect (https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable)
4. it wasn't using any of the parametrization features built into BDN but rather having separate benchmark method per test case

For 2 & 3 we could most probably implement a validator and I am sure that nobody would merge a failing PR, however I am not sure about simple best practices like 4.

I am aware of the fact that the write access is often a bottleneck for the perf repo contributors and in a long term it should definitely change.

Since we have recently invested more in detecting the regressions (@DrewScoggins automation, manual perf runs) perhaps we could book  two days in this release cycle to:
* have a talk about Microbenchmark Design Guidelines
* have every member of the .NET libraries Team look at their areas and implement missing benchmarks (we can detect regressions only for areas that have benchmarks, some have none)
* review the benchmarks and grant the write access to everyone who participated?

@danmoseley @jeffhandley what do you think?

adamsitnik:
> With respect to access -- perhaps near term, we could add a few more people like @stephentoub

👍 

> Seems like we would first need a way to understand where the specific interesting gaps are. We don't want low value tests. Couple ideas: where customers report regressions that we missed; where we have taken changes for performance improvements, which aren't covered. Perhaps we should request that any benchmarks exist before we merge performance improvements.

We already do some of these things and we should definitely always do that. But there are areas with no benchmarks where we introduce non-perf changes and sometimes accidentally regress the performance. Usually we find out about that after new .NET version is released, when the customers are reporting the regression. That is what I would like to avoid.

Perhaps we could use Azure Profiler data and see which components are used frequently and ensure that they are all covered?

DrewScoggins:
I am fully onboard with adding @stephentoub, and want to reiterate what Adam was saying. I think that getting more people who have a really solid understanding of the design guidelines for performance tests, and as a result having write access in this repo is a good thing. I also wonder if we could use that time and effort to look over our existing test assets and see if there are places that we could improve the test quality. Better data in will mean better data out.

All that to say, I have added Stephen to the dotnet-perf group, so he should have write access now.

stephentoub:
Thanks for the write access.  And I'm fully on board with the idea of helping devs to better understand how to write good perf tests, but FWIW, this discussion suggests we hold the tests in dotnet/performance to a higher standard than the official .NET product source living in dotnet/runtime.  That, to me, doesn't make sense, and basically tells folks they're not trusted.

danmoseley:
We could create a PR template with a little comment like

[ ] If you're adding or modifying benchmarks, have you verified the changes comply with https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md ? 

And then if folks make a mistake - we teach/learn.

danmoseley:
created https://github.com/dotnet/performance/pull/2258

danmoseley:
linking https://github.com/dotnet/runtime/pull/64289

EgorBo:
Improvements on arm64: https://github.com/dotnet/perf-autofiling-issues/issues/3572

danmoseley:
@DrewScoggins @adamsitnik  , with the new reminder and link added to the PR template, are you comfortable with giving write access to those in dotnet/runtime that have write access, as a trial?

DrewScoggins:
Yeah, I am fine with that. What group should I add to the permission list to get this done?

danmoseley:
We will have to ask @terrajobst . There is also the question about folks on, eg., the aspnetcore team. But one thing at a time perhaps.

adamsitnik:
> with the new reminder and link added to the PR template, are you comfortable with giving write access to those in dotnet/runtime that have write access, as a trial?

image

Let's give it a try ;)

PR #2258, opened 2022-02-11
Adding PR template to help ensure microbenchmarks are maintained correctly

danmoseley:
This appears in the form when any PR is created; it's commented, so if they don't delete it, it won't show up in the final PR.

the readme.md next to the microbenchmarks is almost all about how to run them: I moved a small part that was actually guidelines into the guidelines doc, so it's together.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,13 @@
+<!-- Thank you for submitting a pull request to our repo.
+
+     It's critical that our microbenchmarks remain efficient, reliable and accurate.
+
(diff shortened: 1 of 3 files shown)


DrewScoggins:
Is it standard to have the PR guidance to be in an XML comment block?

danmoseley:
> Is it standard to have the PR guidance to be in an XML comment block?

I don't know of any "standard". Personally, I find it super noisy to work in repos where issues or PR's have lots of template content in them (`[ x ] I have verified this builds successfully` etc). In dotnet/runtime, we put all the guidance in comments. Even if they don't remember to delete them, they don't appear in the result.

let's see how it goes.

adamsitnik (commented):
It looks great, thank you @danmoseley !

PR #2259, opened 2022-02-11
Update Maui/Mobile pre.py's to simplify running different apps

LoopedBard3:
Update Maui/Mobile pre.py's to simplify running different apps. 

Removes the default app name for each folder and instead takes in the app folder name as an argument to run pre run functionality against. This should make it more straightforward to add and remove tests as their names will not need to be added here, instead only being set in the android_scenarios.proj file.


LoopedBard3:
Yes, getting up it ATM.

LoopedBard3:
Runtime correlating PR is up https://github.com/dotnet/runtime/pull/65231

DrewScoggins (approved):
LGTM

PR #663, opened 2019-07-16
redesign Clear benchmarks into CreateAddAndClear benchmarks

adamsitnik:
it's hard to benchmark a method that requires a clean state, takes nanoseconds to run and much more time to setup. Good example is `List<ValueType>.Clear` which just sets the list index to 0 and is forced to inline:

https://github.com/dotnet/coreclr/blob/8fa6bd4aaa949bbfd3dd263e7327996ab8fa7966/src/System.Private.CoreLib/shared/System/Collections/Generic/List.cs#L295-L311

I think that the best thing we can do with this problem right now is to redesign `Clear` benchmarks into `CreateAddAndClear`. If `Clear` is going to regress, the benchmark is still going to show it. 

@billwert if you like this idea I am going to redesign the `Remove` benchmarks in the same way

PR #2268, opened 2022-02-18
Add some additional micro benchmark coverage for immutable collections.

madelson:
See https://github.com/dotnet/runtime/issues/14477#issuecomment-1037693750 for context.

Adding these additional benchmarks will obviously have runtime cost, so I wanted to start with just adding immutable collections benchmarks to a few files to get some feedback / a sense of whether or not this is the intended direction.


danmoseley:
@adamsitnik maybe can give feedback.

src/benchmarks/micro/libraries/System.Collections/AddRemove/AddRemoveSteadyState.cs
```diff
@@ -73,5 +77,23 @@ public void Stack()
             T item = _stack.Pop();
             _stack.Push(item);
         }
+
+        [GlobalSetup(Target = nameof(ImmutableQueue))]
+        public void SetupImmutableQueue() => _immutableQueue = ValuesGenerator.ArrayOfUniqueValues<T>(Count).Aggregate(ImmutableQueue<T>.Empty, (q, v) => q.Enqueue(v));
```
adamsitnik, line 82: there is no need to use LINQ here

```suggestion
        public void SetupImmutableQueue() => _immutableQueue = Immutable.ImmutableQueue.CreateRange<T>(ValuesGenerator.ArrayOfUniqueValues<T>(Count));
```

src/benchmarks/micro/libraries/System.Collections/AddRemove/AddRemoveSteadyState.cs
```diff
@@ -73,5 +77,23 @@ public void Stack()
...
+
+        [Benchmark]
+        public void ImmutableQueue()
+        {
+            _immutableQueue = _immutableQueue.Dequeue(out T item).Enqueue(item);
+        }
+
+        [GlobalSetup(Target = nameof(ImmutableStack))]
+        public void SetupImmutableStack() => _immutableStack = ValuesGenerator.ArrayOfUniqueValues<T>(Count).Aggregate(ImmutableStack<T>.Empty, (q, v) => q.Push(v));
```
adamsitnik, line 91:
```suggestion
        public void SetupImmutableStack() => _immutableStack = Immutable.ImmutableStack.CreateRange<T>(ValuesGenerator.ArrayOfUniqueValues<T>(Count));
```

src/benchmarks/micro/libraries/System.Collections/Dictionary/DictionarySequentialKeys.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
+using System.Collections.Immutable;
```
adamsitnik, line 6: the existing benchmarks that you can find in this file has been added in https://github.com/dotnet/performance/pull/938 to validate https://github.com/dotnet/coreclr/pull/27299 which required a very specific benchmark coverage of the `Dictionary<TKey, Value>` APIs. We don't need such detailed coverage for other dictionary types as they already have good overall coverage. Please revert all the changes in this file.

src/benchmarks/micro/libraries/System.Collections/AddRemove/AddRemoveSteadyState.cs
```diff
@@ -4,6 +4,8 @@
 
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
```
adamsitnik, line 8:
```suggestion
```

src/benchmarks/micro/libraries/System.Collections/Add/AddGivenSize.cs
```diff
@@ -133,5 +134,71 @@ public ObservableCollection<T> ObservableCollection()
             }
             return collection;
         }
+
+        [Benchmark]
+        public ImmutableArray<T> ImmutableArray()
```
adamsitnik, line 139: These benchmarks overall look good, but we should move them out of this class. This class is called Add**GivenSize** because it tests adding new items to collections that support specifying initial capacity. Like `new List<T>(Size)`. We had benchmarks for `AddDefaultSize` (added in https://github.com/dotnet/performance/pull/92), but we have later decided to redesign them to `CreateAddAndClear` (https://github.com/dotnet/performance/pull/663) so they can also cover `Clear`. Please move the new benchmarks from this type to `CreateAddAndClear` and add a `Clear` method invocation to them.

adamsitnik (commented):
@madelson big thanks for your contribution! Please take a look at my comments.

madelson:
@adamsitnik quickly reviewing the remaining benchmarks, I see the following places where we could add additional immutable collections coverage. **Are any of these of interest?**

TryAddDefaultSize
* SortedDictionary
* SortedList
* ImmutableDictionary
* ImmutableSortedDictionary

AddRemoveFromDifferentThreads/SameThread
* ImmutableQueue + ImmutableInterlocked
* ImmutableStack + ImmutableInterlocked

Perf_Dictionary
* SortedDictionary
* SortedList
* ImmutableDictionary
* ImmutableSortedDictionary

IndexerSet
* ImmutableArray
* ImmutableList
* ImmutableDictionary
* ImmutableSortedDictionary

Perf_SortedSet
* ImmutableSortedSet

CopyTo
* ImmutableList

CreateAddAndRemove
* All immutable collections

Sort
* ImmutableArray
* ImmutableList

src/benchmarks/micro/libraries/System.Collections/Add/AddGivenSize.cs
```diff
@@ -4,6 +4,7 @@
 
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Immutable;
```
adamsitnik, line 7: nit: this is not needed anymore

```suggestion
```

adamsitnik (approved):
LGTM, thank you @madelson !

adamsitnik:
> TryAddDefaultSize

It's a good idea to add these (I was not aware of the existence of the `TryAdd` extension methods before you have asked this question).

> AddRemoveFromDifferentThreads/SameThread

The original benchmarks are quite noisy ([source](https://pvscmdupload.blob.core.windows.net/reports/allTestHistory%2frefs%2fheads%2fmain_x64_Windows%2010.0.18362%2fSystem.Collections.Concurrent.AddRemoveFromDifferentThreads(Int32).ConcurrentQueue(Size%3a%202000000).html)):

image

and I am not sure whether we should promote immutable collections for multi-threaded mutable use cases (cc @stephentoub). I would rather say no.

> Perf_Dictionary

The `Perf_Dictionary` type contains two benchmarks: cloning and `ContainsValue`. You could two dedicated types for that and add benchmarks for each collection that supports it and even remove `Perf_Dictionary`.

> IndexerSet

the immutable collections are throwing for mutable methods:

https://github.com/dotnet/runtime/blob/0a684281f661f5520ad7f358a5056a2eed5b50ba/src/libraries/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.cs#L30

and in general we are not benchmarking edge case scenarios like this (https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-not-unit-tests)

> Perf_SortedSet

It's a good idea to create similar type for ImmutableSortedSet

> CopyTo
> CreateAddAndRemove
> Sort

👍 

@madelson I am going to merge this PR, please send more benchmarks in new PRs. Thank you for your contribution!

madelson:
Thanks @adamsitnik !

> the immutable collections are throwing for mutable methods:

Sorry I should have clarified. The useful equivalent of indexer set for immutable collections is the `SetItem` method (`e. g. list = list.SetItem(2, "a")`. I think it would be reasonable to make that part of the IndexerSet benchmark but it could also be its own thing. Thoughts?

> I am not sure whether we should promote immutable collections for multi-threaded mutable use cases

There is the `ImmutableInterlocked` class which directly supports this use-case if we did want to support that. Happy to skip this one if you don't think it is important, though.

PR #2271, opened 2022-02-22
Added --disable-animations option to device startup testing.

LoopedBard3:
Added --disable-animations option to device startup testing. This allows for the turning off of device animations when running the startup tests on android devices. This has run successfully with the latest runtime with or without the animation flag.

### DIFF
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -62,6 +62,7 @@ class Runner:
         parseonlyparser.add_argument('--package-path', help='Location of test application', dest='packagepath')
         parseonlyparser.add_argument('--package-name', help='Classname of application', dest='packagename')
         parseonlyparser.add_argument('--startup-iterations', help='Startups to run (1+)', type=int, default=5, dest='startupiterations')
+        parseonlyparser.add_argument('--disable-animations', help='Disable Android device animations', action='store_true', dest='animationsdisabled')
         self.add_common_arguments(parseonlyparser)
 
         # inner loop command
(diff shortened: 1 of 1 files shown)


DrewScoggins:
If I remember right @jonathanpeppers mentioned that we need to do a reboot after making these changes for them to take effect. It doesn't look like we do that in the current change.

LoopedBard3:
I will double check locally to see if setting the value in the same way as disabling them reenables the animations.

src/scenarios/shared/runner.py
```diff
@@ -339,6 +341,69 @@ def run(self):
...
+            cmdline = [
+                adb.stdout.strip(),
+                'shell', 'settings', 'get', 'global', 'transition_animation_scale'
+            ]
+            RunCommand(cmdline, verbose=True).run()
+            cmdline = [
+                adb.stdout.strip(),
+                'shell', 'settings', 'get', 'global', 'animator_duration_scale'
+            ]
+            RunCommand(cmdline, verbose=True).run()
```
DrewScoggins, line 404: It doesn't look like we do any checking here of the values. Is the goal to just have them printed out to the console so we can see? I feel like it would be better to fail the tests if we are unable to get them configured properly.

LoopedBard3: That is a better idea, let me see what I can do for getting that implemented.

jonathanpeppers: I had to reboot in my case, because I didn't know what the original values were.

If you get can get the values, and restore them at the end, that would also work.

jonathanpeppers: It's also probably worthwhile to turn these off for other startup tests. I don't see how you'd notice a 10ms regression in the Mono runtime, if it has the splash animation on.

PR #2274, opened 2022-02-23
Add MauiVersion, perf side

LoopedBard3:
Add necessary logic for handling MauiVersion being passed in when running ci_setup.py.

### DIFF
--- a/scripts/ci_setup.py
+++ b/scripts/ci_setup.py
@@ -185,7 +185,7 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         default='',
         required=False,
         type=str,
-        help='Version of Maui to pass'
+        help='Version of Maui used to build app packages'
     )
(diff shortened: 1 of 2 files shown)


scripts/ci_setup.py
```diff
@@ -179,6 +179,14 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         help='Configurations used in the build in key=value format'
     )
 
+    parser.add_argument(
+        '--maui-version',
+        dest='maui_version',
+        default='',
+        required=False,
+        type=str,
+        help='Version of Maui to pass'
```
DrewScoggins, line 188: Can you update this to be slightly more descriptive? Something like Version of Maui used to build app packages.

DrewScoggins (approved):

DrewScoggins (approved):

PR #2279, opened 2022-02-25
[MAUI]Update Device Startup Scenario Naming

LoopedBard3:
Changed startup scenario name to match normal scenario naming scheme of using the scenario name provided in the scenarios.proj file. Update


DrewScoggins (approved):
LGTM

PR #2283, opened 2022-03-01
Change NoPGO runs to OSR runs for a short time

DrewScoggins:
This changes the NoPGO runs, which our of low importance, over to OSR runs so that we can get a weeks worth of data before this tries to go in for preview 3.


LoopedBard3 (approved):
LGTM

AndyAyersMS (approved):
Thanks!

PR #2295, opened 2022-03-08
[MAUI] Enable iOS ipa file size measurement

LoopedBard3:
Adds support for measuring ipa files to the Maui and iOS pre.py files. This also includes the addition of passing in the name of the folder/file you want to use in testing. This is a sibling commit to https://github.com/dotnet/runtime/pull/66358.

### DIFF
--- a/src/scenarios/helloios/pre.py
+++ b/src/scenarios/helloios/pre.py
@@ -23,6 +23,8 @@ args = parser.parse_args()
 
 name = args.name
 namezip = '%s.zip' % (name)
+if not os.path.exists(PUBDIR):
+    os.mkdir(PUBDIR)
 if not os.path.exists(name):
     getLogger().error('Cannot find %s' % (name))
     exit(-1)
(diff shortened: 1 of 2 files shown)


src/scenarios/helloios/pre.py
```diff
@@ -1,10 +1,53 @@
...
+        zip.extractall(os.path.join('.', PUBDIR))
+
+    baseDir = os.path.join(PUBDIR, 'Payload')
+    appFolder = os.listdir(baseDir) # Should only be one
+    if len(appFolder) > 1:
+        getLogger.error("More than one app folder found in payload! Should only be one.")
+        exit(-1)
+
+    appFolderPath = os.path.join(baseDir, appFolder[0])
+    for subfile in os.listdir(appFolderPath):
```
DrewScoggins, line 45: Why are we moving the extracted files after unzipping rather than just measuring the size where we unpack them?

LoopedBard3: The main reason was because the structure of the unzipped ipa is folder.app/{interesting files}. I can play around with it a little bit more to and see if I can get the dirs argument to set the directory to parse to work instead though.

LoopedBard3: I double checked some assumptions I made about the ipa file structure and they were incorrect so I removed the moving section to match measuring the whole zip file.

DrewScoggins (approved):
LGTM

LoopedBard3:
The failing sections are failing in main, merging.

PR #2306, opened 2022-03-11
Runtime issue 14477 more immutable benchmarks

madelson:
See https://github.com/dotnet/runtime/issues/14477#issuecomment-1037693750 and https://github.com/dotnet/performance/pull/2268#issuecomment-1062799822 for context.

### DIFF
--- a/src/benchmarks/micro/libraries/System.Collections/Add/TryAddDefaultSize.cs
+++ b/src/benchmarks/micro/libraries/System.Collections/Add/TryAddDefaultSize.cs
@@ -33,6 +33,26 @@ namespace System.Collections
                 collection.TryAdd(uniqueValues[i], uniqueValues[i]);
             return collection;
         }
+
+        [Benchmark]
+        public SortedDictionary<T, T> SortedDictionary()
+        {
+            var collection = new SortedDictionary<T, T>();
+            var uniqueValues = _uniqueValues;
+            for (int i = 0; i < uniqueValues.Length; i++)
+                collection.TryAdd(uniqueValues[i], uniqueValues[i]);
+            return collection;
+        }
+
+        [Benchmark]
+        public SortedList<T, T> SortedList()
+        {
+            var collection = new SortedList<T, T>();
+            var uniqueValues = _uniqueValues;
+            for (int i = 0; i < uniqueValues.Length; i++)
+                collection.TryAdd(uniqueValues[i], uniqueValues[i]);
+            return collection;
+        }
 #endif
 
         [Benchmark]
--- a/src/benchmarks/micro/libraries/System.Collections/CopyTo.cs
+++ b/src/benchmarks/micro/libraries/System.Collections/CopyTo.cs
@@ -21,6 +21,7 @@ namespace System.Collections
         private T[] _array;
         private List<T> _list;
         private ImmutableArray<T> _immutablearray;
+        private ImmutableList<T> _immutableList;
         private T[] _destination;
 
         [GlobalSetup(Targets = new[] { nameof(Array), nameof(Span), nameof(ReadOnlySpan), nameof(Memory), nameof(ReadOnlyMemory) })]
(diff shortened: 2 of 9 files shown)


src/benchmarks/micro/libraries/System.Collections/Add/TryAddDefaultSize.cs
```diff
@@ -33,6 +33,26 @@ public class TryAddDefaultSize<T>
...
+
+        [Benchmark]
+        public SortedList<T, T> SortedList()
+        {
+            var collection = new SortedList<T, T>();
+            var uniqueValues = _uniqueValues;
+            for (int i = 0; i < uniqueValues.Length; i++)
+                collection.TryAdd(uniqueValues[i], uniqueValues[i]);
+            return collection;
+        }
```
madelson, line 55: I didn't end up adding TryAdd benchmarks for the immutable collections here because they just use the TryAdd extension which does not work for them.

src/benchmarks/micro/libraries/System.Collections/CopyTo.cs
```diff
@@ -21,6 +21,7 @@ public class CopyTo<T>
         private T[] _array;
         private List<T> _list;
         private ImmutableArray<T> _immutablearray;
+        private ImmutableList<T> _immutableList;
```
madelson, line 24: All collection shave the CopyTo method (e. g. dictionaries, sets, etc) but since the original file only went with lists and arrays I also took that approach. Happy to add the rest of the collections if desired

src/benchmarks/micro/libraries/System.Collections/Dictionary/Perf_Dictionary.cs
```diff
@@ -25,22 +25,5 @@ public void InitializeContainsValue()
 
         [Benchmark]
         public object Clone() => new Dictionary<int, int>(_dict);
```
madelson, line 27: The original recommendation was to remove this method and make a Clone benchmark for all collections.

There is already a CtorFromCollection benchmark which largely covers this case and includes immutable collections.

What that benchmark doesn't test is the optimization immutable collections can have where `CreateRange` on another instance of the same type of immutable collection can just return the input. However, that doesn't seem worth benchmarking as it can easily be unit-tested.

src/benchmarks/micro/libraries/System.Collections/Indexer/IndexerSet.cs
```diff
@@ -64,6 +69,30 @@ public List<T> List()
...
+
+        [GlobalSetup(Target = nameof(ImmutableList))]
+        public void SetupImmutableList() => _immutableList = Immutable.ImmutableList.CreateRange(ValuesGenerator.ArrayOfUniqueValues<T>(Size));
+
+        [Benchmark]
+        public ImmutableList<T> ImmutableList()
+        {
+            var immutableList = _immutableList;
+            for (int i = 0; i < immutableList.Count; i++)
+                immutableList = immutableList.SetItem(i, default);
```
madelson, line 92: SetItem is not technically an indexer set, but in other benchmarks (e. g. the Ctor benchmarks) there was precedent for including the immutable equivalent of a mutable method. So, I felt like it made sense to include this here rather than creating a separate SetItem benchmark. Let me know if you disagree.

src/benchmarks/micro/libraries/System.Collections/Sort.cs
```diff
@@ -59,6 +62,20 @@ public class Sort<T> where T : IComparable<T>
...
+
+        [Benchmark]
+        public ImmutableArray<T> ImmutableArray() => _immutableArray.Sort();
+
+        [GlobalSetup(Target = nameof(ImmutableList))]
+        public void SetupImmutableList() =>
+            _immutableList = Immutable.ImmutableList.CreateRange(GenerateValues());
+
+        [Benchmark]
+        public ImmutableList<T> ImmutableList() => _immutableList.Sort();
```
madelson, line 77: Most of this benchmark assumes mutable sort, so it's setup involves populating N lists/arrays upfront and then sorting each of them on subsequent iterations. That doesn't make sense for immutable collections, but will the fixed invocation count throw off the measurement at all here?

madelson:
@adamsitnik any concerns with these?

madelson:
@adamsitnik anything else to do here?

PR #2313, opened 2022-03-16
Add Parameter Filter

Tohron:
This PR adds the option to filter by parameters when running benchmarks.  For example, using the command
dotnet run -c Release -f net7.0 --filter *Perf_Basic*
will run all the benchmark scenarios defined in Perf.Basic, whereas running
dotnet run -c Release -f net7.0 --filter *Perf_Basic* --parameter-filter SkipValidation:True Formatted:False
will run only the scenarios where SkipValidation is set to True and Formatted is set to False.  The filter can handle boolean, integer, and string parameters.


Tohron:
Merges into wrong branch, closing...

PR #2314, opened 2022-03-16
Add Parameter Filter

Tohron:
This PR adds the option to filter by parameters when running benchmarks.  For example, using the command
dotnet run -c Release -f net7.0 --filter *Perf_Basic*
will run all the benchmark scenarios defined in Perf.Basic, whereas running
dotnet run -c Release -f net7.0 --filter *Perf_Basic* --parameter-filter SkipValidation:True Formatted:False
will run only the scenarios where SkipValidation is set to True and Formatted is set to False.  The filter can handle boolean, integer, and string parameters.


src/benchmarks/micro/Properties/Resources.Designer.cs
```diff
@@ -0,0 +1,63 @@
+﻿//------------------------------------------------------------------------------
```
LoopedBard3, line 1: Is this file part of the change or just a VS one?

src/benchmarks/micro/Properties/Resources.resx
```diff
@@ -0,0 +1,101 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
LoopedBard3, line 1: Is this file part of the change or just a VS one?

src/benchmarks/micro/Properties/launchSettings.json
```diff
@@ -0,0 +1,10 @@
+{
```
LoopedBard3, line 1: Is this file part of the change or just a VS one?

src/tests/harness/BenchmarkDotNet.Extensions.Tests/PartitionFilterTests.cs
```diff
@@ -85,7 +85,7 @@ public void NoBenchmarksAreOmitted_RealData()
                 for (int partitionIndex = 0; partitionIndex < PartitionCount; partitionIndex++)
                 {
                     PartitionFilter filter = new(PartitionCount, partitionIndex);
-
+                    IEnumerable<BenchmarkCase> tmp = GetAllBenchmarks(parsedConfig, runnable).SelectMany(benchmark => benchmark.BenchmarksCases);
```
LoopedBard3, line 88: Is this tmp variable used elsewhere? If not, we should probably remove it.

src/harness/BenchmarkDotNet.Extensions/CommandLineOptions.cs
```diff
@@ -53,6 +54,44 @@ public static List<string> ParseAndRemoveStringsParameter(List<string> argsList,
...
+
+            if (parameterIndex + 1 < argsList.Count)
+            {
+                while (parameterIndex + 1 < argsList.Count && !argsList[parameterIndex + 1].StartsWith("-"))
+                {
+                    string[] pair = argsList[parameterIndex + 1].Split(':');
+                    object value;
+                    int i_result;
+                    bool b_result;
+                    if (Int32.TryParse(pair[1], out i_result))
```
DrewScoggins, line 70: If you use this approach to compare the types of the parameters, you will need to make sure that every type that we can use here is covered. Otherwise, if we ever have a type like byte, or long this piece of logic would fail.

LoopedBard3:
@Tohron It also looks like, at least Roslyn, is running into an issue related to this change:
```
[2022/03/16 19:35:08][INFO] $ pushd "C:\h\w\C26C09EE\p\src\benchmarks\real-world\Roslyn"
[2022/03/16 19:35:08][INFO] $ dotnet run --project C:\h\w\C26C09EE\p\src\benchmarks\real-world\Roslyn\CompilerBenchmarks.csproj --configuration Release --framework net7.0 --no-restore --no-build -- --anyCategories roslyn --iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart --stopOnFirstError true --packages C:\h\w\C26C09EE\p\artifacts\packages --buildTimeout 1200
[2022/03/16 19:35:11][INFO] Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
[2022/03/16 19:35:11][INFO]    at ParameterFilter.Predicate(BenchmarkCase benchmarkCase) in C:\h\w\C26C09EE\p\src\harness\BenchmarkDotNet.Extensions\ParameterFilter.cs:line 23
[2022/03/16 19:35:11][INFO]    at BenchmarkDotNet.Running.BenchmarkConverter.<>c__DisplayClass14_1.<GetFilteredBenchmarks>b__1(IFilter filter)
[2022/03/16 19:35:11][INFO]    at System.Linq.Enumerable.All[TSource](IEnumerable`1 source, Func`2 predicate)
[2022/03/16 19:35:11][INFO]    at BenchmarkDotNet.Running.BenchmarkConverter.<>c__DisplayClass14_0.<GetFilteredBenchmarks>b__0(BenchmarkCase benchmark)
[2022/03/16 19:35:11][INFO]    at System.Linq.Enumerable.WhereEnumerableIterator`1.ToArray()
[2022/03/16 19:35:11][INFO]    at System.Collections.Immutable.ImmutableArray.CreateRange[T](IEnumerable`1 items)
[2022/03/16 19:35:11][INFO]    at BenchmarkDotNet.Running.BenchmarkConverter.MethodsToBenchmarksWithFullConfig(Type type, MethodInfo[] benchmarkMethods, IConfig config)
[2022/03/16 19:35:11][INFO]    at BenchmarkDotNet.Running.BenchmarkConverter.TypeToBenchmarks(Type type, IConfig config)
[2022/03/16 19:35:11][INFO]    at BenchmarkDotNet.Running.TypeFilter.<>c__DisplayClass1_0.<Filter>b__0(Type type)
[2022/03/16 19:35:11][INFO]    at System.Linq.Enumerable.SelectListIterator`2.MoveNext()
[2022/03/16 19:35:11][INFO]    at System.Linq.Enumerable.WhereEnumerableIterator`1.ToArray()
[2022/03/16 19:35:11][INFO]    at BenchmarkDotNet.Running.TypeFilter.Filter(IConfig effectiveConfig, IEnumerable`1 types)
[2022/03/16 19:35:11][INFO]    at BenchmarkDotNet.Running.BenchmarkSwitcher.RunWithDirtyAssemblyResolveHelper(String[] args, IConfig config, Boolean askUserForInput)
[2022/03/16 19:35:11][INFO]    at BenchmarkDotNet.Running.BenchmarkSwitcher.Run(String[] args, IConfig config)
[2022/03/16 19:35:11][INFO]    at Program.<Main>$(String[] args) in C:\h\w\C26C09EE\p\src\benchmarks\real-world\Roslyn\Program.cs:line 23
[2022/03/16 19:35:11][INFO]    at Program.<Main>(String[] args)
[2022/03/16 19:35:17][INFO] $ popd
```

src/benchmarks/micro/Properties/launchSettings.json
Tohron: This is the file used to store launch settings (like command line arguments & environment variables) for testing.  Other people will probably use it.

DrewScoggins (changes requested):
We should also think about under what conditions we will want this filter to be used. Does it make sense to run this without the `--filter` argument as well?

src/benchmarks/micro/Properties/launchSettings.json
DrewScoggins: But every person will have their own launch settings and will want to set them locally. As a rule we do not check-in things like this.

src/harness/BenchmarkDotNet.Extensions/CommandLineOptions.cs
```diff
@@ -53,6 +54,44 @@ public static List<string> ParseAndRemoveStringsParameter(List<string> argsList,
...
+        public static List<string> ParseAndRemovePairsParameter(List<string> argsList, string parameter, out Dictionary<string, object> pairValues)
+        {
+            int parameterIndex = argsList.IndexOf(parameter);
+            pairValues = new Dictionary<string, object>();
+
+            if (parameterIndex + 1 < argsList.Count)
+            {
+                while (parameterIndex + 1 < argsList.Count && !argsList[parameterIndex + 1].StartsWith("-"))
+                {
+                    string[] pair = argsList[parameterIndex + 1].Split(':');
```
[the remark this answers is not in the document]
Tohron: No, if that is a valid concern, we will need to identify which character/character combo works as a valid separator between a parameter name and a value.  Spaces are already being used to separate parameter blocks.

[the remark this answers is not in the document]
LoopedBard3: Do we know which characters would not be legal in the name? Maybe one of those could be used as the separator.

src/benchmarks/micro/Properties/launchSettings.json
LoopedBard3: We should add this to the .gitignore file as I think VS autogenerates.

Tohron: Added the change to .gitignore. You can check to see if I got the syntax right

.gitignore
```diff
@@ -55,6 +55,7 @@ bld/
 
 # Visual Studio 2017 auto generated files
 Generated\ Files/
+**/*.Properties/launchSettings.json
```
LoopedBard3, line 58: I think having the . after the * will cause it not to match to the launch settings

Tohron: I was copying the syntax used later with
**/*.DesktopClient/ModelManifest.xml
Is that not valid here?

LoopedBard3: I am not 100% sure, it may be. One way to find out would be to readd the launchsettings file in the same location and see if it is ignored.

Tohron: Seems I did need to remove the dot.  Not sure if the reference I used was also invalid, but I'll leave it for now.

src/harness/BenchmarkDotNet.Extensions/CommandLineOptions.cs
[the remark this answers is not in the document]
DrewScoggins: The problem is not the name of the parameter, but the value. For strings it can be almost anything. For now, we can leave it as is, and if we come across a value that has this we can fix it then.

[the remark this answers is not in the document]
sblom: `Split()` takes a second argument that will limit the number of groups to split into.

`.Split(':',2)` will preserve ALL of whatever's on the right side of the colon.

Illustration: https://dotnetfiddle.net/EAuVmo

adamsitnik:
I wonder if we should just extend BDN's `--filter` with arguments and parameter support.  We could make it optional as it would be a breaking change.

https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Filters/GlobFilter.cs#L14

DrewScoggins:
> I wonder if we should just extend BDN's `--filter` with arguments and parameter support. We could make it optional as it would be a breaking change.
> 
> https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Filters/GlobFilter.cs#L14

For sure, I think that this is probably the right place for this to eventually wind up, but we wanted to just roll out the feature in a slightly more lightweight way first. Don't have to worry about as large an impact when it's just in our repo :)

AndyAyersMS:
Finally got around to using this and it works great. Agree though that supporting it in BDN would be nice.

PR #2315, opened 2022-03-16
[MAUI] Update android startup scenario for better consistency.

LoopedBard3:
Set default number of mobile startup runs to 10, and added delay after startup run, and added a minimum screen timeout so we don't accidentally have the screen timing out during testing. Setting the default number of mobile startup runs to 10 will create less noise and more closely matches the profile.ps1 file that is manually used by the Maui team. Adding the delay after the initial run will give the phone time to do post install processing. Manually setting the screen timeout to be at least 2 minutes ensures the whole test will run for the default startup run number, the timeout is reset back to its previous number and only changed when the default is lower than the desired 2 minutes.

### DIFF
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -61,7 +61,7 @@ class Runner:
         parseonlyparser.add_argument('--device-type', choices=['android','ios'],type=str.lower,help='Device type for testing', dest='devicetype')
         parseonlyparser.add_argument('--package-path', help='Location of test application', dest='packagepath')
         parseonlyparser.add_argument('--package-name', help='Classname of application', dest='packagename')
-        parseonlyparser.add_argument('--startup-iterations', help='Startups to run (1+)', type=int, default=5, dest='startupiterations')
+        parseonlyparser.add_argument('--startup-iterations', help='Startups to run (1+)', type=int, default=10, dest='startupiterations')
         parseonlyparser.add_argument('--disable-animations', help='Disable Android device animations', action='store_true', dest='animationsdisabled')
         self.add_common_arguments(parseonlyparser)
 
@@ -342,7 +342,7 @@ ex: C:\repos\performance;C:\repos\runtime
             RunCommand(cmdline, verbose=True).run()
 
             # Get animation values
-            getLogger().info("Getting animation values")
+            getLogger().info("Getting Values we will need set specifically")
             cmdline = [
                 adb.stdout.strip(),
                 'shell', 'settings', 'get', 'global', 'window_animation_scale'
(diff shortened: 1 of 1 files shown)


src/scenarios/shared/runner.py
```diff
@@ -514,6 +528,8 @@ def run(self):
                     getLogger().exception("Failed to get past permission screen, run locally to see if enough next button presses were used.")
                     exit(-1)
 
+            time.sleep(10) # Add delay to ensure app is fully installed and give it some time to settle
```
sblom, line 531: Are you convinced that there's no way to directly observe the completion of whatever we're waiting for here? Totally fine if you are, but statements like this one read as "expect to change this number after eventual intermittent failures" to me.

src/scenarios/shared/runner.py
```diff
@@ -361,14 +361,21 @@ def run(self):
...
+            getLogger().info(f"Retrieved values window {window_animation_scale_cmd.stdout.strip()}, transition {transition_animation_scale_cmd.stdout.strip()}, animator {animator_duration_scale_cmd.stdout.strip()}, screen timeout {screen_off_timeout_cmd.stdout.strip()}")
 
             # Make sure animations are set to 1 or disabled
-            getLogger().info("Setting animation values")
+            getLogger().info("Setting needed values")
             if(self.animationsdisabled):
                 animationValue = 0
             else:
                 animationValue = 1
+            minimumTimeoutValue = 120000 # Set the timeout minimum to 2 minutes
```
sblom, line 378: [minor]
Slightly different (comment) style that survives future code edits better
`minimumTimeoutValue = 2 * 60 * 1000 # milliseconds`

or if python 3.6+ is guaranteed
`minimumTimeoutValue = 2 * 60_000 # milliseconds`

or 120 is a recognizable enough number to even justify
`minimumTimeoutValue = 120_000 # milliseconds`

sblom (approved):

src/scenarios/shared/runner.py
jonathanpeppers: Yes, this is through trial an error, we've observed numbers are more stable if you leave an Android app on the screen for a bit before doing a startup test run.

I honestly don't see this number needing to change in the future. Nothing would fail if it is lower or higher -- we just might get more sporadic number if it is lower.

jonathanpeppers, line 531: @LoopedBard3 it's not clear to me here, is it sleeping 10 seconds between each run?

Is there a way we can launch the first time, wait 10 seconds, throw that number out. Then do the normal 10 run average?

LoopedBard3: That is what is happening here. The main tests take place from line 534 to 539 and have their own 3 second gap between those tests.

jonathanpeppers: Ok sounds good, then. 👍

LoopedBard3:
The last two failures are consistent with main and this shouldn't affect anything in them, merging.

PR #2320, opened 2022-03-17
Fixes for wasm builds

radical:
`Add --dotnet-path to the ci scripts`

For wasm builds, we use a custom dotnet populated with the bits from a
`runtime` build. And this is passed to bdn via `--cli` argument.

The project is built by `benchmarks_ci.py` first, and then run, which
then builds the project again via BDN, using the dotnet from the
`--cli` argument.

But this means that the project gets built with the "default" dotnet
first, and then the custom one next, which can result in a corrupted
build. Using `--dotnet-path`, same as `--cli` avoids this.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -108,6 +108,21 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         help='Product repository.'
     )
 
+    def __is_valid_dotnet_path(dp: str) -> str:
+        if not os.path.isdir(dp):
+            raise ArgumentTypeError('Path {} does not exist'.format(dp))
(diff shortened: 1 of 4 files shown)


radical:
cc @adamsitnik @DrewScoggins @LoopedBard3 @lewing

radical:
Failures are same as mentioned in https://github.com/dotnet/performance/pull/2323#issuecomment-1072923660 and https://github.com/dotnet/performance/pull/2323#issuecomment-1072923817 .

DrewScoggins:
LGTM. We should also keep a close eye on this after it gets merged to make sure we don't see any unintended consequences in the runtime repo runs.

DrewScoggins (approved):

PR #2322, opened 2022-03-18
Update packages to 7.0-preview2

maryamariyan:
cc @adamsitnik @joperezr

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "6.0.100"
+    "dotnet": "7.0.100-preview.2.22153.17"
   },
   "msbuild-sdks": {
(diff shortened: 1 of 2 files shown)


rick-gwu (approved):
Looks good

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "6.0.100"
+    "dotnet": "7.0.100-preview.2.22153.17"
```
joperezr, line 3: Not very familiar with the policies of this repo and how they handle bumping the SDK, so some maintainer should make sure this is ok. Also, at least with dotnet/runtime, these updates are usually made by arcade automatic PRs, but not sure how this repo works

PR #2324, opened 2022-03-20
Add Perf_Guid::EqualsNotSame benchmark

EgorBo:
I noticed that we don't have a benchmark for a fairly common case - when we compare two unequal guids

For reference https://github.com/dotnet/runtime/pull/66889


adamsitnik (approved):
LGTM, thank you @EgorBo !

PR #2328, opened 2022-03-23
Add NativeAOT (and CoreRT) support for micro benchmarks

adamsitnik:
This PR contains 4 changes:

-  BDN update to include https://github.com/dotnet/BenchmarkDotNet/pull/1960 and https://github.com/dotnet/BenchmarkDotNet/pull/1955
- using `[AotFilter]` to filter out benchmarks that are not supported with AOT
- using `AppContext.BaseDirectory` instead `typeof(X).Assembly.Location` to get AOT-friendly location
- python script changes

I have tested these changes locally on Windows for both NativeAOT and CoreRT and all benchmarks are working fine. I'll report back once I have Linux results.

I've added `nativeaot7.0` and `corert6.0` monikers to the supported framework list. I did not want to add another optional argument like `--aot` because it would make it impossible to compare multiple frameworks using `-f`: 

```cmd
py .\scripts\benchmarks_ci.py -f net6.0 corert6.0 net7.0 nativeaot7.0 --filter *
```

With this PR we will be able to compare .NET 6 with CoreRT vs .NET 7 with NativeAOT and ensure that there are no regressions.

It's also possible to specify `ILCompilerVersion`:

```cmd
py .\scripts\benchmarks_ci.py -f nativeaot7.0 --filter * --bdn-arguments "--ilCompilerVersion $here" 
```

So we will be able to do things like Preview X vs Preview Y.

cc @jkotas @MichalStrehovsky @hez2010 @Beau-Gosse-dev @jeffhandley @agocke


adamsitnik:
@DrewScoggins could you please upload `0.13.1.1737` to our internal NuGet feed? Would you mind if I add one CI leg for NativeAOT microbenchmarks?

jkotas:
> corert6.0

We have retired the corert repo in 2020. It is confusing that this called corert. I assume that the bits are not coming from corert repo. Can we call it nativeaot6.0?

adamsitnik:
> We have retired the corert repo in 2020. It is confusing that this called corert. I assume that the bits are not coming from corert repo. Can we call it nativeaot6.0?

The `corert6.0` is using the retired bits on purpose, I want to use it as a baseline for perf comparison.

|              | Microsoft.DotNet.ILCompiler | NuGet feed                                                                                   |
| ------------ | --------------------------- | -------------------------------------------------------------------------------------------- |
| corert6.0    | 6.0.0-\*                    | https://pkgs.dev.azure.com/dnceng/public/\_packaging/dotnet-experimental/nuget/v3/index.json |
| nativeaot7.0 | 7.0.0-\*                    | https://pkgs.dev.azure.com/dnceng/public/\_packaging/dotnet7/nuget/v3/index.json             |

I've already spotted one perf regression, which is most likely related to printing to console on Windows.

jkotas:
I am ok with having an option to use the .NET 6 experimental bits. My point is these bits are coming from NativeAOT experimental project in runtimelab. They are not coming from https://github.com/dotnet/corert repo. It is confusing to call them `corert6.0`.

DrewScoggins:
No go ahead and add that leg. I assume this will only be on the CI side, and not the collection side, right?

src/benchmarks/micro/Serializers/Xml_FromStream.cs
```diff
@@ -18,6 +19,7 @@ namespace MicroBenchmarks.Serializers
     [GenericTypeArguments(typeof(XmlElement))]
     [GenericTypeArguments(typeof(SimpleStructWithProperties))]
     [GenericTypeArguments(typeof(ClassImplementingIXmlSerialiable))]
+    [AotFilter("Currently not supported due to missing metadata.")]
```
hez2010, line 22: Issues caused by missing metadata can be fixed with `rd.xml`.

adamsitnik: @hez2010 great point! I've opened https://github.com/dotnet/performance/issues/2338 to track it

PR #2329, opened 2022-03-23
Longer strings for string.Replace(char, char) benchmarks

gfoidl:
Cf. https://github.com/dotnet/runtime/pull/67049#issuecomment-1076521522

/cc: @danmoseley

### DIFF
--- a/src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
@@ -163,7 +163,7 @@ namespace System.Tests
             yield return new object[] { "Hello", 'l', '!' };    // Contains two 'l'
             yield return new object[] { "Hello", 'a', 'b' };    // Contains one 'a'
             yield return new object[] { "This is a very nice sentence", 'z', 'y' }; // 'z' does not exist in the string
-            yield return new object[] { "This is a very nice sentence", 'i', 'I' }; // 'i' occuress 3 times in the string
+            yield return new object[] { "This is a very nice sentence", 'i', 'I' }; // 'i' occurs 3 times in the string
             yield return new object[] { new string('-', 100), '-', '+' };
             yield return new object[] { new string('-', 1000), '-', '+' };
         }
--- a/src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
@@ -164,8 +164,8 @@ namespace System.Tests
             yield return new object[] { "Hello", 'a', 'b' };    // Contains one 'a'
             yield return new object[] { "This is a very nice sentence", 'z', 'y' }; // 'z' does not exist in the string
             yield return new object[] { "This is a very nice sentence", 'i', 'I' }; // 'i' occurs 3 times in the string
-            yield return new object[] { new string('-', 100), '-', '+' };
-            yield return new object[] { new string('-', 1000), '-', '+' };
+            yield return new object[] { PerfUtils.CreateRandomString(100, seed: 42), 'b', '+' };    // b occurs 8 times in the string
+            yield return new object[] { PerfUtils.CreateRandomString(1000, seed: 42), 'b', '+' };   // b occurs 42 times in the string
         }
 
         [Benchmark]
--- a/src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
@@ -154,12 +154,19 @@ namespace System.Tests
             => s.TrimEnd(c);
 
         [Benchmark]
-        [Arguments("Hello", 'l', '!')] // Contains two 'l'
-        [Arguments("Hello", 'a', 'b')] // Contains one 'a'
-        [Arguments("This is a very nice sentence", 'z', 'y')] // 'z' does not exist in the string
-        [Arguments("This is a very nice sentence", 'i', 'I')] // 'i' occuress 3 times in the string
+        [ArgumentsSource(nameof(ReplaceArguments))]
         public string Replace_Char(string text, char oldChar, char newChar)
             => text.Replace(oldChar, newChar);
+
+        public static IEnumerable<object[]> ReplaceArguments()
+        {
+            yield return new object[] { "Hello", 'l', '!' };    // Contains two 'l'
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
```diff
@@ -154,13 +154,20 @@ public string TrimEnd_CharArr(string s, char[] c)
...
         [Benchmark]
-        [Arguments("Hello", 'l', '!')] // Contains two 'l'
-        [Arguments("Hello", 'a', 'b')] // Contains one 'a'
-        [Arguments("This is a very nice sentence", 'z', 'y')] // 'z' does not exist in the string
-        [Arguments("This is a very nice sentence", 'i', 'I')] // 'i' occuress 3 times in the string
+        [ArgumentsSource(nameof(ReplaceArguments))]
         public string Replace_Char(string text, char oldChar, char newChar)
             => text.Replace(oldChar, newChar);
 
+        public static IEnumerable<object[]> ReplaceArguments()
```
gfoidl, line 161: Verifyed that the method names are displayed the same way, so that the reporting tool has correlation. Is this enough to be on the safe side here?

```
|          Method |                 text | oldChar | newChar |       Mean |      Error |    StdDev |
|---------------- |--------------------- |-------- |-------- |-----------:|-----------:|----------:|
| Replace_CharNew |  ----(...)---- [100] |       - |       + |  32.553 ns |  21.554 ns | 1.1815 ns |
| Replace_CharNew | ----(...)---- [1000] |       - |       + | 202.048 ns | 178.291 ns | 9.7727 ns |
|    Replace_Char |                Hello |       a |       b |   5.256 ns |   4.023 ns | 0.2205 ns |
| Replace_CharNew |                Hello |       a |       b |   5.367 ns |   2.158 ns | 0.1183 ns |
|    Replace_Char |                Hello |       l |       ! |  17.491 ns |   4.643 ns | 0.2545 ns |
| Replace_CharNew |                Hello |       l |       ! |  18.092 ns |  12.480 ns | 0.6841 ns |
|    Replace_Char | This (...)tence [28] |       i |       I |  36.695 ns |  91.723 ns | 5.0277 ns |
| Replace_CharNew | This (...)tence [28] |       i |       I |  30.385 ns |  25.207 ns | 1.3817 ns |
|    Replace_Char | This (...)tence [28] |       z |       y |  11.598 ns |   6.366 ns | 0.3489 ns |
| Replace_CharNew | This (...)tence [28] |       z |       y |  12.132 ns |  10.304 ns | 0.5648 ns |
```
The ones with suffix _New_ are the methods with the ArgumentsSource, the other ones with Arguments only (as in main-branch).

src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
```diff
@@ -154,13 +154,20 @@ public string TrimEnd_CharArr(string s, char[] c)
...
+        {
+            yield return new object[] { "Hello", 'l', '!' };    // Contains two 'l'
+            yield return new object[] { "Hello", 'a', 'b' };    // Contains one 'a'
+            yield return new object[] { "This is a very nice sentence", 'z', 'y' }; // 'z' does not exist in the string
+            yield return new object[] { "This is a very nice sentence", 'i', 'I' }; // 'i' occuress 3 times in the string
+            yield return new object[] { new string('-', 100), '-', '+' };
+            yield return new object[] { new string('-', 1000), '-', '+' };
```
gfoidl, line 168: Just picked random larger numbers.

gfoidl (commented):
Notes for review.

src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
danmoseley: Would it be more realistic to not replace all the chars in the string? And also have them in varying offsets. Eg. something like this (making it up)
```
yield return new object[] { string.Join("", Enumerable.Range(1, 300).ToArray()), '1', 'x' }
```

src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
```diff
@@ -154,13 +154,20 @@ public string TrimEnd_CharArr(string s, char[] c)
...
+        [ArgumentsSource(nameof(ReplaceArguments))]
         public string Replace_Char(string text, char oldChar, char newChar)
             => text.Replace(oldChar, newChar);
 
+        public static IEnumerable<object[]> ReplaceArguments()
+        {
+            yield return new object[] { "Hello", 'l', '!' };    // Contains two 'l'
+            yield return new object[] { "Hello", 'a', 'b' };    // Contains one 'a'
+            yield return new object[] { "This is a very nice sentence", 'z', 'y' }; // 'z' does not exist in the string
+            yield return new object[] { "This is a very nice sentence", 'i', 'I' }; // 'i' occuress 3 times in the string
```
danmoseley, line 166:
```suggestion
            yield return new object[] { "This is a very nice sentence", 'i', 'I' }; // 'i' occurs 3 times in the string
```

gfoidl: Haha, copy & pasted the type too. Thanks!

src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
gfoidl: > to not replace all the chars in the string?

Solely for the vectorized case it doesn't matter -- it's branchless and the exact same operations are executed in that case.
But you're right that for the scalar case there's a difference, as the branch predictor will do its job.

Will put same random chars (with constant seed?) into string.

gfoidl: --> 4d629c93bfd5c23082aff940e053821a0e0281ab

danmoseley (approved):
LGTM once we get an answer to q above. thanks for adding these.

src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
DrewScoggins: Sorry, I thought I had already responded here. As long as the test name or the arguments have not changed values we should have no issues here. And from what I can tell nothing has changed for the existing test cases. So this LGTM.

DrewScoggins (approved):

PR #2343, opened 2022-03-31
Update skip labels for certain tests

DrewScoggins:
These tests are taking over 600 seconds to complete one operation. This is
causing the entire job to timeout and fail. Removing these tests from
rotation until we figure out what is going on.


LoopedBard3 (approved):
LGTM, outside of this we should probably look into having a complete disable category so we don't have to worry about which specific run types we disable if we want a test completely disabled, or something along those lines.

DrewScoggins:
Sorry, my comment was poorly worded. These are only failing on the exact places where I am disabling them. So either Interpreter or Mono AOT. They will need to continually be off for those configurations. Although @naricc @SamMonoRT these might be good places to look at from a performance perspective, as we don't expect these to be this bad.

PR #2345, opened 2022-04-02
[MAUI] Add capabilities to measure fully drawn instead of first frame drawn.

LoopedBard3:
Add capabilities to measure fully drawn instead of first frame drawn for the android Maui tests. This will also be used by iOS once it is integrated.


LoopedBard3:
Test run was successful.

sblom:
Seems alright. I can't tell from just the diff context: is this extensible to measuring multiple events? (As we start to move more and more into scenario space, measuring various waypoints along a scenario becomes useful.)

LoopedBard3:
This is a start towards being more extensible, in this case it hardcodes to saving either the first frame startup or time to reportFullyDrawn time while changing to the general flow to eventually add custom parsing if desired.

PR #2346, opened 2022-04-02
Multiple improvements, to help with blazor scenarios

radical:
1. improve blazor scenarios' `pre.py` scripts to install, or uninstall workloads only when needed
2. Skip AOT'ing assemblies which get oomkill'ed on CI while compiling with `emcc`
3. Add a new `--not-in-lab` param to help with supporting perf runs for public pull requests

### DIFF
--- a/scripts/ci_setup.py
+++ b/scripts/ci_setup.py
@@ -161,6 +161,15 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         help='Directory to install dotnet to'
     )
 
+    parser.add_argument(
+        '--not-in-lab',
+        dest='in_lab',
+        required=False,
+        action='store_false',
+        default=True,
+        help='Indicates that this is not running in perflab'
+    )
+
     def __is_valid_dotnet_path(dp: str) -> str:
         if not os.path.isdir(dp):
             raise ArgumentTypeError('Path {} does not exist'.format(dp))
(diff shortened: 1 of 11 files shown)


radical:
cc @adamsitnik @LoopedBard3 @DrewScoggins  @lewing

scripts/ci_setup.py
```diff
@@ -161,6 +161,15 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         help='Directory to install dotnet to'
     )
 
+    parser.add_argument(
+        '--not-in-lab',
+        dest='in_lab',
+        required=False,
+        action='store_false',
+        default=True,
```
adamsitnik, line 169: is my understanding correct that if we set the default value to `True`, the current default behavior is going to change? and we will need to update some CI configuration somewhere?

adamsitnik (commented):
Overall LGTM, but I am not sure if we should set `--not-in-lab` to true by default

scripts/ci_setup.py
radical: Sorry, this is a bit confusing. I'm using a negated backing variable here `in_lab`, so the default is `in_lab=true`, which IIUC, is the default behavior. And passing `--not-in-lab` would set `store_false`, and invert the behavior.

I was trying to avoid a `if not not_in_lab` condition, but the way I ended up using this, it's not a concern: `'1' if args.in_lab else '0'`

So, I propose that I change this to be `not_in_lab`, and default that to false. Would that be agreeable?

adamsitnik (approved):
LGTM, thank you @radical !

PR #2347, opened 2022-04-03
Added Benchmark for SpanHelpers.Contains {byte, char}

gfoidl:
Cf. https://github.com/dotnet/runtime/pull/67492#issuecomment-1086758553

I created a new file for this benchmark, as the current benchmarks for [`Span`](https://github.com/dotnet/performance/blob/a37c8e3e4c72d584934e97a1fdc0a4785ddb8674/src/benchmarks/micro/libraries/System.Memory/Span.cs#L18) didn't fit here really well -- though I'm not sure if my benchmarks here are over-engineered.

For e.g. `SpanHelpers.IndexOf` there is a benchmark (https://github.com/dotnet/performance/blob/a37c8e3e4c72d584934e97a1fdc0a4785ddb8674/src/benchmarks/micro/libraries/System.Memory/Span.cs#L69), but this one has (maybe) same flaws:
* it tests only one length (constant 512)
* it doesn't take into account the different pathes that are gone due to vectorization
* it doesn't cover the edge-cases from the vectorized pathes
* BUT: they run fast

At least for the PR https://github.com/dotnet/runtime/pull/67492 I covered these cases, but the execution time for the benchmarks rises.

So I'm not sure what's the right balance here. Especially if we take ARM also into account, then there maybe different cut-offs for the edge-cases making the combinations grow up.

On one hand I think the benchmark should cover edge-cases -- here the perf-critical ones -- too, but on the other hand they should give an good enough estimate on how well the code is performing.

PS: If this should be discussed in an issue rather than this PR here, please create an issue, edit this PR, whatever is best.


src/benchmarks/micro/libraries/System.Memory/SpanHelpers.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public static IEnumerable<object> LengthValues()
+        {
+            // The values for the length take into account the different cut-offs
+            // in the vectorized pathes.
+
+            if (typeof(T) == typeof(byte))
+            {
+                // Vectorization is done on 2 * Vector<byte>.Count => 64 byte elements
+                yield return 63;    // one less the vectorization threshould
```
gfoidl, line 33: At the moment these values are based on `Vector<T>.Count` on x64 with AVX2 enabled.
So for e.g. ARM where `Vector<byte>.Count` is 16 (instead of 32 as on x64) the values should be halfed.

Via `ParamsSource` this could be taken into account. Shall I update to this?
(Didn't do now, because of the discussion in the PR's description -- i.e. I'd like to an outcome from this first).

danmoseley: @adamsitnik might suggest.

danmoseley: @adamsitnik could you advise, so we can get this merged?

adamsitnik: Using `Vector<byte>.Count` seems a good idea in this particular example. cc @tannergooding

tannergooding: if the algorithm is gated on `2 * Vector<byte>.Count` then I agree we should be explicit since that size can vary based on hardware.

src/benchmarks/micro/libraries/System.Memory/SpanHelpers.cs
```diff
@@ -0,0 +1,81 @@
...
+        private T* _searchSpace;
+        private T _value;
+
+        [ParamsSource(nameof(LengthValues))]
+        public int Length { get; set; }
+
+        public static IEnumerable<object> LengthValues()
+        {
+            // The values for the length take into account the different cut-offs
+            // in the vectorized pathes.
```
danmoseley: > [!NOTE]
> This suggestion was AI/Copilot-generated.

Typo: `pathes` → `paths`

```suggestion
            // in the vectorized paths.
```

src/benchmarks/micro/libraries/System.Memory/SpanHelpers.cs
danmoseley: > [!NOTE]
> This suggestion was AI/Copilot-generated.

Typo: `one less the vectorization threshould` → `one less than the vectorization threshold`

```suggestion
                yield return 63;    // one less than the vectorization threshold
```

src/benchmarks/micro/libraries/System.Memory/SpanHelpers.cs
```diff
@@ -0,0 +1,81 @@
...
+                yield return 31;
+                yield return 32;
+                yield return 33;
+                yield return 47;
+                yield return 100;
+            }
+            else
+            {
+                throw new NotSupportedException();
+            }
```
danmoseley: > [!NOTE]
> This suggestion was AI/Copilot-generated.

Hardcoded values assume `Vector<byte>.Count == 32` (256-bit SIMD), but this varies by hardware (128-bit on some ARM, 512-bit with AVX-512). Using `Vector<T>.Count` expressions ensures the benchmark targets the correct vectorization thresholds on all platforms, and also simplifies the code by removing the per-type branches.

```suggestion
            // The values for the length take into account the different cut-offs
            // in the vectorized paths.
            int V = Vector<T>.Count;

            yield return 2 * V - 1;    // one less than the vectorization threshold
            yield return 2 * V;        // exactly two vectorized operations
            yield return 2 * V + 1;    // one element more than standard vectorized loop
            yield return 3 * V - 1;    // one element less than another iteration of the standard vectorized loop
            yield return 100;
```

src/benchmarks/micro/libraries/System.Memory/SpanHelpers.cs
danmoseley: > [!NOTE]
> This suggestion was AI/Copilot-generated.

Typo: `pathes` → `paths`

```suggestion
            // in the vectorized paths.
```

src/benchmarks/micro/libraries/System.Memory/SpanHelpers.cs
danmoseley: > [!NOTE]
> This suggestion was AI/Copilot-generated.

Typo: `one less the vectorization threshould` → `one less than the vectorization threshold`

```suggestion
                yield return 63;    // one less than the vectorization threshold
```

src/benchmarks/micro/libraries/System.Memory/SpanHelpers.cs
danmoseley: > [!NOTE]
> This suggestion was AI/Copilot-generated.

Hardcoded values assume `Vector<byte>.Count == 32` (256-bit SIMD), but this varies by hardware (128-bit on some ARM, 512-bit with AVX-512). Using `Vector<T>.Count` expressions ensures the benchmark targets the correct vectorization thresholds on all platforms, and also simplifies the code by removing the per-type branches.

```suggestion
            // The values for the length take into account the different cut-offs
            // in the vectorized paths.
            int V = Vector<T>.Count;

            yield return 2 * V - 1;    // one less than the vectorization threshold
            yield return 2 * V;        // exactly two vectorized operations
            yield return 2 * V + 1;    // one element more than standard vectorized loop
            yield return 3 * V - 1;    // one element less than another iteration of the standard vectorized loop
            yield return 100;
```

danmoseley:
@gfoidl sorry this was neglected. do you want to make the updates?

I guess iti s https://github.com/danmoseley/performance/commit/bbafe4ad but I do not have write access to this rpo so cannot modify this PR for you.

If you would rather instead I can make a new PR and get it done that way

gfoidl:
@danmoseley changes are pushed.

danmoseley:
I'm probably being stupid, but is it intended that the  value is in the buffer (eg idk at the end)? Or it's fine that it's not there?

PR #2348, opened 2022-04-04
prefix produced tar file name with architecture

adamsitnik:
It's useful to be able to determine the arch without needing to extract the files first.

Sample output:

```log
 Results were created in the following folder:                                      
   .\artifacts\bin\MicroBenchmarks\Release\net7.0\BenchmarkDotNet.Artifacts\results 
 Results were collected into the following tar archive:                             
   .\artifacts\x86-2022-04-04-09-57-net7.0-preview3.tar.gz                          
```


adamsitnik:
@dakersnar ping

dakersnar (approved):
LGTM. Also, my GitHub notifications work now, thanks for the test!

PR #2349, opened 2022-04-04
For every PR run only the affected CI legs

adamsitnik:
I am going to add description once I get it working


DrewScoggins:
This seems good to me. I don't know if I have permission to add pipelines either. I normally just ping the first responders and have them make any pipeline changes. @sblom @LoopedBard3 any thoughts here?

LoopedBard3:
This makes sense to me. I think that having them separate will be better as the scenarios flow grows. Based on what I have seen about adding a pipeline, we will need the FRs or DNCENG to add it.

azure-pipelines-micro.yml
```diff
@@ -0,0 +1,178 @@
...
+    - src/benchmarks/micro/*
+    - src/benchmarks/real-world/*
+    - src/harness/*
+    - src/tools/Reporting/*
+    - scripts/*
+    - eng/*
+    - azure-pipelines-micro.yml
+    exclude:
+    - scripts/benchmarks_monthly.py
+
```
LoopedBard3, line 24: Does this need a schedule? I know it was on a schedule before and don't know if the files will update enough to trigger from here.

azure-pipelines-micro.yml
```diff
@@ -0,0 +1,178 @@
...
+    - container: ubuntu_x64_build_container
+      image: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-20210311173856-047508b
+    - container: centos_x64_build_container
+      image: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-20220107135107-9b5bbc2
+
+# Trigger builds for PR's targeting main
+pr:
+  branches:
+    include:
+    - main
```
LoopedBard3, line 12: Do we only want this running for main?

radical:
Is there a fallback needed for "this change doesn't fall into any of the pipelines", so .. run all?

adamsitnik:
I won't have the time to work on it anytime soon, closing.

PR #2355, opened 2022-04-06
Implement iOS startup time tracing

akoeplinger:


LoopedBard3 (approved):
LGTM

src/scenarios/shared/runner.py
```diff
@@ -588,6 +590,184 @@ def run(self):
...
+                    # ignore the warmup iteration
+                    getLogger().info(f'Warmup iteration took {totalTimeMilliseconds}')
+                else:
+                    # TODO: this isn't really a COLD run, we should have separate measurements for starting the app right after install
+                    launchState = 'COLD'
+                    allResults.append(f'LaunchState: {launchState}\nTotalTime: {int(totalTimeMilliseconds)}\nTimeToMain: {int(timeToMainMilliseconds)}\n\n')
+
+            # Done with testing, uninstall the app
+            getLogger().info("Uninstalling app")
+            uninstallAppCmd = xharnesscommand() + [
```
premun, line 748: Theoretically could go into `finally` since we raise exceptions in the loop.

It is not a big deal to leave the app behind as the bundle ID will always be the same and the app will always be re-installed in future

src/scenarios/shared/runner.py
```diff
@@ -588,6 +590,184 @@ def run(self):
...
+                    getLogger().info("Removed: " + os.path.join(const.TMPDIR, logarchive))
+                    rmtree(logarchive)
+
+            getLogger().info("Checking device state.")
+            cmdline = xharnesscommand() + ['apple', 'state']
+            adb = RunCommand(cmdline, verbose=True)
+            adb.run()
+
+            getLogger().info("Installing app on device.")
+            installCmd = xharnesscommand() + [
```
premun, line 616: You could simplify the code a bit by doing `apple run --no-wait` and killing the app.

You could remove the cold boot and could remove the install too. But I understand if you have plans to move to those no-wait commands later and this just works for now

src/scenarios/shared/runner.py
```diff
@@ -588,6 +590,184 @@ def run(self):
...
+            getLogger().info("Completed install.")
+
+            allResults = []
+            for i in range(self.startupiterations + 1): # adding one iteration to account for the warmup iteration
+                getLogger().info("Waiting 10 secs to ensure we're not getting confused with previous app run.")
+                time.sleep(10)
+
+                getLogger().info(f"Collect startup data for iteration {i}.")
+                runCmdTimestamp = datetime.now()
+                runCmd = xharnesscommand() + [
```
premun, line 635: Similar comment - if you used `--no-wait`, you could then grep the XHarness `just-run-...log` directly (it will have the mlaunch stdout with the PID). And you could skip the `log collect`.

You would however probably have to nuke the XHarness log in between the runs as it is appending (I think)

src/scenarios/shared/runner.py
```diff
@@ -588,6 +590,184 @@ def run(self):
...
+                    f'--killdev={app_pid}',
+                ]
+                killCmdCommand = RunCommand(killCmd, verbose=True)
+                killCmdCommand.run()
+
+                # Process Data
+
+                # There are four watchdog events from SpringBoard during an application startup:
+                #
+                # [application<net.dot.maui>:770] [realTime] Now monitoring resource allowance of 20.00s (at refreshInterval -1.00s)
```
premun, line 673: Will this be in the stdout of the app too? It would again then be easier to do `just-run --no-wait` and grep it from `net.dot.maui.log`

premun (commented):
Didn't mean to intrude but I was curious in how we measure the startup in iOS.

The new `--no-wait` could simplify the code a bit I think but I understand that this is working and is totally fine.

src/scenarios/shared/runner.py
akoeplinger: no, this is raised by the SpringBoard process, not the app itself.

src/scenarios/shared/runner.py
akoeplinger: Yeah using log collect was more reliable for me (might be fixed now).

src/scenarios/shared/runner.py
akoeplinger: Good point, will fix.

src/scenarios/shared/runner.py
```diff
@@ -327,7 +329,7 @@ def run(self):
                     getLogger().info("Removed: " + os.path.join(const.TRACEDIR, file))
                     os.remove(file)
 
-            cmdline = xharnesscommand() + [self.devicetype, 'state', '--adb']
+            cmdline = xharnesscommand() + ['android', 'state', '--adb']
```
DrewScoggins, line 332: Is there a reason that we are moving to the hardcoded value here, when we know that the value will be android because of the elif block?

src/scenarios/shared/runner.py
```diff
@@ -588,6 +590,184 @@ def run(self):
...
+
+                logarchive_filename = os.path.join(const.TMPDIR, f'iteration{i}.logarchive')
+                getLogger().info(f"Waiting 5 secs to ensure app with PID {app_pid} is fully started.")
+                time.sleep(5)
+                collectCmd = [
+                    'sudo',
+                    'log',
+                    'collect',
+                    '--device',
+                    '--start', runCmdTimestamp.strftime("%Y-%m-%d %H:%M:%S"),
```
DrewScoggins, line 654: Couldn't we just get this time from the device itself, so that way we would not be reliant on the two clocks being in sync?

sblom (approved):

src/scenarios/shared/runner.py
akoeplinger: @DrewScoggins sorry, didn't see this comment until now. There's not an easy way to get the time from the device (at least that I know of).

One thing that might be a better option would be to do an initial `log collect` and look at the timestamp of the last event in there as our starting point. We can do that if it ever becomes a problem, iPhones/Macs usually do a good job of keeping the time in sync with a timeserver as long as they have Internet connectivity.

src/scenarios/shared/runner.py
akoeplinger: basically because of that, it is always `android` in that block so using a variable is just slightly harder to read IMO.

premun: Also `apple/wasm state --adb` is not a valid command.

That being said, you could now utilize the `android adb` command directly and this call and some other below can be removed. You don't need to now get ADB's location from `adb.stdout.strip()` and call
```
xharnesscommand() + ['android', 'adb', '--', 'shell', 'vm', 'size']
```

I'd even maybe add `adbcommand()` that would do `xharnesscommand() + ['android', 'adb', '--']`

PR #2359, opened 2022-04-08
Add win forms startup

DrewScoggins:
This is the initial bring up for WinForms startup and SOD tests.

### DIFF
--- a/eng/performance/scenarios.proj
+++ b/eng/performance/scenarios.proj
@@ -45,11 +45,19 @@
     <Scenario Include="New VB Console Template">
       <PayloadDirectory>$(ScenariosDir)emptyvbconsoletemplate</PayloadDirectory>
     </Scenario>
-    <Scenario Include="WinForms Large">
+    <UIScenario Include="WinForms Large">
       <PayloadDirectory>$(ScenariosDir)windowsformslarge</PayloadDirectory>
-    </Scenario>
+    </UIScenario>
   </ItemGroup>
 
+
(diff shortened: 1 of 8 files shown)


eng/performance/scenarios.proj
```diff
@@ -45,8 +45,12 @@
     <Scenario Include="New VB Console Template">
       <PayloadDirectory>$(ScenariosDir)emptyvbconsoletemplate</PayloadDirectory>
     </Scenario>
+    <Scenario Include="WinForms Large">
+      <PayloadDirectory>$(ScenariosDir)windowsformslarge</PayloadDirectory>
+    </Scenario>
```
LoopedBard3, line 50: Should this be conditioned on the run occurring on windows? At the very least the startup test should be as I don't think winforms runs on Linux.

eng/performance/scenarios.proj
```diff
@@ -45,8 +45,12 @@
...
   </ItemGroup>
 
+
```
LoopedBard3, line 53: (Not here specifically) It also appears that the runs are failing due to the target framework being overwritten to just net7.0, not net7.0-windows.

LoopedBard3 (changes requested):

eng/performance/scenarios.proj
LoopedBard3: idk if targets windows is the condition we want to base on, but it should work for now at least.

LoopedBard3 (approved):
LGTM

PR #2367, opened 2022-04-13
blazor_scenarios: Fix issues caused by `dotnet` being shared by all the workitems

radical:
.. because of being in the helix correlation payload directory.

1. 
```
blazor_scenarios: Add new --readonly-dotnet arg to pre, and post
.. commands, when the dotnet installation is shared, and should not be
modified. Workload install/uninstall is one such modification.

This is useful for runtime runs, where the dotnet used already has the
workload installed, and is shared by all the work items via the helix
correlation payload, and thus should not be modified.
```

2. Some improvements to ensure that native relinking is used only when intended.

### DIFF
--- a/src/scenarios/blazor/pre.py
+++ b/src/scenarios/blazor/pre.py
@@ -12,11 +12,10 @@ from test import EXENAME
 
 precommands = PreCommands()
 
-if precommands.has_workload:
-    if precommands.framework == 'net6.0' or precommands.framework == 'net7.0':
-        subprocess.run(["dotnet", "workload", "uninstall", "wasm-tools"])
-    else:
-        subprocess.run(["dotnet", "workload", "uninstall", "microsoft-net-sdk-blazorwebassembly-aot"])
+if precommands.framework == 'net6.0' or precommands.framework == 'net7.0':
+    precommands.uninstall_workload("wasm-tools")
+else:
+    precommands.uninstall_workload("microsoft-net-sdk-blazorwebassembly-aot")
 
 setup_loggers(True)
 precommands.new(template='blazorwasm',
--- a/src/scenarios/blazor/pre.py
+++ b/src/scenarios/blazor/pre.py
@@ -23,4 +23,6 @@ precommands.new(template='blazorwasm',
                 bin_dir=const.BINDIR,
                 exename=EXENAME,
                 working_directory=sys.path[0])
-precommands.execute()
+# ensure that we don't use the workload, which would trigger native
+# relinking
+precommands.execute(['-p:WasmNativeWorkload=false'])
--- a/src/scenarios/blazorminapp/pre.py
+++ b/src/scenarios/blazorminapp/pre.py
@@ -17,6 +17,9 @@ precommands.new(template='blazorwasm',
                 exename=EXENAME,
                 working_directory=sys.path[0])
 
+# This project uses native relinking
+
+# We want to use native relinking which requires the workload
 precommands.install_workload('wasm-tools')
 f = open(os.path.join(os.getcwd(), "app", "emptyblazorwasmtemplate.csproj"), 'r')
 outFileText = ""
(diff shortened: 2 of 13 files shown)


radical:
cc @adamsitnik @DrewScoggins @LoopedBard3

src/scenarios/shared/postcommands.py
```diff
@@ -1,8 +1,31 @@
...
+    '''
+    Handles any post run cleanup
+    '''
+
+    def __init__(self):
+        parser = ArgumentParser()
+
+        parser.add_argument('--readonly-dotnet',
+                            dest='readonly_dotnet',
+                            default=False,
```
DrewScoggins, line 16: nit: This is not strictly needed as the default of the `store_true` action is false.

src/scenarios/blazor/pre.py
```diff
@@ -12,16 +12,17 @@
...
 setup_loggers(True)
 precommands.new(template='blazorwasm',
                 output_dir=const.APPDIR,
                 bin_dir=const.BINDIR,
                 exename=EXENAME,
                 working_directory=sys.path[0])
-precommands.execute()
+# ensure that we don't use the workload, which would trigger native
+# relinking
+precommands.execute(['-p:WasmNativeWorkload=false'])
```
DrewScoggins, line 28: We can already pass this flag by using the `--msbuild` argument and updating the `blazor_scenarios.proj`. Was there a reason that you added this new functionality?

radical: I wanted to set here for the project, instead of passing it from whoever is building the project, to ensure that the relinking/not-relinking thing is enforced by the project itself.

src/scenarios/blazorminapp/pre.py
```diff
@@ -16,15 +16,22 @@
...
+precommands.install_workload('wasm-tools')
 f = open(os.path.join(os.getcwd(), "app", "emptyblazorwasmtemplate.csproj"), 'r')
 outFileText = ""
 for line in f.readlines():
     if "<PropertyGroup>" in line:
         outFileText += line
         outFileText += "    <BlazorEnableTimeZoneSupport>false</BlazorEnableTimeZoneSupport>" + os.linesep
         outFileText += "    <InvariantGlobalization>true</InvariantGlobalization>" + os.linesep
+        # this will trigger native relinking, and fail the build if the workload is not available
+        outFileText += "    <WasmBuildNative>true</WasmBuildNative>" + os.linesep
```
DrewScoggins, line 32: @eerhardt Just wanted you to double check that this is the right goo to put in the csproj so that we are doing the minapp scenario correctly.

src/scenarios/blazor/pre.py
DrewScoggins: OK. I think that this is a fine way to do it for now, but I would like to think deeper on how we can do this in a more general way.

src/scenarios/blazorminapp/pre.py
eerhardt: Why didn't we need these properties before?

radical: Background:

- we recently started building these blazor apps with a sdk that has the `wasm-tools` workload (based on artifacts from a runtime build) already installed.
- Before this PR, the pre/post scripts would install/uninstall the workload, but we don't want to do that

When publishing a blazor project:
1. with the workload installed, native-relinking is used
2. without the workload, no native-relinking

- Using `WasmNativeWorkload=false` can simulate a workload not being available, so we use that to get behavior (2).

- For blazorminapp project we want relinking, for which we don't need to set anything extra.
  - BUT if due to any problem the sdk doesn't have the workload installed, then `publish` will still build successfully, but now silently not using native relinking.
  - Setting `WasmBuildNative=true` causes the workload to be *required*, thus failing the build if a workload is not installed.
  - So, essentially, this is just to enforce the intention of relinking

DrewScoggins (approved):
LGTM, seems you just took the changes from main.

radical:
> LGTM, seems you just took the changes from main.

Sorry, I didn't understand that. Do you mean that I should merge main with this?

DrewScoggins:
Sorry, I just meant that the only changes in the most recent commit were the changes that came from main.

radical:
Could somebody please merge this? I don't have the permissions.

PR #2372, opened 2022-04-13
ResultsComparer improvements

adamsitnik:
* Updated ordering (arm64 is now more important than x64)
* Stop displaying base and diff runtime versions for each benchmark, print preview number in the Legend
* For results recognized as "Noise", display them as "Noise" (not "Same" like we did so far), and use `-` in the ratio column


dakersnar (commented):
Looks good. I'm a fan of the "noise" addition.

PR #2374, opened 2022-04-14
use machine architecture as default SDK architecture

adamsitnik:
we can use [platform.machine()](https://docs.python.org/3/library/platform.html) to get default architecture and don't require the users to specify architectures other than `x64` (current default) in explicit way.

cc @AndyAyersMS 

Tested on:

- [x] Windows 10 arm64
- [x] Windows 11 AMD x64
- [x] Windows 10 Intel x64
- [x] Ubuntu 1804 Intel x64
- [x] macOS 12.2.1 x64
- [x] ubuntu 16.04 arm64
- [x] ubuntu 18.04 arm64

### DIFF
--- a/scripts/benchmarks_monthly.py
+++ b/scripts/benchmarks_monthly.py
@@ -6,6 +6,7 @@ for a .NET preview release, making it easier to contribute to our
 monthly manual performance runs.
 '''
 
+from performance.common import get_machine_architecture
 from performance.logger import setup_loggers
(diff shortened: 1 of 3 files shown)


LoopedBard3 (approved):
LGTM, sdk_scenarios are broken on main and this shouldn't affect them negatively.

jeffhandley (approved):
Very cool!

PR #2376, opened 2022-04-14
Fix capture when time is over 1 second and added an error if more tha…

LoopedBard3:
Fix capture when time is over 1 second and added an error if more than one s (for seconds) shows up. There was a bug found where the ms time in the android capture would not be left padded causing the removal of the second to become that paddings. For example, 1s15ms -> 115 when it should be 1015.


LoopedBard3:
This should go in ASAP.

DrewScoggins (approved):
LGTM

PR #2377, opened 2022-04-14
Improved error conditions when running android startup. This included…

LoopedBard3:
Improved error conditions when running android startup. This included changing sys.exit(1)s to Exceptions with messages so that we don't just quit out of the run if being run by a higher script, and put the bulk of the test in a try block so that set parameters are always reset if possible.


DrewScoggins (approved):
LGTM

PR #2378, opened 2022-04-15
Support an argument to resume MicroBenchmarks from a previous run

jeffhandley:
I've long wanted to be able to interrupt a microbenchmarks run and then run it again to resume from where it left off. This will be especially useful for running manual runs, enabling folks to start a run at night, and if it's not complete before the next morning, the run can be resumed the next night.

To achieve this:

1. The MicroBenchmarks console application recognizes a `--resume` argument
    * When specified, all existing report files are iterated
    * For each report file, the file is renamed to have `-resume` added into the file name, avoiding potential collisions while doing so
    * Then for each JSON report, the benchmarks are deserialized from the report file, and the distinct set of completed benchmarks is collected
    * That list of completed benchmarks is turned into a series of `--exclusion-filter` arguments to exclude those benchmarks from the run
2. The `micro_benchmarks.py` script recognizes and relays the `--resume` argument to `dotnet run`
3. The `benchmarks_ci.py` script recognizes the `--resume` argument so it can be relayed to `micro_benchmarks.py`
4. The `benchmarks_monthly.py` script recognizes the `--resume` argument
    * When specified, the existing results are not deleted before the run
    * And then the argument is relayed to `benchmarks_ci.py`

I've contemplated a few different ways to achieve this. This approach is my first working attempt. I'm open to alternative approaches if others have different ideas for how to accomplish this.

### DIFF
--- a/scripts/benchmarks_monthly.py
+++ b/scripts/benchmarks_monthly.py
@@ -75,7 +75,13 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
         '--no-clean',
         dest='no_clean',
         action='store_true',
-        help='Do not clean the SDK and results directories before execution')
+        help='Do not clean the SDK installations before execution')
+
+    parser.add_argument(
+        '--resume',
+        dest='resume',
+        action='store_true',
+        help='Resume a previous run from existing benchmark results')
 
     parser.add_argument(
         '--dry-run',
@@ -129,7 +135,7 @@ def __main(args: list) -> int:
         resultsPath = os.path.join(rootPath, 'artifacts', 'bin', 'MicroBenchmarks', 'Release', moniker, 'BenchmarkDotNet.Artifacts', 'results')
 
         if not args.no_clean:
-            # Delete any preexisting SDK and results, which allows
+            # Delete any preexisting SDK installations, which allows
             # multiple versions to be run from a single command
             if os.path.isdir(sdkPath):
                 log('rmdir -r ' + sdkPath)
@@ -137,16 +143,19 @@ def __main(args: list) -> int:
                 if not args.dry_run:
                     shutil.rmtree(sdkPath)
 
+        benchmarkArgs = ['--skip-logger-setup', '--filter', args.filter, '--architecture', args.architecture, '-f', version['tfm']]
+
+        if 'build' in version:
+            benchmarkArgs += ['--dotnet-versions', version['build']]
+
+        if args.resume:
+            benchmarkArgs += ['--resume']
+        else:
             if os.path.isdir(resultsPath):
                 log('rmdir -r ' + resultsPath)
 
                 if not args.dry_run:
                     shutil.rmtree(resultsPath)
-
-        benchmarkArgs = ['--skip-logger-setup', '--filter', args.filter, '--architecture', args.architecture, '-f', version['tfm']]
-
-        if 'build' in version:
-            benchmarkArgs += ['--dotnet-versions', version['build']]
 
         if args.bdn_arguments:
             if version['tfm'].startswith('nativeaot'):
--- a/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
@@ -99,7 +99,7 @@ namespace BenchmarkDotNet.Extensions
                 return new string[0];
 
             // Get all existing report files, of any export type; order by descending filename length to avoid rename collisions
-            return artifacts.GetFiles($"*-report-*", SearchOption.AllDirectories)
+            var existingBenchmarks = artifacts.GetFiles($"*-report-*", SearchOption.AllDirectories)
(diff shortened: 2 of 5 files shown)


danmoseley:
I haven't thought about this yet, but would it make sense for BDN to support --resume? Then less work is needed in these scripts perhaps.

dakersnar:
Not super familiar with this codebase yet, but from what I can tell this looks useful. 

How does a user pause a current run? With ^C? If so, can we guarantee the result files will be left in a "stable" state such that the run can be resumed, considering ^C can be pressed at any point?

src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
```diff
@@ -83,5 +92,63 @@ private static DisassemblyDiagnoser CreateDisassembler()
...
+                            var benchmarks = result.Benchmarks.Select(benchmark =>
+                            {
+                                var nameParts = new[] { benchmark.Namespace, benchmark.Type, benchmark.Method };
+                                return string.Join(".", nameParts.Where(part => !string.IsNullOrEmpty(part)));
+                            }).Distinct();
+
+                            return benchmarks;
+                        }
+                        catch (JsonSerializationException)
+                        {
```
danmoseley, line 134: If the last test run ended abruptly such that it left invalid JSON, it seems that we will just fail to resume. 

Another option is, if there's such an exception, grep the file for a distinctive token such as `BenchmarkDotNetVersion` to confirm it's a relevant JSOn file. If it's present, then emit the JSON exception, which will identify the corrupted file.

scripts/benchmarks_monthly.py
```diff
@@ -75,7 +75,13 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
...
         dest='no_clean',
         action='store_true',
-        help='Do not clean the SDK and results directories before execution')
+        help='Do not clean the SDK installations before execution')
+
+    parser.add_argument(
+        '--resume',
+        dest='resume',
+        action='store_true',
+        help='Resume a previous run from existing benchmark results')
```
danmoseley, line 84: attempt to .. ?

danmoseley (approved):

jeffhandley:
> I haven't thought about this yet, but would it make sense for BDN to support --resume? Then less work is needed in these scripts perhaps.

Perhaps. I'd like to hear @adamsitnik's thoughts on that. Since BDN is extensible in terms of the results exporters, I wasn't sure I could get it to universally be able to resume, since this is relying on the JSON exporter, which these scripts can assume are being produced. But we could also have BDN lay down a separate artifact that records the benchmarks that are being executed and record progress against them.

> How does a user pause a current run?

Yeah, `^C` will kill the run. But I have also been plagued with machines spontaneously rebooting, despite Windows Update being paused--and this addresses that too. It also addresses the situation when a benchmark hangs and the run has to be terminated one way or another. I've been using this with the following command to resume where the run left off, while skipping the benchmarks that were hanging:

```ps
py scripts/benchmarks_monthly.py net7.0-preview3 --device-name jeffhandley-pc-win11 --resume --no-clean --bdn-arguments "--exclusion-filter '*FileStream*'"
``

src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
jeffhandley: What I was thinking here is that if the run ended abruptly and invalid JSON was left behind, then we will discard those completed benchmarks, allowing them to run again and produce valid JSON.

scripts/benchmarks_monthly.py
jeffhandley: It _will_ resume from the previous results--running any benchmarks where there aren't existing, valid results. It's possible it will re-run some, but only when their results were incomplete/corrupt/invalid.

jeffhandley:
I've been using this the past couple days with much success. It's already improved the following scenarios for me:

1. A machine that insists on rebooting spontaneously, allowing it to continue from where it was interrupted
2. The FileStream benchmarks hanging and needing to halt the process, resume, but skip those benchmarks (`--resume --bdn-arguments "--exclusion-filter '*FileStream*'")
3. Having a run on one Ubuntu version that was 30 minutes from being complete, needing to leave home for the day, and wanting to start the other Ubuntu version before I left
    * In this case, I hit `^C`, stopped the run with 30 minutes left, then I launched the run on the other WSL distro, let it run for the day and finish, and then resume the other for it to finish its last 30 minutes

We can consider this pretty well tested and certainly valuable. It's just a matter of whether it's implemented with the best approach, or if we should move it to BDN or another layer.

danmoseley:
Sounds good - I'll use it

jeffhandley:
@adamsitnik / @dakersnar -- could this be merged before we kick off the preview 4 run? We could then entertain refactoring this down into BenchmarkDotNet in the future.

dakersnar (approved):
Seems like a very useful feature. I'm not knowledgeable enough to judge the implementation details of this change, but your high-level descriptions seem sound.

PR #2383, opened 2022-04-19
Add support for the s390x architecture

stefan-sf-ibm:
Since .NET 6 architecture s390x is officially supported. Thus, adding support for s390x here, too.

### DIFF
--- a/src/scenarios/shared/util.py
+++ b/src/scenarios/shared/util.py
@@ -47,6 +47,8 @@ def getruntimeidentifier():
 
     if 'aarch64' in platform.machine() or os.environ.get('PERFLAB_BUILDARCH') == 'arm64':
         rid += 'arm64'
+    elif platform.machine() == 's390x':
+        rid += 's390x'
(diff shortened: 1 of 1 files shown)


adamsitnik (commented):
Hi @stefan-sf-ibm 

Big thanks for your contribution!

Do you happen to know if dotnet install scripts (https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) happen to support `s390x` as well?

If so, could you please extend the following logic as well:

https://github.com/dotnet/performance/blob/fbda0909df85ecce0c37a21c6208ded1849e34d3/scripts/performance/common.py#L22-L34

https://github.com/dotnet/performance/blob/54bab8cbbb7011dcb45b211dd724c79362a4ed4c/scripts/dotnet.py#L848-L853

stefan-sf-ibm:
The .NET SDK binaries for s390x are not distributed via Azure so the dotnet install scripts download x64 binaries which in the end fails of course.

PR #2388, opened 2022-04-20
Update GCPerfSim to enable targeting .NET > 5

asundheim:
Modifies the preprocessor directives from `#if NET5_0` to `#NET5_0_OR_GREATER` to allow targeting later versions while still enabling POH functionality.

Also adds the `-ee` alternate flag, which is mentioned a few times in the doc comment but wasn't actually an option.


cshung (approved):
Looks good to me.

PR #2389, opened 2022-04-20
Wpf startup

DrewScoggins:
This adds WPF startup capabilities to the performance lab. We also add two tests:

1. dotnet new wpf
2. ShowFormComplex


LoopedBard3 (approved):
LGTM, assuming it runs

src/tools/ScenarioMeasurement/Startup/WPFParser.cs
```diff
@@ -7,17 +9,101 @@ public class WPFParser : IParser
...
+                    }
+                });
+
+                source.Dynamic.AddCallbackForProviderEvent("Microsoft-Windows-WPF", "WClientUcePresent/Stop", evt =>
+                {
+                    //Then when we see a Present/Stop after the flag has been set
+                    if (doneParsingXaml && pid.HasValue && evt.ProcessID == pid && evt.ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        lastKnownTime = evt.TimeStampRelativeMSec; //Grab the current time
+                        doneParsingXaml = false; //And reset the flag
```
LoopedBard3, line 84: It does not appear that we are using this anywhere else, am I missing something?

LoopedBard3 (changes requested):
It looks like the runs have returned invalid values for the WPF startup value:
```
 Metric         |Average        |Min            |Max
[2022/04/21 00:04:23][INFO] ---------------|---------------|---------------|---------------
[2022/04/21 00:04:23][INFO] WPF Startup    |-125.702 ms    |-125.702 ms    |-125.702 ms
[2022/04/21 00:04:23][INFO] Time on Thread |323.438 ms     |323.438 ms     |323.438 ms
```

I am assuming the startup time should not be negative? The test also only ran once, we should test with more. Looking at the code nothing obvious pops out as to why the time is negative but something is going wrong.

PR #2393, opened 2022-04-22
Maui win UI startup testing

LoopedBard3:
This adds Maui Desktop WinUI Startup testing to the pre/test/post format of scenarios. This is more of a general testing/get something working thing so comments and ideas are appreciated. Testing around the no-restore option and skip-manifest-update are still in the works but comments are useful.

### DIFF
--- a/eng/performance/scenarios.proj
+++ b/eng/performance/scenarios.proj
@@ -100,6 +100,19 @@
     </HelixWorkItem>
   </ItemGroup>
 
+  <ItemGroup>
+    <HelixWorkItem Include="SOD - Desktop Maui Unpackaged" Condition="'$(TargetsWindows)' == 'true'">
+      <PayloadDirectory>$(ScenariosDir)mauidesktop</PayloadDirectory>
+      <PreCommands>$(Python) pre.py publish -c Release -f $(PERFLAB_Framework)-windows10.0.19041.0</PreCommands>
+      <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
+    </HelixWorkItem>
+    <HelixWorkItem Include="Device Startup - Desktop Maui Default" Condition="'$(TargetsWindows)' == 'true'">
+      <PayloadDirectory>$(ScenariosDir)mauidesktop</PayloadDirectory>
+      <PreCommands>$(Python) pre.py publish -c Release -f $(PERFLAB_Framework)-windows10.0.19041.0</PreCommands>
+      <Command>$(Python) test.py startup --scenario-name &quot;%(Identity)&quot;</Command>
+    </HelixWorkItem>
+  </ItemGroup>
+
   <!--
     This is useful for local testing to print the produced helix items
     To use this when you are changing how items are produced, uncomment the target
--- a/eng/performance/scenarios.proj
+++ b/eng/performance/scenarios.proj
@@ -100,28 +100,26 @@
     </HelixWorkItem>
   </ItemGroup>
 
-  <!-- SOD & Device Startup for Windows Desktop MAUI -->
-  <ItemGroup>
-    <HelixWorkItem Include="SOD - Desktop Maui Unpackaged" Condition="'$(TargetsWindows)' == 'true'">
+  <ItemGroup Condition="'$(TargetsWindows)' == 'true'">
+    <!-- SOD & Device Startup for Windows Desktop MAUI -->
+    <HelixWorkItem Include="SOD - Desktop Maui Unpackaged">
       <PayloadDirectory>$(ScenariosDir)mauidesktop</PayloadDirectory>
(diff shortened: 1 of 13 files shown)


LoopedBard3:
Generally looking for feedback on alternatives for the no restore and skip manifest update changes.

src/tools/ScenarioMeasurement/Startup/WinUIParser.cs
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    if (!pid.HasValue) // we're currently in a measurement interval
+                        return;
+
+                    if (evt.NewProcessID != pid && evt.OldProcessID != pid)
+                        return; // but this isn't our process
+
+                    if (evt.OldProcessID == pid) // this is a switch out from our process
+                    {
+                        if (ins.TryGetValue(evt.OldThreadID, out var value)) // had we ever recorded a switch in for this thread?
```
caaavik-msft, line 55: According to [this guide](https://github.com/microsoft/perfview/blob/570cc0364c7e2407222f84600aebb9d9cb2f6007/documentation/TraceEvent/TraceEventProgrammersGuide.md), all these event handlers are fired asynchronously, so I think there are likely to be lots of race conditions here with how this code is set up. In addition, the `List` and `Dictionary` types are not thread-safe, and `+=` is also not atomic either.

At the bottom of the guide it says that if you are just parsing a file and are not handling live streams of events, you should use the `TraceLog` type instead as it lets you iterate over all the events in a for loop rather than needing to set up event handlers. I would say you should switch it to use that instead.

caaavik-msft (changes requested):

LoopedBard3:
Got a successful run: https://dev.azure.com/dnceng/public/_build/results?buildId=1812334&view=results, not sure why the check runs are not showing it.

src/tools/ScenarioMeasurement/Startup/WinUIParser.cs
caaavik-msft: I have learnt from another PR that my comment here was wrong and that it is thread-safe, so resolving the comment

caaavik-msft (approved):

eng/performance/scenarios.proj
```diff
@@ -100,6 +100,34 @@
...
   </ItemGroup>
 
+  <!-- SOD & Device Startup for Windows Desktop MAUI -->
+  <ItemGroup>
+    <HelixWorkItem Include="SOD - Desktop Maui Unpackaged" Condition="'$(TargetsWindows)' == 'true'">
+      <PayloadDirectory>$(ScenariosDir)mauidesktop</PayloadDirectory>
+      <PreCommands>$(Python) pre.py publish -c Release -f $(PERFLAB_Framework)-windows10.0.19041.0</PreCommands>
+      <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
+    </HelixWorkItem>
+    <HelixWorkItem Include="Device Startup - Desktop Maui Default" Condition="'$(TargetsWindows)' == 'true'">
```
jonathanpeppers, line 110: If you want to avoid the duplicate `Condition`, you can move it up to the `<ItemGroup>`:

```suggestion
  <ItemGroup Condition="'$(TargetsWindows)' == 'true'">
    <HelixWorkItem Include="SOD - Desktop Maui Unpackaged">
      <PayloadDirectory>$(ScenariosDir)mauidesktop</PayloadDirectory>
      <PreCommands>$(Python) pre.py publish -c Release -f $(PERFLAB_Framework)-windows10.0.19041.0</PreCommands>
      <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
    </HelixWorkItem>
    <HelixWorkItem Include="Device Startup - Desktop Maui Default">
```

eng/performance/scenarios.proj
```diff
@@ -100,6 +100,34 @@
...
+  </ItemGroup>
+
+  <!-- SOD & Device Startup for Windows Desktop MAUI Blazor -->
+  <ItemGroup>
+    <HelixWorkItem Include="SOD - Desktop Maui Blazor Unpackaged" Condition="'$(TargetsWindows)' == 'true'">
+      <PayloadDirectory>$(ScenariosDir)mauiblazordesktop</PayloadDirectory>
+      <PreCommands>$(Python) pre.py publish -c Release -f $(PERFLAB_Framework)-windows10.0.19041.0</PreCommands>
+      <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
+    </HelixWorkItem>
+    <HelixWorkItem Include="Device Startup - Desktop Maui Blazor Default" Condition="'$(TargetsWindows)' == 'true'">
```
jonathanpeppers, line 124: Same here, you can move the `Condition` up.

jonathanpeppers (approved):
I don't think I can comment on any of the python code, but the rest LGTM. 👍

PR #2398, opened 2022-04-26
Add upload return value of 0

LoopedBard3:
Add return value of 0 so that upload code checks work as they were written. Python by default returns None type instead of 0.


LoopedBard3:
If you search for calls to upload, some store the returned value and assume 0 is success, but we haven't been returning zero as the python default return value is None.

DrewScoggins (approved):
LGTM

PR #2402, opened 2022-04-29
Perf update framework var name

LoopedBard3:
We are seeing warnings about _Framework being a read-only variable in the pipelines. Currently they still allow overwriting, but it may change in the future. This is an initial attempt to get ahead of that.


DrewScoggins (approved):
LGTM

PR #2406, opened 2022-05-03
Mark these as NoAOT to allow for an AOT Arm64 run from Nov 2021

DrewScoggins:
This change is temporary and should be reverted once the run is under way.


LoopedBard3 (approved):
LGTM

PR #2408, opened 2022-05-03
Fix UnsupportedOS issue

DrewScoggins:
This was introduced by https://github.com/dotnet/runtime/pull/68432. We need to mark the
test as not runnable on the platforms where we are missing this property


LoopedBard3 (approved):
LGTM presuming it fixes the issue.

DrewScoggins:
I tested this locally and it seemed to work. Going to wait for the CI to finish to make sure though.

PR #2409, opened 2022-05-03
Revert "Revert "Mark these as NoAOT to allow for an AOT Arm64 run from Nov 2021""

DrewScoggins:
Reverts dotnet/performance#2407


LoopedBard3 (approved):
LGTM for round 2

PR #2411, opened 2022-05-03
Revert "Mark these as NoAOT to allow for an AOT Arm64 run from Nov 2021

DrewScoggins:
Reverts dotnet/performance#2409


LoopedBard3:
Does this still need to be merged?

PR #2422, opened 2022-05-09
Create script to contribute data gathered using `benchmarks-monthly.py`

sblom:
This is the upload script for contributing perf data from a monthly performance run. The other piece of this system is an Azure Function that stores contributed files in an Azure Storage Blob container.

In its current state, `py perfcontrib.py filename.tar.gz` will ask you to authenticate using your work account and will then upload the BDN tgz file that you named on the command line.

Next steps include:
* Wire upload() into `benchmarks-monthly.py`.

/cc @jeffhandley @adamsitnik

### DIFF
--- a/scripts/perfcontrib.py
+++ b/scripts/perfcontrib.py
@@ -1,0 +1,78 @@
+from json import loads, dumps
+from urllib.request import urlopen, Request
+from urllib.parse import urlencode
+import time
+
+def main() -> None:
+    tenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47"
+    appId = "c2fe4cd0-be4a-468b-aa4f-078c67dcab6e"
+    uploadService = "https://perfcontrib.azurewebsites.net/"
+    
+    authUrl = f"https://login.microsoftonline.com/{tenantId}"
+    
+    # The Url to my Function. This is the Url I am trying to access. It includes the base name above and the actual Function.
+    uploadEndpoint = f"{uploadService}/api/UploadMonthlyReport"
+    
+    # Details of Url from https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-device-code
+    # I need 'profile' scope for oid
+    # I need 'openid' scope to get the id_token else that is not returned
+    authBody = {
+        "tenant": tenantId,
+        "client_id": appId,
+        "scope": "User.Read openid profile",
+    }
+    
+    authBodyEncoded = urlencode(authBody).encode()
+
+    with urlopen(Request(f"{authUrl}/oauth2/v2.0/devicecode", data = authBodyEncoded)) as response:
+        item = loads(response.read().decode('utf-8'))
+
+    print(item["message"])
+    
+    authBody2 = {
+        "tenant": tenantId, 
+        "grant_type": "urn:ietf:params:oauth:grant-type:device_code",
+        "client_id": appId,
+        "device_code": item["device_code"]
+    }
+    
+    authBody2Encoded = urlencode(authBody2).encode()
+    
+    authStatus = "waiting"
+    response2 = None
+    while (authStatus == "waiting"):
+        # Try to get the access token. if we encounter an error check the reason. 
+        # If the reason is we are waiting then sleep for some time. 
+        # If the reason is the user has declined or we timed out then quit.  
+        try:
+            with urlopen(Request(f"{authUrl}/oauth2/v2.0/token", data = authBody2Encoded)) as response:
+                item = loads(response.read().decode('utf-8'))
+            authStatus = "done"
+        except:
+            time.sleep(10)
+
+    # Based on https://docs.microsoft.com/en-us/azure/app-service/configure-authentication-customize-sign-in-out#client-directed-sign-in
+    print("Thanks.")
+    authBody3 = {
+        "access_token": item["id_token"]
+    }
+
+    authBody3Json = dumps(authBody3).encode()
+
+    with urlopen(Request(f"{uploadService}/.auth/login/aad",
+                data = authBody3Json,
+                headers = {"Content-Type": "application/json"})) as response:
+        item = loads(response.read().decode('utf-8'))
+
+    print("Uploading.")
+   
+    $response3.authenticationToken
+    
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/performance?pullRequest=2422) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/performance?pullRequest=2422) before we can accept your contribution.<br/><br/>:x: sblom [sign now](https://cla.dotnetfoundation.org/dotnet/performance?pullRequest=2422)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/performance?pullRequest=2422) it.</sub>

scripts/perfcontrib.py
```diff
@@ -0,0 +1,116 @@
...
+                    headers = { "X-ZUMO-AUTH": token })) as response:
+                print("Using cached credentials.")
+        except HTTPError as error:
+            token = None
+
+    if not token:
+        token = authenticate()
+    
+    if token:
+        with open(os.path.expanduser("~/.perfcontrib/token"), "w") as tokenfile:
```
danmoseley, line 44: I believe best practice is to never let tokens touch the disk (?) ... is this token expected to be public? In which case, protecting it obviously doesn't matter.

sblom: That's a good bias; here are some thoughts regarding this specific case. This token isn't particularly sensitive (see **Background** below). Not storing it at all means that perf friends will have to log in to AAD using the **Device Token flow** (see below) for each monthly perf run they wish to contribute.

If we're willing to complicate `benchmarks-monthly.py` to require installation of python dependencies, `msal` and `msal-extensions` (which is what the Azure CLI uses for its similar login flow) has an implementation of cross-platform data encryption for its token store (DPAPI on Windows, KeyChain on macOS, etc).

A wilder option, I suppose, would be to find a way to persist this token in an in-memory service for the duration of a couple of perf data contributions.

For what it's worth, users will have to log in each month on each device from which they're contributing data, so it's entirely possible that this is the only exceptional month where most people are uploading two or more datasets for which this token cache would have any impact.

**Background**
The token in question is a user- (and device-) specific jwt that authenticates you to only the perfcontrib endpoint for 30 days. It's issued and managed by Azure Funtions ASP.NET-style EasyAuth (in this case, kinda like a session cookie, but for a CLI).

Here's an example of the complete claims that it contains:
```json
{
  "stable_sid": "sid:153afaa8a107203cd2eb9d6aeb2f40ca",
  "sub": "sid:4e1d5eff56c861acbad180778084720d",
  "idp": "aad",
  "ver": "3",
  "nbf": 1652119044,
  "exp": 1654711044,
  "iat": 1652119044,
  "iss": "https://perfcontrib.azurewebsites.net/",
  "aud": "https://perfcontrib.azurewebsites.net/"
}
```

**Device Token flow**
image

LoopedBard3 (approved):

scripts/perfcontrib.py
sblom: > Is it possible to shorten the validity time that these tokens are valid when creating them?

Perhaps. I'll investigate that on the Azure Functions EasyAuth side of things.

scripts/perfcontrib.py
```diff
@@ -0,0 +1,115 @@
...
+        "tenant": tenantId, 
+        "grant_type": "urn:ietf:params:oauth:grant-type:device_code",
+        "client_id": appId,
+        "device_code": devicecodeResponse["device_code"]
+    }
+    
+    authBody2Encoded = urlencode(authBody2).encode()
+    
+    authStatus = "waiting"
+    while (authStatus == "waiting"):
```
caaavik-msft, line 73: Should there be a max number of iterations so this doesn't go forever?

scripts/perfcontrib.py
```diff
@@ -0,0 +1,115 @@
...
+        # Try to get the access token. if we encounter an error check the reason.
```
caaavik-msft, line 74: I don't see any code here that checks the reason if there is an error, but it does make sense to be added if it does provide useful error information on failures.

scripts/perfcontrib.py
```diff
@@ -0,0 +1,115 @@
...
+        # Try to get the access token. if we encounter an error check the reason. 
+        # If the reason is we are waiting then sleep for some time. 
+        # If the reason is the user has declined or we timed out then quit.  
+        try:
+            with urlopen(Request(f"{aadUrl}/oauth2/v2.0/token", data = authBody2Encoded)) as response:
+                tokenResponse = loads(response.read().decode('utf-8'))
+            authStatus = "done"
+        except Exception as ex:
+            print("waiting")
+            time.sleep(10)
```
caaavik-msft, line 83: What kinds of failures are we expecting here? Are they just transient failures that we just need to keep retrying until it works? Or are there other failures modes to consider that sleeping and retrying doesn't make sense?

caaavik-msft (approved):

jeffhandley (approved):
Thanks for getting us moving in this direction, @sblom!

scripts/perfcontrib.py
sblom: The only failure I'm really expecting here is that the user hasn't finished logging in using the device code yet.

The design of the device code flow is to print out the message something like:

> To sign in, use a web browser to open then page https://microsoft.com/devicelogin and enter the code XXXXXYYYYZZ to authenticate

and then wait for the user to complete that flow.

Since this is a script running in the terminal of a technical audience, "press Ctrl-C" is the only other terminating condition that I'm realistically expecting, and that's probably not worth handling explicitly.

Will tighten up the comments.

scripts/perfcontrib.py
sblom: I think I'll use a loop count of {Ctrl-C}. :)

Setting a cap here means that the user is running against an arbitrary clock to get their browser opened, pointed at the device code page, and logged in. I don't think it buys us any improvements to user experience.

PR #2427, opened 2022-05-11
[ResultsComparer] make sure the data is not duplicated

adamsitnik:


dakersnar (commented):
Thanks for the fix, Adam!

PR #2430, opened 2022-05-11
[ResultsComparer] ensure correctness of the decompressed data

adamsitnik:
* define file name based on serialized title, as we can't rely on TarArchiveEntry.Key which is limited to 100 chars and incomplete for some files
* no conflicts (use more recent results)

More details:

So far to get all `.json` files from given archive we were using `entry.Key`:

https://github.com/dotnet/performance/blob/b2dfbc4051b30b700116f0e57fc66daaa8bd2c0a/src/tools/ResultsComparer/Data.cs#L55

the problem is that it's limited to 100 characters (a bug), and not every `.json` file was ending with `.json`. Example:

```log
2022-04-01-00-44-nativeaot6.0/Microsoft.AspNetCore.Server.Kestrel.Performance.PipeThroughputBenchmar
```

I did not want to spend too much time on it, but it's most likely caused by these lines:

https://github.com/adamhathcock/sharpcompress/blob/3d0a18b05dbd6260212557fd2d3d217b17ecb82f/src/SharpCompress/Common/Tar/Headers/TarHeader.cs#L122

the solution is to open the archive stream, deserialize it and read all the data on our own. In case of conflict, choose more recent results. (this solves the problem with duplicated from `--resume` or running for preview3 using different BenchmarkDotNet versions)

### DIFF
--- a/src/tools/ResultsComparer/Data.cs
+++ b/src/tools/ResultsComparer/Data.cs
@@ -22,22 +22,20 @@ namespace ResultsComparer
         {
             using Stream mainZipStream = zip.OpenRead();
             using ZipArchive mainArchive = ZipArchive.Open(mainZipStream);
-            string folderPath;
-            HashSet<string> savedFiles = new();
+            Dictionary<string, BdnResult> savedFiles = new();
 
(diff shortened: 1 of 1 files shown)


dakersnar (approved):
Thanks Adam, I'll regenerate the data once this merges!

PR #2432, opened 2022-05-11
System.Formats.Tar benchmarks

carlossanlop:
System.Formats.Tar was added in net7.0-preview4 so it's time to add benchmarks.

These benchmarks call `TarFile` methods. They internally call `TarReader` and `TarWriter`, so I didn't think it would be necessary to create benchmarks exclusive for those types.

@adamsitnik @dakersnar @jeffhandley

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -111,7 +111,6 @@
     <Compile Remove="runtime\Math\Functions\Double\Atanh.cs" />
     <Compile Remove="runtime\Math\Functions\Double\Cbrt.cs" />
     <Compile Remove="runtime\Math\Functions\Single\**\*.cs" />
-    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '3.1')">
@@ -133,7 +132,6 @@
     <Compile Remove="libraries\Common\ConnectedStreams.cs" />
     <Compile Remove="libraries\Common\StreamBuffer.cs" />
     <Compile Remove="libraries\Common\MultiArrayBuffer.cs" />
-    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '5.0')">
@@ -142,7 +140,6 @@
     <Compile Remove="libraries\System.Text.Json\Serializer\WritePreservedReferences.cs" />
     <Compile Remove="libraries\System.Net.Security\SslStreamTests.Context.cs" />
     <Compile Remove="libraries\System.Formats.Cbor\*.cs" />
-    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0')">
@@ -150,6 +147,9 @@
     <Compile Remove="libraries\Common\AlignedMemory.cs" />
     <Compile Remove="libraries\System.IO.FileSystem\Perf.RandomAccess.cs" />
     <Compile Remove="libraries\System.IO.FileSystem\Perf.RandomAccess.NoBuffering.cs" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '7.0')">
     <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -56,6 +56,7 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemVersion)" />
     <PackageReference Include="System.Drawing.Common" Version="$(SystemVersion)" />
     <PackageReference Include="System.Formats.Cbor" Version="$(SystemVersion)" Condition="'$(TargetFramework)' == 'net5.0' or '$(TargetFramework)' == 'net6.0' or '$(TargetFramework)' == 'net7.0'" />
+    <PackageReference Include="System.Formats.Tar" Version="$(SystemVersion)" Condition="'$(TargetFramework)' == 'net7.0'" />
     <PackageReference Include="System.IO.Pipelines" Version="$(SystemVersion)" />
     <PackageReference Include="System.Memory" Version="4.5.4" />
     <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
@@ -111,6 +112,7 @@
     <Compile Remove="runtime\Math\Functions\Double\Atanh.cs" />
     <Compile Remove="runtime\Math\Functions\Double\Cbrt.cs" />
     <Compile Remove="runtime\Math\Functions\Single\**\*.cs" />
+    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '3.1')">
@@ -132,6 +134,7 @@
     <Compile Remove="libraries\Common\ConnectedStreams.cs" />
     <Compile Remove="libraries\Common\StreamBuffer.cs" />
     <Compile Remove="libraries\Common\MultiArrayBuffer.cs" />
+    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '5.0')">
@@ -140,6 +143,7 @@
     <Compile Remove="libraries\System.Text.Json\Serializer\WritePreservedReferences.cs" />
     <Compile Remove="libraries\System.Net.Security\SslStreamTests.Context.cs" />
     <Compile Remove="libraries\System.Formats.Cbor\*.cs" />
+    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0')">
@@ -147,6 +151,7 @@
     <Compile Remove="libraries\Common\AlignedMemory.cs" />
     <Compile Remove="libraries\System.IO.FileSystem\Perf.RandomAccess.cs" />
     <Compile Remove="libraries\System.IO.FileSystem\Perf.RandomAccess.NoBuffering.cs" />
+    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(OS)' != 'Windows_NT'">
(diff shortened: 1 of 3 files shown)


src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -56,6 +56,7 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemVersion)" />
     <PackageReference Include="System.Drawing.Common" Version="$(SystemVersion)" />
     <PackageReference Include="System.Formats.Cbor" Version="$(SystemVersion)" Condition="'$(TargetFramework)' == 'net5.0' or '$(TargetFramework)' == 'net6.0' or '$(TargetFramework)' == 'net7.0'" />
+    <PackageReference Include="System.Formats.Tar" Version="$(SystemVersion)" Condition="'$(TargetFramework)' == 'net7.0'" />
```
tannergooding, line 59: Just checking locally, it looks like this package reference is implicitly available just by targeting .NET 7 Preview 4

carlossanlop: Thanks for the help offline. This was causing the restore problem. Removing the whole line fixed it.

Now I got other errors. Investigating them...

carlossanlop: Ok the other errors were just build failures I couldn't see before, or exceptions thrown due to incorrect setup/cleanup. All fixed now.

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -140,13 +142,15 @@
...
     <Compile Remove="libraries\System.Formats.Cbor\*.cs" />
+    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0')">
     <Compile Remove="libraries\System.Collections\PriorityQueue\Perf_PriorityQueue.cs" />
     <Compile Remove="libraries\Common\AlignedMemory.cs" />
     <Compile Remove="libraries\System.IO.FileSystem\Perf.RandomAccess.cs" />
     <Compile Remove="libraries\System.IO.FileSystem\Perf.RandomAccess.NoBuffering.cs" />
+    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
```
adamsitnik, line 153: this compile remove does not work for net6.0. I think it would be better to just exclude it for everything that is prior to .NET 7:

```xml
  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '7.0')">
    <Compile Remove="libraries\System.Formats.Tar\*.cs" />
  </ItemGroup>
```

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
```diff
@@ -0,0 +1,93 @@
...
+        private static readonly string _inputDirPath = Path.Combine(_rootDirPath, "inputdir");
+        private static readonly string _outputDirPath = Path.Combine(_rootDirPath, "outputdir");
+        private static readonly string _testDirPath = Path.Combine(_inputDirPath, "testdir");
+        private static readonly string _testFilePath = Path.Combine(_inputDirPath, "file.txt");
+        private static readonly string _inputTarFilePath = Path.Combine(_rootDirPath, "input.tar");
+        private static readonly string _outputTarFilePath = Path.Combine(_rootDirPath, "output.tar");
+        private MemoryStream _memoryStream = null;
+
+
+        // Setup and Cleanup
```
adamsitnik, line 35: nit: this comment seems to be redundant

```suggestion
```

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
```diff
@@ -0,0 +1,93 @@
...
+        public void TarFile_CreateFromDirectory_Path() => TarFile.CreateFromDirectory(sourceDirectoryName: _inputDirPath, destinationFileName: _outputTarFilePath, includeBaseDirectory: false);
+
+        [Benchmark]
+        public void TarFile_CreateFromDirectory_Stream() => TarFile.CreateFromDirectory(sourceDirectoryName: _inputDirPath, destination: _memoryStream, includeBaseDirectory: false);
+
+        [Benchmark]
+        public void TarFile_ExtractToDirectory_Path() => TarFile.ExtractToDirectory(sourceFileName: _inputTarFilePath, destinationDirectoryName: _outputDirPath, overwriteFiles: false);
+
+        [Benchmark]
+        public void TarFile_ExtractToDirectory_Stream() => TarFile.ExtractToDirectory(source: _memoryStream, destinationDirectoryName: _outputDirPath, overwriteFiles: false);
```
adamsitnik, line 80: `_memoryStream` is always going to be empty here. Should we read some real file into it and make sure that we measure more "real-life" scenario rather than edge case?

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                Directory.Delete(_outputDirPath, recursive: true);
+            }
+            Directory.CreateDirectory(_outputDirPath);
+            CleanupMemoryStream();
+            _memoryStream = new MemoryStream();
+        }
+
+        [Benchmark]
+        public void TarFile_CreateFromDirectory_Path() => TarFile.CreateFromDirectory(sourceDirectoryName: _inputDirPath, destinationFileName: _outputTarFilePath, includeBaseDirectory: false);
```
adamsitnik, line 71: the `TarFile` prefix is redundant, as these methods belong to the `Perf_TarFile` class and their ID already contains the `TarFile` word:

```diff
- System.Formats.Tar.Tests.Perf_TarFile.TarFile_CreateFromDirectory_Path
+ System.Formats.Tar.Tests.Perf_TarFile.CreateFromDirectory_Path
```

```suggestion
        public void CreateFromDirectory_Path() => TarFile.CreateFromDirectory(sourceDirectoryName: _inputDirPath, destinationFileName: _outputTarFilePath, includeBaseDirectory: false);
```

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
```diff
@@ -0,0 +1,93 @@
...
+
+
+        // Setup and Cleanup
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            Directory.CreateDirectory(_rootDirPath);
+            Directory.CreateDirectory(_inputDirPath);
+            Directory.CreateDirectory(_testDirPath);
```
adamsitnik, line 42: I have learned this very recently, but when you want to create `./a/b/c` and `./a` does not exist, then `Directory.CreateDirectory` is going to create it for you (same goes for `./a/b` and `./a/b/c`)

```suggestion
            Directory.CreateDirectory(_testDirPath);
```

adamsitnik (commented):
@carlossanlop thank you for your contribution! PTAL at my comments.

BTW to run them locally I am using the following command that downloads latest SDK and does everything for me:

```cmd
py .\scripts\benchmarks_ci.py -f net7.0 --filter *Perf_TarFile*
```

src/benchmarks/micro/MicroBenchmarks.csproj
carlossanlop: Strange. Do you know why it does not work?

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
```diff
@@ -0,0 +1,93 @@
...
+            Directory.Delete(_rootDirPath, recursive: true);
+            CleanupMemoryStream();
+        }
+        
+        [IterationSetup]
+        public void SetupIteration()
+        {
+            if (File.Exists(_outputTarFilePath))
+            {
+                File.Delete(_outputTarFilePath);
```
danmoseley, line 59: Nit, no need to check File.Exists. Delete will not throw if it doesn't. Different to Directory.Delete I believe.

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+            if (Directory.Exists(_outputDirPath))
+            {
+                Directory.Delete(_outputDirPath, recursive: true);
+            }
+            Directory.CreateDirectory(_outputDirPath);
+            CleanupMemoryStream();
+            _memoryStream = new MemoryStream();
```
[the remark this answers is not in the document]
carlossanlop: Yeah I saw the comment for `IterationSetup` to avoid using it for microbenchmarks, but I also noticed we use it _everywhere_ in microbenchmarks :D so I wanted to get feedback here, in case that documentation was obsolete somehow.

Why does `IterationSetup` make the results less stable? Is it because the performance of those methods is counted towards the final report instead of skipped?

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
carlossanlop: Strange, I was seeing an exception when attempting to delete the file but it was not found. Let me check the implementation again.

src/benchmarks/micro/MicroBenchmarks.csproj
carlossanlop: And to make sure I understand your request correctly: You want me to remove all the `Compile Remove` entries that I added to the existing ItemGroups, not just this one, and then add the new one you suggested. Correct?

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
carlossanlop: You're right, it's in the summary of that API 😄 : `Creates all directories and subdirectories in the specified path unless they already exist.`

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
carlossanlop: I'll end up removing this whole method since Adam mentioned it is not good for microbenchmarking. I'll still need to delete the file, so I'll verify what I said above.

src/benchmarks/micro/MicroBenchmarks.csproj
adamsitnik: > Do you know why it does not work?

I am not 100% sure, but I think that `&lt; '6.0'` means "everything less than 6.0", but not "equal 6.0". 

> Correct?

exactly!

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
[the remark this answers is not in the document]
adamsitnik: > Why does IterationSetup make the results less stable?

It's because when `IterationSetup` is used, BDN performs a single invocation per iteration.

So it's sth like:

```cs
IterationSetup();
Stopwatch.Start();
Benchmark();
Stopwatch.Stop();
```

while without `IterationSetup`, it performs multiple operations:

```cs
Stopwatch.Start();
for (long i = 0; i < invocationCount; i++)
   Benchmark();
Stopwatch.Stop();
```

This heavily affects [The Central Limit Theorem](https://learning.oreilly.com/library/view/pro-net-benchmarking/9781484249413/html/437795_1_En_4_Chapter.xhtml#:-:text=The%20Central%20Limit%20Theorem) which BDN relies on. And `Stopwatch` resolution is limited, so for nano-benchmarks, the difference between start and stop timestamps can be equal zero.

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
carlossanlop: Ah, that was a bug. Thanks for noticing, I'll do a File.OpenRead of the tar file created in setup.

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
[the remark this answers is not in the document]
carlossanlop: I got rid of IterationSetup, I used your suggestions.

src/benchmarks/micro/MicroBenchmarks.csproj
carlossanlop: I think that was the case. You'll let me know if it wasn't. 🙂

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
```diff
@@ -0,0 +1,69 @@
...
+        Created on demand:
+                outputdir/
+                output.tar
+        */
+        private static readonly string _rootDirPath = FileUtils.GetTestFilePath();
+        private static readonly string _inputDirPath = Path.Combine(_rootDirPath, "inputdir");
+        private static readonly string _outputDirPath = Path.Combine(_rootDirPath, "outputdir");
+        private static readonly string _testDirPath = Path.Combine(_inputDirPath, "testdir");
+        private static readonly string _testFilePath = Path.Combine(_inputDirPath, "file.txt");
+        private static readonly string _inputTarFilePath = Path.Combine(_rootDirPath, "input.tar");
```
dakersnar, line 30: I think `input.tar` is missing from your comment above

dakersnar (approved):
Looks good!

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
```diff
@@ -0,0 +1,69 @@
...
+        Always exist:
+            root/
+                inputdir/
+                        testdir/
+                        file.txt
+
+        Created on demand:
+                outputdir/
+                output.tar
+        */
```
carlossanlop, line 24:
```suggestion
        /*
        Always exist:
            root/
                inputdir/
                        testdir/
                        file.txt
                input.tar

        Created on demand:
                outputdir/
                output.tar
        */
```

src/benchmarks/micro/libraries/System.Formats.Tar/Perf.TarFile.cs
carlossanlop: Good catch. Fixed.

adamsitnik (approved):
LGTM, thank you @carlossanlop !

PR #2437, opened 2022-05-13
Add benchmark for `File.Move`

deeprobin:
This PR is used to introduce a benchmark to measure the performance of `File.Move`.
See https://github.com/dotnet/runtime/issues/65092

/cc @adamsitnik

### DIFF
--- a/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.File.cs
+++ b/src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.File.cs
@@ -29,6 +29,31 @@ namespace System.IO.Tests
         private Dictionary<int, string> _filesToRead;
         private string[] _linesToAppend;
         private Dictionary<int, string> _textToAppend;
+        private string _moveSource;
+        private string _moveDestination;
+
+        [IterationSetup(Target = nameof(Move))]
+        public void SetupMove()
+        {
+            _moveSource = FileUtils.GetTestFilePath();
+            _moveDestination = FileUtils.GetTestFilePath();
+
+            using var file = File.Create(_moveSource);
+            var data = ValuesGenerator.Array<byte>(OneMibibyte);
+            file.Write(data, 0, data.Length);
+        }
+
+        [Benchmark]
+        public void Move() => File.Move(_moveSource, _moveDestination);
+
+        [IterationCleanup(Target = nameof(Move))]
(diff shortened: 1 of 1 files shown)


carlossanlop:
@adamsitnik should we consider categorizing System.IO benchmarks by the filesystem they are running against? The results might differ a lot if the drive is NTFS, EXT4, EXFAT, etc., even within the same OS and .NET version.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.File.cs
```diff
@@ -29,6 +29,31 @@ public class Perf_File
...
+
+        [Benchmark]
+        public void Move() => File.Move(_moveSource, _moveDestination);
+
+        [IterationCleanup(Target = nameof(Move))]
+        public void CleanupMove()
+        {
+            if (File.Exists(_moveDestination))
+            {
+                File.Delete(_moveDestination);
```
danmoseley, line 54: I believe File.Delete does not need protecting with File.Exists

deeprobin: I have taken the procedure from other methods in the class.

src/benchmarks/micro/libraries/System.IO.FileSystem/Perf.File.cs
```diff
@@ -29,6 +29,31 @@ public class Perf_File
         private Dictionary<int, string> _filesToRead;
         private string[] _linesToAppend;
         private Dictionary<int, string> _textToAppend;
+        private string _moveSource;
+        private string _moveDestination;
+
+        [IterationSetup(Target = nameof(Move))]
```
adamsitnik, line 35: we try not to use IterationSetup if possible: https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#iterationsetup

In case of move benchmark, we could just swap the files:

```cs
private ulong counter;

[Benchmark]
public void Move()
{
     if (counter++ % 2 == 0)
         File.Move(_moveSource, _moveDestination);
     else
         File.Move(_moveDestination, _moveSource);
}
```

danmoseley:
@deeprobin  do you plan to address feedback?

danmoseley:
Closing as old.

Issue #2442, opened 2022-05-17
`System.Formats.Tar.Tests.Perf_TarFile` benchmarks failing on linux with `System.Formats.Tar is not supported on this platform`


radical:
These are failing on the internal perf pipeline for wasm jobs on `ubuntu.1804.amd64.tiger.perf`:

```
// BeforeAnythingElse

// Benchmark Process Environment Information:
// Runtime=.NET Core (Mono) 7.0.0-ci, Wasm NativeAOT
// GC=Non-concurrent Workstation
// Job: Job-ANWXEH(PowerPlanMode=00000000-0000-0000-0000-000000000000, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, WarmupCount=1)


System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.PlatformNotSupportedException: System.Formats.Tar is not supported on this platform.
   at System.Formats.Tar.TarFile.CreateFromDirectory(String , String , Boolean )
   at System.Formats.Tar.Tests.Perf_TarFile.Setup()
   at BenchmarkDotNet.Engines.EngineFactory.CreateReadyToRun(EngineParameters engineParameters)
   at BenchmarkDotNet.Autogenerated.Runnable_44.Run(IHost host, String benchmarkName)
   at System.Reflection.RuntimeMethodInfo.InvokeNonEmitUnsafe(Object , IntPtr* , Span`1 , BindingFlags )
   --- End of inner exception stack trace ---
   at System.Reflection.MethodInvoker.InvokeUnsafe(Object , IntPtr* , Span`1 , BindingFlags )
   at System.Reflection.RuntimeMethodInfo.Invoke(Object , BindingFlags , Binder , Object[] , CultureInfo )
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at BenchmarkDotNet.Autogenerated.UniqueProgramName.AfterAssemblyLoadingAttached(String[] args)
// AfterAll
No Workload Results were obtained from the run.
// Benchmark Process 32629 has exited with code 255.

// ** Remained 81 (64.3%) benchmark(s) to run. Estimated finish 2022-05-16 19:32 (0h 24m from now) **
// ***** BenchmarkRunner: Finish  *****

// * Export *
[...]

PR #2443, opened 2022-05-17
Disable Perf.TarFile for wasm, as the library is not supported on wasm

radical:
Fixes https://github.com/dotnet/performance/issues/2442


adamsitnik (approved):
LGTM. Thanks for the fix and apologies for breaking WASM !

PR #2445, opened 2022-05-18
Allow for uploading partial benchmark results when some runs fail

caaavik-msft:
This change ensures that if running the microbenchmarks returns a non-zero status code, that it will still upload any other successful results that it found. If there were any unsuccessful results, then the build pipeline will still be marked as failing so that any failures can be investigated.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -259,10 +259,11 @@ def __main(args: list) -> int:
 
     # Run micro-benchmarks
     if not args.build_only:
+        run_contains_errors = False
         upload_container = UPLOAD_CONTAINER
         try:
             for framework in args.frameworks:
-                micro_benchmarks.run(
+                is_success = micro_benchmarks.run(
                     BENCHMARKS_CSPROJ,
                     args.configuration,
                     framework,
@@ -270,6 +271,11 @@ def __main(args: list) -> int:
                     verbose,
                     args
                 )
+
+                if not is_success:
+                    getLogger().warn(f"Benchmark run for framework '{framework}' contains errors")
+                    run_contains_errors = True
+
             globpath = os.path.join(
                 get_artifacts_directory() if not args.bdn_artifacts else args.bdn_artifacts,
                 '**',
@@ -288,6 +294,10 @@ def __main(args: list) -> int:
             if upload_code != 0:
                 sys.exit(upload_code)
         # TODO: Archive artifacts.
+
+        # Still return 1 so that the build pipeline shows failures even though there were some successful results
+        if run_contains_errors:
+            sys.exit(1)
 
 if __name__ == "__main__":
     __main(sys.argv[1:])
--- a/src/harness/BenchmarkDotNet.Extensions/Extensions.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/Extensions.cs
@@ -17,10 +17,20 @@ namespace BenchmarkDotNet.Extensions
                 return 1;
 
             // if anything has failed, it's an error
-            if (summaries.Any(summary => summary.HasCriticalValidationErrors || summary.Reports.Any(report => !report.BuildResult.IsBuildSuccess || !report.AllMeasurements.Any())))
+            if (summaries.Any(summary => summary.HasAnyErrors()))
                 return 1;
 
             return 0;
+        }
+
+        public static bool HasAnyErrors(this Summary summary)
+        {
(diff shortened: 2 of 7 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/performance?pullRequest=2445) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/performance?pullRequest=2445) before we can accept your contribution.<br/><br/>:x: caaavik-msft [sign now](https://cla.dotnetfoundation.org/dotnet/performance?pullRequest=2445)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/performance?pullRequest=2445) it.</sub>

src/harness/BenchmarkDotNet.Extensions/Extensions.cs
```diff
@@ -17,10 +17,20 @@ public static int ToExitCode(this IEnumerable<Summary> summaries)
                 return 1;
 
             // if anything has failed, it's an error
-            if (summaries.Any(summary => summary.HasCriticalValidationErrors || summary.Reports.Any(report => !report.BuildResult.IsBuildSuccess || !report.AllMeasurements.Any())))
+            if (summaries.Any(summary => summary.HasAnyErrors()))
```
DrewScoggins, line 20: Does this widen the meaning of what we consider a failure from before? I am just not familiar with this method.

caaavik-msft: no, the method is implemented just below it, I just refactored the code a bit so that I could use it from the reporter code

src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -108,7 +110,8 @@ public string GetJson()
                 build,
                 os,
                 run,
-                tests
+                tests,
+                criticalErrors = HasCriticalValidationErrors
```
caaavik-msft, line 114: I added this line in here to say that if there are any critical validation errors, then they will be surfaced as a field "criticalErrors" in the run JSON. The value is true if there are critical errors and false if there aren't. I noticed that we have a few critical errors configured in the project such as having too many benchmarks in one category.

DrewScoggins, line 114: I am not sure that this will work like we want. If I understand correctly, if any of the individual benchmarks have a failure, than all of the result json files will have this critical errors field marked true. However, parts of the run that occurred on another machine will have this value set to false, and then depending on which result json we end up processing last will dictate whether we mark that run as having failures or not. 

To my thinking we could either fix this, by storing this information at the test level, or making a change in the uploader to respect the fact that the critical error value has been set true. I think I would prefer storing this at the test level, as tests executed on another machine that did not have errors should not really be thought of as suspect in any way.

caaavik-msft: HasCriticalValidationErrors is a field that is tied to the Summary object, not individual reports. So from what I can tell there is no way to associate it to a single test run.

caaavik-msft: Here is the source code for `HasCriticalValidationErrors`: https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Reports/Summary.cs#L74

I think it makes sense we can't associate it to a single test run, since some of the validators apply to multiple test runs (e.g. ones that check if there are too many benchmarks in one category)

DrewScoggins: OK, that makes sense, but we still will need to apply the attribute at the test level when we are putting it into the json, or we will still have the same issue of either clobbering the correct value for the critical error attribute, or treating every test from a run as having critical errors when they only occurred in a small subset of the tests.

Could solve this by just iterating over all of the tests in the GetJson method and adding this failure to the additional data fields that we already carry with each test.

scripts/benchmarks_ci.py
```diff
@@ -289,5 +295,9 @@ def __main(args: list) -> int:
                 sys.exit(upload_code)
```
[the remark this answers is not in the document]
LoopedBard3: Nvm, line above causes this to only run on failure, so it would fail either way.

LoopedBard3 (approved):
LGTM

PR #2446, opened 2022-05-19
Raise error when SOD scenario does not have any files sized

LoopedBard3:
Raise error when SOD scenario does not have any files sized. In this case, it raises a ValueError as the count of the files found is 0 where we can expect at least 1 file. To accomplish this, I kept the change in the python script side of the code as it allows for uses of the SOD tool where no files may be expected, while catching the 0 files case for our testing. This also ensures that the error gets raised properly through the python.

### DIFF
--- a/src/scenarios/shared/sod.py
+++ b/src/scenarios/shared/sod.py
@@ -70,8 +70,9 @@ class SODWrapper(object):
 
         sod_command = RunCommand(sod_args, verbose=True)
         sod_command.run()
-        zero_size_regex = f"{scenarioname} - Count\s*\|\s*0.000 count" # Checks if the overall count is zero
-        if re.match(zero_size_regex, sod_command.stdout) != None:
+        
+        zero_size_regex = rf"({(scenarioname or 'Empty Scenario Name')}\s*-\s*Count\s*\|\s*0\.000\s*count)" # Checks if the overall count is zero    
+        if re.search(zero_size_regex, sod_command.stdout) != None:
             raise ValueError(f'No files found for sizing in scenario {scenarioname}')
  
         if artifact:
--- a/src/scenarios/shared/sod.py
+++ b/src/scenarios/shared/sod.py
@@ -71,7 +71,7 @@ class SODWrapper(object):
         sod_command = RunCommand(sod_args, verbose=True)
         sod_command.run()
         zero_size_regex = f"{scenarioname} - Count\s*\|\s*0.000 count" # Checks if the overall count is zero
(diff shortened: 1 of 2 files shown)


LoopedBard3:
Failure case was tested for functionality locally and successfully through the ValueError when 0 files are counted.

src/scenarios/shared/sod.py
```diff
@@ -67,7 +68,12 @@ def runtests(self, scenarioname, dirs, artifact=None):
         ]
         sod_args += dirs.split(';')
 
-        RunCommand(sod_args, verbose=True).run()
+        sod_command = RunCommand(sod_args, verbose=True)
+        sod_command.run()
+        
+        zero_size_regex = rf"({(scenarioname or 'Empty Scenario Name')}\s*-\s*Count\s*\|\s*0\.000\s*count)" # Checks if the overall count is zero
```
caaavik-msft, line 74: The `(scenarioname or 'Empty Scenario Name')` part is duplicated here and on a previous line, maybe assign this at the start of the function instead

src/scenarios/shared/sod.py
```diff
@@ -67,7 +68,12 @@ def runtests(self, scenarioname, dirs, artifact=None):
         ]
         sod_args += dirs.split(';')
 
-        RunCommand(sod_args, verbose=True).run()
+        sod_command = RunCommand(sod_args, verbose=True)
+        sod_command.run()
+        
+        zero_size_regex = rf"({(scenarioname or 'Empty Scenario Name')}\s*-\s*Count\s*\|\s*0\.000\s*count)" # Checks if the overall count is zero    
+        if re.search(zero_size_regex, sod_command.stdout) != None:
```
caaavik-msft, line 75: I'm not sure if searching stdout for a specific log is durable enough since it means that someone who changes the log ever in the sod executable may break this script without realising it. Instead, are we able to get this information by reading any of the files that are produced as output from the sod executable?

src/scenarios/shared/sod.py
caaavik-msft, line 74: The parentheses at the start and end of the regex are redundant since the body only contains one alternative

src/scenarios/shared/sod.py
LoopedBard3: Latest push switches from the regex/console output parsing to reading the json.

src/scenarios/shared/sod.py
LoopedBard3: Changing to the json parsing removed the duplicate usage.

caaavik-msft (approved):

PR #2447, opened 2022-05-19
Add GitHub API call backoff for getting commit date

LoopedBard3:
Add GitHub API call backoff for getting commit date. We are currently hitting issues where occasionally the fetch of the commit date is not occurring successfully; the goal is to give git more time to get the proper data so we don't fail because we tried to get the data to quickly. Because of the difficulty of reproducing this error, this change is set to hopefully mitigate it when it occurs, but also provide more information for debugging if it doesn't.

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -622,7 +622,7 @@ def get_commit_date(
 
     build_timestamp = None
     item = None
-    retrycount = 0
+    retrycount = -1
     success = 0
     while success == 0 and retrycount <= 4:
         try:
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -622,9 +622,9 @@ def get_commit_date(
 
     build_timestamp = None
     item = None
-    retrycount = -1
+    retrycount = 0
     success = 0
-    while success == 0 and retrycount <= 4:
+    while success == 0 and retrycount <= 5:
         try:
             with urlopen(url) as response:
                 getLogger().info("Commit: %s", url)
@@ -634,7 +634,7 @@ def get_commit_date(
         except URLError:
             retrycount += 1
             getLogger().warning(f"URL Error trying to get commit date from {url}, Attempt {retrycount}")
-            sleep(30)
+            sleep(60)
 
     if not build_timestamp:
         if not item:
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -623,26 +623,22 @@ def get_commit_date(
     build_timestamp = None
     item = None
     retrycount = 0
-    success = 0
-    while success == 0 and retrycount <= 5:
+    sleep_time = 10 # Start with 10 second sleep timer
+    for retrycount in range(5):
         try:
             with urlopen(url) as response:
                 getLogger().info("Commit: %s", url)
                 item = loads(response.read().decode('utf-8'))
                 build_timestamp = item['commit']['committer']['date']
-                success = 1
-        except URLError:
-            retrycount += 1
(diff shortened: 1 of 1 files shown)


scripts/dotnet.py
```diff
@@ -621,9 +621,10 @@ def get_commit_date(
         url = urlformat % (owner, repo, commit_sha)
 
     build_timestamp = None
+    item = None
     retrycount = 0
     success = 0
-    while success == 0 and retrycount <= 3:
+    while success == 0 and retrycount <= 5:
```
caaavik-msft, line 627: Not really caused by the changes in this PR, but this should be replace-able with a for loop over range(6) and using a `break` instead of the `success` variable.

scripts/dotnet.py
```diff
@@ -632,10 +633,16 @@ def get_commit_date(
                 success = 1
         except URLError:
             retrycount += 1
+            getLogger().warning(f"URL Error trying to get commit date from {url}, Attempt {retrycount}")
+            sleep(60)
```
caaavik-msft, line 637: 60 seconds seems like a long time to sleep, maybe do an exponential backoff instead

scripts/dotnet.py
```diff
@@ -632,10 +633,16 @@ def get_commit_date(
...
 
     if not build_timestamp:
-        raise RuntimeError(
-            'Could not get timestamp for commit %s' % commit_sha)
+        if not item:
+            raise RuntimeError(
+                'Could not get timestamp for commit %s' % commit_sha)
+        else:
+            raise RuntimeError(
+                'Could not get timestamp for commit %s, data retrieved:\n %s' % (commit_sha, item))
```
caaavik-msft, line 645: Is this a real scenario we have seen? Trying to understand why the API would return successfully but not contain the build timestamp

LoopedBard3: Since success is not the only breakout factor currently, we may not have a timestamp if we run out of retries and quit after failing.

LoopedBard3: You are right, I had been still figuring out where the exception might be thrown and it is thrown before item gets assigned. I don't think we have seen the case with success but no datetime, although it is hard to know for sure.

scripts/dotnet.py
LoopedBard3: Reimplemented into a backoff with times (in seconds) 10, 20, 40, and 80.

LoopedBard3: This provides 2 shortish, 1 medium, and 1 long wait. Waiting the 80 secs is still faster than rerunning if it works.

scripts/dotnet.py
```diff
@@ -621,17 +621,20 @@ def get_commit_date(
         url = urlformat % (owner, repo, commit_sha)
 
     build_timestamp = None
+    item = None
```
caaavik-msft, line 624: `item = None` and `retrycount = 0` lines can be removed as they are local to the for loop. Also to make things a bit nicer to read, you could move the for loop into a function `get_build_timestamp` and that function can just return the timestamp from inside the for loop, meaning no need to have a `build_timestamp = None` variable

caaavik-msft (approved):

scripts/dotnet.py
LoopedBard3: Followed the local variable guidance, glad you caught that! For the loop, I am going to keep it as is for now with potentially changing to match the function format you brought up in the future.

LoopedBard3:
Approved, and CentOS is failing on main also, merging.

PR #2466, opened 2022-06-05
Fix the resume feature by moving the files in an independent loop

jeffhandley:
I tracked down why errors occur when launching the `benchmarks_monthly.py` script using `--resume`. I had an enumeration where I was moving files inside the enumeration over the files in the directory, and then I was filtering from within that enumeration as well.

This fix separates the enumeration, first getting the list of all report files that need to be renamed, then renaming them all, and then getting the list of JSON reports that contain benchmarks that can be skipped because of the resume. By moving the files outside that enumeration, the error no longer occurs.

### DIFF
--- a/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
@@ -99,56 +99,52 @@ namespace BenchmarkDotNet.Extensions
                 return new string[0];
 
             // Get all existing report files, of any export type; order by descending filename length to avoid rename collisions
-            var existingBenchmarks = artifacts.GetFiles($"*-report-*", SearchOption.AllDirectories)
-                .OrderByDescending(resultFile => resultFile.FullName.Length)
+	    var toRename = artifacts.GetFiles("*-report-*", SearchOption.AllDirectories)
+		    .Where(resultFile => !resultFile.FullName.Contains("-resume-report-") || File.Exists(resultFile.FullName.Replace("-resume-report-", "-report-")))
+		    .OrderByDescending(resultFile => resultFile.FullName.Length);
+
+	    foreach (var resultFile in toRename)
+            {
+		// Prepend the report name with -resume, potentially multiple times if multiple reports for the same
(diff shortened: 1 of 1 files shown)


src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
```diff
@@ -99,56 +99,52 @@ private static IEnumerable<string> GetBenchmarksToResume(DirectoryInfo artifacts
                 return new string[0];
 
             // Get all existing report files, of any export type; order by descending filename length to avoid rename collisions
-            var existingBenchmarks = artifacts.GetFiles($"*-report-*", SearchOption.AllDirectories)
-                .OrderByDescending(resultFile => resultFile.FullName.Length)
+	    var toRename = artifacts.GetFiles("*-report-*", SearchOption.AllDirectories)
+		    .Where(resultFile => !resultFile.FullName.Contains("-resume-report-") || File.Exists(resultFile.FullName.Replace("-resume-report-", "-report-")))
```
dakersnar, line 103: Is this removal of the "-resume-report-" part of the name accounting for a user resuming multiple times?

dakersnar (approved):
Looks good, thanks for the fix!

src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
jeffhandley: Yeah, this is doing:

1. Get all files matching `*-report-*`, which would include `-*resume-report-*`
2. Filter to those that _do not_ already have `-resume-report-` in their name (so they are just `-report-`), **OR** those that _do_ already have `-resume-report-` in the name, and there's also a `-report-` file sitting behind it waiting to get renamed

When there's already a `-report-` and `-resume-report`, then the `-resume-report` file will get renamed to `-resume-resume-report` and `-report-` will get renamed to `-resume-report-` as a result. But if there's a `-resume-report-` and there isn't a `-report-` to match, then we don't need to rename it to `-resume-resume-report-` because there won't be a collision.

PR #2468, opened 2022-06-06
Add release/6.0 to performance-ci runs and add Maui WinUI

LoopedBard3:
Adds release/6.0 branch to our daily and PR runs (when the PR is against release/6.0).

### DIFF
--- a/src/scenarios/shared/precommands.py
+++ b/src/scenarios/shared/precommands.py
@@ -88,7 +88,8 @@ class PreCommands:
             exename: str,
             working_directory: str,
             language: str = None,
-            no_https: bool = False):
+            no_https: bool = False,
+            no_restore: bool = True):
         'makes a new app with the given template'
         self.project = CSharpProject.new(template=template,
                                  output_dir=output_dir,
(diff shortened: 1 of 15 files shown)


LoopedBard3:
This is in part the release/6.0 version of #2393.
Pertaining to the comments left on #2393, I am going to use the same format for the parser as other parsers for now so that they are all consistent with each other, but improving the general format is definitely something to be explored.

src/scenarios/shared/precommands.py
```diff
@@ -184,6 +199,18 @@ def add_event_source(self, file: str, line: str, trace_statement: str):
         filepath = os.path.join(projpath, file)
         insert_after(filepath, line, trace_statement)
 
+    def install_workload(self, workloadid: str, install_args: list = ["--skip-manifest-update"]):
+        'Installs the workload, if needed'
+        if not self.has_workload:
```
caaavik-msft, line 204: rather than requiring the user to pass in something saying the workload has already been installed, could we check it using `dotnet workload list`?

caaavik-msft (approved):

src/scenarios/shared/precommands.py
LoopedBard3: Great find, I think that approach would work. It looks like the has-workload was added here: https://github.com/dotnet/performance/pull/2346 and may be something we can investigate modifying to be a capability check.

Issue #2454, opened 2022-05-25
MicroBenchmarks.dll fails with "The specified network password is not correct."


kunalspathak:
We are seeing failures in our superpmi collection as soon as we start executing MicroBenchmarks. Wondering if perf lab is also seeing these failures and what is the fix? This is blocking our ability to do superpmi collection on benchmarks.

```
[18:22:58] C:\h\w\97FE0875\w\9B61093D\u\performance>C:\h\w\97FE0875\w\9B61093D\u\performance\tools\dotnet\x64\dotnet C:\h\w\97FE0875\w\9B61093D\u\performance\artifacts\MicroBenchmarks.dll  --filter "*" --corerun C:\h\w\97FE0875\p\superpmi\CoreRun.exe --partition-count 30 --partition-index 11 --envVars COMPlus_JitName:superpmi-shim-collector.dll  COMPlus_ZapDisable:1  COMPlus_ReadyToRun:0 --iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart 
[18:22:58] Unhandled exception. System.TypeInitializationException: The type initializer for 'System.Net.Security.Tests.SslStreamTests' threw an exception.
[18:22:58]  ---> Internal.Cryptography.CryptoThrowHelper+WindowsCryptographicException: The specified network password is not correct.
[18:22:58]    at System.Security.Cryptography.X509Certificates.CertificatePal.FilterPFXStore(ReadOnlySpan`1 rawData, SafePasswordHandle password, PfxCertStoreFlags pfxCertStoreFlags)
[18:22:58]    at System.Security.Cryptography.X509Certificates.CertificatePal.FromBlobOrFile(ReadOnlySpan`1 rawData, String fileName, SafePasswordHandle password, X509KeyStorageFlags keyStorageFlags)
[18:22:58]    at System.Security.Cryptography.X509Certificates.X509Certificate..ctor(Byte[] rawData, String password, X509KeyStorageFlags keyStorageFlags)
[18:22:58]    at System.Net.Test.Common.Configuration.Certificates.GetCertificate(String certificateFileName) in C:\h\w\97FE0875\w\9B61093D\u\performance\src\benchmarks\micro\libraries\System.Net.Http\Configuration.Certificates.cs:line 27
[...]

PR #2469, opened 2022-06-06
Wrap certification initialization with try-catch

kunalspathak:
From last few weeks we are seeing issue in superpmi where the `MicroBenchmarks.dll` execution fails on startup. This is before it starts executing any test. The code does static fields initialization and few fields rely on certificates to be present on the machine. For some reason, this is not the case on the test machines where `superpmi collect` pipeline is run and because of this failure, we are unable to do any collection for benchmarks.

This PR wraps the certificate code with `try-catch` so even if it fails, it will just fail the respective benchmarks test instead of the entire `MicroBenchmark.dll`.

Related: https://github.com/dotnet/performance/issues/2454

### DIFF
--- a/src/benchmarks/micro/libraries/System.Net.Http/Configuration.Certificates.cs
+++ b/src/benchmarks/micro/libraries/System.Net.Http/Configuration.Certificates.cs
@@ -36,7 +36,8 @@ namespace System.Net.Test.Common
                                  certificateFileName)),
                          CertificatePassword,
                          X509KeyStorageFlags.DefaultKeySet);
-                } catch (Exception)
+                }
+                catch (Exception)
                 {
                     return null;
                 }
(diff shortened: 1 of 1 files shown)


kunalspathak:
cc: @BruceForstall 
@DrewScoggins - can we get this in so we can unblock our superpmi collections?

src/benchmarks/micro/libraries/System.Net.Http/Configuration.Certificates.cs
```diff
@@ -24,15 +24,23 @@ public static class Certificates
...
+                         File.ReadAllBytes(
+                             Path.Combine(
+                                 AppContext.BaseDirectory,
+                                 "libraries",
+                                 "System.Net.Http",
+                                 certificateFileName)),
+                         CertificatePassword,
+                         X509KeyStorageFlags.DefaultKeySet);
+                } catch (Exception)
+                {
```
gfoidl, line 40: Nit:
```suggestion
                } 
                catch (Exception)
                {
```

PR #2477, opened 2022-06-10
Reboot iOS device if we detect device is in a blocked state

akoeplinger:


LoopedBard3 (approved):
LGTM, only pipeline failure appears to be Out of Memory Exception and are not from this change.

src/scenarios/shared/runner.py
```diff
@@ -691,7 +692,37 @@ def run(self):
...
+                            'apple',
+                            'mlaunch',
+                            '--',
+                            '--rebootdev',
+                        ]
+                        rebootCmdCommand = RunCommand(rebootCmd, verbose=True)
+                        rebootCmdCommand.run()
+
+                        getLogger().info("Waiting 30 secs for the device to boot.")
+                        time.sleep(30)
```
caaavik-msft, line 713: Is there any way to check if the reboot has completed? if so it might be better to change this to a loop that checks every 5 seconds if the reboot is complete, just in case it is faster than 30 seconds.

akoeplinger: There's no good way unfortunately. Given this happens quite rarely I don't think it's an issue if we waste a few seconds when the phone reboots more quickly.

LoopedBard3:
The two errors are unrelated to this, although the changes are not tested here. If we have a successful run of these changes, given that the retry changes are external to this and may not have fixed the run the restart occurred in, I think this is good to merge to hopefully get us in a better position while waiting for the final external updates.

PR #2479, opened 2022-06-14
Base64Tests: Use valid input data

a74nh:
The tests were using random generated chars. This would cause the Decode
functions to exit early (usually after a single iteration).

### DIFF
--- a/src/benchmarks/micro/libraries/System.Buffers/Base64Tests.cs
+++ b/src/benchmarks/micro/libraries/System.Buffers/Base64Tests.cs
@@ -52,22 +52,22 @@ namespace System.Buffers.Text.Tests
         [GlobalSetup(Target = nameof(Base64Decode))]
         public void SetupBase64Decode()
         {
-            _encodedBytes = ValuesGenerator.Array<byte>(NumberOfBytes);
+            _encodedBytes = ValuesGenerator.ArrayBase64EncodingBytes(NumberOfBytes);
             _decodedBytes = new byte[Base64.GetMaxEncodedToUtf8Length(NumberOfBytes)];
         }
 
         [Benchmark]
         public OperationStatus Base64Decode() => Base64.DecodeFromUtf8(_encodedBytes, _decodedBytes, out _, out _);
 
-        [GlobalSetup(Target = nameof(Base64DecodeDetinationTooSmall))]
-        public void SetupBase64DecodeDetinationTooSmall()
+        [GlobalSetup(Target = nameof(Base64DecodeDestinationTooSmall))]
+        public void SetupBase64DecodeDestinationTooSmall()
         {
-            _encodedBytes = ValuesGenerator.Array<byte>(NumberOfBytes);
+            _encodedBytes = ValuesGenerator.ArrayBase64EncodingBytes(NumberOfBytes);
             _decodedBytes = new byte[Base64.GetMaxEncodedToUtf8Length(NumberOfBytes) - 1];
         }
 
         [Benchmark]
-        public OperationStatus Base64DecodeDetinationTooSmall() => Base64.DecodeFromUtf8(_encodedBytes, _decodedBytes, out _, out _);
+        public OperationStatus Base64DecodeDestinationTooSmall() => Base64.DecodeFromUtf8(_encodedBytes, _decodedBytes, out _, out _);
 
 #if !NETFRAMEWORK // API added in .NET Core 2.1
         [GlobalSetup(Target = nameof(ConvertTryFromBase64Chars))]
--- a/src/benchmarks/micro/libraries/System.Buffers/Base64Tests.cs
+++ b/src/benchmarks/micro/libraries/System.Buffers/Base64Tests.cs
@@ -69,6 +69,16 @@ namespace System.Buffers.Text.Tests
         [Benchmark]
         public OperationStatus Base64DecodeDestinationTooSmall() => Base64.DecodeFromUtf8(_encodedBytes, _decodedBytes, out _, out _);
 
+        [GlobalSetup(Target = nameof(Base64DecodeInvalidData))]
+        public void SetupBase64DecodeInvalidData()
+        {
+            _encodedBytes = ValuesGenerator.Array<byte>(NumberOfBytes);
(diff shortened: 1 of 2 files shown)


a74nh:
Spotted this whilst implementing https://github.com/dotnet/runtime/pull/70654 and https://github.com/dotnet/runtime/pull/70336

src/benchmarks/micro/libraries/System.Buffers/Base64Tests.cs
```diff
@@ -52,22 +52,22 @@ public void SetupConvertToBase64CharArray()
         [GlobalSetup(Target = nameof(Base64Decode))]
         public void SetupBase64Decode()
         {
-            _encodedBytes = ValuesGenerator.Array<byte>(NumberOfBytes);
```
kunalspathak, line 55: Does it make sense to also keep the ones that operates on invalid data? Just to reflect real world scenario?

kunalspathak: @DrewScoggins  should be able to confirm

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -71,6 +71,32 @@ public static T[] Array<T>(int count)
...
+
+        public static byte[] ArrayBase64EncodingBytes(int count)
+        {
+            var result = new byte[count];
+
+            var random = new Random(Seed);
+
+            for (int i = 0; i < result.Length; i++)
+            {
+                int index = (byte)random.Next(0, s_encodingMap.Length - 1);    // Do not pick '='
```
kunalspathak, line 93: Not sure I understand the comment `// Do not pick '='`. I don't see it in the `s_encodingMap`

danmoseley: I just read it as "= is often seen in base64 output, but it's just a delimiter, don't include it in the bytes"

kunalspathak: > the 2nd parameter is an exclusive upper bound

That's what I meant.

danmoseley: oh - ha - got it, we have the same observation

a74nh: Yeah, agreed that looks wrong.
I just copied the table and function from the library tests in dotnet:

https://github.com/dotnet/runtime/blob/41419131095d36fb5b811600ad0dab3b0d804269/src/libraries/System.Memory/tests/Base64/Base64TestHelper.cs#L16

https://github.com/dotnet/runtime/blob/41419131095d36fb5b811600ad0dab3b0d804269/src/libraries/System.Memory/tests/Base64/Base64TestHelper.cs#L75

I wonder if an older version of the code had an = sign on the end of the array.

Either way, I'll update this to:
int index = (byte)random.Next(0, s_encodingMap.Length);

And I guess I could fix it in dotnet by squeezing it into my intrinsic patch.

danmoseley: yeah, I was going to ask whether you'd mind fixing it there too. might be easiest to throw it up in its own PR. thanks!

src/benchmarks/micro/libraries/System.Buffers/Base64Tests.cs
danmoseley: We don't write perf tests for failure cases, because we assume they're rare, and often times just bugs that need to be fixed. This is noted in the guidelines -- https://github.com/dotnet/performance/blob/f6066b426c20a5519254b61f9b60153757dc600f/docs/microbenchmark-design-guidelines.md#L288

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
a74nh: It seemed a little small to put in it's own PR, but that's fine, I'll do that

src/benchmarks/micro/libraries/System.Buffers/Base64Tests.cs
a74nh: This isn't going to throw an exception. But, yes, all this is testing is "a small number of vector loops and then fallback to scalar"

I'll remove the failure case I added (unless anyone objects)

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
a74nh: Removed from this PR

kunalspathak (approved):
LGTM. Thanks for doing this.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -71,6 +71,32 @@ public static T[] Array<T>(int count)
...
+
+        public static byte[] ArrayBase64EncodingBytes(int count)
+        {
+            var result = new byte[count];
+
+            var random = new Random(Seed);
+
+            for (int i = 0; i < result.Length; i++)
+            {
+                int index = (byte)random.Next(0, s_encodingMap.Length);
```
danmoseley, line 93:
```suggestion
                int index = random.Next(0, s_encodingMap.Length);
```

danmoseley:
@LoopedBard3  the centos runs are failing:
```

2022/06/20 00:02:53][INFO] System.TypeInitializationException: A type initializer threw an exception. To determine which type, inspect the InnerException's StackTrace property.
[2022/06/20 00:02:53][INFO]  ---> EETypeRva:0x031F34D0(System.Reflection.MissingRuntimeArtifactException): This object cannot be invoked because no code was generated for it: 'BenchmarkDotNet.Jobs.EnvironmentMode.PlatformCharacteristic'.
[2022/06/20 00:02:53][INFO]    at System.Reflection.Runtime.FieldInfos.RuntimeFieldInfo.get_FieldAccessor() + 0x11e
```
As this is not caused by this change, nor is this coverage critical for this change, I will merge

danmoseley:
Thanks @a74nh

PR #2490, opened 2022-06-17
Add MauiVersion to Maui Desktop testing and Cleanup Maui Desktop testing

LoopedBard3:
Adds MauiVersion to Maui Desktop testing and Cleanup Maui Desktop testing.

### DIFF
--- a/src/tools/Reporting/Reporting/Reporter.cs
+++ b/src/tools/Reporting/Reporting/Reporter.cs
@@ -89,6 +89,12 @@ namespace Reporting
                 TimeStamp = DateTime.Parse(environment.GetEnvironmentVariable("PERFLAB_BUILDTIMESTAMP")),
             };
             build.AdditionalData["productVersion"] = environment.GetEnvironmentVariable("DOTNET_VERSION");
+
+            // Additional Data we only want populated if it is available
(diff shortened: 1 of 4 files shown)


src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -89,6 +89,12 @@ private void Init()
                 TimeStamp = DateTime.Parse(environment.GetEnvironmentVariable("PERFLAB_BUILDTIMESTAMP")),
             };
             build.AdditionalData["productVersion"] = environment.GetEnvironmentVariable("DOTNET_VERSION");
+
+            // Additional Data we only want populated if it is available
+            if (environment.GetEnvironmentVariable("MAUI_VERSION") != null)
```
caaavik-msft, line 94: nit: would be cleaner using a pattern match rather than a null check I think
```csharp
if (environment.GetEnvironmentVariable("MAUI_VERSION") is string mauiVersion)
{
    build.AdditionalData["mauiVersion"] = mauiVersion;
}
```

caaavik-msft (approved):

LoopedBard3:
These runs seem to be working again based on this morning's pipeline runs, so the failure should not keep occurring once merged.

PR #2492, opened 2022-06-20
[wasm] Fix wasm perf runs

radical:
Disable strict version checks for emsdk. Because of the way runtime
consumes emsdk updates right now, it can get into a situation where the
emsdk version doesn't match between runtime, and the workload pack.

This will fix the perf wasm jobs failing with: 
`error : Emscripten version mismatch. The runtime pack in /home/helixbot/work/9E1608B3/p/dotnet/packs/Microsoft.NETCore.App.Runtime.Mono.browser-wasm/7.0.0-ci expects 'emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.7 (48a16209b1a0de5efd8112ce6430415730008d18)', but emcc being used has version 'emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.12 (38d1292ba2f5b4a7c8518931f5ae6f97ef0f6827)'. This might cause build failures. `

cc @adamsitnik @LoopedBard3 @DrewScoggins

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.Wasm.targets
+++ b/src/benchmarks/micro/MicroBenchmarks.Wasm.targets
@@ -4,7 +4,10 @@
     <WasmBuildOnlyAfterPublish>true</WasmBuildOnlyAfterPublish>
 
     <!-- this will fail the build if the emcc versions don't match -->
-    <_WasmStrictVersionMatch>true</_WasmStrictVersionMatch>
+    <!-- disabling this because of the way runtime consumes emsdk updates,
+         it can get into a situation where the emsdk version doesn't match
+         between runtime, and the workload pack. -->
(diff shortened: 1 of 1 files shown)


radical:
https://github.com/dotnet/runtime/pull/70991 fixes the mismatch, but the change here will help when it gets out of sync again, which happens sometimes due to all the updates not happening together.

LoopedBard3 (approved):
LGTM, the pipeline error does not appear to be related as it is occurring in other branches.

caaavik-msft (approved):

adamsitnik (approved):

adamsitnik:
Should this PR be merged or is there something that needs to be done first?

PR #2515, opened 2022-06-28
Add `mauiblazordesktop` scenario

TanayParikh:
~~Adds the `mauiblazordesktop` scenario. How can we integrate with the `__MAUI_Blazor_WebView_OnAfterRender__` console "magic string", as we're doing on [other platforms](https://github.com/dotnet/runtime/pull/68685/files#diff-58d65effbc166c8c040b8e44f90ce3504a61b52916dca244c60b91ed8b547a41R239)?~~

~~We can probably update `pre.py` to actually write out the `__MAUI_Blazor_WebView_OnAfterRender__` ([this is how we do it for other platforms](https://github.com/dotnet/runtime/blob/dcb62e3d07fe29629abd63721002824e9d3fa7e2/eng/pipelines/coreclr/templates/build-perf-maui-apps.yml#L222-L244)) ([we'll probably have to do something Windows compatible](https://stackoverflow.com/a/23530712)). We'll need to update the collection logic to ensure it only considers the app started once that string is printed to the console.~~

### DIFF
--- a/eng/performance/scenarios.proj
+++ b/eng/performance/scenarios.proj
@@ -76,6 +76,7 @@
     </HelixWorkItem>
   </ItemGroup>
 
+  <!-- SOD & Device Startup for Windows Desktop MAUI -->
   <ItemGroup>
     <HelixWorkItem Include="SOD - Desktop Maui Unpackaged" Condition="'$(TargetsWindows)' == 'true'">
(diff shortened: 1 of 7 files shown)


TanayParikh:
Closing out in favor of https://github.com/dotnet/performance/pull/2526.

PR #2520, opened 2022-07-05
Add support for listing uploads in perfcontrib

sblom:


sblom:
> I'm assuming this only lists uploads from the user, right?

Yep.

PR #2540, opened 2022-07-29
Adding benchmarks for the new single/double math APIs

tannergooding:
These benchmarks follow the existing pattern used for the other math tests. This pattern was setup to avoid noise and other issues that BDN currently has with such small/highly optimized functions (such as https://github.com/dotnet/BenchmarkDotNet/issues/1802).


src/benchmarks/micro/runtime/Math/Functions/Single/SinCosPi.cs
```diff
@@ -0,0 +1,43 @@
...
+        public void SinCosPi() => SinCosPiTest();
+
+        public static void SinCosPiTest()
+        {
+            float resultX = 0.0f, resultY = 0.0f, value = -1.57079633f;
+
+            for (int iteration = 0; iteration < MathTests.Iterations; iteration++)
+            {
+                value += sinCosPiDelta;
+                // (float sinResult, float cosResult) = float.SinCosPi(value);
```
tannergooding, line 28: These functions should be updated to use the actual `SinCosPi` function once we move onto the next preview.

It is beneficial for them to exist beforehand to show the difference between calling individually vs the helper that reduces overall logic.

DrewScoggins (approved):
LGTM

src/benchmarks/micro/runtime/Math/Functions/Double/Exp2M1.cs
```diff
@@ -0,0 +1,38 @@
...
+using BenchmarkDotNet.Attributes;
+
+namespace System.MathBenchmarks
+{
+    public partial class Double
+    {
+        // Tests double.Exp2M1(double) over 5000 iterations for the domain -1, +1
+
+        private const double exp2M1Delta = 0.0004;
+        private const double exp2M1ExpectedResult = 410.85643799078713;
```
DrewScoggins, line 15: This value is quite different than the value in Exp2.cs

src/benchmarks/micro/runtime/Math/Functions/Double/Exp2.cs
```diff
@@ -0,0 +1,38 @@
...
+using BenchmarkDotNet.Attributes;
+
+namespace System.MathBenchmarks
+{
+    public partial class Double
+    {
+        // Tests double.Exp2(double) over 5000 iterations for the domain -1, +1
+
+        private const double exp2Delta = 0.0004;
+        private const double exp2ExpectedResult = 5410.8564379907702;
```
DrewScoggins, line 15: This value is different than Exp2M1.cs

tannergooding: I see the confusion.

This is not "Apple M1", but rather a function `Exp2M1` which represents `Exp2 Minus One` (`Exp(2) - 1`) but computed more accurately.

PR #2541, opened 2022-07-29
Add System.Text.Json SerializeToWriter/DeserializeFromReader benchmarks

eiriktsarpalis:
Adds a couple of System.Text.Json serialization benchmarks for writing directly to a `Utf8JsonWriter` targeting a prepopulated `IBufferWriter<byte>` and reading directly from `Utf8JsonReader`. Existing serialization benchmarks spend a substantial part of their time/allocations in buffer copying/UTF-8 transcoding, so this should provide clearer visibility in the performance of the raw serialization loop.

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text.Json/Serializer/ReadJson.cs
+++ b/src/benchmarks/micro/libraries/System.Text.Json/Serializer/ReadJson.cs
@@ -57,6 +57,14 @@ namespace System.Text.Json.Serialization.Tests
         [Benchmark]
         public T DeserializeFromUtf8Bytes() => JsonSerializer.Deserialize<T>(_utf8Serialized);
 
+        [BenchmarkCategory(Categories.Libraries, Categories.JSON)]
+        [Benchmark]
+        public T DeserializeFromReader()
+        {
+            Utf8JsonReader reader = new Utf8JsonReader(_utf8Serialized);
+            return JsonSerializer.Deserialize<T>(ref reader);
+        }
+
         [BenchmarkCategory(Categories.Libraries, Categories.JSON, Categories.NoWASM)]
         [Benchmark]
         public async Task<T> DeserializeFromStream()
--- a/src/benchmarks/micro/libraries/System.Text.Json/Serializer/WriteJson.cs
+++ b/src/benchmarks/micro/libraries/System.Text.Json/Serializer/WriteJson.cs
@@ -6,6 +6,7 @@ using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Attributes.Filters;
 using MicroBenchmarks;
 using MicroBenchmarks.Serializers;
+using System.Buffers;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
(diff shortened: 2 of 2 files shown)


src/benchmarks/micro/libraries/System.Text.Json/Serializer/ReadJson.cs
```diff
@@ -57,6 +57,14 @@ public async Task Setup()
         [Benchmark]
         public T DeserializeFromUtf8Bytes() => JsonSerializer.Deserialize<T>(_utf8Serialized);
 
+        [BenchmarkCategory(Categories.Libraries, Categories.JSON)]
```
adamsitnik, line 60: nit: this attribute could be applied to the entire class and then there would be no need to apply it to each benchmark separately

src/benchmarks/micro/libraries/System.Text.Json/Serializer/WriteJson.cs
```diff
@@ -67,5 +82,42 @@ public async Task SerializeToStream()
 
         [Benchmark]
         public string SerializeObjectProperty() => JsonSerializer.Serialize(_objectWithObjectProperty);
+
+        private sealed class ArrayBufferWriter : IBufferWriter<byte>
+        {
+            private int _offset = 0;
+            private byte[] _buffer = new byte[128];
```
adamsitnik, line 89: nit: if there is any particular reason why this value is set to 128 (like "this is what we do in BCL") it would be nice to add a comment here.

src/benchmarks/micro/libraries/System.Text.Json/Serializer/WriteJson.cs
```diff
@@ -67,5 +82,42 @@ public async Task SerializeToStream()
 
         [Benchmark]
         public string SerializeObjectProperty() => JsonSerializer.Serialize(_objectWithObjectProperty);
+
+        private sealed class ArrayBufferWriter : IBufferWriter<byte>
```
adamsitnik, line 86: have we considered adding such type to BCL? it's not the first time I see us re-implementing it

adamsitnik (approved):
LGTM, thank you!

CI failure is unrelated (I am going to send fix in a few minutes)

src/benchmarks/micro/libraries/System.Text.Json/Serializer/WriteJson.cs
eiriktsarpalis: No real reason, it just anticipates the expected output size of this specific benchmark. It doesn't have any real bearing on the benchmark itself, since the buffer will resize to the required size on the first run.

src/benchmarks/micro/libraries/System.Text.Json/Serializer/WriteJson.cs
eiriktsarpalis: I don't think this would be a good implementation for BCL. A canonical IBufferWriter would avoid resizing and use chunks from a buffer pool.

src/benchmarks/micro/libraries/System.Text.Json/Serializer/ReadJson.cs
eiriktsarpalis: I figured I'd just stay consistent with what we already have rather than change it right now.

PR #2544, opened 2022-08-01
fix NativeAOT benchmarks

adamsitnik:
annotate CryptoConfig.CreateFromName benchmarks with DynamicDependency attribute to fix NativeAOT benchmarks

@MichalStrehovsky it used to work, I am just letting you know in case there was so unintentional breaking change (which I doubt)


MichalStrehovsky (approved):
The API is annotated as `RequiresUnreferencedCode` so it's not expected to work without extra work:

https://github.com/dotnet/runtime/blob/fc7577bea2baaa61685ca80257b860151daf5567/src/libraries/System.Security.Cryptography/ref/System.Security.Cryptography.cs#L516-L517

PR #2547, opened 2022-08-04
Add support for decimals in ValuesGenerator.

### LINKED ISSUE #2546: Add support for System.Decimal / decimal to the ValuesGenerator
As suggested [in the documentation ](https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#always-the-same-input-data) it would be nice to add support for additional data types to the `ValuesGenerator`. I noticed also, while looking at the code and comparing to the documentation that the list of supported types in the docs doesn't match the list of types actually supported in the code, so fixing that documentation at the same time seems reasonable.

IDisposable:
As suggested in the documentation, added support for the decimal type.
Also corrected the documentation to properly enumerate the supported types.
Fixes #2546

### DIFF
--- a/src/benchmarks/micro/libraries/Common/AlignedMemory.cs
+++ b/src/benchmarks/micro/libraries/Common/AlignedMemory.cs
@@ -24,7 +24,7 @@ namespace System
         }
 
         public static unsafe AlignedMemory Allocate(uint length, uint alignment)
-            => new AlignedMemory(NativeMemory.AlignedAlloc(length, alignment), (int)length);
+            => new AlignedMemory(NativeMemory.AlignedAlloc((UIntPtr)length, (UIntPtr)alignment), (int)length);
 
         public bool IsDisposed => _disposed;
 
--- a/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
@@ -14,19 +14,31 @@ namespace BenchmarkDotNet.Extensions
     {
         private const int Seed = 12345; // we always use the same seed to have repeatable results!
 
+        /// <summary>
+        /// Returns a T value that is NOT the default(T) (typically zero) value.
+        /// For bool, there's only one choice (true/1).
+        /// For byte/sbyte, will never return byte.MaxValue (256) so there are a maximum of 254 values.
+        /// </summary>
         public static T GetNonDefaultValue<T>()
         {
-            if (typeof(T) == typeof(byte)) // we can't use ArrayOfUniqueValues for byte
-                return Array<T>(byte.MaxValue).First(value => !value.Equals(default));
+            if (typeof(T) == typeof(byte) || typeof(T) == typeof(sbyte)) // we can't use ArrayOfUniqueValues for byte/sbyte (but they have the same range)
+                return Array<T>(byte.MaxValue).First(value => !value.Equals(default(T)));
             else
-                return ArrayOfUniqueValues<T>(2).First(value => !value.Equals(default));
+                return ArrayOfUniqueValues<T>(2).First(value => !value.Equals(default(T)));
         }
 
         /// <summary>
-        /// does not support byte because there are only 256 unique byte values
+        /// Returns an array of the requested size where each entry is a distinct value. 
+        /// For byte and sbyte support a maximum count of 255 because there are only 256
+        /// unique byte values. For bool, there are only 2 values so throw for more requested
         /// </summary>
         public static T[] ArrayOfUniqueValues<T>(int count)
         {
+             if (count > 2 && typeof(T) == typeof(bool))
+                throw new ArgumentOutOfRangeException("count", "Cannot exceed 2 for bool values");
+             if (count > 255 && (typeof(T) == typeof(byte) || typeof(T) == typeof(sbyte)))
+                throw new ArgumentOutOfRangeException("count", "Cannot exceed 255 for byte or sbyte values");
+
             // allocate the array first to try to take advantage of memory randomization
             // as it's usually the first thing called from GlobalSetup method
             // which with MemoryRandomization enabled is the first method called right after allocation
@@ -50,6 +62,11 @@ namespace BenchmarkDotNet.Extensions
             return result;
         }
 
+        /// <summary>
+        /// Returns an array of the requested size where each entry is a random value and values could repeat 
+        /// For byte and sbyte values are built from Random.NextBytes, for other types GenerateValue is used
+        /// to generate a random value in the appropriate range
+        /// </summary>
         public static T[] Array<T>(int count)
         {
             var result = new T[count];
@@ -82,6 +99,10 @@ namespace BenchmarkDotNet.Extensions
             52, 53, 54, 55, 56, 57, 43, 47          //4..9, +, /
         };
 
+        /// <summary>
+        /// Returns an array of bytes of the requested size where each entry is a random value 
+        /// from the legal subset of ASCII characters used in Base-64 encoded values
+        /// </summary>
         public static byte[] ArrayBase64EncodingBytes(int count)
         {
             var result = new byte[count];
@@ -97,8 +118,18 @@ namespace BenchmarkDotNet.Extensions
             return result;
         }
 
+        /// <summary>
+        /// Returns a Dictionary of the requested size where each entry is a has a distinct key value and
+        /// the stored values are randomly generated. 
+        /// GenerateValue is used to generate a random value in the appropriate range for both the key and value
+        /// </summary>
         public static Dictionary<TKey, TValue> Dictionary<TKey, TValue>(int count)
         {
+            if (count > 2 && typeof(TKey) == typeof(bool))
+                throw new ArgumentOutOfRangeException("count", "Cannot exceed 2 for Dictionary<bool, TValue>");
+            if (count > 255 && (typeof(TKey) == typeof(byte) || typeof(TKey) == typeof(sbyte)))
+                throw new ArgumentOutOfRangeException("count", "Cannot exceed 255 for Dictionary<byte, TValue> or Dictionary<sbyte, TValue>");
+
             var dictionary = new Dictionary<TKey, TValue>();
 
             var random = new Random(Seed);
@@ -114,6 +145,11 @@ namespace BenchmarkDotNet.Extensions
             return dictionary;
         }
 
+        /// <summary>
+        /// Returns an array of the requested size where each entry is a random string value.
+        /// The values are random length strings within the range requested and are composed
+        /// a random assortment of the letters a..z, A..Z, and digits 0..9
+        /// </summary>
         public static string[] ArrayOfStrings(int count, int minLength, int maxLength)
         {
             var random = new Random(Seed);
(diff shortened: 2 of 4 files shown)


IDisposable:
This is my first contribution here, so please let me know if I need to do more than simply document the change. There don't seem to be a ton of [tests around ValuesGenerator here](https://github.com/dotnet/performance/blob/main/src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs), but I could add some specifically for `decimal` support, or even more verifying that all the supported base-types at least generate some arrays.

Also, while there are opportunities to use the generation of `decimal` values in this project (e.g. in [Perf.Decimal](https://github.com/dotnet/performance/blob/main/src/benchmarks/micro/libraries/System.Runtime/Perf.Decimal.cs), it would likely change all existing timings so I did NOT make those changes.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -182,5 +184,23 @@ private static Guid GenerateRandomGuid(Random random)
             random.NextBytes(bytes);
             return new Guid(bytes);
         }
+
+        private static decimal GenerateRandomDecimal(Random random)
```
IDisposable, line 188: This is the [Jon Skeet method](https://stackoverflow.com/a/609529) cleaned up to reflect the style in this codebase. I don't think that the fact that these values are not uniformly distributed should matter here (given how we use a subset of ASCII in the string tests), but if desired can [Rasmus Faber variant](https://stackoverflow.com/a/610228)

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -33,7 +33,7 @@ public static T[] ArrayOfUniqueValues<T>(int count)
             // of random-sized memory by BDN engine
             T[] result = new T[count];
 
-            var random = new Random(Seed);
```
IDisposable, line 36: Some leading and trailing white-space and embedded spaces after casts were in this file. The VSCode save cleaned those up to match the project code style, but I can revert the white-space changes if desired.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -182,5 +184,23 @@ private static Guid GenerateRandomGuid(Random random)
...
+            byte scale = (byte)random.Next(29);
+            bool sign = random.Next(2) == 0;
+            return new decimal(random.NextInt32(),
+                               random.NextInt32(),
+                               random.NextInt32(),
+                               sign,
+                               scale);
+        }
+
+        private static int NextInt32(this Random random)
```
IDisposable, line 199: Per Jon Skeet, the `Random.Next(int)` method will never return a _negative_ value, those eliminating an entire half of the bit-space of an `int`. This code will return an integer across the entire `int` value set.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
```diff
@@ -128,6 +128,8 @@ public static string[] ArrayOfStrings(int count, int minLength, int maxLength)
 
         private static T GenerateValue<T>(Random random)
         {
+            if (typeof(T) == typeof(byte))
+                return (T)(object)(byte)random.Next(byte.MinValue, byte.MaxValue);
```
IDisposable, line 132: Technically this is never going to generate a return value of `byte.MaxValue` because `Random.Next` upper limit is exclusive, but we're not obsessing about full coverage in consumers of  this helper.

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
```diff
@@ -72,5 +72,115 @@ public void GetNonDefaultValueReturnsNonDefaultValue()
...
+        [Fact]
+        public void SupportsDouble()
+        {
+            Supports<double>();
+        }
+
+        [Fact]
+        public void SupportsBool()
+        {
+            SupportsArray<bool>();
```
IDisposable, line 139: `bool` is a special case, as there's really no useful answer to the non-default or unique values methods.

IDisposable:
Added test coverage for the supported types and fixed a compilation error on newer platforms in in AlignedMemory.

IDisposable:
@adamsitnik is there anything else I need to do to contribute? I have the CLA done.

src/benchmarks/micro/libraries/Common/AlignedMemory.cs
```diff
@@ -24,7 +24,7 @@ private unsafe AlignedMemory(void* memory, int length)
         }
 
         public static unsafe AlignedMemory Allocate(uint length, uint alignment)
-            => new AlignedMemory(NativeMemory.AlignedAlloc(length, alignment), (int)length);
+            => new AlignedMemory(NativeMemory.AlignedAlloc((UIntPtr)length, (UIntPtr)alignment), (int)length);
```
IDisposable, line 27: Builds in some environments were generating warning about the argument types being wrong for `NativeMemory.AlignedAlloc`, so added the casts. Did NOT think we should change the public signature of the `Allocate` method exposed, but if we want to do that, it's a sweeping and breaking change.

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
```diff
@@ -72,5 +72,115 @@ public void GetNonDefaultValueReturnsNonDefaultValue()
             Assert.NotEqual(default(int), ValuesGenerator.GetNonDefaultValue<int>());
             Assert.NotEqual(default(string), ValuesGenerator.GetNonDefaultValue<string>());
         }
+
```
IDisposable, line 75: Add all the missing tests that assert that we'll be able to generate values using the exposed public methods. For each type (other than `bool`) we will call `Supports<T>();` of the correct type which internally will do the following to assert behavior of the public methods. For `bool` there's only two unique values, so we only assert the array construction, not the ones that test distinct or non-default values.
```
SupportsArray<T>();
SupportsNonDefaultValue<T>();
SupportsArrayOfUniques<T>();
```

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
danmoseley: Why not just add one to the upper bound in those cases?

IDisposable: Certainly possible that would work, just not consistent with the existing behavior... I'll respect any decision on that, just let me know.

danmoseley: I bet the existing cases (short, char etc) are not missing a +1 intentionally. As you say, it may not matter, but it would be the expectation of a naive caller. @adamsitnik  can say.

adamsitnik, line 132: this method was missing `byte` support on purpose. Most of its callers don't support `byte` as they try to generate collections of unique values:

https://github.com/dotnet/performance/blob/3e63b12c2cfe8d3f4f8b6083e49ab28b5111cedd/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs#L26

Since `byte` gives us only 256 unique values, we can't generate a collection of more than 256 unique bytes and loops like this:

https://github.com/dotnet/performance/blob/3e63b12c2cfe8d3f4f8b6083e49ab28b5111cedd/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs#L106-L112

are never going to finish.

You should either remove the byte support or modify all the callers to throw when a unique collection of more than 256 bytes was requested to be created (my preffered solution).

adamsitnik: > Random.Next upper limit is exclusive

To be honest we just missed it when this was being implemented. You can change it for the new types that are being added here, but we can't change it for types that are already supported. But most likely keeping the consistency is going to be the best.

Why we can't change it for the types that were already supported? https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#always-the-same-input-data So if we change it, a method that was generating `int[]` for `Array.Sort` could start generating different data (example: `2 int.MaxValue 1` instead `1 3 2` ) and result in the benchmarks doing more/less work and showing up as a performance regression/improvement.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
adamsitnik: Could you please add a comment to this method explaining what it does and why we can't start using it in `GenerateValue<T>` now?

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
```diff
@@ -72,5 +72,115 @@ public void GetNonDefaultValueReturnsNonDefaultValue()
...
+        [Fact]
+        public void SupportsByte()
+        {
+            Supports<byte>();
+        }
```
adamsitnik, line 80: nit: you could use expressions for all the one-liners:

```suggestion
        public void SupportsByte() => Supports<byte>();
```

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
```diff
@@ -72,5 +72,115 @@ public void GetNonDefaultValueReturnsNonDefaultValue()
...
+        {
+            Supports<Guid>();
+        }
+
+        private static void SupportsArray<T>()
+        {
+            var array = ValuesGenerator.Array<T>(10);
+            Assert.NotNull(array);
+            Assert.Equal(10, array.Length);
+        }
```
adamsitnik, line 165: nit: missing new line

```suggestion
        }

```

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
```diff
@@ -72,5 +72,115 @@ public void GetNonDefaultValueReturnsNonDefaultValue()
...
+        {
+            var value = ValuesGenerator.GetNonDefaultValue<T>();
+            Assert.NotEqual(default(T), value);
+        }
+
+        private static void Supports<T>()
+        {
+            SupportsArray<T>();
+            SupportsNonDefaultValue<T>();
+            SupportsArrayOfUniques<T>();
```
adamsitnik, line 183: could you please also add a method for testing `Dictionary` support?

https://github.com/dotnet/performance/blob/3e63b12c2cfe8d3f4f8b6083e49ab28b5111cedd/src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs#L100

adamsitnik (approved):
@IDisposable big thanks for your contribution! Overall, the code looks great, it just needs some minor polishing before merging.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
IDisposable: Will add the exception throw

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
IDisposable: Since the dictionary has a `TKey` and a `TValue`, but only the type of the `TKey` matters to the value generation logic, do you have a preference for the type used for `TValue`? 

- `TValue` could be same type as `TVal`, and the value stored would be the key's value
- `TValue` could be `string` and be the  string representation of the key value generated, or a random `string` value
- `TValue` could just be `int` and be the `(int)` cast of the key value generated, or a random `int` value

For now, unless you express another choice, will do the first option.

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
IDisposable: Fixed

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
IDisposable: Good idea!

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
IDisposable: Done

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
IDisposable: Done

src/benchmarks/micro/libraries/Common/AlignedMemory.cs
IDisposable: No objections here, so marking resolved

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
IDisposable: Addressed desire for commenting WHY in the code.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
IDisposable: No objections, resolving.

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
IDisposable: Done

src/harness/BenchmarkDotNet.Extensions/ValuesGenerator.cs
IDisposable: Added comments to the code to warn NOT changing things that have been public before.

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
IDisposable: Turns out that with some tweaks we _can_ reasonably treat bools like any other type. Added tests to ensure we throw if someone asks for too many distinct values (like we do for bytes)

src/tests/harness/BenchmarkDotNet.Extensions.Tests/UniqueValuesGeneratorTests.cs
IDisposable: Extended the tests to include bool and also added `Dictionary<TKey, TValue>` tests for all types (with `TValue` being the same type as `TKey` and ALSO `string`)

adamsitnik:
@IDisposable thank you!

PR #2556, opened 2022-08-11
First step in secret management

DrewScoggins:
This adds the registry and the tooling for doing automatic secret management.

### DIFF
--- a/.config/settings.json
+++ b/.config/settings.json
@@ -1,6 +1,6 @@
 {
-    
+    "KeyVaultUri": "https://dotnetperfkeyvault.vault.azure.net/",
     "Storage": {
-        "test-key": "[vault(testk-key)]"
+        "test-key": "[vault(test-key)]"
     }
 }
--- a/.config/settings.json
+++ b/.config/settings.json
@@ -1,0 +1,6 @@
+{
+    
+    "Storage": {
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -579,18 +585,48 @@ jobs:
...
-      osVersion: 20H1
-      architecture: arm64
-      pool:
-        vmImage: windows-2019
-      kind: scenarios
-      machinePool: Tiger
-      queue: Windows.10.Arm64.Perf.Surf
-      projectFile: scenarios.proj
-      channels: # for public jobs we want to make sure that the PRs don't break any of the supported channels
-        - main
```
LoopedBard3, line 596: Can you confirm that we do want to remove this run?

DrewScoggins: This is something that is just for testing. I assume that this will constantly be changing based on what people have been testing.

.config/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    
+    "Storage": {
+        "test-key": "[vault(testk-key)]"
```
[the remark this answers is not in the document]
DrewScoggins: It should be test-key. I will fix this.

PR #2557, opened 2022-08-15
add binary xml version to xml benchmarks

Daniel-Svensson:
As suggested by @danmoseley in https://github.com/dotnet/runtime/pull/73336#issuecomment-1214465324

This should help protect performance for binary xml scenarios such as corewcf.

Apart from [PR mentioned above](https://github.com/dotnet/runtime/pull/73336) this should also cover the changes in
- https://github.com/dotnet/runtime/pull/73332
- https://github.com/dotnet/runtime/pull/71478

And generally make Datacontractserializer changes easier to spot since binary xml has lower overhead than text based version

### DIFF
--- a/src/benchmarks/micro/Serializers/Xml_FromStream.cs
+++ b/src/benchmarks/micro/Serializers/Xml_FromStream.cs
@@ -91,7 +91,7 @@ namespace MicroBenchmarks.Serializers
 
         [GlobalCleanup]
         public void Cleanup()
-		  {
+        {
             xmlDictionaryReader.Dispose();
             memoryStream.Dispose();
         }
--- a/src/benchmarks/micro/Serializers/Xml_FromStream.cs
+++ b/src/benchmarks/micro/Serializers/Xml_FromStream.cs
@@ -92,8 +92,8 @@ namespace MicroBenchmarks.Serializers
         [GlobalCleanup]
         public void Cleanup()
         {
-            xmlDictionaryReader.Dispose();
-            memoryStream.Dispose();
+            xmlDictionaryReader?.Dispose();
+            memoryStream?.Dispose();
         }
     }
 }
--- a/src/benchmarks/micro/Serializers/Xml_FromStream.cs
+++ b/src/benchmarks/micro/Serializers/Xml_FromStream.cs
@@ -26,6 +26,8 @@ namespace MicroBenchmarks.Serializers
         private XmlSerializer xmlSerializer;
         private DataContractSerializer dataContractSerializer;
         private MemoryStream memoryStream;
+        private byte[] memoryBytes;
+        private XmlDictionaryReader xmlDictionaryReader;
 
         [GlobalSetup(Target = nameof(XmlSerializer_))]
         public void SetupXmlSerializer()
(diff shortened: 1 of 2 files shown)


danmoseley:
@StephenMolloy could you review? I'm not familiar with the API.

src/benchmarks/micro/Serializers/Xml_FromStream.cs
```diff
@@ -63,9 +65,35 @@ public T DataContractSerializer_()
...
+            return (T)dataContractSerializer.ReadObject(xmlDictionaryReader);
+        }
+
+      // YAXSerializer is not included in the benchmarks because it does not allow to deserialize from stream (only from file and string)
 
         [GlobalCleanup]
-        public void Cleanup() => memoryStream.Dispose();
+        public void Cleanup()
+        {
+            xmlDictionaryReader.Dispose();
```
StephenMolloy, line 95: >xmlDictionaryReader.Dispose() [](http://example.com/codeflow?start=12&length=29)

Check for null? I don't think this gets initialized if running the other benchmarks here.

src/benchmarks/micro/Serializers/Xml_FromStream.cs
```diff
@@ -63,9 +65,35 @@ public T DataContractSerializer_()
...
+
+            memoryBytes = memoryStream.ToArray();
+            xmlDictionaryReader = XmlDictionaryReader.CreateBinaryReader(memoryBytes, XmlDictionaryReaderQuotas.Max);
+        }
+
+        [BenchmarkCategory(Categories.Libraries)]
+        [Benchmark(Description = nameof(XmlDictionaryReader))]
+        public T DataContractSerializer_BinaryXml_()
+        {
+            ((IXmlBinaryReaderInitializer)xmlDictionaryReader).SetInput(memoryBytes, 0, memoryBytes.Length, null, XmlDictionaryReaderQuotas.Max, null, null);
```
StephenMolloy, line 86: >xmlDictionaryReader).SetInput [](http://example.com/codeflow?start=42&length=29)

Consider using [IterationSetup] for this step.

src/benchmarks/micro/Serializers/Xml_ToStream.cs
```diff
@@ -52,9 +54,23 @@ public void DataContractSerializer_()
...
+            dataContractSerializer.WriteObject(xmlDictionaryWriter, value);
+        }
+
         // YAXSerializer is not included in the benchmarks because it does not allow to serialize to stream (only to file and string)
 
         [GlobalCleanup]
-        public void Dispose() => memoryStream.Dispose();
+        public void Dispose()
+        {
+            xmlDictionaryWriter.Dispose();
```
StephenMolloy, line 72: >xmlDictionaryWriter. [](http://example.com/codeflow?start=12&length=20)

null check

src/benchmarks/micro/Serializers/Xml_ToStream.cs
```diff
@@ -52,9 +54,23 @@ public void DataContractSerializer_()
             dataContractSerializer.WriteObject(memoryStream, value);
         }
 
+        [BenchmarkCategory(Categories.Libraries)]
+        [Benchmark(Description = nameof(XmlDictionaryWriter))]
+        public void DataContractSerializer_BinaryXml_()
+        {
+            memoryStream.Position = 0;
+            ((IXmlBinaryWriterInitializer)xmlDictionaryWriter).SetOutput(memoryStream, null, null, ownsStream: false);
```
StephenMolloy, line 62: >xmlDictionaryWriter).SetOutput [](http://example.com/codeflow?start=42&length=30)

Again, consider using [IterationSetup] for this.

StephenMolloy (approved):
A couple small comments. But looks fine otherwise.

src/benchmarks/micro/Serializers/Xml_FromStream.cs
danmoseley: Right, this will fail if you select a subset of benchmarks

Daniel-Svensson: fixed

src/benchmarks/micro/Serializers/Xml_FromStream.cs
Daniel-Svensson: I did so earlier for my own benchmarks but that was not very successful for this case.

I think it might be fair game to benchmark with the SetInput call since that is needed for any real world scenario (corewcf , OpenRiaServices etc) where you want to read one "xml document" at a time but pool readers/writers for reuse.

When using iterationsetup the method itself should be quite long running (i think either above guidance or some other guidance says 200ms+)  so it must process lots of objects. 

However I have prepared an alternative version at https://github.com/Daniel-Svensson/dotnet_performance/tree/remote_setoutput with commit 
https://github.com/dotnet/performance/commit/128fcdfac56bdf3b42f26b754dc8ce64de05d2c4 if you rather want that one. It will instead read N (100 but I am open for any value) similar objects inside a single "root" element. So the overhead will go from <= 1% to maybe 0,01% of the execution time.

src/benchmarks/micro/Serializers/Xml_ToStream.cs
Daniel-Svensson: That can be removed, but then it will emit slight different output (no start of document) for most calls.
I did that change int the separate branche mentioned above

src/benchmarks/micro/Serializers/Xml_FromStream.cs
Daniel-Svensson: @danmoseley Do you prefer the current version which mimics the end user scenario or should the binary xml version of the benchmarks serialize many more objects than the other benchmark versions ?

unless the ci-failures are related to the changes here I think this is ready for possible merge.

danmoseley: This is @StephenMolloy  area to answer..

Daniel-Svensson:
@StephenMolloy, do you think the current version looks fine and is ready to merge?

 Most of your comments are fixed, except iterationsetup  (see comment above on why).

PR #2558, opened 2022-08-15
Remove base diff

dakersnar:
When publishing the monthly perf report, it is necessary to generate a version of the report without the raw base and diff data, only publicly showing the ratio between them. This PR adds a `--ratio-only` flag to `benchmarks_monthly`, which generates the final report without the base and diff columns.

### DIFF
--- a/src/tools/ResultsComparer/MultipleInputsComparer.cs
+++ b/src/tools/ResultsComparer/MultipleInputsComparer.cs
@@ -44,7 +44,7 @@ namespace ResultsComparer
                 Console.WriteLine($"## {benchmarkResults.Key}");
                 Console.WriteLine();
 
-                var table; 
+                var table = null;
                 if (args.RatioOnly) 
                 {
                     var data = benchmarkResults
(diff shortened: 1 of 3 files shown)


src/tools/ResultsComparer/MultipleInputsComparer.cs
```diff
@@ -40,23 +44,43 @@ internal static void Compare(MultipleInputsOptions args)
...
-                        AllocatedDiff = GetAllocatedDiff(result.diffResult, result.baseResult),
-                        OperatingSystem = Stats.GetSimplifiedOSName(result.baseEnv.OsVersion),
-                        Architecture = result.baseEnv.Architecture,
-                        ProcessorName = result.baseEnv.ProcessorName,
-                        Modality = Helper.GetModalInfo(result.baseResult) ?? Helper.GetModalInfo(result.diffResult),
-                    })
-                    .ToArray();
-
-                var table = data.ToMarkdownTable().WithHeaders(Headers);
+                Table table = null;
```
adamsitnik, line 47: I wish I knew a simpler way to remove these columns.

adamsitnik (approved):
LGTM, thank you @dakersnar !

PR #2566, opened 2022-08-17
Fixes https://github.com/dotnet/performance/issues/2444

### LINKED ISSUE #2444: GC benchmarks setup fails on Windows 11
On Windows 11, the following gc benchmarks setup step seems to fail (doesn't fail on Windows 10):

```
cd path\to\performance\src\benchmarks\gc
python . setup
```

The root cause looks to be a call to `performance\src\benchmarks\gc\src\exec\env\out\Release\get_host_info.exe` which gives a different sort of output only on Windows 11:
```
Invalid relationship 7
Invalid relationship 7
Invalid relationship 7
Invalid relationship 7
Invalid relationship 7
Invalid relationship 7
Invalid relationship 7
Invalid relationship 7
numa_nodes:
  -
    numa_node_number: 0
    cpu_group_number: 0
    ranges:
      - { lo: 0, hi: 15 }
n_physical_processors: 8
n_logical_processors: 16
caches:
  l1: { n_caches: 16, total_bytes: 655360 }
  l2: { n_caches: 8, total_bytes: 4194304 }
  l3: { n_caches: 1, total_bytes: 16777216 }
```

Note the "**Invalid relationship 7**" text that's only produced on Windows 11 (those extra lines don't get printed out on Win10). This extra text seems to make the yaml parser fail later on.

That "**7**" refers to LOGICAL_PROCESSOR_RELATIONSHIP enum value of 7, which is **RelationProcessorModule** and looks to be undocumented (https://docs.microsoft.com/en-us/windows/win32/api/winnt/ne-winnt-logical_processor_relationship). The benchmark doesn't have a specific case to handle this relationship (falls into the default case) in get_host_info.cpp and prints these "Invalid relationship" lines as a result.

Wondering if the GC team has also run into this issue on Win11 and if there's a recommended way to correct this issue.

Thanks

mrsharm:
Fixes #2444 by accounting for the ``RelationProcessorModule`` case prevalent in Windows 11.

### DIFF
--- a/src/benchmarks/gc/src/exec/env/get_host_info.cpp
+++ b/src/benchmarks/gc/src/exec/env/get_host_info.cpp
@@ -191,6 +191,7 @@ static CacheStats getCacheStats()
             }
             case RelationProcessorPackage:
             case RelationGroup:
+            case RelationProcessorModule:
                 break;
 
             default:


mangod9 (approved):

Maoni0:
thanks for fixing this @mrsharm!

PR #2567, opened 2022-08-18
Fixup Brotli compression tests

DrewScoggins:
As seen here, https://github.com/dotnet/runtime/issues/73391#issuecomment-1205619431, this test had a bug where in
the work that was being done by the underlying test was not being correctly measured. This fix resolves that.

### DIFF
--- a/src/benchmarks/micro/libraries/System.IO.Compression/Brotli.cs
+++ b/src/benchmarks/micro/libraries/System.IO.Compression/Brotli.cs
@@ -13,7 +13,7 @@ namespace System.IO.Compression
         private const int Window = 22;
 
         public override Stream CreateStream(Stream stream, CompressionMode mode) => new BrotliStream(stream, mode);
-        public override Stream CreateStream(Stream stream, CompressionLevel level) => new BrotliStream(stream, level);
+        public override Stream CreateStream(Stream stream, CompressionLevel level) => new BrotliStream(stream, level, leaveOpen: true);
 
         [Benchmark]
         public Span<byte> Compress_WithState()
--- a/src/benchmarks/micro/libraries/System.IO.Compression/CompressionStreamPerfTestBase.cs
+++ b/src/benchmarks/micro/libraries/System.IO.Compression/CompressionStreamPerfTestBase.cs
@@ -59,7 +59,7 @@ namespace System.IO.Compression
         {
             CompressedFile.CompressedDataStream.Position = 0; // all benchmarks invocation reuse the same stream, we set Postion to 0 to start at the beginning
 
-            var compressor = CreateStream(CompressedFile.CompressedDataStream, level);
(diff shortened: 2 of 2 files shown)


src/benchmarks/micro/libraries/System.IO.Compression/CompressionStreamPerfTestBase.cs
```diff
@@ -59,7 +59,7 @@ public void Compress()
         {
             CompressedFile.CompressedDataStream.Position = 0; // all benchmarks invocation reuse the same stream, we set Postion to 0 to start at the beginning
 
-            var compressor = CreateStream(CompressedFile.CompressedDataStream, level);
+            using var compressor = CreateStream(CompressedFile.CompressedDataStream, level);
```
stephentoub, line 62: It's less of an issue, but https://github.com/dotnet/performance/blob/ccbeb4ed946e0a7bb676c58f6b5841740c616873/src/benchmarks/micro/libraries/System.IO.Compression/CompressionStreamPerfTestBase.cs#L71 should also be disposing of this.

src/benchmarks/micro/libraries/System.IO.Compression/Brotli.cs
```diff
@@ -13,7 +13,7 @@ public class Brotli : CompressionStreamPerfTestBase
         private const int Window = 22;
 
         public override Stream CreateStream(Stream stream, CompressionMode mode) => new BrotliStream(stream, mode);
-        public override Stream CreateStream(Stream stream, CompressionLevel level) => new BrotliStream(stream, level);
+        public override Stream CreateStream(Stream stream, CompressionLevel level) => new BrotliStream(stream, level, leaveOpen: true);
```
stephentoub, line 16: The same thing needs to be done in these:
https://github.com/dotnet/performance/blob/ccbeb4ed946e0a7bb676c58f6b5841740c616873/src/benchmarks/micro/libraries/System.IO.Compression/CompressionStreamPerfTestBase.cs#L11-L28

I think it'd also be cleaner if both CreateStream overloads always set leaveOpen: true, so it doesn't differ by which is invoked in the tests.

rzikm:
@adamsitnik I updated the PR (with Drew's permission), I think this should give us more representative measurements. Can you please take a look?

rzikm:
> The changes LGTM, but we need to keep in mind that the fix may create an influx of perf improvements/regressions reported by the bot

That is expected, but given that previous measurements were not accurate/representative of the real thing (as part of the work wasn't done) we should do it.

adamsitnik:
> > The changes LGTM, but we need to keep in mind that the fix may create an influx of perf improvements/regressions reported by the bot
> 
> That is expected, but given that previous measurements were not accurate/representative of the real thing (as part of the work wasn't done) we should do it.

Absolutely, I just wanted to give @DrewScoggins a heads up (he owns the process of automatic regression/improvement detection).

PR #2568, opened 2022-08-18
Fix secret error

DrewScoggins:


DrewScoggins:
I screwed up the timing (didn't use 24 hr time in the crontab). This fixes that along with updating a secret that I put in wrong.

LoopedBard3 (approved):
LGTM

PR #2573, opened 2022-08-23
Add F# compiler service scenario

0101:
This is a current snapshot of F# compiler service project. 

We would like to measure how long it takes to build and possibly other stats as part of https://github.com/dotnet/fsharp/issues/13606

Also to be able to catch any issues quickly.


src/scenarios/fsharpcompilerservice/pre.py
```diff
@@ -0,0 +1,14 @@
+'''
+pre-command
+'''
+import os
+import sys
+from performance.logger import setup_loggers
+from shared import const
+from shared.precommands import PreCommands
+from test import EXENAME
```
IDisposable, line 9: This seems unused.

DrewScoggins:
Failures are unrelated. Going to merge this

Issue #2575, opened 2022-08-24
[perf pipeline] `Perf_BigInteger.ModPow` failing with `System.ArgumentException: Destination is too short. (Parameter 'destination')`


radical:
[Log from runtime-wasm-perf](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-74097-merge-452b23c3e6b4475d86/wasm.x64.micro.net7.0.Partition5/1/console.8998d92c.log?helixlogtype=result) run on https://github.com/dotnet/runtime/pull/74097:

```
// **************************
// Benchmark: Perf_BigInteger.ModPow: Job-BJXMHI(PowerPlanMode=00000000-0000-0000-0000-000000000000, Runtime=Wasm, Toolchain=Wasm, InvocationCount=1, IterationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=0) [arguments=1024,1024,64 bits]
// *** Execute ***
// Launch: 1 / 1
// Execute: /home/helixbot/.jsvu/v8 --expose_wasm test-main.js -- --run 95a3fee8-6d92-4c8b-8059-369b73cd756f.dll --benchmarkName "System.Numerics.Tests.Perf_BigInteger.ModPow(arguments: 1024,1024,64 bits)" --job "PowerPlanMode=00000000-0000-0000-0000-000000000000, Runtime=Wasm, Toolchain=Wasm, InvocationCount=1, IterationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=0" --benchmarkId 67  in /datadisks/disk1/work/A6C10953/w/B67009A6/e/performance/artifacts/bin/for-running/MicroBenchmarks/95a3fee8-6d92-4c8b-8059-369b73cd756f/bin/net7.0/browser-wasm/AppBundle
Failed to set up high priority. Make sure you have the right permissions. Message: Permission denied
[...]

PR #2578, opened 2022-08-25
temporarily disable failing BigInteger benchmark

adamsitnik:
https://github.com/dotnet/performance/issues/2575 turned out to be a product bug. Disable it before the product is fixed and the fix flows to the SDK installer.

cc @tannergooding


danmoseley (approved):

tannergooding (approved):
LGTM, thanks!

The fix for this should be: https://github.com/dotnet/runtime/pull/74585

PR #2581, opened 2022-08-29
Basic linker benchmarks

jtschuster:
This is the first basic linker benchmarks. Both benchmarks start by publishing an untrimmed net6.0 hello world app with the dotnet sdk on the path. BasicBenchmark runs the trimmer on the output .dll's using reflection to get the `Main` method of `Mono.Linker.Driver`. MSBuildBenchmark shells out to `dotnet publish --no-build /p:PublishTrimmed=true` and times how long that takes.

I'm not sure if this is the best way to go about this, so I wanted to make a draft to get thoughts.


azure-pipelines.yml
```diff
@@ -600,6 +600,39 @@ jobs:
         inlineScript: |
           Get-ChildItem .vault-config/*.yaml |% { dotnet secret-manager synchronize $_}
 
+  # Ubuntu 1804 x64 ILLink Benchmarks
```
DrewScoggins, line 603: I realized that you will also want to add this to the public correctness run section as well. You will want your stuff run to make sure that the tests don't get broken.

azure-pipelines.yml
```diff
@@ -600,6 +600,39 @@ jobs:
...
+  - template: /eng/performance/benchmark_jobs.yml
+    parameters:
+      osName: windows
+      osVersion: 22H1
+      kind: illink
+      architecture: x64
+      pool:
+        vmImage: windows-2019
+      machinePool: Tiger
+      queue: Ubuntu.1804.Amd64.Tiger.Perf # using a dedicated private Helix queue (perftigers)
```
DrewScoggins, line 630: This queue name needs to be updated to the actual Windows queue name, `Windows.10.Amd64.19H1.Tiger.Perf`

eng/Versions.props
```diff
@@ -13,6 +13,7 @@
     </MicrosoftNETSdkInternalPackageVersion>
     <MicrosoftWindowsDesktopAppRuntimewinx64PackageVersion>5.0.4</MicrosoftWindowsDesktopAppRuntimewinx64PackageVersion>
     <MicrosoftAspNetCoreAppRuntimewinx64PackageVersion>5.0.4</MicrosoftAspNetCoreAppRuntimewinx64PackageVersion>
+    <MicrosoftNetILLinkTasksVersion>7.0.100-1.22057.1</MicrosoftNetILLinkTasksVersion>
```
DrewScoggins, line 16: Did you also use darc to add this dependency to our subscription? That way we can ensure that this value is automatically updated going forward.

src/benchmarks/real-world/ILLink/BasicBenchmark.cs
```diff
@@ -0,0 +1,237 @@
...
+    public void LinkHelloWorldGlobalCleanup()
+    {
+        Directory.Delete(_publishOutputFolder, recursive: true);
+    }
+
+    [Benchmark]
+    [BenchmarkCategory("ILLink")]
+    public object LinkHelloWorld()
+    {
+        Mono.Linker.WarnVersion x = Mono.Linker.WarnVersion.Latest;
```
DrewScoggins, line 65: Could we do the reflection setup in the GlobalSetup, or is there a reason that we need to do this for every operation? In general, you want to have as little code unrelated to the value you are trying to measure in the benchmark method.

src/benchmarks/real-world/ILLink/MSBuildBenchmark.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        _iterationPublishPath = Utilities.PublishSampleProject(_projectFilePath, "--no-build", "/p:PublishTrimmed=true", $"/p:_LinkSemaphore=\"{_linkSemaphore}\"");
+        return _iterationPublishPath;
+    }
+
+    [IterationCleanup(Targets = new[] { nameof(LinkHelloWorld) })]
+    public void LinkHelloWorldIterationCleanup ()
+    {
+        Directory.Delete(_iterationPublishPath, recursive: true);
+        File.Delete(_linkSemaphore);
```
DrewScoggins, line 44: If you delete the `_linkSemaphore` here subsequent iterations will not have the file, since the `GlobalSetup` method will not be run again. You might want to move the sempahore creation into an IterationSetup method.

jtschuster: The `_linkSemaphore` is used as the signal for incremental builds to avoid running the linker, so it needs to be deleted after every iteration.

src/benchmarks/real-world/ILLink/BasicBenchmark.cs
jtschuster: Good idea, I'll move that into the setup.

eng/Versions.props
jtschuster: I did not yet, I'll add that

azure-pipelines.yml
```diff
@@ -293,6 +293,39 @@ jobs:
...
+  - template: /eng/performance/benchmark_jobs.yml
+    parameters:
+      osName: ubuntu
+      osVersion: 1804
+      kind: illink
+      architecture: x64
+      pool:
+        vmImage: ubuntu-latest
+      machinePool: Tiger
+      queue: Ubuntu.1804.Amd64.Tiger.Perf # using a dedicated private Helix queue (perftigers)
```
DrewScoggins, line 306: This queue and the one for the Windows job need to match the other queue names in the public correctness section. We cannot use the queues from the private jobs area, as those queues are only available for internal runs.

DrewScoggins:
It looks like the benchmarks are running, but looking at the output I see that we are only running 1 iteration, with no warmup. It might be best to take a look at the docs for BDN here, https://benchmarkdotnet.org/articles/guides/choosing-run-strategy.html, and move over to the monitoring strategy, with about 5-10 iterations. This way you will get more reliable results over time.

src/benchmarks/real-world/ILLink/ILLinkBenchmarks.csproj
```diff
@@ -0,0 +1,46 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<OutputType>Exe</OutputType>
+		<TargetFrameworks Condition="'$(TargetFrameworks)' == ''">net6.0;net7.0</TargetFrameworks>
```
DrewScoggins, line 7: You need to add this line above the current `<TargetFrameworks>` tag

`<TargetFrameworks>$(PERFLAB_TARGET_FRAMEWORKS)</TargetFrameworks>`

This will ensure in the lab that you are only building against the one TFM that you care about at that time. It will prevent the kind of errors that we are seeing now.

jtschuster:
It looks like there's still an issue with the SDK looking for Microsoft.NETCoreApp.Ref version 6.0.10. Other tests are failing the same way and it does the same locally with a hello world app targeting net6.0, so it might be an SDK issue.

DrewScoggins:
Looking at the actual errors, it looks like we are seeing a hang on Windows. Are you seeing this when you try and run the tests locally? On Ubuntu we are seeing the Nuget failures, but those are related to the fact that the HelloWorld.csproj is trying to target net6.0.

jtschuster:
Yes, I was able to recreate the hang locally using the ci script. There was a deadlock associated with waiting for the publish process to exit before reading the redirected standard output. Finishing a fix for that now, but it looks like the same sdk issue (searching for 6.0.10) still exists.

src/benchmarks/real-world/ILLink/SampleProject/HelloWorld.template.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
+		<OutputType>exe</OutputType>
+		<TargetFramework>net6.0</TargetFramework>
```
DrewScoggins, line 4: I think if this gets updated to target net7.0 this test should start working and we can get this checked in.

DrewScoggins (approved):
LGTM

PR #2583, opened 2022-08-30
Add release/7.0 channel to channel map

LoopedBard3:
Add release/7.0 channel to channel map and remove preview releases as their values do not match what they would be.


caaavik-msft (approved):

LoopedBard3:
Failures are related to ongoing issues in centos and maui desktop scenarios unrelated to this, merging to fix runtime pipeline.

PR #2586, opened 2022-08-31
Added F# compiler service scenario (2)

0101:


0101:
@DrewScoggins the empty counter issue didn't reappear, so can we merge this now?

PR #2592, opened 2022-09-02
F# real world benchmarks

0101:
Added some benchmarks using current snapshots of [FSharpPlus](https://github.com/fsprojects/FSharpPlus) and [FsToolkit.ErrorHandling](https://github.com/demystifyfp/FsToolkit.ErrorHandling).


0101:
@DrewScoggins what do you think about this one? I tried to trigger the added benchmarks via a new job in `pr.yml` - not sure if that was successful or where would see the results.

azure-pipelines.yml
```diff
@@ -276,6 +292,23 @@ jobs:
       channels: # for ML.NET jobs we want to check .NET Core 3.1 and 5.0 only
         - main
 
+  # Ubuntu 1804 x64 F# benchmarks
```
DrewScoggins, line 295: You should also add this and the other Windows section above in the Scheduled Private job section that starts on line 552. This way the tests will actually run in the lab and generate results. You only have them added to the public correctness section, which means they will only run as a part of PRs.

DrewScoggins:
I left a comment about a little more work that needs to happen. Also I am fairly certain that `pr.yml` is deprecated and just never got deleted. The jobs that you added in `azure_pipelines.yml` are actually just PR jobs.

azure-pipelines.yml
0101: Got it, added. I took some of the parameters from other jobs in that section, please check if it's correct.

DrewScoggins (approved):
LGTM

PR #2598, opened 2022-09-06
Modify ci_setup get date backoff to be based off of the GitHub counter reset

LoopedBard3:
Modify ci_setup get date backoff to be based off of the counter reset time instead of arbitrarily waiting since the Api call counter only resets once an hour. This should solve some issues with hitting rate limits in our pipelines. The get_commit_date method has been tested locally an appears to work properly.

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -16,6 +16,7 @@ from shutil import rmtree
 from stat import S_IRWXU
 from subprocess import CalledProcessError, check_output
 from sys import argv, platform
+import time
 from typing import Tuple
 from urllib.error import URLError
 from urllib.parse import urlparse
@@ -613,6 +614,7 @@ def get_commit_date(
 
     url = None
     urlformat = 'https://api.github.com/repos/%s/%s/commits/%s'
+    ratelimiturl = 'https://api.github.com/rate_limit'
     if repository is None:
         # The origin of the repo where the commit belongs to has changed
         # between release. Here we attempt to naively guess the repo.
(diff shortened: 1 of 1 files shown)


scripts/dotnet.py
```diff
@@ -625,18 +627,22 @@ def get_commit_date(
...
-            getLogger().warning(f"URL Error trying to get commit date from {url}; Reason: {error.reason}; Attempt {retrycount}")
-            sleep(sleep_time)
-            sleep_time = sleep_time * 2
+            with urlopen(ratelimiturl) as response:
+                item = loads(response.read().decode('utf-8'))
+                reset_timestamp = item['resources']['core']['reset']
+                remaining_calls = item['resources']['core']['remaining']
+                seconds_to_retry = reset_timestamp - time.time() + 10 # Add 10 seconds to be safe
+                getLogger().warning(f"URL Error trying to get commit date from {url}; Reason: {error.reason}; Attempt {retrycount}; reset time: {reset_timestamp} (Unix Time); time to wait: {seconds_to_retry} seconds; remaining calls: {remaining_calls}")
+                if(remaining_calls == 0 and retrycount == 0):
```
DrewScoggins, line 644: So this seems to only wait if `remaining_calls` is equal to zero. I assume that means that we have hit the retry limit from the GitHub API. Are we sure that we don't want to have any kind of sleep in the case where we are not at the limit? Or am I just missing something here?

LoopedBard3: We should probably add a sleep of some sort, I will add one for 30 seconds. I think it could go either way with adding a sleep as if we still have remaining calls, something else likely failed, instead of having exceeded the API rate limit.

DrewScoggins: Yeah, I think that we really want this retry and backoff for two things. One is to deal with transient network failures. And in that case we just want to wait some set amount of time and try again. However, if we are being rate limited then we will need to use the more complex call that we have in this PR. I wonder if it would be worthwhile to break it out into those two paths?

PR #2609, opened 2022-09-16
Add dictionary tests for non-string reference type

stephentoub:


stephentoub:
This has been here for a long time. Anyone still want it?
cc: @eiriktsarpalis

src/benchmarks/micro/libraries/System.Collections/TryGetValue/TryGetValueFalse.cs
```diff
@@ -15,6 +15,7 @@ namespace System.Collections
 {
     [BenchmarkCategory(Categories.Libraries, Categories.Collections, Categories.GenericCollections)]
     [GenericTypeArguments(typeof(int), typeof(int))] // value type
+    [GenericTypeArguments(typeof(SmallClass), typeof(SmallClass))] // reference type
```
eiriktsarpalis, line 18: It would be nice if we could add a few large structs as well, tuples are fairly common dictionary keys.

eiriktsarpalis (approved):

src/benchmarks/micro/libraries/System.Collections/TryGetValue/TryGetValueFalse.cs
stephentoub: Done

danmoseley (approved):

eiriktsarpalis:
@caaavik-msft @DrewScoggins could we get help merging this PR? Thanks.

PR #2614, opened 2022-09-21
Issue #75070 (Dotnet Runtime) - Add a performance microbenchmark for …

RyanThomas73:
…the new PriorityQueue DequeueEnqueue method to help monitor for performance regressions

Serves as a performance regression check benchmark for the new `PriorityQueue` `DequeueEnqueue` public api method added as part of this dotnet runtime issue https://github.com/dotnet/runtime/issues/75070

### DIFF
--- a/src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
+++ b/src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
@@ -106,5 +106,30 @@ namespace System.Collections.Tests
                 queue.Dequeue();
             }
         }
+
+        [Benchmark]
+        public void KMaxElements_DequeueEnqueue_PerformanceRegressionBenchmark()
+        {
+            const int k = 5;
+            var queue = _priorityQueue;
+            var items = _items;
+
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.Collections/PriorityQueue/Perf_PriorityQueue.cs
```diff
@@ -106,5 +106,30 @@ public void K_Max_Elements()
                 queue.Dequeue();
             }
         }
+
+        [Benchmark]
+        public void KMaxPriorityQueueSize_DequeueEnqueue_PerformanceRegressionBenchmark()
```
eiriktsarpalis, line 111: IIRC this isn't really computing any max or min elements, since unlike `EnqueueDequeue`, `DequeueEnqueue` is never comparing the extracted min heap entry with the inserted entry. The `PerformanceRegressionBenchmark` suffix is redundant, since that's what all benchmarks in this repo are intended for. Recommend keeping the benchmark name ambiguous:

```suggestion
        public void DequeueEnqueue()
```

eiriktsarpalis (approved):

PR #2616, opened 2022-09-22
Support for explicit output for build and publish commands.

cincuranet:


src/scenarios/shared/precommands.py
```diff
@@ -177,12 +182,10 @@ def execute(self, build_args: list = []):
...
         if self.operation == BUILD:
             self._restore()
-            self._build(configuration=self.configuration, framework=self.framework, build_args=build_args)
+            self._build(configuration=self.configuration, framework=self.framework, output=self.output, build_args=build_args)
         if self.operation == PUBLISH:
             self._restore()
-            self._publish(configuration=self.configuration,
-                          runtime_identifier=self.runtime_identifier,
-                          framework=self.framework, build_args=build_args)
+            self._publish(configuration=self.configuration, runtime_identifier=self.runtime_identifier, framework=self.framework, output=self.output, build_args=build_args)
```
DrewScoggins, line 188: Is there a reason you moved this all to one line?

DrewScoggins (approved):
I assume this change is so we can deploy/build directly into the location we want rather than having to copy it there after we build into the pub folder?

Either way this change LGTM

src/scenarios/shared/precommands.py
cincuranet: IIRC, consistency with rest of the file.

PR #2622, opened 2022-09-29
Allow for the passing of the runtime commit time to use in ci_setup.

LoopedBard3:
Allow for the passing of the runtime commit time to use in ci_setup. This is to allow for bypassing of calling to github for the commit timestamp.

### DIFF
--- a/scripts/ci_setup.py
+++ b/scripts/ci_setup.py
@@ -5,9 +5,8 @@ from logging import getLogger
 
 import os
 import sys
+import datetime
 
-
-from dateutil import parser as date_parser
 from subprocess import check_output
 
 from performance.common import get_repo_root_path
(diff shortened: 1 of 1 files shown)


scripts/ci_setup.py
```diff
@@ -309,7 +317,11 @@ def __main(args: list) -> int:
         target_framework_moniker = dotnet.FrameworkAction.get_target_framework_moniker(framework)
         dotnet_version = dotnet.get_dotnet_version(target_framework_moniker, args.cli) if args.dotnet_versions == [] else args.dotnet_versions[0]
         commit_sha = dotnet.get_dotnet_sdk(target_framework_moniker, args.cli) if args.commit_sha is None else args.commit_sha
-        source_timestamp = dotnet.get_commit_date(target_framework_moniker, commit_sha, repo_url)
+        if(args.commit_time is not None):
+            parsed_timestamp = datetime.datetime.strptime(args.commit_time, '%Y-%m-%d %H:%M:%S %z').astimezone(datetime.timezone.utc)
```
DrewScoggins, line 321: We should double check this, but I am pretty sure that we currently get the time in the pacific time zone.

DrewScoggins: I was just wrong. We store the time in UTC

scripts/ci_setup.py
```diff
@@ -309,7 +317,11 @@ def __main(args: list) -> int:
         target_framework_moniker = dotnet.FrameworkAction.get_target_framework_moniker(framework)
         dotnet_version = dotnet.get_dotnet_version(target_framework_moniker, args.cli) if args.dotnet_versions == [] else args.dotnet_versions[0]
         commit_sha = dotnet.get_dotnet_sdk(target_framework_moniker, args.cli) if args.commit_sha is None else args.commit_sha
-        source_timestamp = dotnet.get_commit_date(target_framework_moniker, commit_sha, repo_url)
+        if(args.commit_time is not None):
```
DrewScoggins, line 320: Should we instead fallback to grabbing the date from Github if we see a parse error? Otherwise, I don't really see a strong need to keep around the other code path.

LoopedBard3: Yes, I think if the commit_time is not included or fails to parse that is the correct approach.

DrewScoggins (approved):
LGTM

PR #2626, opened 2022-10-03
Added: MicroBenchmark for object.Equals() for DateTime

Sewer56:
This PR adds a simple microbenchmark for the `object.Equals()` implementation of `System.DateTime`.  
It's a companion to the following PR https://github.com/dotnet/runtime/pull/76559 .


cincuranet:
Thanks.

PR #2632, opened 2022-10-05
Fix links for public builds.

cincuranet:


LoopedBard3:
.Net 5 is no longer officially supported per https://github.com/dotnet/core/blob/main/releases.md. I think it would be worth it to remove the reference in this update, since we are making the change.

LoopedBard3 (approved):
LGTM, with or without 5.0.1 link removal

PR #2635, opened 2022-10-06
Remove .NET 5 links, because it's no longer officially supported

cincuranet:
Based on a suggestion from #2632.


LoopedBard3 (approved):
LGTM

PR #2638, opened 2022-10-07
Add ability to fetch an internal RC2 build using a secret "--internal-build-key" argument fetched from a keyvault

dakersnar:
This solution is a bit hacky, but users in the correct security group should be able to run the perf report with an internal build using the following workflow:

1. `$value = az keyvault secret show --vault-name dnceng-partners-kv --name dotnetbuilds-internal-container-read-token --query "value"`
2. `$value = $value.replace('"', "'")`
3. `py .\scripts\benchmarks_monthly.py net7.0-rc2 --architecture x64 --device-name drew-x64 --azure-feed-url https://dotnetbuilds.blob.core.windows.net/internal --internal-build-key $value`
> 

I haven't tested this on Linux or Mac, might need to adjust it as needed.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -41,7 +41,8 @@ def init_tools(
         dotnet_versions: str,
         target_framework_monikers: list,
         verbose: bool,
-        key: str = None) -> None:
+        azure_feed_url: str = None,
+        internal_build_key: str = None) -> None:
     '''
     Install tools used by this repository into the tools folder.
     This function writes a semaphore file when tools have been successfully
@@ -58,7 +59,8 @@ def init_tools(
         channels=channels,
         versions=dotnet_versions,
         verbose=verbose,
-        key=key
+        azure_feed_url=azure_feed_url,
+        internal_build_key=internal_build_key
     )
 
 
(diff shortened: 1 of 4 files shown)


adamsitnik (commented):
The code looks good, but when I am trying to follow the steps, I am getting:

```log
ERROR: The user, group or application '$removedForSecurityReasons' does not have secrets get permission on key vault '$anotherName'.
```

Also, when the key is printed to the log:

```log
[2022/10/07 09:54:48][INFO] $ powershell.exe -NoProfile -ExecutionPolicy Bypass D:\projects\forks\internal-build\tools\dotnet\x64\dotnet-install.ps1 -InstallDir D:\projects\forks\internal-build\tools\dotnet\x64 -Architecture x64 -AzureFeed https://dotnetbuilds.blob.core.windows.net/internal -FeedCredential secretKey -Version 7.0.100-rc.2.22477.23
```

We should either not print it when it's provided, anonymize it (replace with sth like "secret") or at least tell the users that they should not copy-paste entire log content to GH

dakersnar:
@adamsitnik I believe permissions are still propagating, so that first issue should be resolved soon.

Good point about the log, I forgot about that detail. Let me update the script to not print the key.

jeffhandley (approved):
LGTM but I wasn't able to test because I don't have the Azure CLI installed right now

PR #2642, opened 2022-10-10
Use latest docker containers

adiaaida:
This change moves all of the docker images to the latest tag as part of https://github.com/dotnet/arcade/issues/10377.


LoopedBard3 (approved):
Final error is unrelated, LGTM

PR #2629, opened 2022-10-04
Fix wasm builds

radical:
## NuGet.config: add dotnet8 feed

Building blazor scenarios with 8.0 sdk (used by `dotnet/runtime` `main`)
fails:

```
blazorminappaot/app/emptyblazorwasmtemplate.csproj:
error nu1102: Unable to find package Microsoft.AspNetCore.Components.WebAssembly with version (>= 8.0.0-alpha.1.22479.18)
error nu1102:   - Found 2091 version(s) in dotnet6 [ Nearest version: 6.0.0-rtm.21520.4 ]
error nu1102:   - Found 1610 version(s) in dotnet7 [ Nearest version: 7.0.0-rtm.22503.15 ]
error nu1102:   - Found 579 version(s) in dotnet5 [ Nearest version: 5.0.0-rtm.20512.8 ]
error nu1102:   - Found 58 version(s) in dotnet-public [ Nearest version: 7.0.0-rc.1.22427.2 ]
error nu1102:   - Found 1 version(s) in dotnet-tools [ Nearest version: 5.0.0-rc.2.20513.11 ]
error nu1102:   - Found 0 version(s) in benchmark-dotnet-prerelease
error nu1102:   - Found 0 version(s) in dotnet-eng
error nu1102:   - Found 0 version(s) in dotnet3.1-transport
error nu1102:   - Found 0 version(s) in dotnet3.1
error nu1102:   - Found 0 version(s) in dotnet5-transport
```

Add the new dotnet8 feed.

## Use RollForward=LatestMajor for MicroBenchmarks project

The project is built for `net7.0`, but run with `8.0` sdk, which fails
with:

```
App: /home/helixbot/work/ADE4096E/w/B4B209A5/e/performance/artifacts/bin/for-running/MicroBenchmarks/MicroBenchmarks.dll
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '7.0.0-rc.1.22426.10' (x64)
.NET location: /home/helixbot/work/ADE4096E/p/dotnet/

The following frameworks were found:
  8.0.0-alpha.1.22478.11 at [/home/helixbot/work/ADE4096E/p/dotnet/shared/Microsoft.NETCore.App]
  8.0.0-alpha.1.22502.5 at [/home/helixbot/work/ADE4096E/p/dotnet/shared/Microsoft.NETCore.App]

Learn about framework resolution:
https://aka.ms/dotnet/app-launch-failed

To install missing framework, download:
[...]

PR #2643, opened 2022-10-10
Fix parameter passing into dotnet.build.

cincuranet:


src/scenarios/shared/precommands.py
```diff
@@ -276,8 +276,7 @@ def _build(self, configuration: str, framework: str = None, output: str = None,
                            [framework],
                            output is None,
                            None,
-                           *['--output', output] if output else [],
-                           *build_args)
+                           (['--output', output] if output else []) + build_args)
```
DrewScoggins, line 279: Just curious, how was this failing before?

caaavik-msft: I suspect the problem is that `*['--output', output] if output else []` is the same as `(*['--output', output]) if output else []`, The array unpacking is only applying to the first case and not the 2nd, so if `output` is empty then instead of passing in no parameters, it passes in an empty array

caaavik-msft: looks like this commit changed the method to take a list rather than a star-args list of args https://github.com/dotnet/performance/pull/2629/files#diff-bb114884f249cbe31f5bc1d08f1fa3ab45f9229180b94880a0f0cc76467337ee

caaavik-msft (approved):

src/scenarios/shared/precommands.py
cincuranet: Yes, the change from star-args to plain list.

PR #2658, opened 2022-10-18
Add Environment namespace to fix specialfolderoptions not found.

LoopedBard3:
Microbenchmarks are hitting specialfolderoptions not found errors from merging in: https://github.com/dotnet/performance/pull/2654. This adds the beginning namespace and should fix the issue.


LoopedBard3:
It looks like this has fixed the micro runs. scenario failures are unrelated. Merging to get it in before the next runtime pipeline run.

PR #2661, opened 2022-10-21
Additional reflection invoke benchmarks

steveharter:
These are useful for comparing the 4-parameter to 5-parameter invoke since we do additional allocations starting with 5 parameters.


buyaa-n (approved):

LoopedBard3:
Failures are unrelated to these changes as the change only affects micro benchmarks. Merging.

PR #2668, opened 2022-10-25
Refactoring into ScenarioDirectoryName.

cincuranet:
To keep consistency with `scenarios.proj` (and even maybe future split between building and measuring).

### DIFF
--- a/eng/performance/sdk_scenarios.proj
+++ b/eng/performance/sdk_scenarios.proj
@@ -41,86 +41,108 @@
   
   <ItemGroup>
     <SDKWorkItem Include="@(Framework -> 'SDK Console Template %(Identity)')" Exclude="*2.1;*3.0;*3.1;*5.0;*6.0">
-      <PayloadDirectory>$(ScenariosDir)emptyconsoletemplate</PayloadDirectory>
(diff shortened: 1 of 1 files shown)


eng/performance/sdk_scenarios.proj
```diff
@@ -41,86 +41,108 @@
   
   <ItemGroup>
     <SDKWorkItem Include="@(Framework -> 'SDK Console Template %(Identity)')" Exclude="*2.1;*3.0;*3.1;*5.0;*6.0">
-      <PayloadDirectory>$(ScenariosDir)emptyconsoletemplate</PayloadDirectory>
+      <ScenarioDirectoryName>emptyconsoletemplate</ScenarioDirectoryName>
+      <PayloadDirectory>$(ScenariosDir)%(ScenarioDirectoryName)</PayloadDirectory>
```
caaavik-msft, line 45: I am not sure if it will work, but can this PayloadDirectory definition be moved into an ItemDefinitionGroup so that it doesn't need to be copy-pasted everywhere?

caaavik-msft (approved):

eng/performance/sdk_scenarios.proj
cincuranet: That will be part of bigger refactoring I'm doing in #2665.

PR #2679, opened 2022-11-01
Vector128 microbenchmarks

matouskozak:
Vector128 microbenchmarks. Both non-generics and generics versions. 

  
Local testing setup:
```
BenchmarkDotNet=v0.13.2.1940-nightly, OS=ubuntu 20.04
Intel Core i7-8665U CPU 1.90GHz (Coffee Lake), 1 CPU, 8 logical and 4 physical cores
.NET SDK=7.0.100-rc.1.22431.12
  [Host]     : .NET 7.0.0 (7.0.22.42610), X64 RyuJIT AVX2
  Job-ZRCHKB : .NET 8.0.0 (42.42.42.42424), X64 RyuJIT SSE4.2

PowerPlanMode=00000000-0000-0000-0000-000000000000  Runtime=MonoAOTLLVM  Toolchain=MonoAOTLLVM
IterationTime=250.0000 ms  MaxIterationCount=20  MinIterationCount=15
WarmupCount=1
```

Time needed to run new microbenchmarks locally using Mono AOT compilation with MiniJIT fallback:

| Microbenchmark         | Num benchmarks | Run time (hh:mm:ss) | Global time (hh:mm:ss) |
| ------------------------ | -------------------: |  ----------------------:| -------------------------:|   
| `Perf_Vector128`          |                          12 |                    00:01:33 |                        00:09:55 |
| `Perf_Vector128Float`  |                          48 |                    00:06:44 |                        00:15:07 |
| `Perf_Vector128Int`      |                          48 |                     00:06:47|                        00:15:23 |
| `Perf_Vector128Of<T>` |                        480 |                              XX |                         01:05:18 |

|  X X `<T>`  |  Run time (hh:mm:ss) |
|------------ | -----------------------:|
|`Byte`          |                     00:05:44|
|`Double`     |                     00:05:29|
|`Int16`        |                      00:05:29|
|`Int32`        |                      00:05:37|
|`Int64`        |                      00:05:33|
|`SByte`       |                      00:05:21|
|`Single`      |                      00:05:25|
|`UInt16`     |                      00:05:29|
|`UInt32`     |                      00:05:34|
|`UInt64`     |                      00:05:42|

### DIFF
--- a/src/benchmarks/micro/libraries/System.Runtime.Intrinsics/Perf_Vector128.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime.Intrinsics/Perf_Vector128.cs
@@ -1,0 +1,364 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Extensions;
+using MicroBenchmarks;
+
+namespace System.Runtime.Intrinsics.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.SIMD, Categories.JIT)]
+    public class Perf_Vector128
+    {
+        private static readonly Vector128<double> _vectorDouble = Vector128.Create(1.0, 2.0);
+        private static readonly Vector128<float> _vectorFloat = Vector128.Create(1.0F, 2.0F, 3.0F, 4.0F);
+        private static readonly Vector128<long> _vectorLong = Vector128.Create(1, 2);
+        private static readonly Vector128<ulong> _vectorULong = Vector128.Create((ulong)1, 2);
+        private static readonly Vector128<int> _vectorInt = Vector128.Create(1, 2, 3, 4);
+        private static readonly Vector128<uint> _vectorUInt = Vector128.Create((uint)1, 2, 3, 4);
+        
+        [Benchmark]
+        public Vector128<double> FloorDoubleBenchmark() => Vector128.Floor(_vectorDouble);
+
+        [Benchmark]
+        public Vector128<float> FloorFloatBenchmark() => Vector128.Floor(_vectorFloat);
+
+        [Benchmark]
+        public Vector128<double> CeilingDoubleBenchmark() => Vector128.Ceiling(_vectorDouble);
+
+        [Benchmark]
+        public Vector128<float> CeilingFloatBenchmark() => Vector128.Ceiling(_vectorFloat);
+
+        [Benchmark]
+        public Vector128<double> ConvertLongToDoubleBenchmark() => Vector128.ConvertToDouble(_vectorLong);
+
+        [Benchmark]
+        public Vector128<double> ConvertULongToDoubleBenchmark() => Vector128.ConvertToDouble(_vectorULong);
+
+        [Benchmark]
+        public Vector128<int> ConvertFloatToIntBenchmark() => Vector128.ConvertToInt32(_vectorFloat);
+
+        [Benchmark]
+        public Vector128<long> ConvertDoubleToLongBenchmark() => Vector128.ConvertToInt64(_vectorDouble);
+
+        [Benchmark]
+        public Vector128<float> ConvertIntToFloatBenchmark() => Vector128.ConvertToSingle(_vectorInt);
+
+        [Benchmark]
+        public Vector128<float> ConvertUIntToFloatBenchmark() => Vector128.ConvertToSingle(_vectorUInt);
+
+        [Benchmark]
+        public Vector128<uint> ConvertFloatToUIntBenchmark() => Vector128.ConvertToUInt32(_vectorFloat);
+        
+        [Benchmark]
+        public Vector128<ulong> ConvertDoubleToULongBenchmark() => Vector128.ConvertToUInt64(_vectorDouble);
+    }
+
+    [BenchmarkCategory(Categories.Libraries, Categories.SIMD, Categories.JIT)]
+    public class Perf_Vector128Float
+    {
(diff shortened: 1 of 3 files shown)


fanyang-mono (approved):

matouskozak:
Checked the failing CI lanes and they don't seem to be related to my change as they were also failing on main at that time e.g., here https://dev.azure.com/dnceng-public/public/_build/results?buildId=67891&view=logs&j=c031d5b3-1a95-5cd3-3bbd-1614cd88da98&t=da2cf600-eb7e-5b2b-f54c-73752134e5f3.

fanyang-mono:
@tannergooding Could you please review this?

tannergooding:
This also needs review/sign-off from someone like @adamsitnik or @DrewScoggins.

sblom:
This is great, @matouskozak!

I'm a little concerned about the 1 hour 45 minute total run-time for these tests.

Is there a meaningful subset that we can put into regular rotation that would catch most categories of potential perf regressions here using a fraction of that total time? (See [the microbenchmark design guidelines](https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-not-unit-tests) for some thoughts on this.)

@tannergooding, do you have any thoughts on ways to subset this? (both the generic type list and also perhaps operators with clustered perf behaviors)

tannergooding:
> @tannergooding, do you have any thoughts on ways to subset this? (both the generic type list and also perhaps operators with clustered perf behaviors)

This is a similar problem as with `Vector<T>` where they are core perf primitives that are critical to track perf on for various runtimes.

We could remove the duplicated sets between the `Perf_Vector128T<T>` and `Perf_Vector128Single`/`Perf_Vector128Int32` tests, but otherwise I think the ideal thing is to push for the parallelization support I've raised before. We could also look at sub setting some behind an "outerloop" perf run to differentiate between "critical" and "important".

cincuranet:
@sblom We're now running micro benchmarks in 5 partitions. Maybe we can add few more to split the work more and use simply more machines.

matouskozak:
> We could remove the duplicated sets between the `Perf_Vector128T<T>` and `Perf_Vector128Single`/`Perf_Vector128Int32` tests, but otherwise I think the ideal thing is to push for the parallelization support I've raised before. We could also look at sub setting some behind an "outerloop" perf run to differentiate between "critical" and "important".

I think we could also try to reduce the set of types tested in `Perf_Vector128Of<T>` (`Perf_Vector128T<T>` before, I renamed it to follow the vector naming). But this will be difficult as many APIs use different code paths depending on the type.

matouskozak:
Additional time-saving can be done by running the non-generic microbenchmark together. I tested it locally, and the global time was 26:54, whereas when run separately, it takes around 40 mins.

fanyang-mono:
@sblom Are you okay to merge this? Or is there a way to parallelize the run of these microbenchmarks?

sblom:
I don't think the "parallelize microbenchmarks" experiment is ready.

If we take [@matouskozak's update](https://github.com/dotnet/performance/pull/2679#issuecomment-1303940354), I guess this is okay if we really are exposed to the risk of any single one of these operations regressing with no sign of the regression on any others.

(Admitting my non-expertise here, I remain very surprised that we don't expect, for example `|` and `&` to be impacted with a correlation of approximately 1.0 in future perf regressions.)

kotlarmilos:
Hey @fanyang-mono is there an impact presented by Vector128 operations?

tannergooding:
> (Admitting my non-expertise here, I remain very surprised that we don't expect, for example | and & to be impacted with a correlation of approximately 1.0 in future perf regressions.)

@sblom, most of these APIs are treated intrinsically by the compiler and generate very specific assembly, often 1-3 instructions. In most hardware, even operations like `&` and `|` are "per type" since there may need to be other functionality, such as `masking`, done to fully support the various SIMD factors that can come into play.

For integers you often get `byte` (`int8`/`uint8`), `word` (`int16`/`uint16`), and `dword` (`int32`/`uint32`) everywhere with `qword` (`int64`/`uint64`) being a mix of "works everywhere" and "64-bit only" behavior. For some APIs, like `add`, the sign doesn't "normally" matter. For other APIs, like `greater than`, it does. The exact semantics of which matters is dependent on the target architecture and what ISAs are supported.

For floating-point, its often a separate co-processor/pipeline and will end up getting `float` and `double` separately.

This means that we have between 6 and 10 different types that have to be validated. Most of these types has a core usage:
* `byte` - UTF-8 string processing
* `word` - UTF-16 string processing (.NET default)
* `dword` - Native Target for most .NET integer handling
* `float` - Graphics Processing, Image Manipulation, Machine Learning

We then also have:
* `qword` - Native indexing and pointer support on 64-bit
* `double` - Similar scenarios as float, also the .NET default for many years

For the integer sizes, `byte`/`word` typically prefer `unsigned` where-as `dword` typically prefers `signed`. But we have a decent mix of both throughout, particularly when it comes to comparisons so we can emit more optimal sequences when the input values are known to fit a given constraint (such as being definitely positive).

fanyang-mono:
> Hey @fanyang-mono is there an impact presented by Vector128 operations?

What kind of impact? I don't quite understand your question.

kotlarmilos:
> > Hey @fanyang-mono is there an impact presented by Vector128 operations?
> 
> What kind of impact? I don't quite understand your question.

Speedup achieved by using vector optimizations, i.e. results with and without the changes

fanyang-mono:
@kotlarmilos There microbenchmarks once being merged will be run on the daily bases against `dotnet/runtime`. We will know when future change causes improvements or regression. @matouskozak also ran them locally previously against the changes he made for Vector128 on AMD64. And these microbnenchmarks did prove the impact of his changes.

fanyang-mono:
@sblom Shall we merge this? It seems that everyone agrees that these microbenchmarks are important to have.

sblom:
> Shall we merge this?

Alright.

sblom (approved):

Issue #2684, opened 2022-11-02
Redirect job send to runtime queues (.rt queues) to general use queue


AlitzelMendez:
Hi team,

I am from dnceng (dotnet engineering services) team, and we realize that the dotnet/performance team is sending jobs to a runtime queue, the runtime queues are all the queues that ends with ".rt". This is the yaml in where this job in being send incorrect: https://github.com/dotnet/performance/blob/main/azure-pipelines.yml#L274

That queue is only for runtime use, the best thing to do is to send that job a non-runtime queue, in this case the best match would be **centos.7.amd64.open** (removing the .rt part) 

I would really appreciate if you can do this change.

Let me know if you have any question.

Thanks @LoopedBard3 for redirecting to the right place to report this.

LoopedBard3:
cc: @caaavik-msft @sblom @cincuranet Future general guidance on not using .rt pipelines when we go at add them in the future. I will fix this quickly.

Issue #2646, opened 2022-10-11
Performance-CI CentOS runs failing with NotSupportedException


LoopedBard3:
Performance-ci CentOs runs are failing with the following error:
```
[2022/10/11 21:01:56][INFO] // Found 2 benchmarks:
[2022/10/11 21:01:56][INFO] //   ConfigurationBinderBenchmarks.Get: Job-RWGRKX(PowerPlanMode=00000000-0000-0000-0000-000000000000, Runtime=NativeAOT 7.0, Toolchain=Latest ILCompiler, InvocationCount=1, IterationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=0) [ConfigurationProvidersCount=16, KeysCountPerProvider=10]
[2022/10/11 21:01:56][INFO] //   ConfigurationBinderBenchmarks.Get: Job-RWGRKX(PowerPlanMode=00000000-0000-0000-0000-000000000000, Runtime=NativeAOT 7.0, Toolchain=Latest ILCompiler, InvocationCount=1, IterationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=0) [ConfigurationProvidersCount=32, KeysCountPerProvider=40]
[2022/10/11 21:01:56][INFO] 
[2022/10/11 21:01:56][INFO] // **************************
[2022/10/11 21:01:56][INFO] // Benchmark: ConfigurationBinderBenchmarks.Get: Job-RWGRKX(PowerPlanMode=00000000-0000-0000-0000-000000000000, Runtime=NativeAOT 7.0, Toolchain=Latest ILCompiler, InvocationCount=1, IterationCount=1, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=0) [ConfigurationProvidersCount=16, KeysCountPerProvider=10]
[2022/10/11 21:01:56][INFO] // *** Execute ***
[2022/10/11 21:01:56][INFO] // Launch: 1 / 1
[...]

PR #2685, opened 2022-11-02
Removed .rt from the CentOS.

LoopedBard3:
Removed .rt from the CentOS. Fixes: #2684

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -271,7 +271,7 @@ jobs:
       pool: 
         vmImage: ubuntu-latest
       machinePool: Open
-      queue: centos.7.amd64.open.rt
+      queue: centos.7.amd64.open
       container: centos_x64_build_container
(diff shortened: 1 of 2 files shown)


cincuranet (approved):

adamsitnik (commented):
@LoopedBard3 thank you for the fix and apologies for introducing the issue in the first place!

While you are at it you could mark the benchmark that fails for NativeAOT as unsupported by applying `[AotFilter]` to it:

https://github.com/dotnet/performance/blob/6e1585f375b41e1486420f264b0704d5f24d3694/src/benchmarks/micro/runtime/System.Reflection/Activator.cs#L26

https://github.com/dotnet/performance/blob/c44fc429e6e62a16d381526dbc298cf8191e85a0/src/benchmarks/micro/libraries/Microsoft.Extensions.Configuration/ConfigurationBinderBenchmarks.cs#L16-L17

which would also fix #2646

PR #2689, opened 2022-11-04
Add pipeline jobs for bepuphysics and ImageSharp

caaavik-msft:


cincuranet (approved):

azure-pipelines.yml
```diff
@@ -259,6 +259,38 @@ jobs:
...
+    parameters:
+      osName: windows
+      osVersion: 21H1
+      kind: bepuphysics
+      architecture: x64
+      pool:
+        vmImage: windows-2019
+      machinePool: Open
+      queue: Windows.10.Amd64.Client21H1.Open
+      csproj: src\benchmarks\real-world\bepuphysics2\DemoBenchmarks.csproj
```
kunalspathak, line 273: As seen in [README.md](https://github.com/dotnet/performance/tree/main/src/benchmarks/real-world/bepuphysics2), we might need to run the shorter version of the benchmarks (which takes 12 minutes) instead of default version (that takes 45 minutes). To run shorter version, you need to pass `-f *CollisionBatcherTaskBenchmarks.* *GroupedCollisionTesterBenchmarks.* *GatherScatterBenchmarks.* *OneBodyConstraintBenchmarks.* *TwoBodyConstraintBenchmarks.* *ThreeBodyConstraintBenchmarks.* *FourBodyConstraintBenchmarks.* *SweepBenchmarks.* *ShapeRayBenchmarks.* *ShapePileBenchmark.* *RagdollTubeBenchmark.*`

caaavik-msft: This is achieved using the Benchmark Categories I added, you will see I only added the category to these test in the filter, is this approach fine? that also means that to run just the shorter version locally you need to run `--anyCategories BepuPhysics`

kunalspathak: Ah ok.

kunalspathak (approved):
LGTM

PR #2696, opened 2022-11-04
Try forcing wasm to use 7.0 channel.

LoopedBard3:
A less impactful replacement for https://github.com/dotnet/performance/pull/2695. Fixes a wasm build issue by forcing wasm to target net7.0 instead of net8.0, while leaving the rest of the runs to use net8.0.

### DIFF
--- a/scripts/channel_map.py
+++ b/scripts/channel_map.py
@@ -28,7 +28,7 @@ class ChannelMap():
         },
         '7.0': {
             'tfm': 'net7.0',
-            'branch': '7.0.1xx',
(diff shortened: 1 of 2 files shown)


caaavik-msft (approved):

LoopedBard3:
cc @radical, here is the test run that seems to show this fixing the wasm issue: https://dev.azure.com/dnceng/internal/_build/results?buildId=2037728&view=results. Please ping if this is the incorrect run or the fix did not work like it did in the test run.

PR #2702, opened 2022-11-07
Fix the file name

kunalspathak:
Fix the file name of image file used by ImageSharp since it is case sensitive on linux.


kunalspathak:
@caaavik-msft

caaavik-msft (approved):

PR #2705, opened 2022-11-08
Update markdowns to net8.0.

LoopedBard3:
Updated markdowns to net8.0. This mostly consisted of changing net7.0 usages to net8.0, but also some changing of missed net6.0 and adding net8.0 to lists.

@adamsitnik Would it be appropriate to also replace all of the 7.0 references in https://github.com/dotnet/performance/blob/main/src/tools/ResultsComparer/README.md to net8.0? For example, change the net7.0-preview3 on line 53 to net8.0.


docs/benchmarkdotnet.md
```diff
@@ -4,29 +4,31 @@ BenchmarkDotNet is the benchmarking tool that allows to run benchmarks for .NET,
...
-- [Regressions](#regressions)
-- [Profiling](#profiling)
-- [Disassembly](#disassembly)
-- [Private Runtime Builds](#private-runtime-builds)
-  - [Running In Process](#running-in-process)
-  - [CoreRun](#corerun)
-  - [dotnet cli](#dotnet-cli)
-  - [Private CLR Build](#private-clr-build)
-  - [Private CoreRT Build](#private-corert-build)
+- [BenchmarkDotNet](#benchmarkdotnet)
```
adamsitnik, line 7: this document is dedicated to BenchmarkDotNet, what do we gain by adding one more level here?

image

docs/benchmarking-workflow-dotnet-runtime.md
```diff
@@ -8,6 +8,8 @@
     - [Code Organization](#code-organization)
     - [dotnet runtime Prerequisites for CLR](#dotnet-runtime-prerequisites-for-clr)
     - [dotnet runtime Prerequisites for wasm](#dotnet-runtime-prerequisites-for-wasm)
+      - [Run the benchmarks with the interpreter](#run-the-benchmarks-with-the-interpreter)
+      - [Run the benchmarks with AOT](#run-the-benchmarks-with-aot)
```
adamsitnik, line 12: great idea to add these items here 👍

docs/sdk-scenarios.md
```diff
@@ -45,19 +45,21 @@ Run precommand to create a new console template.
...
-| .NET 5 SDK   | x             | x             | x      |        |
-| .NET 6 SDK   | x             | x             | x      | x      |
+|              | netcoreapp2.1 | netcoreapp3.1 | net5.0 | net6.0 | net7.0 | net8.0 |
+|--------------|---------------|---------------|--------|--------|--------|--------|
+| .NET 2.1 SDK | x             |               |        |        |        |        |
+| .NET 3.1 SDK | x             | x             |        |        |        |        |
+| .NET 5 SDK   | x             | x             | x      |        |        |        |
+| .NET 6 SDK   | x             | x             | x      | x      |        |        |
+| .NET 7 SDK   | x             | x             | x      | x      | x      |        |
+| .NET 8 SDK   | x             | x             | x      | x      | x      | x      |
```
adamsitnik, line 62: thanks for adding the missing `net7.0` 👍

adamsitnik (approved):
LGTM, thank you for keeping the docs up to date @LoopedBard3 !

> Would it be appropriate to also replace all of the 7.0 references in https://github.com/dotnet/performance/blob/main/src/tools/ResultsComparer/README.md to net8.0? 

The net7.0-preview3 is used just as an example there, we can keep it the way it is (changing it would require changing the `Base V` and `Diff V` columns with some non-existing version numbers as we don't have any .NET 8 previews yet).

adamsitnik:
@mrsharm PTAL at the GC part

docs/benchmarkdotnet.md
LoopedBard3: Fixed this, my linter automatically added it since it is a header.

docs/benchmarkdotnet.md
```diff
@@ -17,10 +17,10 @@ BenchmarkDotNet is the benchmarking tool that allows to run benchmarks for .NET,
 - [Reading the Results](#reading-the-results)
   - [Reading the Histogram](#reading-the-histogram)
   - [Reading Memory Statistics](#reading-memory-statistics)
-- [Multiple Runtimes](#multiple-runtimes)
-- [Regressions](#regressions)
 - [Profiling](#profiling)
 - [Disassembly](#disassembly)
+- [Multiple Runtimes](#multiple-runtimes)
```
mrsharm, line 22: Is this change necessary? Seems like we are just moving these links few lines below.

mrsharm:
LGTM for the most part! Thanks for the heads up, @adamsitnik @LoopedBard3!

PR #2717, opened 2022-11-10
Update benchmarks_monthly.py for net7.0 GA

dakersnar:
@adamsitnik I'm mirroring how 6.0 was set up in this file. Should this work for 7.0 GA, or do I still need to find a specific build number like we did for the previews?


dakersnar:
This works locally, so I'm going to merge it.

PR #2727, opened 2022-11-14
don't build Vector128 benchmarks for older runtimes

adamsitnik:
I am trying to run the benchmarks for .NET 6 vs 7 comparison and I am getting a lot of compiler errors.


dakersnar (approved):
@tannergooding is this expected when comparing against .NET 6? Is there anything else we want to exclude (Vector64, Vector256, Vector<T>, etc?)

@adamsitnik once this gets merged, I'll send a follow up email asking people to pull again. Thanks for the find.

PR #2735, opened 2022-11-16
Clean up the azure-pipelines.yml file

caaavik-msft:
As I was getting the real-world benchmarks set up for the private runs, I found it was difficult to keep track of all the different pipelines and all the different environments that they run on. This PR creates a template called build_machine_matrix.yaml which allows you to apply a template for a list of build machines, significantly reducing duplication and putting the configuration of the machines in a central location.

This PR also adds private pipeline runs for the real world benchmarks

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -128,10 +128,10 @@ jobs:
         - ubuntu-x64
       isPublic: true
       jobParameters:
-          kind: micro
-          csproj: src\benchmarks\micro\MicroBenchmarks.csproj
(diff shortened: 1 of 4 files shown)


cincuranet (approved):

LoopedBard3 (approved):
LGTM

caaavik-msft:
> LGTM, although it looks like the RS5 x64 run has a potentially related issue.

Ah, thanks for catching that, fixed in the most recent commit

PR #2763, opened 2022-11-29
Upload the trace file and related info in case of failure (for easier debugging/fixing later)

cincuranet:


src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -332,9 +332,20 @@ static void checkArg(string arg, string name)
...
+                {
+                    var counters = parser.Parse(traceFilePath, processName, pids, commandLine);
+                    CreateTestReport(scenarioName, counters, reportJsonPath, logger);
+                }
+                catch
+                {
+                    logger.Log($"{nameof(processName)} = {processName}");
+                    logger.Log($"{nameof(pids)} = {string.Join(", ", pids)}");
+                    logger.Log($"{nameof(commandLine)} = {commandLine}");
+                    logger.Log(Convert.ToBase64String(File.ReadAllBytes(traceFilePath)));
```
DrewScoggins, line 346: Rather than write this out to the log, we should copy the file to the `HELIX_WORKITEM_UPLOAD_ROOT` so that it is automatically picked up by Helix and saved for the workitem. Below I have linked the documentation where this comes from.

https://github.com/dotnet/arcade/blob/20511c59bfded2dd75953c05bc59ef7ffbc264c0/src/Microsoft.DotNet.Helix/Sdk/Readme.md#common-helix-client-environment-variables

cincuranet: Right. Maybe less convenient, but better design.

LoopedBard3 (approved):
LGTM with a successful test run.

cincuranet:
The _ubuntu 1804 x64 sdk_scenarios Open main_ failing is actually the driving factor and was failing before. Merging.

PR #2772, opened 2022-12-06
Augment System.Uri tests

MihaZupan:
This change also modifies the existing benchmarks (not sure how much we care in this case):
- `Ctor` now takes arguments
- `Unescape` now takes arguments, so I took the opportunity to rename it if we're already breaking history
- The Uri used by `GetComponents` is longer, which will show up as a regression (it wasn't really doing much before)


stephentoub (approved):
Thanks.

PR #2774, opened 2022-12-07
dotnet.py: `--output` is no longer supported with `dotnet build/publish`

radical:
.. with .sln files. Instead, pass `/p:PublishDir=..` to the build command.

This changed in https://github.com/dotnet/sdk/pull/29065, and broke `dotnet-runtime-perf` pipeline's blazor scenarios.

```
$ dotnet publish /home/helixbot/work/A8850905/p/performance/src/scenarios/blazorpizza/app/BlazingPizza.sln --configuration Release --output pub /p:NuGetPackageRoot=/home/helixbot/work/A8850905/w/A9A608EA/u/artifacts/packages/ /p:UseSharedCompilation=false /p:BuildInParallel=false /m:1 --framework net7.0 /p:_TrimmerDumpDependencies=true -bl:./traces/blazor_publish.binlog -p:WasmNativeWorkload=false
MSBuild version 17.5.0-preview-22601-03+a2490dd3f for .NET

/home/helixbot/work/A8850905/p/dotnet/sdk/8.0.100-alpha.1.22606.3/Current/SolutionFile/ImportAfter/Microsoft.NET.Sdk.Solution.targets(36,5):
    error NETSDK1194: The "--output" option isn't supported when building a solution. [/home/helixbot/work/A8850905/p/performance/src/scenarios/blazorpizza/app/BlazingPizza.sln]
```


radical:
Running with this change in https://dev.azure.com/dnceng-public/public/_build/results?buildId=104271&view=logs&j=9331bb7a-642b-57f2-cf5f-0083d34246f4 with a test PR - https://github.com/dotnet/runtime/pull/79328 .

radical:
Without this the `dotnet-runtime-perf` pipeline is broken for blazor scenarios.

LoopedBard3 (approved):
LGTM, once the test run finishes. Does this also need to be backported to release7.0?

radical:
> Does this also need to be backported to release7.0?

The change seems to be only in the 8.0 sdk.

zzvara:
@LoopedBard3 @radical this is also an issue in .NET 7. We upgraded our CI build image running on Azure DO, which bumped to a newer .NET 7 patch version. The same build where only the underlying build image has changed a patch version of .NET 7 now fails.

```
/usr/share/dotnet/sdk/7.0.200/Current/SolutionFile/ImportAfter/Microsoft.NET.Sdk.Solution.targets(36,5): Error NETSDK1194: The "--output" option isn't supported when building a solution.
```

PR #2789, opened 2022-12-13
[Release/7.0] Add Maui Mobile Android Benchmarks

LoopedBard3:
This adds Maui Mobile scenarios for android, including android default, android podcast app, and blazor android. This also updates how the dotnet version commit date is retrieved to minimize throttling from running in public pools.

### DIFF
--- a/eng/performance/maui_scenarios_android.proj
+++ b/eng/performance/maui_scenarios_android.proj
@@ -15,6 +15,10 @@
     <PreparePayloadWorkItemBaseDirectory Condition="'$(TargetsWindows)' != 'true'">$(CorrelationPayloadDirectory)$(PreparePayloadOutDirectoryName)/</PreparePayloadWorkItemBaseDirectory>
   </PropertyGroup>
 
+  <Target Name="RemoveDotnetFromCorrelationStaging" BeforeTargets="BeforeTest">
+    <Message Text="Removing Dotnet from Correlation Staging" Importance="high" />
+    <RemoveDir Directories="$(CorrelationPayloadDirectory)\dotnet" />
+  </Target>
 
   <ItemGroup>
     <MAUIAndroidScenario Include="Maui Android Default">
--- a/eng/performance/maui_scenarios_android.proj
+++ b/eng/performance/maui_scenarios_android.proj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">
 
   <Import Project="Scenarios.Common.props" />
+  
   <PropertyGroup>
     <IncludeXHarnessCli>true</IncludeXHarnessCli>
     <MicrosoftDotNetXHarnessCLIVersion>1.0.0-prerelease.21566.2</MicrosoftDotNetXHarnessCLIVersion>
--- a/eng/performance/maui_scenarios_android.proj
+++ b/eng/performance/maui_scenarios_android.proj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">
 
   <Import Project="Scenarios.Common.props" />
-  
+
   <PropertyGroup>
     <IncludeXHarnessCli>true</IncludeXHarnessCli>
     <MicrosoftDotNetXHarnessCLIVersion>1.0.0-prerelease.21566.2</MicrosoftDotNetXHarnessCLIVersion>
@@ -34,7 +34,7 @@
 
   <ItemGroup>
     <PreparePayloadWorkItem Include="@(MAUIAndroidScenario)">
-      <Command>$(Python) pre.py -f $(PERFLAB_Framework)-android -o $(PreparePayloadWorkItemBaseDirectory)%(PreparePayloadWorkItem.ScenarioDirectoryName) -r android-arm64 --self-contained</Command>
+      <Command>$(Python) pre.py publish -f $(PERFLAB_Framework)-android -o $(PreparePayloadWorkItemBaseDirectory)%(PreparePayloadWorkItem.ScenarioDirectoryName) -r android-arm64 --self-contained</Command>
       <WorkingDirectory>%(PreparePayloadWorkItem.PayloadDirectory)</WorkingDirectory>
     </PreparePayloadWorkItem>
   </ItemGroup>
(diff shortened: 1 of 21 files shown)


LoopedBard3:
Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2066069&view=results

scripts/dotnet.py
```diff
@@ -612,7 +613,7 @@ def get_commit_date(
         raise ValueError('.NET Commit sha was not defined.')
 
     url = None
-    urlformat = 'https://api.github.com/repos/%s/%s/commits/%s'
+    urlformat = 'https://github.com/%s/%s/commit/%s.patch'
```
caaavik-msft, line 616: Could we add a comment with a sample URL? It's hard to review without knowing what the response looks like

src/scenarios/mauiandroid/test.py
```diff
@@ -1,11 +1,26 @@
...
 if __name__ == "__main__":
+    try:
+        with open(f'pub/{MAUIVERSIONFILE}', 'r') as f:
+            maui_version = f.read()
+            if("sha" not in maui_version or "azdo" not in maui_version):
+                raise ValueError(f"MAUI_VERSION does not contain sha and azdo indicating failure to retrieve or set the value. MAUI_VERSION: {maui_version}")
+            else:
+                print(f"Found MAUI_VERSION {maui_version}")
+                os.environ["MAUI_VERSION"] = maui_version
+        os.remove(f'pub/{MAUIVERSIONFILE}')
```
caaavik-msft, line 19: What's the purpose of removing the file?

src/scenarios/mauiandroid/pre.py
```diff
@@ -3,47 +3,58 @@
...
+os.remove(f"{output_file_partial_path}.aab")
+
+# Copy the MauiVersion to a file so we have it on the machine
+result = subprocess.run(['powershell', '-Command', rf'Get-ChildItem .\{const.APPDIR}\obj\Release\net7.0-android\android-arm64\linked\Microsoft.Maui.dll | Select-Object -ExpandProperty VersionInfo | Select-Object ProductVersion | Select-Object -ExpandProperty ProductVersion'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+maui_version = result.stdout.decode('utf-8').strip()
+print(f'MAUI_VERSION: {maui_version}')
+if("sha" not in maui_version or "azdo" not in maui_version):
+    raise ValueError(f"MAUI_VERSION does not contain sha and azdo indicating failure to retrieve or set the value. MAUI_VERSION: {maui_version}")
+with open(f'{args.output_dir}/{MAUIVERSIONFILE}', 'w') as f:
+    f.write(maui_version)
```
caaavik-msft, line 60: Is passing data via files the only approach for sending variables from pre to test? There isn't any other way we can pass arguments in or maybe use environment variables?

src/scenarios/mauiblazorandroid/pre.py
```diff
@@ -1,49 +1,99 @@
...
+# Remove the aab files as we don't need them, this saves space
+output_file_partial_path = f"com.companyname.{str.lower(EXENAME)}"
+if args.output_dir:
+    output_file_partial_path = os.path.join(args.output_dir, output_file_partial_path)
 
+os.remove(f"{output_file_partial_path}-Signed.aab")
+os.remove(f"{output_file_partial_path}.aab")
 
+# Copy the MauiVersion to a file so we have it on the machine
+result = subprocess.run(['powershell', '-Command', rf'Get-ChildItem .\{const.APPDIR}\obj\Release\net7.0-android\android-arm64\linked\Microsoft.Maui.dll | Select-Object -ExpandProperty VersionInfo | Select-Object ProductVersion | Select-Object -ExpandProperty ProductVersion'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
```
caaavik-msft, line 93: There is a lot of duplication here in mauiandroid, mauiandroidpodcast, and mauiblazorandroid. Is it possible instead to store the powershell command as a separate script, then run that script from python instead? that way we don't need to update the script in multiple places.

src/scenarios/mauiandroid/pre.py
```diff
@@ -3,47 +3,58 @@
...
+# Remove the aab files as we don't need them, this saves space
+output_file_partial_path = f"com.companyname.{str.lower(EXENAME)}"
+if args.output_dir:
+    output_file_partial_path = os.path.join(args.output_dir, output_file_partial_path)
+
+os.remove(f"{output_file_partial_path}-Signed.aab")
+os.remove(f"{output_file_partial_path}.aab")
+
+# Copy the MauiVersion to a file so we have it on the machine
+result = subprocess.run(['powershell', '-Command', rf'Get-ChildItem .\{const.APPDIR}\obj\Release\net7.0-android\android-arm64\linked\Microsoft.Maui.dll | Select-Object -ExpandProperty VersionInfo | Select-Object ProductVersion | Select-Object -ExpandProperty ProductVersion'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
```
caaavik-msft, line 54: I may be wrong as I don't use Powershell much, but I think you can remove the `Select-Object ProductVersion` since all it is doing is trimming the object down to one property, when you can just expand the property on the original object

caaavik-msft (changes requested):

src/scenarios/mauiandroid/pre.py
```diff
@@ -3,47 +3,58 @@
...
-    copyfile(apkname, os.path.join(PUBDIR, apkname))
-
-
-
+        help='capture of the output directory')
+args, unknown_args = parser.parse_known_args()
+
+# Download what we need
+with open ("MauiNuGet.config", "wb") as f:
+    f.write(requests.get(f'https://raw.githubusercontent.com/dotnet/maui/net7.0/NuGet.config', allow_redirects=True).content)
```
cincuranet, line 29: I suppose this will get (forward)ported to `main` eventually and hence the `net7.0` here (and below) should come from variable.

src/scenarios/mauiandroidpodcast/pre.py
```diff
@@ -0,0 +1,54 @@
...
+        required=False,
+        type=str,
+        help='capture of the output directory')
+args, unknown_args = parser.parse_known_args()
+
+# Download what we need
+with open ("MauiNuGet.config", "wb") as f:
+    f.write(requests.get(f'https://raw.githubusercontent.com/dotnet/maui/net7.0/NuGet.config', allow_redirects=True).content)
+
+subprocess.run(['git', 'clone', 'https://github.com/microsoft/dotnet-podcasts.git', '-b', 'net7.0', '--single-branch', '--depth', '1'])
```
cincuranet, line 28: Did you considered using submodule?

src/scenarios/mauiblazorandroid/pre.py
cincuranet: I agree. And even parts of Python code are duplicated in `pre.py`/`test.py`. This could be refactored, once in final stretch, to separate files.

cincuranet (changes requested):

src/scenarios/mauiandroid/test.py
LoopedBard3: The reason for the file being removed is to ensure it is not included in any Size On Disk tests. If I find a good workaround to not use a file, then this will not be necessary.

src/scenarios/mauiandroid/pre.py
LoopedBard3: Since the pre and test files are run on different machines, I did not find a good way to pass the version as a variable. Although I am open to suggestions and am looking to see if it would be possible to get an environment variable from the pre.py run, save it, and pass it later in a MSBuild task.

cincuranet: Given the `proj` file is executed twice effectively, probably only sharing something through pipeline variable is an option (using `echo vso...`). But I don't think that's the way to go.

I think the file is OK. Maybe even create it as a JSON file, so it's ready for more values in the future (ultimate solution would be to have encapsulated helper for this).

Other option would be to encode the MAUI version into i.e. directory name (in the payload) and take it from there. Similar to what we do with report files in blob storage and all the different configurations. It's more hacky and limited than file, but also more straightforward.

I personally would vote for (JSON) file.

src/scenarios/mauiandroidpodcast/pre.py
LoopedBard3: I had not, but looking into submodules more, I think that the submodule commands would be overdoing what we need and come with additional cost of either merging the submodule each time the podcast app updates or having to call git multiple times to add and then init the submodule. However, since I have not used it much before I may just be misunderstanding the docs.

src/scenarios/mauiandroid/pre.py
LoopedBard3: I like this idea of having a generic Version JSON file. I will generalize this to use JSON, then down the line the same file can be used for any other version info we want to include from the build stage.

LoopedBard3: For the current iteration, I added a shared set of version management functions to create a JSON file from a dictionary, and to read and save variables from this JSON file to the environment. This format minimizes the generation needed as the python json import provides most of the functionality. @caaavik-msft let me know if this seems like a reasonable middle ground.

src/scenarios/mauiandroid/pre.py
LoopedBard3: This has been fixed

eng/performance/maui_scenarios_android.proj
```diff
@@ -0,0 +1,79 @@
...
+      <WorkingDirectory>%(PreparePayloadWorkItem.PayloadDirectory)</WorkingDirectory>
+    </PreparePayloadWorkItem>
+  </ItemGroup>
+
+<!-- We only run the android tests from Windows machines (at least for now) -->
+  <ItemGroup>
+    <HelixWorkItem Include="@(MAUIAndroidScenario -> 'SOD - %(Identity) APK Size')">
+      <PreCommands>echo on; xcopy %HELIX_CORRELATION_PAYLOAD%\$(PreparePayloadOutDirectoryName)\%(HelixWorkItem.ScenarioDirectoryName) %HELIX_WORKITEM_ROOT%\pub\ /E /I /Y</PreCommands>
+      <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
+      <Timeout>00:30</Timeout>
```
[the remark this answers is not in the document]
LoopedBard3: This has been fixed

scripts/dotnet.py
LoopedBard3: This has been updated.

src/scenarios/mauiandroidpodcast/pre.py
cincuranet: You can track tip of branch with submodule. But my comment was purely question. I don't think either is superior solution and both work for me.

cincuranet (approved):

LoopedBard3:
Working test run since this only runs on private builds: https://dev.azure.com/dnceng/internal/_build/results?buildId=2068946&view=results

LoopedBard3:
Successful test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2095104&view=results

LoopedBard3:
Errors are unrelated to these changes, merging.

PR #2816, opened 2023-01-06
[wasm] Update to build for tfm=net8.0

radical:
- Blazor scenarios: Use 8.0.0-* nugets when building for net8.0 .
  - Without this change, when building for net7/net8, the aspnetcore assemblies were used from 6.0.0 packages.
- Also, remove workarounds added earlier when we were targeting net7.0 but with the 8.0 sdk. This allows building for net8.0 with the 8.0 sdk.
- Add a `blazor_scenarios` job for `ubuntu`
- Add binlogs for blazor scenarios

### DIFF
--- a/eng/performance/blazor_scenarios.proj
+++ b/eng/performance/blazor_scenarios.proj
@@ -1,11 +1,18 @@
 <Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">
 
-  <ItemGroup>
-    <HelixCorrelationPayload Include="$(CorrelationPayloadDirectory)">
-      <PayloadDirectory>%(Identity)</PayloadDirectory>
-    </HelixCorrelationPayload>
-  </ItemGroup>
-  
+  <PropertyGroup>
+    <LogDirectory Condition="'$(TargetsWindows)' == 'true'">%HELIX_WORKITEM_UPLOAD_ROOT%\</LogDirectory>
+    <LogDirectory Condition="'$(TargetsWindows)' != 'true'">%24{HELIX_WORKITEM_UPLOAD_ROOT}/</LogDirectory>
+
+    <PerfLabTargetFrameworksEnvVar Condition="'$(TargetsWindows)' == 'true'">%PERFLAB_TARGET_FRAMEWORKS%</PerfLabTargetFrameworksEnvVar>
+    <PerfLabTargetFrameworksEnvVar Condition="'$(TargetsWindows)' != 'true'">%24{PERFLAB_TARGET_FRAMEWORKS}</PerfLabTargetFrameworksEnvVar>
+    <PublishArgs>--msbuild "/p:_TrimmerDumpDependencies=true" --msbuild /warnaserror:NU1602,NU1604 --msbuild-static AdditionalMonoLinkerOptions=%27&quot;%24(AdditionalMonoLinkerOptions) --dump-dependencies&quot;%27 --binlog $(LogDirectory)blazor_publish.binlog</PublishArgs>
+    <PublishCommand>$(Python) pre.py publish $(PublishArgs)</PublishCommand>
+
+    <PizzaAppPubLocation Condition="'$(TargetsWindows)' == 'true'">pub\wwwroot</PizzaAppPubLocation>
+    <PizzaAppPubLocation Condition="'$(TargetsWindows)' != 'true'">pub/wwwroot</PizzaAppPubLocation>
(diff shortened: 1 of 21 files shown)


src/scenarios/blazorpizza/src/Directory.Build.props
```diff
@@ -1,8 +1,12 @@
 <Project>
-    <PropertyGroup>
+    <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+        <AspNetCoreVersion Condition="'$(AspNetCoreVersion)' == ''">8.0.0-*</AspNetCoreVersion>
+        <BlazorVersion Condition="'$(AspNetCoreVersion)' == ''">8.0.0-*</BlazorVersion>
+        <SystemNetHttpJsonVersion>8.0.0-*</SystemNetHttpJsonVersion>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(TargetFramework)' != 'net8.0'">
```
lewing, line 7: Is there a reason to not set net7.0 correctly here and in release/7.0 as well and maybe add a warning for net9?

lewing (commented):
Looks good to me but it might be nice to indicate the build this is first used in, I expect size regressions and I'm not sure what the best way to isolate them specifically is

src/scenarios/blazorpizza/src/Directory.Build.props
radical: I don't know what the existing cases are, or what depends on these older versions, so I went with preserving older behavior except for the cases that I know about.

If there isn't anything depending on this, then we can set it to net7.0, and the blazor, and STJ versions to match.

radical:
I'll add the blazor package versions to the report in a follow up PR.

radical:
The failures are known issues. The Windows/micro_net462 will be fixed by https://github.com/dotnet/performance/pull/2836 .

eng/performance/blazor_scenarios.proj
```diff
@@ -1,11 +1,18 @@
...
-    </HelixCorrelationPayload>
-  </ItemGroup>
-  
+  <PropertyGroup>
+    <LogDirectory Condition="'$(TargetsWindows)' == 'true'">%HELIX_WORKITEM_UPLOAD_ROOT%\</LogDirectory>
+    <LogDirectory Condition="'$(TargetsWindows)' != 'true'">%24{HELIX_WORKITEM_UPLOAD_ROOT}/</LogDirectory>
+
+    <PerfLabTargetFrameworksEnvVar Condition="'$(TargetsWindows)' == 'true'">%PERFLAB_TARGET_FRAMEWORKS%</PerfLabTargetFrameworksEnvVar>
+    <PerfLabTargetFrameworksEnvVar Condition="'$(TargetsWindows)' != 'true'">%24{PERFLAB_TARGET_FRAMEWORKS}</PerfLabTargetFrameworksEnvVar>
+    <PublishArgs>--msbuild "/p:_TrimmerDumpDependencies=true" --msbuild /warnaserror:NU1602,NU1604 --msbuild-static AdditionalMonoLinkerOptions=%27&quot;%24(AdditionalMonoLinkerOptions) --dump-dependencies&quot;%27 --binlog $(LogDirectory)blazor_publish.binlog</PublishArgs>
```
ilonatommy, line 9: Passing the same arg to python's argparser multiple times has an effect of ignoring all args but the last one. I doubt `_TrimmerDumpDependencies=true` was set.
E.g. running `python3 ./pre.py publish -f net7.0 --readonly-dotnet --has-workload  --msbuild "/p:HybridGlobalization=true" --msbuild "/p:RunAOTCompilation=true" --msbuild /warnaserror:NU1602,NU1604` results in:
`dotnet publish /workspaces/performance/src/scenarios/blazorpizza/app/BlazingPizza.sln --configuration Release /p:NuGetPackageRoot=/workspaces/performance/artifacts/packages/ /p:RestorePackagesPath=/workspaces/performance/artifacts/packages/ /p:UseSharedCompilation=false /p:BuildInParallel=false /m:1 /p:PublishDir=/workspaces/performance/src/scenarios/blazorpizza/pub --framework net7.0 /warnaserror:NU1602,NU1604 -p:WasmNativeWorkload=false /p:EnableWindowsTargeting=true`
and running with e.g. `--msbuild "/p:HybridGlobalization=true" --msbuild "/p:RunAOTCompilation=true" ` results in `/p:RunAOTCompilation=true -p:WasmNativeWorkload=false /p:EnableWindowsTargeting=true`

Issue #2819, opened 2023-01-09
Cleanup pipelines from .NET Core 3.1.x runs


cincuranet:
3.1.x is now out of support.

PR #2822, opened 2023-01-09
[release/3.1.4xx] Disable pipelines

cincuranet:
Related to #2821.
Part of #2819.


LoopedBard3 (commented):
Would it make sense to set the trigger to 'trigger: none' instead, so that it is easier to tell that it is disabled?

cincuranet:
Yeah, that's probably better idea. I wanted to keep the structure, but I guess that can be found in history if needed.

LoopedBard3 (approved):
LGTM

PR #2842, opened 2023-01-16
Set value into registry to prevent PDN checking for updates

### LINKED ISSUE #2841: PDN without update checks
At the moment PDN scenario fails from time to time because there's new alpha or RTM version and the update dialog is "blocking" the main window.

cincuranet:
Workaround by setting `Updates/LastCheckTimeUtc` to "recent" value.

Successful run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2089541&view=results.

Fixes #2841.


LoopedBard3 (approved):
LGTM

PR #2851, opened 2023-01-20
Fix `Desktop Maui Blazor Default` to generate `PerfLabGenericEventSource` `Startup` event

cincuranet:

PR #2860, opened 2023-01-24
Install WebView on the testing machine, not CI.

cincuranet:
Porting #2851 also into `main` to have it there when MAUI on `main` is ready again.


LoopedBard3 (approved):
LGTM, good find

PR #2861, opened 2023-01-24
Is NuGetPackageRoot even needed?

cincuranet:


cincuranet:
Yes, it is. :) Closing.

PR #2870, opened 2023-01-27
Run ILLink only on x64

cincuranet:
Fixes failure after #2864. [Here](https://github.com/dotnet/performance/blob/main/src/benchmarks/real-world/ILLink/Utilities.cs#L27) is the check.


LoopedBard3 (approved):
LGTM

PR #2871, opened 2023-01-30
SOD formatting

cincuranet:


LoopedBard3 (approved):
LGTM

DrewScoggins:
Sorry, I was slow here, but our typical code convention is to use `string` for declaring variables, and to use `String` for all of the static functions and properties.

cincuranet:
Do you want me to revert the string part?

DrewScoggins:
Yeah, sorry I forgot to add that part in. I just want to make sure that we keep consistent style across our code.

cincuranet:
#2875 corrects this. I think `.editorconfig` is (over)due. Because the `string` and `String` usage is inconsistent across the codebase.

PR #2877, opened 2023-01-30
Add Android Memory Consumption testing capabilities.

LoopedBard3:
This adds the necessary support to the runner.py file and a new memory consumption scenario measurement tool to support testing Android memory consumption testing.


src/tools/ScenarioMeasurement/MemoryConsumption/AndroidMemoryParser.cs
```diff
@@ -0,0 +1,84 @@
...
+            var ussMins = new List<double>();
+            var ussAvgs = new List<double>();
+            var ussMaxs = new List<double>();
+            var rssMins = new List<double>();
+            var rssAvgs = new List<double>();
+            var rssMaxs = new List<double>();
+            var captureCounts = new List<double>();
+            // 52MB-52MB-52MB/45MB-45MB-45MB/138MB-138MB-138MB over 1
+            // PSS: min {pss[0]}, avg {pss[1]}, max {pss[2]}; USS: min {uss[0]}, avg {uss[1]}, max {uss[2]}; RSS: min {rss[0]}, avg {rss[1]}, max {rss[2]}; Number: {CaptureNumber}
+            Regex memoryConsumptionPattern = new Regex(@"(?<pssMin>\d+)MB-(?<pssAvg>\d+)MB-(?<pssMax>\d+)MB\/(?<ussMin>\d+)MB-(?<ussAvg>\d+)MB-(?<ussMax>\d+)MB\/(?<rssMin>\d+)MB-(?<rssAvg>\d+)MB-(?<rssMax>\d+)MB\s*over\s*(?<captureNumber>\d+)");
```
cincuranet, line 42: Better to use `Regex.Matches` down below, instead of creating your own instance manually.

src/tools/ScenarioMeasurement/MemoryConsumption/IParser.cs
```diff
@@ -0,0 +1,14 @@
...
+using System;
+
+namespace ScenarioMeasurement
+{
+    public interface IParser
+    {
+        void EnableUserProviders(ITraceSession user);
+        void EnableKernelProvider(ITraceSession kernel);
+        IEnumerable<Counter> Parse(string mergeTraceFile, string processName, IList<int> pids, string commandLine);
+        IEnumerable<Counter> Parse(string mergeTraceDirectory, string mergeTraceFilter, string processName, IList<int> pids, string commandLine) => throw new NotImplementedException();
```
cincuranet, line 12: Is this needed for something? I don't see it ever implemented/overridden. Maybe for now, it can be removed and when time comes, it will be implemented.

src/tools/ScenarioMeasurement/MemoryConsumption/MemoryConsumption.cs
```diff
@@ -0,0 +1,392 @@
...
+            if (!skipMeasurementIteration)
+            {
+                // Run trace session
+                using (var traceSession = TraceSessionManager.CreateSession("MemoryConsumptionSession", traceName, traceDirectory, logger))
+                {
+                    traceSession.EnableProviders(parser);
+                    for (int i = 0; i < iterations; i++)
+                    {
+                        logger.LogIterationHeader($"Iteration {i}");
+                        (bool Success, List<int> Pids) iterationResult;
```
cincuranet, line 192: Declaration and assignment can be one line (or at least seems so).

src/tools/ScenarioMeasurement/MemoryConsumption/MemoryConsumption.cs
```diff
@@ -0,0 +1,392 @@
...
+                    }
+                }
+            }
+
+            return (failed ? -1 : 0);
+        }
+
+        private static IProcessHelper CreateProcHelper(string command, string args, bool runWithExit, Logger logger)
+        {
+            IProcessHelper procHelper;
```
cincuranet, line 266: I would return directly from the if-else and skip the variable.

src/tools/ScenarioMeasurement/MemoryConsumption/MemoryConsumption.cs
```diff
@@ -0,0 +1,392 @@
...
+                {
+                    pids.Add(runResult.Pid);
+                }
+            }
+
+            if (runResult.Proc != null)
+            {
+                logger.LogStepHeader("Stopping process");
+                runResult.Proc.Kill();
+                Thread.Sleep(2000);
```
cincuranet, line 345: Use constant for `2000`.

src/tools/ScenarioMeasurement/MemoryConsumption/ParserUtility.cs
```diff
@@ -0,0 +1,132 @@
...
+                        return CompareResult.Mismatch;
+                    }
+                }
+            }
+            return CompareResult.Match;
+        }
+
+        public static bool MatchSingleProcessID(TraceEvent evt, TraceSourceManager source, int pid)
+        {
+            return MatchProcessID(evt, source, new List<int> { pid});
```
cincuranet, line 112: nit: Formatting.

src/tools/ScenarioMeasurement/MemoryConsumption/ParserUtility.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Data;
+using System.Diagnostics;
+using System.IO;
+
+
+namespace ScenarioMeasurement
+{
+    class ParserUtility
+    {
+
```
cincuranet, line 13: nit: Formatting.

src/tools/ScenarioMeasurement/MemoryConsumption/TraceSourceManager.cs
```diff
@@ -0,0 +1,81 @@
...
+        public event Action<TraceEvent> ProcessStart { add { parser.ProcessStart += value; } remove { parser.ProcessStart -= value; } }
+        public event Action<TraceEvent> ProcessStop { add { parser.ProcessEndGroup += value; } remove { parser.ProcessEndGroup -= value; } }
+        public event Action<TraceEvent> ContextSwitch { add { parser.ThreadCSwitch += value; } remove { parser.ThreadCSwitch -= value; } }
+
+        public WindowsKernelParser(ETWTraceEventSource source)
+        {
+            parser = new KernelTraceEventParser(source);
+        }
+    }
+
```
cincuranet, line 80: nit: Formatting.

src/tools/ScenarioMeasurement/MemoryConsumption/AndroidMemoryParser.cs
```diff
@@ -0,0 +1,84 @@
...
+using Reporting;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace ScenarioMeasurement
+{
+
```
cincuranet, line 11: nit: Formatting.

cincuranet (changes requested):

src/tools/ScenarioMeasurement/MemoryConsumption/IParser.cs
LoopedBard3: This is a remnant from adding device startup parsing that appears to not have been kept, removed here.

src/scenarios/shared/memoryconsumption.py
```diff
@@ -0,0 +1,165 @@
...
+    def parsetraces(self, traits: TestTraits):
+        directory = TRACEDIR
+        if traits.tracefolder:
+            directory = TRACEDIR + '/' + traits.tracefolder
+            getLogger().info("Parse Directory: " + directory)
+
+        memoryconsumption_args = [
+            self.memoryconsumptionpath,
+            '--app-exe', traits.apptorun,
+            '--parse-only',
```
DrewScoggins, line 68: Should this be controlled by a test trait? In the future if we end up running memory tests that run locally we would not want to default to `--parse-only`

src/scenarios/shared/runner.py
```diff
@@ -319,6 +340,87 @@ def run(self):
...
+                ]
+
+                clearLogsCmd = [
+                    androidHelper.adbpath,
+                    'logcat',
+                    '-c'
+                ]
+
+                allResults = []
+                for i in range(self.testiterations):
```
DrewScoggins, line 380: I know that this is how we currently do this for startup, but why can't we move the execution of the remote commands into the memory consumption tool, and possibly also bring that back for the startup side. I honestly can't remember if we took this approach last time out of convenance or if there was a technical reason why we had to use that approach.

src/tools/ScenarioMeasurement/MemoryConsumption/AndroidMemoryParser.cs
```diff
@@ -0,0 +1,84 @@
...
+            var pssMaxs = new List<double>();
+            var ussMins = new List<double>();
+            var ussAvgs = new List<double>();
+            var ussMaxs = new List<double>();
+            var rssMins = new List<double>();
+            var rssAvgs = new List<double>();
+            var rssMaxs = new List<double>();
+            var captureCounts = new List<double>();
+            // 52MB-52MB-52MB/45MB-45MB-45MB/138MB-138MB-138MB over 1
+            // PSS: min {pss[0]}, avg {pss[1]}, max {pss[2]}; USS: min {uss[0]}, avg {uss[1]}, max {uss[2]}; RSS: min {rss[0]}, avg {rss[1]}, max {rss[2]}; Number: {CaptureNumber}
```
DrewScoggins, line 41: If the format of the output from the Android devices changes to we have to update both the regex here as well as the one in `runner.py`? If possible I would like to make that not true so we can limit the complexity of changes here in the future.

DrewScoggins (commented):
I left some comments around design, specifically where we are doing the execution and parsing of the data from the devices. We should figure out if the approach that we took for startup still makes sense.

src/scenarios/shared/runner.py
LoopedBard3: After thinking about this, I think at least part of the reason for having the remote commands in the python instead of the tool is that the tools are currently setup to run processes rather than execute a set of steps (hopefully that makes sense). I am not set in having the execution stuff live in python over the tool, I think it is just a matter of figuring out if we want to support steps outside of executing an exe to live in the tools.

LoopedBard3: One potential solution to this would be to move the runner.py section for the startup and device memory to a different script that we call in these specific cases.

LoopedBard3:
Converted back to draft while waiting for the Maui fix https://github.com/dotnet/performance/pull/2884 to go into main for a rebase.

fanyang-mono (approved):
I mainly looked at the concept of this work, not the implementation details. Thanks for adding this benchmark.

LoopedBard3:
Here is a test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2131093&view=logs&j=d610060c-e82a-5178-ebd9-7afbbedc05c2

PR #2879, opened 2023-01-31
Processing also mstat (for NativeAOT) file as part of SOD.

cincuranet:


src/tools/ScenarioMeasurement/SizeOnDisk/MStatProcessor.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        public sealed record Result(string Name, int Size);
+
+        public List<Result> AssemblyStats { get; private set; }
+        public List<Result> BlobStats { get; private set; }
+
+        public void Process(string path)
+        {
+            var asm = AssemblyDefinition.ReadAssembly(path);
+            var globalType = (TypeDefinition)asm.MainModule.LookupToken(0x02000001);
```
LoopedBard3, line 18: Is this Lookup value universal or specific to the Native AOT test we have? Looks good if it can be used universally.

LoopedBard3 (approved):
LGTM with comment in mind.

src/tools/ScenarioMeasurement/SizeOnDisk/MStatProcessor.cs
MichalStrehovsky: It is universal - MSTAT places the methods with data into what ECMA-335 considers "global methods". Global methods are owned by a special type - the global type (C# compiler generates it as type `<Module>`). This type must be the first type in the module (`02` in this magic number means type definition, `000001` means first entry). This is standardized by the file format of both .NET and MSTAT and will never change (unless we take a breaking change to the file format, last of which happened when generics were introduced to .NET).

PR #2881, opened 2023-02-01
improved stability of MemoryCache benchmarks

adamsitnik:
Don't use strings for MemoryCache keys, as they produce different hash codes for every benchmark run.
This can lead to different internal structure of ConcurrentDictionary used by MemoryCache, which leads to different amount of allocated memory and randomness from GC point of view

I've verified that the benchmark reports the same amount of allocated memory now.


mrsharm (approved):
LGTM! Thanks for fixing this @adamsitnik!

PR #2884, opened 2023-02-06
Update maui net8.0 android build

LoopedBard3:
This fixes issues with the Maui building in net8.0 and disables dotnet podcast building for deprecation issues. The primary impact of note is that the Maui APKs will no longer be built targeting just android-arm64 making them larger and increasing the expected size on disk of the APK by roughly 2x (14.8ish MB to 30.1MB).
@jonathanpeppers Just a heads up to make sure this SOD impact is understood.

### DIFF
--- a/src/scenarios/mauiandroid/pre.py
+++ b/src/scenarios/mauiandroid/pre.py
@@ -23,7 +23,7 @@ precommands.new(template='maui',
                 no_restore=False)
 
 # Build the APK
-precommands.execute([])
+precommands.execute(['--source', 'MauiNuGet.config'])
 
 # Remove the aab files as we don't need them, this saves space
 output_dir = const.PUBDIR
--- a/src/scenarios/mauiandroid/pre.py
+++ b/src/scenarios/mauiandroid/pre.py
@@ -23,7 +23,7 @@ precommands.new(template='maui',
                 no_restore=False)
 
 # Build the APK
-precommands.execute(['--no-restore', '--source', 'MauiNuGet.config'])
+precommands.execute([])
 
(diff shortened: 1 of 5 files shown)


LoopedBard3:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2106195&view=results Here is a successful test run.

src/scenarios/mauiandroid/pre.py
```diff
@@ -23,7 +23,7 @@
                 no_restore=False)
 
 # Build the APK
-precommands.execute(['--no-restore', '--source', 'MauiNuGet.config'])
+precommands.execute([])
```
jonathanpeppers, line 26: Does the build get the `NuGet.config` another way now?

You probably would need at least the dotnet8 feed:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
  </packageSources>
</configuration>
```

LoopedBard3: It does not seem to, but I will add back in what we had before to cover our bases.

LoopedBard3:
Successful build with updated Nuget.config file setup: https://dev.azure.com/dnceng/internal/_build/results?buildId=2107041&view=logs&j=d610060c-e82a-5178-ebd9-7afbbedc05c2

cincuranet (approved):

PR #2896, opened 2023-02-09
Add C# record and polymorphism benchmarks to STJ.

eiriktsarpalis:
Seems like we weren't benchmarking constructor deserialization or polymorphism. Adding these now.


src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -15,7 +15,7 @@
     <DebugSymbols>true</DebugSymbols>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <EnableXlfLocalization>false</EnableXlfLocalization>    
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>10.0</LangVersion>
```
eiriktsarpalis, line 18: FYI this is needed to support record types.

krwq (approved):

PR #2905, opened 2023-02-16
Add FrozenDictionary to collections tests

stephentoub:


stephentoub:
Is there something I need to do to get the build to use preview nuget packages?  These APIs are in System.Collections.Immutable.dll, which ships downlevel, but it appears to be using officially supported packages and thus can't find the APIs downlevel.

cincuranet:
Where is the preview package? Could the feed be added to this https://github.com/dotnet/performance/blob/main/NuGet.config?

stephentoub:
If nothing else, at this point the .NET 8 Preview 1 package is available on nuget.
https://www.nuget.org/packages/System.Collections.Immutable/8.0.0-preview.1.23110.8

cincuranet:
I'll check why `net462` and `nativeaot7.0` are complaining. Rest is green as grass.

cincuranet:
Hmm, for the `nativeaot7.0` `ilc` takes from it's own `packages\runtime.win-x64.microsoft.dotnet.ilcompiler\7.0.3\framework\System.Collections.Immutable.dll`. Not sure how to force it to use the one from NuGet. Any ideas?

cc @MichalStrehovsky

MichalStrehovsky:
> Not sure how to force it to use the one from NuGet. Any ideas?

I don't know of a way short of hacking MSBuild targets. We track this in https://github.com/dotnet/runtime/issues/67080.

Can this be limited to testing with 8.0 NativeAOT?

cincuranet:
> Can this be limited to testing with 8.0 NativeAOT?

Checking...

cincuranet:
Finally all the stars aligned and it's green (Blazor failure is unrelated). Merging.

stephentoub:
Thank you!!

PR #2919, opened 2023-02-28
Switch nativeaot7.0 microbenchmarks to 8.0

cincuranet:
This should eventually unblock #2905.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -160,7 +160,7 @@ jobs:
         csproj: src\benchmarks\micro\MicroBenchmarks.csproj
         runCategories: 'runtime libraries' 
         channels:
-          - nativeaot7.0
+          - nativeaot8.0
 
   # ML.NET benchmarks
   - template: /eng/performance/build_machine_matrix.yml
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -87,6 +87,8 @@ class FrameworkAction(Action):
             return 'net6.0'
         if framework == 'nativeaot7.0':
             return 'net7.0'
+        if framework == 'nativeaot8.0':
+            return 'net8.0'
         else:
             return framework
 
--- a/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
@@ -38,6 +38,11 @@ namespace BenchmarkDotNet.Extensions
                     .DontEnforcePowerPlan(); // make sure BDN does not try to enforce High Performance power plan on Windows
             }
 
+            job = job.WithArguments(new[]
+            {
+                new MsBuildArgument("/p:IlcVerboseLogging=true"),
+            });
+
             var config = ManualConfig.CreateEmpty()
                 .WithBuildTimeout(TimeSpan.FromMinutes(15)) // for slow machines
                 .AddLogger(ConsoleLogger.Default) // log output to console


LoopedBard3 (approved):
LGTM, given tests pass.

cincuranet:
@MichalStrehovsky can you advise on the errors in NativeAOT micro benchmarks on `net8.0`? The logs are [here](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-8ff50ead78f048559a/Partition0/1/console.1ba70da6.log?helixlogtype=result), [here](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-8ff50ead78f048559a/Partition3/1/console.90baa7ff.log?helixlogtype=result) and [here](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-8ff50ead78f048559a/Partition4/1/console.cfd82b00.log?helixlogtype=result).

MichalStrehovsky:
> @MichalStrehovsky can you advise on the errors in NativeAOT micro benchmarks on net8.0? The logs are [here](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-8ff50ead78f048559a/Partition0/1/console.1ba70da6.log?helixlogtype=result), [here](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-8ff50ead78f048559a/Partition3/1/console.90baa7ff.log?helixlogtype=result) and [here](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-8ff50ead78f048559a/Partition4/1/console.cfd82b00.log?helixlogtype=result).

These are all:

```
The command "ilc" exited with code 137.
```

Exit code 137 is very likely the [OOM killer](https://www.baeldung.com/linux/memory-overcommitment-oom-killer). How much memory do the containers running this have?

What is the effective change this PR is doing? Is this switching from building the tests with .NET 7 to .NET 8 daily builds?

Is there a quick way for me to run this locally? We should be using less memory, not more, but it's possible there's something pathological in the new framework libraries, for example.

cincuranet:
> How much memory do the containers running this have?

These are Azure VMs. _Standard_D2a_v4_, aka 2 CPU, 8 GB RAM.

> Is this switching from building the tests with .NET 7 to .NET 8 daily builds?

Correct.

> Is there a quick way for me to run this locally? We should be using less memory, not more, but it's possible there's something pathological in the new framework libraries, for example.

`python3 scripts/benchmarks_ci.py --csproj src/benchmarks/micro/MicroBenchmarks.csproj --incremental no --architecture x64 -f nativeaot8.0 --dotnet-versions 8.0.100-preview.3.23128.7 --cli-source-info args --cli-branch 8.0 --cli-commit-sha 4c9bc7448c243a8954f1a70a6a8961e6187d5dd3 --cli-repository https://github.com/dotnet/core-sdk --cli-source-timestamp 2023-02-28T18:40:08Z '--bdn-arguments=--anyCategories runtime libraries --iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart --stopOnFirstError true --partition-count 5 --partition-index 0'`

MichalStrehovsky:
I was hoping for a way to run the failing build command by itself.

Do I understand it correctly that this is AOT-compiling all the tests in the repo into a single assembly?

Is there a way to add extra things into the generated project that is failing the build?

I'd like to add:

```xml
<ItemGoup>
<IlcArg Include="--verbose" />
</ItemGroup>
```

With that we'd at least know when this is getting killed.

Pretty sure this is a OOM situation. I don't know how to repro locally,

cc @dotnet/ilc-contrib

cincuranet:
You can see the commands BDN uses starting around `2023/02/28 20:31:22` (in [this log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-8ff50ead78f048559a/Partition0/1/console.1ba70da6.log?helixlogtype=result)).

Not sure about the `IlcArg` item. Probably not easily. Regular property would work fine, because that can be easily passed from command line.

MichalStrehovsky:
> You can see the commands BDN uses starting around `2023/02/28 20:31:22` (in [this log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-8ff50ead78f048559a/Partition0/1/console.1ba70da6.log?helixlogtype=result)).

That command doesn't help because it's for a generated project with a generated C# in it. I don't know what's in it.

> Not sure about the `IlcArg` item. Probably not easily. Regular property would work fine, because that can be easily passed from command line.

@adamsitnik do you know if there's a way to inject extra things into the csproj that BDN generates? It looks like the project build is getting OOM-killed. I'd like to at least know when this is happening. It could be narrowed down with the above (https://github.com/dotnet/performance/pull/2919#issuecomment-1453172651) snippet but I don't see a way to inject it into the generated CSPROJ. I have a suspicion where this could be happening (object writing) and would love to be able to confirm because it affects product direction - we had seen OOM during object writing in the past.

Right now I think we'll need to run this on a machine with more RAM. The project seems to be rather large.

adamsitnik:
No, but it's possible to pass any MsBuild property as an argument to `dotnet restore/build/publish` executed by the BDN.

```cs
job = job.WithArguments(new Argument[]
{
    new MsBuildArgument("/p:DebugType=portable") // See https://github.com/dotnet/roslyn/issues/42393
});
```

It can be done here:

https://github.com/dotnet/performance/blob/77261e41b0586cfea9e26dbd98caad76c12640bc/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L33-L34

MichalStrehovsky:
Thanks! Opened https://github.com/dotnet/runtime/pull/83024 to expose this as a property. I'm not aware of a way to pass Items like this.

cincuranet:
I see the `IlcVerboseLogging` merged so I've added it in 747f36d2343a5560d9c14a61d9194b4153c34ecb. I'll re-trigger this build later when the change in runtime is in _daily_.

MichalStrehovsky:
> /azp run

Looks like this is still 8.0.0-preview.2.23127.4 (i.e. February 27th). The runtime -> SDK -> installer flow has been pretty slow lately.

cincuranet:
Yeah. All that's left to do is wait.

MichalStrehovsky:
It does look like this is running out of memory during [object writing](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2919-merge-3fe566618a2c41cba1/Partition0/1/console.49120fe4.log?helixlogtype=result). The log indicates we sent all the data to the LLVM side and the process gets OOM-killed in somewhere around `EmitDebugModuleInfo` or `FinishObjWriter`. Cc @markples 

```
[2023/03/08 22:50:05][INFO]   ILC: Writing 1201695 object nodes...
[2023/03/08 22:50:05][INFO]   ILC: 10%...
[2023/03/08 22:50:05][INFO]   ILC: 20%...
[2023/03/08 22:50:05][INFO]   ILC: 30%...
[2023/03/08 22:50:05][INFO]   ILC: 40%...
[2023/03/08 22:50:05][INFO]   ILC: 50%...
[2023/03/08 22:50:05][INFO]   ILC: 60%...
[2023/03/08 22:50:05][INFO]   ILC: 70%...
[2023/03/08 22:50:05][INFO]   ILC: 80%...
[2023/03/08 22:50:05][INFO]   ILC: 90%...
[2023/03/08 22:50:05][INFO]   ILC: 100%...
[2023/03/08 22:50:05][INFO]   ILC: Finalizing output to '/home/helixbot/work/ADBD09A3/w/9F6408C1/e/artifacts/obj/BenchmarkDotNet.Autogenerated/Release/net8.0/linux-x64/native/d0166487-eca5-4936-bea6-510815f67765.o'...
[2023/03/08 22:50:05][INFO]   /home/helixbot/work/ADBD09A3/t/MSBuildTemphelixbot/tmp842cf9d03fec46c4bab7d621f32b195f.exec.cmd: line 2:  9592 Killed                  "/home/helixbot/work/ADBD09A3/w/9F6408C1/e/artifacts/packages/runtime.linux-x64.microsoft.dotnet.ilcompiler/8.0.0-preview.3.23158.3/tools/ilc" @"/home/helixbot/work/ADBD09A3/w/9F6408C1/e/artifacts/obj/BenchmarkDotNet.Autogenerated/Release/net8.0/linux-x64/native/d0166487-eca5-4936-bea6-510815f67765.ilc.rsp"
```

MichalStrehovsky:
Merged #83181 that might help. We'll need to wait until that propagates to the sdk+installer repo.

MichalStrehovsky:
Merged https://github.com/dotnet/runtime/pull/83249 that might help.

agocke:
Some interesting info here: I managed to get a repro that runs on both Linux and Windows, and on Linux it consumes a lot more memory.

agocke:
OK, it looks like the bulk of the allocations are in the object writer, as Michal suspected. I have some basic stack trace information, and one thing that jumps out are these lines:

<html xmlns="http://www.w3.org/1999/xhtml">
<body>
Line # | Instance | Process | Stack | Stack ID | Ref Count | Size (B) | Count
-- | -- | -- | -- | -- | -- | -- | --
33 |   |   |      objwriter.dll!DwarfMemberFunctionIdTypeInfo::DumpStrings |   |   | 124,780,543 | 4
34 |   |   |   objwriter.dll!llvm::SmallVectorImpl<char>::append<char const *,void> |   |   | 112,197,631 | 1

</body></html>

That looks to be this code here:

```C#
void DwarfMemberFunctionIdTypeInfo::DumpStrings(MCObjectStreamer *Streamer) {
  Streamer->emitBytes(StringRef(Name));
  Streamer->emitIntValue(0, 1);

  MCContext &context = Streamer->getContext();
  LinkageNameSymbol = context.createTempSymbol();
  Streamer->emitLabel(LinkageNameSymbol);
  Streamer->emitBytes(StringRef(LinkageName));
  Streamer->emitIntValue(0, 1);
}
```

My understanding of this code is that we're pulling the current section and emitting the function name and its mangled name into it.

But it's notable that I think pretty much all the function names are in the same section. So we're basically pushing 120 MB of allocations through LLVM's "SmallVectorImpl". I don't know how that's implemented, but it raises a red flag for me that we might not be using this API correctly.

agocke:
>  don't know how that's implemented, but it raises a red flag for me that we might not be using this API correctly.

Hmm, the more I dig, the more I think the API we're not using correctly is LLVM itself, namely the assembly-level API. I'm not sure this API was supposed to be used for the scale we're throwing at it.

agocke:
Ah, you know what, I no longer think 8 GB of memory is unreasonable for this compilation.

The output image is **448 MB**.

> $ du -hs /home/andy/code/performance/artifacts/obj/BenchmarkDotNet.Autogenerated/Release/net8.0/linux-x64/native/764ca92
b-1570-40ea-89f5-a9f67194f16a.o
> 448M    /home/andy/code/performance/artifacts/obj/BenchmarkDotNet.Autogenerated/Release/net8.0/linux-x64/native/764ca92b-1570-40ea-89f5-a9f67194f16a.o

I need to look at the inputs here before trying to optimize ILC.

MichalStrehovsky:
Agreed, 448 MB object file is a pretty large application. After linking/stripping, it will likely be more in the 100-200 MB range, but it's still pretty large. IIRC Benchmark.NET forces partial trimming because it's not compatible with full trimming. It likely contributes to the problem. We had problems with this in the past (https://github.com/dotnet/BenchmarkDotNet/pull/2046).

That said, I retriggered the CI leg after https://github.com/dotnet/runtime/pull/83447 and looks like now it succeeded. The Blazor leg is still failing, but looks unrelated.

🍾

agocke:
It looks like the microbenchmarks are already getting trimmed with TrimMode=full. So it's probably just that the microbenchmarks are pulling in basically the whole framework.

No easy fix for that. The test machine might just eventually need > 8 GB of RAM. I don't think it's unreasonable to take 16 GB of RAM to compile a 500 MB exe.

jkotas:
> split these tests into two assemblies.

Splitting the microbenchmarks into multiple assemblies is the only option that will work long term. One huge binary with all microbenchmarks will keep running into limits. It is not possible to have >2GB binary on Windows due to file format and OS limitations. Once we hit the 2GB limit, no amount of extra RAM on the build machine is going to help.

adamsitnik:
> Splitting the microbenchmarks into multiple assemblies is the only option that will work long term. 

In a sense, we are already doing that. The CI by default uses `PartitionCount = 5`, which means that it splits all the microbenchmarks and it runs them on 5 machines, each of them is building a project that contains 1/5 of all the benchmarks.

https://github.com/dotnet/performance/blob/7eddd9da5bdfbba7809e91846bbb6e6d851f19cb/eng/performance/helix.proj#L25-L34

This is configurable via command line:

```cmd
--partition-count $PartitionCount --partition-index $PartitionCount
```

PR #2921, opened 2023-03-06
Enhancements to GCPerfSim

PeterSolMS:
Enhancements to GCPerfSim:
- Simulate server "requests" where we hold on to a collection of objects for an intermediate amount of time.
- Enable varying the number of threads between phases
- Enable varying the amount of computation vs. allocation between phases
- Add pre-baked object size distribution to become more realistic in this regard

Details:
- Option "-requestAllocMB" or "-ramb" determines the amount of memory allocated per request in MB.
- Option "-requestLiveMB" or "-rlmb" determines the live memory size per request
- Option "-sizeDistribution" or "sizeDist" determines the object size distribution used - only "1" is supported now
- Slightly relaxed "AboutEquals" method.
- Moved "compute" field from PerThreadArgs to Phase to enable varying between phases.
- Support comments at the start of an input text file.


src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -1035,7 +1187,7 @@ public CharSpan TakeWord()
     private static bool IsDigit(char c) =>
         '0' <= c && c <= '9';
     private static bool IsDigitOrDot(char c) =>
```
Maoni0, line 1189: nit - since you added a comma, perhaps call it `IsDigitOrDecimalSeparator`?

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -769,29 +772,180 @@ public override string ToString()
...
+        BuildBucketSpecsFromSizeDistribution(bucketSpecs, sizeSlots, sizeRange, survInterval, reqSurvInterval, pinInterval, finalizableInterval, true);
+    }
+    public static void BuildPOHBucketSpecsFromSizeDistribution(List<BucketSpec> bucketSpecs, uint survInterval, uint reqSurvInterval, uint pinInterval, uint finalizableInterval)
+    {
+        BuildBucketSpecsFromSizeDistribution(bucketSpecs, pohSizeSlots, new SizeRange(24, 10_000_000), survInterval, reqSurvInterval, pinInterval, finalizableInterval, true);
+    }
+}
+
+
 readonly struct Phase
```
Maoni0, line 915: it'd be great to have a sample phase txt checked in

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -769,29 +772,180 @@ public override string ToString()
...
+
+    // an object size distribution for SOH/LOH derived from a real server scenario
+    public static SizeSlot[] sohLohSizeSlots = new SizeSlot[]
+    {
+//        new SizeSlot(        24,  16229766), We are creating some small wrapper objects of 32 and 40 bytes around the
+//        new SizeSlot(        32,  48790439), payload objects - somewhat compensate for this by distorting the distribution
+//        new SizeSlot(        40,  21496250), at the low end
+//        new SizeSlot(        48,  15020039),
+//
+        new SizeSlot(        56,  12134890),
```
mrsharm, line 793: NIT but maybe for later having the ability to read these distributions from a text file might be helpful in case we want to allow the users to specify these.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -1164,6 +1317,10 @@ private static Args ParseFromFile(string fileName)
             {
                 verifyLiveSize = true;
             }
+            else if (word == "compute")
```
mrsharm, line 1320: NIT: maybe doing a case insensitive ordinal comparison here might be both performant and be less error prone in case the users provide the word in a different format: 

```csharp
else if (string.Equals(word, "compute", StringComparison.OrdinalIgnoreCase))
```

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -1405,26 +1586,33 @@ private static Args ParseFromCommandLine(string[] args)
...
+        uint sizeDist = 0;
+
         for (uint i = 0; i < args.Length; ++i)
         {
             switch (args[i])
             {
                 case "-compute":
                 case "-c":
-                    compute = true;
+                    compute = 1000;
```
mrsharm, line 1615: Might be worth adding a line or two about why this is value is hardcoded and the significance of the value?

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
PeterSolMS: Good point - thanks!

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -1462,6 +1650,14 @@ private static Args ParseFromCommandLine(string[] args)
                 case "-tagb":
                     totalAllocBytes = Util.GBToBytes(ParseDouble(args[++i]));
                     break;
+                case "-requestAllocMB":
```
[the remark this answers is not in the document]
PeterSolMS: Will do.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
PeterSolMS: In a sense we already have that capability, because the text file given with "-file" can contain definitions of size buckets. But for many buckets, it's inconvenient because you'd have to specify survivorships and other properties for each bucket. Let's think about which capabilities we really want here...

mrsharm (approved):
LGTM!

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -1377,11 +1554,15 @@ private static Args ParseFromFile(string fileName)
...
     private const uint DEFAULT_POH_FINALIZABLE_INTERVAL = 0;
     private const uint DEFAULT_POH_SURV_INTERVAL = 0;
+    private const uint DEFAULT_REQ_POH_SURV_INTERVAL = 0;
 #endif
 
     private const uint DEFAULT_PINNING_INTERVAL = 100;
     private const uint DEFAULT_FINALIZABLE_INTERVAL = 0;
 
+    private const uint DEFAULT_REQ_SOH_SURV_INTERVAL = 3;
+    private const uint DEFAULT_REQ_LOH_SURV_INTERVAL = 2;
```
Maoni0, line 1564: this seems really tiny for LOH.

PeterSolMS: Oh, you mean fewer objects in LOH should survive till the end of the request?

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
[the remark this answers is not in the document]
PeterSolMS: Done...

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
PeterSolMS: Agreed - done.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
PeterSolMS: Well, we aren't doing this in other places - even the command line option parsing is case sensitive, despite a comment stating the opposite.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
PeterSolMS: Done.

Maoni0:
I looked at a couple of errors and they are very difficult to decipher so I can't tell what exactly are the errors but they seem to be related to running microbenchmarks so they don't seem to have anything to do with Peter's changes, eg, [this one](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-performance-refs-pull-2921-merge-fbefe70b286943cc84/Partition4/1/console.113e0520.log?helixlogtype=result). 

@DrewScoggins, could you please take a look and see if you agree?

cincuranet:
@Maoni0 these failures are indeed unrelated.

This can be merged @PeterSolMS.

Maoni0:
@cincuranet thanks for taking a look! however, it looks like I cannot merge because it seems to require that all test must pass - 

"Required statuses must pass before merging"

is there a way to bypass that?

cincuranet:
Only, probably, owners of this repo can bypass the checks. 

Merged for you.

Maoni0:
thank you @cincuranet! I did have permission to merge before this enforcement was in place. so looks like I'm not given permission to bypass the checks 😅

PR #2923, opened 2023-03-07
Update markdownlint.yml versions

LoopedBard3:
Update markdownlint.yml actions/checkout@v2->v3 and setup-node@v1->v3 with node version 16. Per: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


cincuranet (approved):

LoopedBard3:
This change didn't change any of the testing, merging.

Issue #2931, opened 2023-03-17
Windows Micro Runs failing


LoopedBard3:
Windows micro runs are failing. One found error is the following:
```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
[2023/03/17 09:39:58][INFO]  ---> System.PlatformNotSupportedException: System.Drawing.Common is not supported on this platform.
[2023/03/17 09:39:58][INFO]    at System.Drawing.Image..ctor()
[2023/03/17 09:39:58][INFO]    at System.Drawing.Bitmap..ctor(Int32 width, Int32 height)
[2023/03/17 09:39:58][INFO]    at System.Drawing.Tests.Perf_Graphics_DrawBeziers.Setup() in C:\dotnetbuild\work\B0A8098B\w\B11609C6\e\performance\src\benchmarks\micro\libraries\System.Drawing\Perf_Graphics_DrawBeziers.cs:line 26
[2023/03/17 09:39:58][INFO]    at BenchmarkDotNet.Engines.EngineFactory.CreateReadyToRun(EngineParameters engineParameters)
[2023/03/17 09:39:58][INFO]    at BenchmarkDotNet.Autogenerated.Runnable_50.Run(IHost host, String benchmarkName) in C:\dotnetbuild\work\B0A8098B\w\B11609C6\e\performance\artifacts\bin\MicroBenchmarks\Release\net8.0\76fcf152-5065-49fa-8c49-03d03341cc3e\76fcf152-5065-49fa-8c49-03d03341cc3e.notcs:line 10472
[2023/03/17 09:39:58][INFO]    at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
[2023/03/17 09:39:58][INFO]    at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
[2023/03/17 09:39:58][INFO]    --- End of inner exception stack trace ---
[2023/03/17 09:39:58][INFO]    at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
[2023/03/17 09:39:58][INFO]    at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
[...]

PR #2936, opened 2023-03-21
Removed failing System.Drawing tests from net8.0+ runs.

LoopedBard3:
Fixes: https://github.com/dotnet/performance/issues/2931. 

System.Drawing tests recently started failing because the runtime library code was moved out of the runtime: https://github.com/dotnet/runtime/pull/83356. This PR removes the tests that broke from this change from net8.0+ runs.

Here is a net8.0 test run showing it fixed the windows runs: https://dev.azure.com/dnceng/internal/_build/results?buildId=2139959&view=results


cincuranet (approved):

LoopedBard3:
Failing run is failing on main and due to unrelated changes, merging.

PR #2946, opened 2023-03-29
Update CrankConfiguration.yaml

mrsharm:
Updated the tc and tagb to some acceptable numbers based on the number of cores in the ASPNet machines.


danmoseley:
@mrsharm is this still active? there are merge conflicts

mrsharm:
@danmoseley - no, closing this out since we figured out a cleaner way of getting the number of cores of the machine the scenario is run on. Thanks!

PR #2950, opened 2023-03-31
Half conversion benchmarks

MineCake147E:
Comparing performance difference of https://github.com/dotnet/runtime/pull/81632/.


MineCake147E:
@dotnet-policy-service agree

MineCake147E:
@dotnet-policy-service agree

MineCake147E:
@dotnet-policy-service agree

src/benchmarks/micro/libraries/System/ConvertHalfToSingle.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+
+using BenchmarkDotNet.Attributes;
+
+using MicroBenchmarks;
+
+namespace System
```
tannergooding, line 14: nit: These should instead go under the https://github.com/dotnet/performance/tree/main/src/benchmarks/micro/libraries/System.Runtime folder and follow the same general conventions as `Perf.Double.cs` and `Perf.Single.cs`

This includes adding the required license header, matching the general namespace, and matching the general other conventions used for passing in arguments/etc

We would likewise ideally not write our own loop here but rather let BDN handle doing the looping itself unless absolutely required. -- I don't believe its required here due to the innate overhead in doing `Half<->Single` conversions

MineCake147E: Thank you for your advice about folder hierarchy.
However, input permutation is absolutely necessary.
If we don't permute the input at all, the previous method can greatly take advantage of branch prediction.

tannergooding: Benchmark.NET is design to appropriately measure things here and technology such as DPGO will itself optimize based on branches.

If there are multiple interesting paths then we should simply ensure the most interesting ones are covered by the benchmark as separate runs.

src/benchmarks/micro/libraries/System.Runtime/Perf.Half.cs
```diff
@@ -0,0 +1,149 @@
...
+            Span<float> bA = bufferSourceSingle.AsSpan();
+            Span<Half> bD = bufferDestinationHalf.AsSpan();
+            ref float rsi = ref MemoryMarshal.GetReference(bA);
+            ref Half rdi = ref MemoryMarshal.GetReference(bD);
+            nint i = 0, length = Math.Min(bA.Length, bD.Length);
+            for (; i < length; i++)
+            {
+                Unsafe.Add(ref rdi, i) = (Half)Unsafe.Add(ref rsi, i);
+            }
+        }
```
tannergooding, line 77: We shouldn't be looping or doing other complex logic here unless absolutely necessary. If customization is absolutely required then extensibility points such as `OperationsPerInvoke`, `IterationSetup`, and `IterationCleanup` should be preferred over more logic in the actual benchmarked method. -- That being said, this shouldn't be one of those cases, we really only hit such cases for intrinsics where we're testing a single instruction that takes 1-3 cycles to execute.

Doing "more" here means we aren't measuring the cost of `float->half` conversions, but rather we're measure the cost of `Math.Min`, `AsSpan`, `GetReference`, `looping`, and `float->half`.

We really instead want something like:
```csharp
[ParamsSource(nameof(HalfValues))]
public Half HalfValue;

public IEnumerable<Half> HalfValues => new Half[] { ... };

[Benchmark]
public float HalfToSingle(Half value) => (float)value;

[Params(...)]
public float SingleValue;

[Benchmark]
public Half SingleToHalf(float value) => (Half)value;
```

There are also options like `[Arguments]` and `[ArgumentsSource]` that can work as well: https://benchmarkdotnet.org/articles/features/parameterization.html

tannergooding: CC. @adamsitnik

MineCake147E: The purpose of randomizing the order of the inputs was to reveal the dependence of performance on the order of the inputs.
In practical use, the same value is almost never given as an argument over and over again in succession, as in benchmarking. In most cases, a wide variety of data is given as arguments.
In such situations, branch predictions that rely on the consistency of input values should not be relied upon, and algorithms that use as few branches as possible should be favored in both reality and the benchmark.

MineCake147E: I created an Issue https://github.com/dotnet/BenchmarkDotNet/issues/2314 to replace loops by it.

src/benchmarks/micro/libraries/System.Runtime/Perf.Half.cs
```diff
@@ -0,0 +1,43 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+using BenchmarkDotNet.Attributes;
+
```
adamsitnik, line 12: nit: empty lines

```suggestion
using BenchmarkDotNet.Attributes;
```

src/benchmarks/micro/libraries/System.Runtime/Perf.Half.cs
```diff
@@ -0,0 +1,43 @@
...
+            BitConverter.UInt16BitsToHalf(0x03ff),  //Maximum subnormal number in Half
+            (Half)12345.0f /* same value used by other tests to compare the perf */,
+            BitConverter.UInt16BitsToHalf(0x7dff)   //NaN
+        };
+
+        public static IEnumerable<float> SingleValues => new float[]
+        {
+            6.097555E-05f,
+            12345.0f /* same value used by other tests to compare the perf */,
+            65520.0f,   //Minimum value that is infinity in Half
```
adamsitnik, line 31: nit: inconsistent comments

```suggestion
            BitConverter.UInt16BitsToHalf(0x03ff), // Maximum subnormal number in Half
            (Half)12345.0f, // same value used by other tests to compare the perf
            BitConverter.UInt16BitsToHalf(0x7dff) // NaN
        };

        public static IEnumerable<float> SingleValues => new float[]
        {
            6.097555E-05f,
            12345.0f, // same value used by other tests to compare the perf
            65520.0f // Minimum value that is infinity in Half
```

adamsitnik (approved):
LGTM, thank you for your contribution @MineCake147E!

src/benchmarks/micro/libraries/System.Runtime/Perf.Half.cs
adamsitnik, line 12:
```suggestion
```

adamsitnik:
For some reason I can not apply the suggestions, since they are just style nits I am going to merge the PR right now so we start collecting numbers before merging https://github.com/dotnet/runtime/pull/81632

PR #2955, opened 2023-04-03
[WIP] Add Android BenchmarkDotnet Scenario

LoopedBard3:
This adds the capability to run Android Instrumentation tests. In this case, the target is BDN testing instrumentation on android, but could be other scenarios in the future. This PR also adds the scenario folder for this instrumentation scenario and makes the PerfLabExporter class public so that other apps can reference it when the performance repo is downloaded. Having it public will allow for these apps to output microbenchmark tests in the format we already use.


src/scenarios/shared/androidinstrumentation.py
```diff
@@ -0,0 +1,148 @@
...
+                pullFilesFromDeviceCmd = [
+                    adbpath,
+                    'pull',
+                    defaultDeviceBdnOutputDir,
+                    TRACEDIR
+                ]
+
+                RunCommand(pullFilesFromDeviceCmd, verbose=True).run()
+
+                # Replace the JSON with the correct values
```
cincuranet, line 89: For my own understanding, why the _perf-lab-report_ does not have correct values? Why is i.e. `repo` or `branch` incorrect?

cincuranet (approved):

src/scenarios/shared/androidinstrumentation.py
LoopedBard3: I was unable to find a way to pass through the correct data to the built app at runtime such that it could be set as an environment variable. The environment variables are then used by the PerfLabExporter in the .json it generates, requiring all the values to have something set.

cincuranet: Thanks.

PR #2956, opened 2023-04-04
Update CrankConfiguration.yaml

mrsharm:
Fixed the thread count value to (# of cores - 2) * 2 and added the main GC PerfSim Scenarios.

### DIFF
--- a/src/benchmarks/gc/scenarios/CrankConfiguration.yaml
+++ b/src/benchmarks/gc/scenarios/CrankConfiguration.yaml
@@ -10,27 +10,57 @@ jobs:
     waitForExit: true
     readyStateText: GCPerfSim
     variables:
-      tc: 20
-      tagb: 40
-      tlgb: 5
+      tc: {{ cores |minus: 2 |times: 3 }}
+      tagb: 150 
+      tlgb: 0
       lohar: 0
       pohar: 0
       sohsr: 100-4000
       lohsr: 102400-204800
       pohsr: 100-204800
-      sohsi: 30
+      sohsi: 0
       lohsi: 0
       pohsi: 0
-      sohpi: 50
+      sohpi: 0
       lohpi: 0
       sohfi: 0
       lohfi: 0
       pohfi: 0
       allocType: reference
       testKind: time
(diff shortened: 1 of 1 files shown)


src/benchmarks/gc/scenarios/CrankConfiguration.yaml
```diff
@@ -28,9 +28,9 @@ jobs:
       pohfi: 0
       allocType: reference
       testKind: time
-    arguments: "-tc {{ tc }} -tagb {{ tagb }} -tlgb {{ tlgb }} -lohar {{ lohar }} -lohsr {{ lohsr }} -sohsi {{ sohsi }} -lohsi {{ lohsi }} -sohpi {{ sohpi }} -lohpi {{ lohpi }} -sohfi {{ sohfi }} -lohfi {{ lohfi }} -allocType {{ allocType }} -testKind {{ testKind }}"
+    arguments: "-tc {{ cores | minus : 2 | times: 2 }} -tagb {{ tagb }} -tlgb {{ tlgb }} -lohar {{ lohar }} -lohsr {{ lohsr }} -sohsi {{ sohsi }} -lohsi {{ lohsi }} -sohpi {{ sohpi }} -lohpi {{ lohpi }} -sohfi {{ sohfi }} -lohfi {{ lohfi }} -allocType {{ allocType }} -testKind {{ testKind }}"
```
sebastienros, line 31: This should really be `{{ tc }}` and you should update the default value of `tc` instead

sebastienros: Otherwise you will never be able to set anything else

sebastienros:
1- Have you tried each scenario with crank?
2- Just for your crank education, creating scenarios in this file is useful when you need to reuse common configurations. If you are just varying one variable between two scenarios for the CI only, like to be able to compare multiple values of the same argument then it's not required to create a custom scenario, it can just be set on the command line of the common scenario. Here I am not saying it's wrong, just explaining when it's best to create custom scenarios.
3- Might be useful to create a README file like in aspnet/benchmarks that describes each scenario, what variables are available, and sample crank command lines to reuse.

sebastienros (approved):

cshung (approved):

PR #2960, opened 2023-04-06
More precise SOD measurement for NativeAOT.

### LINKED ISSUE #2958: NativeAOT SOD includes dbg and mstat as top level counters, but only binary is important

cincuranet:
* Ignore MSTAT file(s) in global measurements.
* Do not generated debug/pdb files.

Fixes #2958.


LoopedBard3 (approved):
Looks good.

PR #2962, opened 2023-04-11
Do not include dbg (and pdb) files (though keep producing them to have vanilla experience) into final NativeAOT measurements.

cincuranet:


LoopedBard3 (approved):
LGTM

PR #2970, opened 2023-04-18
Add retries to result uploading

LoopedBard3:
Put the upload blob and queue send message into a looping try catch that tries to do the action 3 times with 5 seconds between attempts. This is to hopefully help with individual jobs from the pipeline failing due to Server Busy errors (API limit reached). This was a quick change so am open to other suggestions.

### DIFF
--- a/scripts/performance/common.py
+++ b/scripts/performance/common.py
@@ -17,6 +17,7 @@ from platform import machine
 
 import os
 import sys
+import time
 
 
 def get_machine_architecture():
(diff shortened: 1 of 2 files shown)


scripts/upload.py
```diff
@@ -30,14 +31,33 @@ def upload(globpath, container, queue, sas_token_env, storage_account_uri):
...
+                except Exception as ex:
+                    getLogger().error('{0}: {1}'.format(type(ex), str(ex)))
+                    time.sleep(5)
+                    continue
+                
+            if uploadedSuccessfully:
+                for i in range(3):
+                    try:
+                        if queue is not None:
+                            queue_client = QueueClient(account_url=storage_account_uri.format('queue'), queue_name=queue, credential=sas_token, message_encode_policy=TextBase64EncodePolicy())
```
cincuranet, line 50: I guess this could be outside the loop. We don't need to retry constructor.

scripts/upload.py
```diff
@@ -30,14 +31,33 @@ def upload(globpath, container, queue, sas_token_env, storage_account_uri):
...
-            with open(infile, "rb") as data:
-                blob_client.upload_blob(data, blob_type="BlockBlob", content_settings=ContentSettings(content_type="application/json"))
-
-            if queue is not None:
-                queue_client = QueueClient(account_url=storage_account_uri.format('queue'), queue_name=queue, credential=sas_token, message_encode_policy=TextBase64EncodePolicy())
-                queue_client.send_message(blob_client.url)
-
-            getLogger().info("upload complete")
+            uploadedSuccessfully = False
+            for i in range(3):
```
cincuranet, line 35: I would probably create a helper "retry" function that would take lambda and have all the looping and catching encapsulated there.

cincuranet (commented):
Is the delay=5s and count=3 based on some measurement or something? The reason I'm asking is because I'm working on similar stuff in _infra_/_perfcommandcore_ and if these values are based on something "real", I would happily ~steal~use same values. :)

scripts/upload.py
LoopedBard3: Nope, These were arbitrarily chosen.

cincuranet: Fair enough.

scripts/upload.py
LoopedBard3: Did this, give the new changes a look when you get a chance please. 👍

scripts/performance/common.py
```diff
@@ -157,6 +158,37 @@ def push_dir(path: str = None) -> None:
     else:
         yield
 
+def retry_on_exception(function, retry_count=3, retry_delay=5, retry_delay_multiplier=1, retry_on_exception=None):
```
cincuranet, line 161: Would it make sense to have `retry_on_exception=Exception` instead code on L178-179?

scripts/upload.py
cincuranet: Love the `retry_on_exception`!

scripts/performance/common.py
LoopedBard3: Updated! I agree that this seems like a better approach.

cincuranet (approved):
Love the `retry_on_exception`. Great work!

LoopedBard3:
Failures are also occurring on main, merging.

PR #2995, opened 2023-04-28
Make it so that setup_loggers will only setup 1 logger globally

LoopedBard3:
Make it so that setup_loggers will only setup 1 logger in a at most once fashion. This will keep child calls to setup_loggers from creating duplicate loggers and duplicate output lines.

One side effect of this is that multiple setup_loggers calls will only have the first verbosity value set.

Open to feedback on whether we think this is how the functionality should work.


cincuranet (approved):
Nice!

PR #3000, opened 2023-05-02
Added PowerShell real-world benchmarks

TIHan:
This is a snapshot of the PowerShell benchmarks found [here](https://github.com/PowerShell/PowerShell/tree/master/test/perf/benchmarks).

We will update these as the PowerShell team updates their benchmarks.

cc @kunalspathak

### DIFF
--- a/src/benchmarks/real-world/PowerShell.Benchmarks/Categories.cs
+++ b/src/benchmarks/real-world/PowerShell.Benchmarks/Categories.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace MicroBenchmarks
+{
+    public static class Categories
+    {
+        /// <summary>
+        /// Benchmarks belonging to this category are executed for CI jobs.
+        /// </summary>
+        public const string Components = "Components";
+
+        /// <summary>
+        /// Benchmarks belonging to this category are executed for CI jobs.
+        /// </summary>
+        public const string Engine = "Engine";
+
+        /// <summary>
+        /// Benchmarks belonging to this category are targeting internal APIs.
+        /// </summary>
+        public const string Internal = "Internal";
+
+        /// <summary>
+        /// Benchmarks belonging to this category are targeting public APIs.
+        /// </summary>
+        public const string Public = "Public";
+    }
+}
--- a/src/benchmarks/real-world/PowerShell.Benchmarks/Engine.Compiler.cs
+++ b/src/benchmarks/real-world/PowerShell.Benchmarks/Engine.Compiler.cs
@@ -1,0 +1,78 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#if NET6_0
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Management.Automation;
+using System.Management.Automation.Language;
+
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace Engine
+{
+    [BenchmarkCategory(Categories.Engine, Categories.Internal)]
+    public class Compiler
+    {
+        private static readonly Dictionary<string, ScriptBlockAst> s_scriptBlocksDict;
+        private static readonly List<string> s_functionNames;
+        private ScriptBlockAst _currentAst;
+
+        static Compiler()
+        {
+            string pattern = string.Format("{0}test{0}perf{0}benchmarks", Path.DirectorySeparatorChar);
+            string location = typeof(Compiler).Assembly.Location;
+            string testFilePath = null;
+
+            int start = location.IndexOf(pattern, StringComparison.Ordinal);
+            if (start > 0)
+            {
+                testFilePath = Path.Join(location.AsSpan(0, start + pattern.Length), "assets", "compiler.test.ps1");
+            }
+
+            var topScriptBlockAst = Parser.ParseFile(testFilePath, tokens: out _, errors: out _);
+            var allFunctions = topScriptBlockAst.FindAll(ast => ast is FunctionDefinitionAst, searchNestedScriptBlocks: false);
+
+            s_scriptBlocksDict = new Dictionary<string, ScriptBlockAst>(capacity: 16);
+            s_functionNames = new List<string>(capacity: 16);
+
+            foreach (FunctionDefinitionAst function in allFunctions)
+            {
+                s_functionNames.Add(function.Name);
+                s_scriptBlocksDict.Add(function.Name, function.Body);
+            }
+        }
+
+        [ParamsSource(nameof(FunctionName))]
+        public string FunctionsToCompile { get; set; }
+
+        public IEnumerable<string> FunctionName() => s_functionNames;
+
+        [GlobalSetup(Target = nameof(CompileFunction))]
+        public void GlobalSetup()
+        {
+            _currentAst = s_scriptBlocksDict[FunctionsToCompile];
+
+            // Run it once to get the C# code jitted.
+            // The first call to this takes relatively too long, which makes the BDN's heuristic incorrectly
+            // believe that there is no need to run many ops in each iteration. However, the subsequent runs
+            // of this method is much faster than the first run, and this causes 'MinIterationTime' warnings
+            // to our benchmarks and make the benchmark results not reliable.
+            // Calling this method once in 'GlobalSetup' is a workaround. 
+            // See https://github.com/dotnet/BenchmarkDotNet/issues/837#issuecomment-828600157
+            CompileFunction();
+        }
+
+        [Benchmark]
+        public bool CompileFunction()
+        {
+            var compiledData = new CompiledScriptBlockData(_currentAst, isFilter: false);
+            return compiledData.Compile(true);
+        }
+    }
+}
+
+#endif
--- a/src/benchmarks/real-world/PowerShell.Benchmarks/PowerShell.Benchmarks.csproj
+++ b/src/benchmarks/real-world/PowerShell.Benchmarks/PowerShell.Benchmarks.csproj
@@ -1,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>$(PERFLAB_TARGET_FRAMEWORKS)</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">net7.0</TargetFrameworks>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <StartArguments>--filter *</StartArguments>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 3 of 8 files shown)


kunalspathak:
> We will update these as the PowerShell team updates their benchmarks.

We want to take the `startup` benchmark for sure (sometime next week), but after that probably we should update it once in a release so our test history maintains stable.

kunalspathak:
Can you add a small `README.md`?

src/benchmarks/real-world/PowerShell.Benchmarks/README.md
```diff
@@ -0,0 +1,11 @@
+# Benchmarks
+
+This project contains a variety of performance tests for different pieces of the library.
```
kunalspathak, line 3: Can you add the commit link that you picked? Will be easier to follow later.

kunalspathak:
@LoopedBard3 @cincuranet @sblom

kunalspathak:
> @kunalspathak Do you want to setup runs, auto-filing, etc.?

yes. We might want to add "startup" benchmark to the real world dashboard (when we add it).

kunalspathak (dismissed):

src/benchmarks/real-world/PowerShell.Benchmarks/Engine.Compiler.cs
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#if NET6_0
```
LoopedBard3, line 4: Do we want to benchmark these benchmarks on newer dotnet versions than Net6.0?

src/benchmarks/real-world/PowerShell.Benchmarks/Categories.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// <summary>
+        /// Benchmarks belonging to this category are executed for CI jobs.
+        /// </summary>
+        public const string Engine = "Engine";
+
+        /// <summary>
+        /// Benchmarks belonging to this category are targeting internal APIs.
+        /// </summary>
+        public const string Internal = "Internal";
```
LoopedBard3, line 21: Are the Internal and Public categories specifically for denoting the API availability or should they also be extended to the run environment? i.e. only run public publicly and internal internally?

LoopedBard3 (commented):
Generally looks good, just a few questions to answer before merging. Also, there are no runs specifically setup currently. We can add the runs in a separate PR or in this PR, an example for adding the runs are https://github.com/dotnet/performance/blob/main/azure-pipelines.yml#L225-L238 for public runs, and https://github.com/dotnet/performance/blob/main/azure-pipelines.yml#L452-L467 for internal runs. Let me know and I can also help out for getting the runs setup 👍.

src/benchmarks/real-world/PowerShell.Benchmarks/Engine.Compiler.cs
TIHan: Good catch, yes we probably want to run these on versions newer than net6.0.

src/benchmarks/real-world/PowerShell.Benchmarks/Categories.cs
TIHan: @adityapatwardhan - Perhaps you could shed some light on this one. We are copying PowerShell's benchmarks over to dotnet/performance verbatim

src/benchmarks/real-world/PowerShell.Benchmarks/Engine.Compiler.cs
kunalspathak: Just to make sure that I understand. In our lab, we always run against current which is `.net8`. Leaving it as .NET 6 will not execute this benchmark?

LoopedBard3: That is correct as a whole. Benchmarks in the main branch are only run against whatever latest is. Each release we make a branch to correspond to that release and hold the tests to be run against that release while it is in support, but we don't generally backport new tests to these branches.

TIHan: It shouldn't. If the project is compiled to a target other than NET6, then these benchmarks will not be included.

src/benchmarks/real-world/PowerShell.Benchmarks/Categories.cs
LoopedBard3: We should add at least the public run verification to this, I will make a PR against the branch to add the runs. 👍

LoopedBard3: @TIHan made the PR here: https://github.com/TIHan/performance/pull/1

LoopedBard3:
@cincuranet @caaavik-msft any reason not to add these as a new runkind (powershell)?

cincuranet:
This has my full supoport.

src/benchmarks/real-world/PowerShell.Benchmarks/Categories.cs
TIHan: @LoopedBard3 Merged the change.

LoopedBard3: Looks like I made a typo in the PR in the azure yaml. "src/benchmarks/real-world/Powershell.Benchmarks/PowerShell.Benchmarks.csproj" needs to have the first Powershell changed to PowerShell. I can make a PR for this, but it may be a quicker change for you to push. There should be 2 locations in the azure-pipelines.yml file.

TIHan: Ok, I fixed it.

LoopedBard3:
Looks like Ubuntu is still hitting an error:
```
Assembly PowerShell.Benchmarks which defines benchmarks references non-optimized System.Management.Automation
If you own this dependency, please, build it in RELEASE.
If you don't, you can disable this policy by using 'config.WithOptions(ConfigOptions.DisableOptimizationsValidator)'.
```

LoopedBard3:
It looks like the ubuntu run is hitting the following error:
```
Assembly PowerShell.Benchmarks which defines benchmarks references non-optimized System.Management.Automation
If you own this dependency, please, build it in RELEASE.
If you don't, you can disable this policy by using 'config.WithOptions(ConfigOptions.DisableOptimizationsValidator)'.
```

Are you aware of whether the dependency is being built in RELEASE?  Otherwise, we can disable the policy as long as we keep the understanding of the potential affect on the benchmarks.

kunalspathak:
@adityapatwardhan might know, but did you find out why it just fails on Ubuntu and not windows?

LoopedBard3:
@kunalspathak I have not found a reason why it only fails on Ubuntu and not Windows, and it is not something that I have any ideas about what could be causing it.

adityapatwardhan:
looking

adityapatwardhan:
The SDK package `Microsoft.PowerShell.SDK` has all assemblies as MSIL, is this the issue?

adityapatwardhan:
Actually this could be the reason:

https://github.com/PowerShell/PowerShell/blob/c444645b0941d73dc769f0bba6ab70d317bd51a9/PowerShell.Common.props#LL188C2-L194C19

kunalspathak:
> Actually this could be the reason:

Thanks @adityapatwardhan . Not sure if I understand "Debugging the issues resolves the problem". Do you know what errors you get if we `Optimize=true`? Also, does that mean that powershell is not optimized at all for non-windows?

adityapatwardhan:
This might be something very old. I am going to try and repro it.

kunalspathak:
> This might be something very old. I am going to try and repro it.

Opened https://github.com/PowerShell/PowerShell/issues/19677 to track it.

TIHan:
I've recently come back from vacation today, is there anything I can do to help with the failures on Ubuntu?

kunalspathak:
FYI- https://github.com/PowerShell/PowerShell/pull/19701

kunalspathak:
> FYI- [PowerShell/PowerShell#19701](https://github.com/PowerShell/PowerShell/pull/19701)

This is merged. What are the next steps remaining here?

LoopedBard3:
> > FYI- [PowerShell/PowerShell#19701](https://github.com/PowerShell/PowerShell/pull/19701)
> 
> This is merged. What are the next steps remaining here?

It looks like the ubuntu run is still failing due to the Release vs Debug issue based on my recent run. The Microsoft.PowerShell.SDK version probably needs to be updated, although it does not look like there has been a recent update pushed out: https://www.nuget.org/packages/Microsoft.PowerShell.SDK/. @adityapatwardhan, do you have an idea of when the next package version with the update will be released or if there is a way to get a version with the changes?

Aside from that, it seems windows defender is stopping it on the public windows, but there was a working windows run before so it should not act as a blocker to merging. Although we may need to remove the public windows run until the issue is fixed.

src/benchmarks/real-world/PowerShell.Benchmarks/PowerShell.Benchmarks.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>$(PERFLAB_TARGET_FRAMEWORKS)</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">net7.0</TargetFrameworks>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <StartArguments>--filter *</StartArguments>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.3.4" />
```
LoopedBard3, line 13:
```suggestion
    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.3.7" />
    <PackageReference Include="Microsoft.Management.Infrastructure" Version="3.0.0-*" />
```

LoopedBard3:
@kunalspathak @TIHan It appears that the packages have been updated to the point that the debug errors are fixed and we should be able to add these benchmarks. I added a suggestion with the necessary package updates based on some of my testing: https://dev.azure.com/dnceng/internal/_build/results?buildId=2284858&view=results. With the package suggestion in and the pipeline run, we should be good to add these benchmarks.

kunalspathak:
Thanks @LoopedBard3 . If the CI is happy, please commit the suggested suggestions and merge the PR.

kunalspathak (dismissed):

kunalspathak (approved):

kunalspathak:
i am assuming the CI failures are unrelated. @TIHan we should add this benchmark in the superpmi collection.
CC: @BruceForstall

TIHan:
Yea that makes sense to me.

LoopedBard3:
> i am assuming the CI failures are unrelated. @TIHan we should add this benchmark in the superpmi collection. CC: @BruceForstall

Is that in addition to here or instead of here?

kunalspathak:
> Is that in addition to here or instead of here?

It is unrelated to this.

LoopedBard3:
> > Is that in addition to here or instead of here?
> 
> It is unrelated to this.

Cool, I will go ahead and merge this in then.

PR #3004, opened 2023-05-03
Update centos7 image to net image

LoopedBard3:
Centos is not being supported in net8 because it doesn't have up to date GLIBCXX files. Initial issue figuring this out: https://github.com/dotnet/sdk/issues/31893, additional information: https://github.com/dotnet/runtime/issues/83428

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -61,92 +61,92 @@ jobs:
 
 - ${{ if or(eq(variables['System.TeamProject'], 'public'), parameters.runPublicJobs) }}:
 
-  # # Scenario benchmarks
-  # - template: /eng/performance/build_machine_matrix.yml
-  #   parameters:
-  #     jobTemplate: /eng/performance/scenarios.yml
-  #     buildMachines:
-  #       - win-x64
-  #       - ubuntu-x64
-  #     isPublic: true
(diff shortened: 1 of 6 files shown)


LoopedBard3:
@adamsitnik CentOS appears to have been chosen as a part of https://github.com/dotnet/performance/pull/2328. I am wondering if you are aware of any specific reasons CentOS was chosen over other containers, such as the ubuntu one this would change the run to? Thanks!

cincuranet (approved):
LGTM, although we should probably notify interested parties for possible changes in numbers from the OS change.

adamsitnik:
> I am wondering if you are aware of any specific reasons CentOS was chosen over other containers, such as the ubuntu one this would change the run to? 

TBH I did not remember the reason, but from the list of my commits of shame where I was trying to get it working I can see that it just had all native dependencies in place:

image

adamsitnik (approved):
LGTM, thank you @LoopedBard3 !

adamsitnik:
> although we should probably notify interested parties for possible changes in numbers from the OS change.

afaik this CI leg was used only for verification that all benchmarks can be compiler and executed for NativeAOT.

adamsitnik:
I'll merge it to unblock other PRs

PR #3016, opened 2023-05-10
Ascii benchmarks

adamsitnik:
benchmarks for https://github.com/dotnet/runtime/pull/85926

cc @gfoidl @GrabYourPitchforks

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text.Encoding/Perf.Ascii.cs
+++ b/src/benchmarks/micro/libraries/System.Text.Encoding/Perf.Ascii.cs
@@ -1,0 +1,106 @@
+﻿using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+using System.Buffers;
+using System.Linq;
+
+namespace System.Text
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public class Perf_Ascii
+    {
+        [Params(
+            6, // non-vectorized code path
+            128)] // vectorized code path
+        public int Size;
+
+        private byte[] _bytes, _sameBytes, _bytesDifferentCase;
+        private char[] _characters, _sameCharacters, _charactersDifferentCase;
+
+        [GlobalSetup]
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/libraries/System.Text.Encoding/Perf.Ascii.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Linq;
+
+namespace System.Text
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public class Perf_Ascii
+    {
+        [Params(
+            6, // non-vectorized code path
+            128)] // vectorized code path
```
gfoidl, line 13: Should there be a size for Vector128/256 too?

The logic for Vec256 and Vec128 is actually the same, just different vector sizes.

gfoidl (approved):
One question, otherwise LGTM 👍🏻

src/benchmarks/micro/libraries/System.Text.Encoding/Perf.Ascii.cs
adamsitnik: > Should there be a size for Vector128/256 too?

You mean the smallest size that gets vectorized? Since we cover both byte and char inputs it would require adding at least few more cases, which would increase the time it takes to run the benchmarks (it's 6.5 minute currently). I would prefer to avoid doing that.

stephentoub (approved):

adamsitnik:
@DrewScoggins Could you please run the tool that generates the html files for https://pvscmdupload.blob.core.windows.net/reports/allTestHistory/TestHistoryIndexIndex.html so the new benchmarks are included?

My motivation is showing the gains from https://github.com/dotnet/runtime/pull/85926 and https://github.com/dotnet/runtime/pull/85266

cc @sblom

adamsitnik:
@cincuranet thank you! please consider myself unblocked

PR #3020, opened 2023-05-11
Add processor affinity option for BDN runs

LoopedBard3:
This adds processor affinity passthrough option for BDN runs that is defined in the yml files. This also sets up Roslyn and Ml.NET to us this affinity parameter. This will likely be quickly iterated on once a path forward for passing environment variables through is figured out.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -351,101 +351,101 @@ jobs:
 # Scheduled runs will run all of the jobs on the PerfTigers, as well as the Arm64 job
 - ${{ if or(and(and(ne(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'Schedule')), not(contains(variables['Build.QueuedBy'], 'Weekly'))), parameters.runScheduledPrivateJobs) }}:
 
-  # # SDK scenario benchmarks
-  # - template: /eng/performance/build_machine_matrix.yml
-  #   parameters:
-  #     jobTemplate: /eng/performance/scenarios.yml
-  #     buildMachines:
-  #       - win-x64
-  #       - win-x86
-  #       - ubuntu-x64
-  #     isPublic: false
-  #     jobParameters:
-  #       kind: sdk_scenarios
-  #       projectFile: sdk_scenarios.proj
-  #       channels:
-  #         - main
+  # SDK scenario benchmarks
+  - template: /eng/performance/build_machine_matrix.yml
+    parameters:
+      jobTemplate: /eng/performance/scenarios.yml
+      buildMachines:
+        - win-x64
+        - win-x86
+        - ubuntu-x64
+      isPublic: false
+      jobParameters:
+        kind: sdk_scenarios
+        projectFile: sdk_scenarios.proj
+        channels:
+          - main
 
-  # # Blazor 3.2 scenario benchmarks
-  # - template: /eng/performance/build_machine_matrix.yml
-  #   parameters:
-  #     jobTemplate: /eng/performance/scenarios.yml
-  #     buildMachines:
-  #       - win-x64
-  #     isPublic: false
-  #     jobParameters:
-  #       kind: blazor_scenarios
-  #       projectFile: blazor_scenarios.proj
-  #       channels:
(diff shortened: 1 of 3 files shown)


LoopedBard3:
Full test run here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2178988&view=results

cincuranet (changes requested):
Please split out the "PERFLAB_DATA" in `Reporter` into separate PR, to have nice/clean history. We can merge that first and then this.

LoopedBard3:
Of course: https://github.com/dotnet/performance/pull/3021.

cincuranet (approved):

LoopedBard3:
FYI @kunalspathak, this gets in the processor affinity for MLNET and ROSLYN tests, expect to see this in the PowerBI.

azure-pipelines.yml
```diff
@@ -447,13 +430,59 @@ jobs:
...
+        - win-arm64
+        - ubuntu-arm64
+      isPublic: false
+      jobParameters:
+        kind: mlnet
+        csproj: src\benchmarks\real-world\Microsoft.ML.Benchmarks\Microsoft.ML.Benchmarks.csproj
+        runCategories: 'mldotnet'
+        channels:
+          - main
+        affinity: '15'
```
kunalspathak, line 463: it would be worth mentioning the binary equivalent of them to specify that we are affinitizing every other cores for x64 because we assume hyper-threading is ON for them. Likewise for arm64.

scripts/ci_setup.py
```diff
@@ -355,6 +361,8 @@ def __main(args: list) -> int:
             out_file.write(variable_format % ('UseSharedCompilation', 'false'))
             out_file.write(variable_format % ('DOTNET_ROOT', dotnet_path))
             out_file.write(variable_format % ('MAUI_VERSION', args.maui_version))
+            if args.affinity:
```
kunalspathak, line 364: @LoopedBard3 or @cincuranet - can one of you confirm how does this get used to set the affinity during benchmark launch?

cincuranet: [This line](https://github.com/dotnet/performance/pull/3020/files#diff-744fe116e315e84456d3107734f7b9f7f57bd567a6599a215a423cb63cf01cf7R97) sets the [parameter](https://github.com/dotnet/performance/pull/3020/files#diff-744fe116e315e84456d3107734f7b9f7f57bd567a6599a215a423cb63cf01cf7R28) for BDN [via command line](https://benchmarkdotnet.org/api/BenchmarkDotNet.ConsoleArguments.CommandLineOptions.html#BenchmarkDotNet_ConsoleArguments_CommandLineOptions_Affinity).

kunalspathak: @adamsitnik - I checked the BDN code, but just want to confirm my understanding. We launch the process, set the affinity mask and then start the benchmark execution, right? Because, `ProcessorAffinity` field of `Process` object can only be set after the process is launched.

adamsitnik: > We launch the process, set the affinity mask and then start the benchmark execution, right?

@kunalspathak that is correct:

start->priority->affinity

https://github.com/dotnet/BenchmarkDotNet/blob/40305b402ff795992ac9d926c8ad1aee853d8492/src/BenchmarkDotNet/Toolchains/Executor.cs#L87

adamsitnik:
> This also sets up Roslyn and Ml.NET to us this affinity parameter.

May I ask what was the reason behind making this change? I am just curious.

kunalspathak:
> > This also sets up Roslyn and Ml.NET to us this affinity parameter.
> 
> May I ask what was the reason behind making this change? I am just curious.

We want to check if Arm64 has performance parity with x64 and we want to measure it in a similar configuration environment. Setting the affinity will make sure that they both are executed with same core count (since hyper threading is ON in x64, it is set to `0x55` and for Arm64 it is `0xF`). Hope that helps!

adamsitnik:
@kunalspathak thanks! It makes sense now ;)

PR #3025, opened 2023-05-17
Add Environment Variable Support to ci_setup and yml files

LoopedBard3:


azure-pipelines.yml
```diff
@@ -234,35 +235,36 @@ jobs:
...
-      buildMachines:
-        - win-x64
-        - ubuntu-x64
-      isPublic: true
-      jobParameters:
-        kind: nativeaot_scenarios
-        projectFile: nativeaot_scenarios.proj
-        channels: # for public jobs we want to make sure that the PRs don't break any of the supported frameworks
-          - main
+        runEnvVars: "DOTNET_GCgen0size=1E00000 DOTNET_GCHeapCount=4 DOTNET_GCTotalPhysicalMemory=200000000"
```
caaavik-msft, line 238: I think it might be better to take this in as an array, you can then join them together using https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#join

LoopedBard3:
Working test run: https://dev.azure.com/dnceng-public/public/_build/results?buildId=277711&view=results. BDN lists the RunEnvVars in its Complus/DOTNET printout. Removing the setup for the PR so we can verify all the values and enable everything in a single separate PR.

scripts/ci_setup.py
```diff
@@ -361,9 +367,14 @@ def __main(args: list) -> int:
...
             out_file.write(variable_format % ('MAUI_VERSION', args.maui_version))
+            out_file.write(path_variable % dotnet_path)
             if args.affinity:
                 out_file.write(variable_format % ('PERFLAB_DATA_AFFINITY', args.affinity))
-            out_file.write(path_variable % dotnet_path)
+            if args.run_env_vars:
+                for env_var in args.run_env_vars:
+                    key, value = env_var.split('=', 1) 
+                    out_file.write(variable_format % (key, value))
+                    out_file.write(variable_format % ("PERFLAB_DATA_" + key, value))
```
kunalspathak, line 377: remind me, but these I believe also gets set in BDN, right? Can you check if it overrides the `HeapCountCharacteristic`?

LoopedBard3: According to https://benchmarkdotnet.org/articles/configs/jobs.html#environment, host process environment setups are used when not specified otherwise. I am not finding anything in the BDN src that would point to otherwise, @adamsitnik if you have any insight off the top of your head it would be useful. I also looked to make sure we are not setting anything in the performance repo and did not find any references to things related to the GC in either the Test files or the benchmarks.extension. We will also verify these EnvVars locally to make sure that they seem to be having an impact before pushing them in a separate PR.

cincuranet (approved):

scripts/ci_setup.py
adamsitnik: > Host process environment setups are used when not specified otherwise

That is true.

eng/performance/scenarios.yml
```diff
@@ -10,7 +10,7 @@ parameters:
   dotnetVersionsLinks: [] # optional alternative to channels that uses 'channel: link' values to scrape the link's json for dotnet_version or version
   projectFile: ''       # required -- project file to build (current choices: scenarios.proj/sdk_scenarios.proj )
   machinePool: ''       # required -- Name of perf machine pool (Tiger, Owl, etc)
-
+  runEnvVars: ''        # optional -- Environment variables to set for the benchmark run in the form of a multi-line object like: "runEnvVars:\n - DOTNET_GCgen0size=1E00000 \n - DOTNET_GCHeapCount=4 \n - DOTNET_GCTotalPhysicalMemory=200000000"
```
caaavik-msft, line 13: For arrays, set the default value to `[]` instead of `''`

eng/performance/scenarios.yml
```diff
@@ -22,6 +22,12 @@ jobs:
         displayName: '${{ parameters.osName }} ${{ parameters.osVersion }} ${{ parameters.architecture }} ${{ parameters.kind }} ${{ parameters.machinePool }}'
         timeoutInMinutes: 320
         variables:
+        - ${{ if ne(parameters.runEnvVars, '')}}:
+          - name: runEnvVarsParam
+            value: "--run-env-vars ${{ join(' ', parameters.runEnvVars)}}"
+        - ${{ if eq(parameters.runEnvVars, '')}}: # if no runEnvVars are specified, don't pass the parameter to ci_setup
```
caaavik-msft, line 28: I would replace this with an `{{ else }}`

caaavik-msft (changes requested):

eng/performance/scenarios.yml
```diff
@@ -22,6 +22,12 @@ jobs:
         displayName: '${{ parameters.osName }} ${{ parameters.osVersion }} ${{ parameters.architecture }} ${{ parameters.kind }} ${{ parameters.machinePool }}'
         timeoutInMinutes: 320
         variables:
+        - ${{ if ne(parameters.runEnvVars, '')}}:
```
caaavik-msft, line 25: When you change the default to `[]`, you can replace this condition with a `ne(length(parameters.runEnvVars), 0)`

eng/performance/scenarios.yml
```diff
@@ -22,6 +22,12 @@ jobs:
...
+          - name: runEnvVarsParam
+            value: "--run-env-vars ${{ join(' ', parameters.runEnvVars)}}"
```
caaavik-msft, line 27: Since both branches of the condition set the same variable name, you can just put the `if` around the `value: ` line rather than having to duplicate `name: runEnvVarsParam`

LoopedBard3:
Feedback updated run: https://dev.azure.com/dnceng-public/public/_build/results?buildId=282307&view=logs&jobId=1c50299c-e5ca-5ffe-6d4f-41d81fafee72&j=1c50299c-e5ca-5ffe-6d4f-41d81fafee72

caaavik-msft (approved):

PR #3026, opened 2023-05-18
First iteration on perf-team's code style

cincuranet:
This is first attempt to put our coding style into a config file. I think we'll fine-tune it as we go.

I'll also follow with commit that changes current code to match the rules once this is merged.


cincuranet:
My plan is to do some solution-wide changes (more like unifications) and then we'll continue finetuning as we go and see stuff.

PR #3032, opened 2023-05-21
Add Convert.FromHex benchmark

hrrrrustic:
See https://github.com/dotnet/runtime/pull/86556#issuecomment-1556276521

### DIFF
--- a/src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Convert.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Convert.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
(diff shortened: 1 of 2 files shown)


danmoseley (dismissed):
not sure whether at some future point we might want to add a range of lengths or not -- more benchmarks has their cost. this is something.

src/benchmarks/micro/libraries/System.Runtime.Extensions/Perf.Convert.cs
```diff
@@ -18,6 +18,7 @@ public class Perf_Convert
         private byte[] _binaryData;
         private char[] _base64CharArray;
         private string _base64String;
+        private string _hexString;
```
danmoseley, line 21:
```suggestion
#if NET5_0_OR_GREATER
        private string _hexString;
#endif
```

danmoseley (approved):

danmoseley:
@LoopedBard3 fyi this failed to download the CLI, so I added some robustification. (Oddly, it needs the retries when I run locally today.)

danmoseley (approved):

danmoseley:
@LoopedBard3 looks like someone other than me needs to sign off -- can you, and merge?

LoopedBard3 (approved):
LGTM

LoopedBard3 (approved):
LGTM

PR #3035, opened 2023-05-22
Add microbenchmarks for Crc32 and Crc64 algorithms

brantburnett:
See https://github.com/dotnet/runtime/pull/86539#issuecomment-1557172918

### DIFF
--- a/src/benchmarks/micro/libraries/System.IO.Hashing/Crc32AppendTests.cs
+++ b/src/benchmarks/micro/libraries/System.IO.Hashing/Crc32AppendTests.cs
@@ -1,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+
+namespace System.IO.Hashing.Tests
+{
+    public class Crc32AppendTests : CrcAppendTests<Crc32>
+    {
+        [Params(16, 128, 10240)]
+        public override int BufferSize { get; set; }
+    }
+}
--- a/src/benchmarks/micro/libraries/System.IO.Hashing/Crc32AppendTests.cs
+++ b/src/benchmarks/micro/libraries/System.IO.Hashing/Crc32AppendTests.cs
@@ -6,7 +6,7 @@ using BenchmarkDotNet.Attributes;
 
 namespace System.IO.Hashing.Tests
 {
-    public class Crc32AppendTests : CrcAppendTests<Crc32>
+    public class Crc32_Perf : Crc_Perf<Crc32>
     {
         [Params(16, 128, 10240)]
         public override int BufferSize { get; set; }
--- a/src/benchmarks/micro/libraries/System.IO.Hashing/Crc64AppendTests.cs
+++ b/src/benchmarks/micro/libraries/System.IO.Hashing/Crc64AppendTests.cs
@@ -1,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+
+namespace System.IO.Hashing.Tests
+{
+    public class Crc64AppendTests : CrcAppendTests<Crc64>
+    {
+        [Params(16, 256, 10240)]
+        public override int BufferSize { get; set; }
+    }
+}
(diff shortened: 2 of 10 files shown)


brantburnett:
@adamsitnik 

Here are the CRC benchmarks per your request. I do see an execution failure in one case in CI, but my attempts to make sense of it have hit a dead end. I don't think it's related to this change, but can't be certain.

danmoseley:
Yea, it's not your change, some fragility in a winforms test.

danmoseley (approved):

danmoseley (approved):

src/benchmarks/micro/libraries/System.IO.Hashing/CrcAppendTests.cs
```diff
@@ -0,0 +1,33 @@
...
+        public void Setup()
+        {
+            Buffer = ValuesGenerator.Array<byte>(BufferSize);
+            Crc = new T();
+        }
+
+        [Benchmark]
+        public void Append()
+        {
+            Crc.Append(Buffer);
```
adamsitnik, line 30: Please excuse me if this is a dumb question (I've become an owner of `System.IO.Hashing` quite recently), but is my understanding correct that `Append` is not having any side-effects that could prolong the execution time of the next invocations and we don't need to call `Crc.Reset()` for every benchmark?

src/benchmarks/micro/libraries/System.IO.Hashing/CrcAppendTests.cs
```diff
@@ -0,0 +1,33 @@
...
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Extensions;
+using MicroBenchmarks;
+
+namespace System.IO.Hashing.Tests
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public abstract class CrcAppendTests<T>
```
adamsitnik, line 12: in the future we might add benchmarks for other methods, so we could make the type name more general
```suggestion
    public abstract class Crc_Perf<T>
```

src/benchmarks/micro/libraries/System.IO.Hashing/Crc64AppendTests.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+
+namespace System.IO.Hashing.Tests
+{
+    public class Crc64AppendTests : CrcAppendTests<Crc64>
```
adamsitnik, line 9:
```suggestion
    public class Crc64_Perf : Crc_Perf<Crc64>
```

src/benchmarks/micro/libraries/System.IO.Hashing/Crc32AppendTests.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+
+namespace System.IO.Hashing.Tests
+{
+    public class Crc32AppendTests : CrcAppendTests<Crc32>
```
adamsitnik, line 9:
```suggestion
    public class Crc32_Perf : Crc_Perf<Crc32>
```

src/benchmarks/micro/libraries/System.IO.Hashing/CrcAppendTests.cs
```diff
@@ -0,0 +1,33 @@
...
+            Buffer = ValuesGenerator.Array<byte>(BufferSize);
+            Crc = new T();
+        }
+
+        [Benchmark]
+        public void Append()
+        {
+            Crc.Append(Buffer);
+        }
+    }
```
adamsitnik, line 32: Would you mind adding benchmarks for other methods? Like [[GetCurrentHash()](https://learn.microsoft.com/en-us/dotnet/api/system.io.hashing.noncryptographichashalgorithm.getcurrenthash?view=dotnet-plat-ext-7.0#system-io-hashing-noncryptographichashalgorithm-getcurrenthash)](https://learn.microsoft.com/en-us/dotnet/api/system.io.hashing.noncryptographichashalgorithm.getcurrenthash?view=dotnet-plat-ext-7.0#system-io-hashing-noncryptographichashalgorithm-getcurrenthash(system-span((system-byte)))) for example?

adamsitnik (commented):
@brantburnett thank you for your contribution! Overall it looks good, but PTAL at my comments.

adamsitnik (commented):
@brantburnett thank you for your contribution! Overall it looks good, but PTAL at my comments.

cincuranet:
@adamsitnik is there anything that prevents this from merging?

cincuranet:
I'm going to merge this as it is now and add the remaining benchmarks myself in separate PR.

PR #3038, opened 2023-05-23
Add affinity and runEnvVars for MLNET and Roslyn private runs.

LoopedBard3:
This combines the previously split MLNET and Roslyn jobs (win/ubuntu + x64/arm64) and sets them to use affinity 85 (01010101) and sets multiple environment variables to setup the GC.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -437,20 +437,6 @@ jobs:
       buildMachines:
         - win-x64
         - ubuntu-x64
-      isPublic: false
-      jobParameters:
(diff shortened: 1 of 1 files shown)


kunalspathak (approved):
Thanks. So this will get enabled immediately this PR is merged?

LoopedBard3:
> Thanks. So this will get enabled immediately this PR is merged?

Yes, once this is merged, the changes will impact future runs.

LoopedBard3:
These changes don't affect any public job runs, merging.

PR #3042, opened 2023-05-24
Add processor affinity options scenarios

LoopedBard3:


LoopedBard3:
Test run demonstrating the E2E working: https://dev.azure.com/dnceng-public/public/_build/results?buildId=285453&view=logs&jobId=85302f16-2cc4-5077-bbf0-683d7d61c604&j=85302f16-2cc4-5077-bbf0-683d7d61c604&t=b12d3e62-1722-50df-a17e-c06a6f3518be

LoopedBard3:
Tested the affinity parameter in the E2E demonstration, but not keeping the test in. Instead will add the new scenario setup in its own PR, especially since there will be more than just affinity changed.

src/tools/ScenarioMeasurement/Util/ProcessHelper.cs
```diff
@@ -265,7 +265,7 @@ public ProcessResult Run()
                 Thread.Sleep(MeasurementDelay * 1000);
                 if (!process.HasExited)
                 {
-                    if(GuiApp)
+                    if (GuiApp)
```
kunalspathak, line 268: As I mentioned offline, a better way to measure the startup is to ready the console output from PDN and remove the code that we have today for `AppStarted`, `AppIdle`, etc.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -135,7 +137,22 @@ static void checkArg(string arg, string name)
 
         logger.Log($"Running {appExe} (args: \"{appArgs}\")");
         logger.Log($"Working Directory: {workingDir}");
-        if(runWithoutExit)
+
+        if (affinity > 0 && (OperatingSystem.IsWindows() || OperatingSystem.IsLinux()))
+        {
```
cincuranet, line 142: Use variable for `Process.GetCurrentProcess()` and reuse it, instead of doing the call again.

cincuranet (changes requested):

src/scenarios/shared/runner.py
```diff
@@ -198,11 +206,21 @@ def parseargs(self):
         if args.scenarioname:
             self.scenarioname = args.scenarioname
 
+        if self.testtype in [const.STARTUP, const.INNERLOOP, const.INNERLOOPMSBUILD, const.DOTNETWATCH, const.SDK, const.CROSSGEN, const.CROSSGEN2] and (args.affinity or os.environ.get('PERFLAB_DATA_AFFINITY')): # Set affinity if doing a Startup based test
```
LoopedBard3, line 209: Would it be better to use getattr to check if affinity exists instead?

LoopedBard3: Attempted to make a good solution using getattr and couldn't come up with one that worked with both affinity passing options.

cincuranet (approved):

PR #3056, opened 2023-05-30
Add PDN as an affinitized Scenario.

LoopedBard3:
Add PDN as an affinitized Scenario. With this setup, both the testname and affinity additional data can be used to differentiate between runs. Also updated the note for affinity in ci_setup and what it functionally does.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -306,229 +306,229 @@ jobs:
           - DOTNET_GCHeapCount=4
           - DOTNET_GCTotalPhysicalMemory=400000000 # 16GB
 
-  # # Maui Android scenario benchmarks
-  # - template: /eng/performance/build_machine_matrix.yml
-  #   parameters:
-  #     jobTemplate: /eng/performance/scenarios.yml
-  #     buildMachines:
-  #       - win-x64-android-arm64-pixel
-  #       - win-x64-android-arm64-galaxy
-  #     isPublic: false
-  #     jobParameters:
-  #       kind: maui_scenarios_android
-  #       projectFile: maui_scenarios_android.proj
-  #       dotnetVersionsLinks:
-  #         main: https://aka.ms/dotnet/sdk/maui/net8.0.json
+  # Maui Android scenario benchmarks
+  - template: /eng/performance/build_machine_matrix.yml
+    parameters:
+      jobTemplate: /eng/performance/scenarios.yml
+      buildMachines:
+        - win-x64-android-arm64-pixel
+        - win-x64-android-arm64-galaxy
+      isPublic: false
+      jobParameters:
+        kind: maui_scenarios_android
+        projectFile: maui_scenarios_android.proj
+        dotnetVersionsLinks:
+          main: https://aka.ms/dotnet/sdk/maui/net8.0.json
 
-  # # Maui iOS scenario benchmarks
-  # - template: /eng/performance/build_machine_matrix.yml
-  #   parameters:
-  #     jobTemplate: /eng/performance/scenarios.yml
-  #     buildMachines:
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -286,6 +286,25 @@ jobs:
...
+      buildMachines:
+        - win-x64
+        - win-arm64
+      isPublic: false
+      jobParameters:
+        kind: scenarios
+        projectFile: scenarios_affinitized.proj
+        channels:
+          - main
+        affinity: '85'
```
kunalspathak, line 302: This should be `0x55` for x64 and `0xF` for arm64.

azure-pipelines.yml
```diff
@@ -286,6 +286,25 @@ jobs:
...
+  # Affinitized Scenario benchmarks (Initially just PDN)
+  - template: /eng/performance/build_machine_matrix.yml
+    parameters:
+      jobTemplate: /eng/performance/scenarios.yml
+      buildMachines:
+        - win-x64
+        - win-arm64
+      isPublic: false
+      jobParameters:
+        kind: scenarios
```
cincuranet, line 298: The `kind` should be something else, isn't it? With this it blends into "regular" scenarios.

cincuranet: I was more worried about new affinitized numbers going into old non-affinitized. But as I'm thinking about it, it's probably not a bad thing. We know about it and it's just kind of configuration setting.

LoopedBard3: For this scenario, the name of the affinitized test is also updated to include the affinity in the name to denote the difference.

azure-pipelines.yml
LoopedBard3: The value is currently passed in as a decimal so the 85 corresponds to 0x55. Other than not specifying the alternate cores, is there a specific reason to use 0xF over 0x55? Using the same for both will make some of the downstream report generation more clean, if possible.

kunalspathak: I honestly not sure if specifying one way or the other makes any difference. lets go with `0x55` for now. Just to confirm these settings are also applicable for ml.net and roslyn, right?

LoopedBard3: Yup, both ml.net and roslyn are already using 0x55.

LoopedBard3:
Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2190983&view=results, PDN tests both passed, unrelated scenarios failed. 
The failures are unrelated and are a mix of:
Windows arm64:
```
<class 'azure.core.exceptions.ServiceResponseError'>: ('Connection aborted.', ConnectionResetError(10054, 'An existing connection was forcibly closed by the remote host', None, 10054, None))
```
Ubuntu arm64:
```
[2023/05/30 19:25:49][INFO] Failed to load /home/helixbot/work/AFFC096A/p/dotnet/host/fxr/8.0.0-preview.5.23272.14/libhostfxr.so, error: /home/helixbot/work/AFFC096A/p/dotnet/host/fxr/8.0.0-preview.5.23272.14/libhostfxr.so: cannot open shared object file: No such file or directory
[2023/05/30 19:25:49][INFO] The library libhostfxr.so was found, but loading it from /home/helixbot/work/AFFC096A/p/dotnet/host/fxr/8.0.0-preview.5.23272.14/libhostfxr.so failed
[2023/05/30 19:25:49][INFO]   - Installing .NET prerequisites might help resolve this problem.
[2023/05/30 19:25:49][INFO]      https://go.microsoft.com/fwlink/?linkid=2063370
```

cincuranet (approved):
LGTM. 

Maybe small question. The non-affinitized version for PDN scenario is still running, do we have clear separation between data we're storing?

PR #3076, opened 2023-06-08
Mitigate Maui android publish failing due to NETSDK1191

LoopedBard3:
Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2196823&view=logs&j=d610060c-e82a-5178-ebd9-7afbbedc05c2

Fixes: https://github.com/dotnet/performance/issues/3045

May impact the SOD tests with only android-arm64 being built for sure now.

PR #3077, opened 2023-06-09
Use no-self-contained for Maui Android Publish to keep SOD consistent

LoopedBard3:
Publishing no-self-contained appears to result in the expected APK file, but allows for the building of all of the RIDs at once. This should keep the SOD results consistent pre and post break. This is a direct follow up to #3076 which used self-contained, meaning only one of the 4 RIDs would be built.

Test Run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2197034&view=results

SOD for Maui Android Default APK for validation:
Pre-break run: 33039638.000 bytes
Test run: 33142038.000 bytes

### DIFF
--- a/eng/performance/maui_scenarios_android.proj
+++ b/eng/performance/maui_scenarios_android.proj
@@ -51,7 +51,7 @@
 
   <ItemGroup>
     <PreparePayloadWorkItem Include="@(MAUIAndroidScenario)">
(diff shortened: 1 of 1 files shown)


caaavik-msft (approved):

LoopedBard3:
The only error in the test run is in the blazor memory consumption test, but appears to be related to device health rather than the blazor app. Merging.

PR #3079, opened 2023-06-12
Switch to only building Maui Android apps for android-arm64

LoopedBard3:
Switch to only building Maui Android apps for android-arm64 to match net7.0 building, and speedup build times. The impact of this will be the SOD results decreasing by a good bit.

Test run from matching PR(https://github.com/dotnet/performance/pull/3076): https://dev.azure.com/dnceng/internal/_build/results?buildId=2196823&view=results


jonathanpeppers (approved):

mkArtakMSFT (approved):
Thanks @LoopedBard3.
@Eilon I assume you have no concerns with this, have you?

caaavik-msft (approved):

Issue #3085, opened 2023-06-15
Microbenchmark uploads hitting blob already exists


LoopedBard3:
Some of our uploads are hitting blob already exist errors causing long run-times when hit in conjunction with our retry strategy. If a blob already exists, we should not retry uploading it, maybe return error instead.

This seems to be in part a reason for some runs to take an extremely long time as the resource does not exist but we keep trying to upload. Specifically, the ubuntu.1804 queue seems to be stuck on a handful of jobs occassionally where the tests finish in 30-60 minutes, but the uploading failing and retrying is causing the time to go to the timeout of 6 hours.

PR #3087, opened 2023-06-15
Add raising exceptions to retry_on_exception and use for uploading if resource already exists 

LoopedBard3:
Add ability to not retry on certain exceptions in retry_on_exception and use this to not retry on ResourceExistsError.
The new raise_exceptions will raise if they are hit rather than retrying, this is useful if there are known non-transient exceptions.

Tested locally to ensure that this Exception captures properly.
Fixes: https://github.com/dotnet/performance/issues/3085

### DIFF
--- a/scripts/performance/common.py
+++ b/scripts/performance/common.py
@@ -158,7 +158,7 @@ def push_dir(path: str = None) -> None:
     else:
         yield
 
-def retry_on_exception(function, retry_count=3, retry_delay=5, retry_delay_multiplier=1, retry_on_exception=Exception):
+def retry_on_exception(function, retry_count=3, retry_delay=5, retry_delay_multiplier=1, retry_exceptions=[Exception], raise_exceptions=[]):
     '''
     Retries the specified function if it throws an exception.
 
@@ -166,7 +166,8 @@ def retry_on_exception(function, retry_count=3, retry_delay=5, retry_delay_multi
     :param retry_count: The number of times to retry the function.
     :param retry_delay: The delay between retries (seconds).
     :param retry_delay_multiplier: The multiplier to apply to the retry delay after failure.
-    :param retry_on_exception: The exception to retry on (Defaults to Exception).
+    :param retry_exceptions: The exception to retry on (Defaults to Exception). (Exclusive use from raise_exceptions)
+    :param raise_exceptions: The exceptions to ignore (Defaults to no exceptions ignored). (Exclusive use from retry_exceptions)
     '''
     if retry_count < 0:
         raise ValueError('retry_count must be >= 0')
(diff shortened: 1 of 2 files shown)


DrewScoggins:
Do we have any idea as to why we are suddenly seeing blobs already existing? It seems like we should try and get at the underlying issue here.

LoopedBard3:
Each of the jobs that I saw that failed in this way were hit in attempts after the first. I have not had a chance to investigate why the first run failed, but I am pretty sure that the results are being uploaded in previous attempts under the same name.

scripts/performance/common.py
```diff
@@ -158,30 +158,39 @@ def push_dir(path: str = None) -> None:
...
+def retry_on_exception(function, retry_count=3, retry_delay=5, retry_delay_multiplier=1, retry_exceptions=[Exception], raise_exceptions=[]):
     '''
     Retries the specified function if it throws an exception.
 
     :param function: The function to execute.
     :param retry_count: The number of times to retry the function.
     :param retry_delay: The delay between retries (seconds).
     :param retry_delay_multiplier: The multiplier to apply to the retry delay after failure.
-    :param retry_on_exception: The exception to retry on (Defaults to Exception).
+    :param retry_exceptions: The exception to retry on (Defaults to Exception). (Exclusive use from raise_exceptions)
```
DrewScoggins, line 169: Can you update this to make it a bit more clear that you can't specify both `raise_exceptions` and `retry_exceptions` at the same time.

DrewScoggins:
When you mention failed and multiple attempts, is that from inside our automation, or is this workitem retries that happen through Helix?

LoopedBard3:
The attempts I am talking about above refer to helix retries, not upload retries. So the workitem is run again.

scripts/performance/common.py
LoopedBard3: Changed the verbiage: "Exclusive use from --_exceptions" -> "Cannot be used with --_exceptions"

scripts/performance/common.py
```diff
@@ -158,30 +158,39 @@ def push_dir(path: str = None) -> None:
...
+            # If using the raise_exceptions list, raise those exceptions no matter what
+            if len(raise_exceptions) > 0 and type(e) in raise_exceptions:
+                raise
+            # If use retry_exceptions list, only retry on those exceptions (or all if Exception is in the list)
+            elif Exception not in retry_exceptions and type(e) not in retry_exceptions:
+                raise
             if i == retry_count - 1:
                 raise
-            getLogger().info('Exception caught: %s', e)
+            getLogger().info('Exception caught {0}: {1}'.format(type(e), str(e)))
```
cincuranet, line 193: nit: I would maybe use f-strings.

cincuranet (approved):

scripts/performance/common.py
LoopedBard3: Fixed

cincuranet (approved):

PR #3109, opened 2023-07-06
Setup android power consumption test capability

LoopedBard3:
Setup the ability to run android power consumption tests. This includes:
* Adding the devicepowerconsumption run type in runner.py.
* Creating the devicepowerconsumption wrapper that runs the android specific flow and builds the power consumption skeleton tool.
* Creating the PowerConsumption tool project with the minimal setup to get the device tests working (copied from the MemoryConsumption tool).
* Minor updates to androidhelper.py to enable immediate return after calling to start the app.

### DIFF
--- a/src/scenarios/shared/devicepowerconsumption.py
+++ b/src/scenarios/shared/devicepowerconsumption.py
@@ -149,6 +149,14 @@ class DevicePowerConsumptionHelper(object):
                 packagename
             ]
 
+            # Gets information about the battery (https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/services/core/java/com/android/server/BatteryService.java;l=1117-1119?q=%22Current%20Battery%20Service%20state%22)
+            getGeneralBatteryInformation = [
+                androidHelper.adbpath,
+                'shell',
+                'dumpsys',
+                'battery'
+            ]
+
             allResults = []
             # Verify that a yepkit board is found
             yepkitCheck = RunCommand(listYepkitBoardsCmd, verbose=True)
@@ -163,10 +171,14 @@ class DevicePowerConsumptionHelper(object):
                 RunCommand(clearBatteryStatsCmd, verbose=True).run()
                 startStats = RunCommand(androidHelper.startappcommand, verbose=True)
                 startStats.run()
+                preExecutionBatteryInfo = RunCommand(getGeneralBatteryInformation, verbose=True)
+                preExecutionBatteryInfo.run()
                 RunCommand(disconnectYepKitPowerCmd, verbose=True).run()
                 time.sleep(runtimeseconds)
                 RunCommand(reconnectYepKitPowerCmd, verbose=True).run()
                 time.sleep(5) # Wait for the phone to reconnect to the commputer
+                postExecutionBatteryInfo = RunCommand(getGeneralBatteryInformation, verbose=True)
+                postExecutionBatteryInfo.run()
                 RunCommand(androidHelper.stopappcommand, verbose=True).run()
                 
                 captureUid = RunCommand(getUidOfPackageCmd, verbose=True)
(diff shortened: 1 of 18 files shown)


cincuranet (approved):

jandupej:
LGTM. As a side note, the unit of energy is Joule, so it is a shame that Android reports that in mAh. This is probably because phones use "fuel gauge" ICs to monitor charge in and out of the battery, and these are Coulomb counters. mAh maps to Joules directly only if battery voltage remains constant, which is not true. However, if the tests do not take long, the battery voltage will not drop by a huge amount, so this should be OK.

fanyang-mono:
The overall logic of capturing Android app power consumption looks good to me. Thanks for enabling this in the perf lab. :)

LoopedBard3:
> LGTM. As a side note, the unit of energy is Joule, so it is a shame that Android reports that in mAh. This is probably because phones use "fuel gauge" ICs to monitor charge in and out of the battery, and these are Coulomb counters. mAh maps to Joules directly only if battery voltage remains constant, which is not true. However, if the tests do not take long, the battery voltage will not drop by a huge amount, so this should be OK.

That makes a lot of sense. I went ahead and found a way to get voltage values for before and after running the app so I could add a Joule estimate to the results. Since we don't have a good way to get the voltage and mAh overtime this is just an estimate that takes the average voltage as if it was constant. This does require the assumption that the voltage only slightly drops but we it seems that we should be able to get a good ballpark for the Joule calculation. The calculations were added here: https://github.com/dotnet/performance/pull/3109/files#diff-fbb5a17c84622b58e4a5fa676b084bf0a4113f56ec1ac8d9f4bdd179e0ebdfc1R73-R78, so let me know if you think the results from this should be useful.

> Also, it would be good to make sure that batches of tests are repeated with the same starting charge level, e.g. by starting the test suite only if the phone is fully charged (this assumes that a fully charged phone has always the same-ish battery voltage). Otherwise the mAh figures could be distorted if there are large differences in initial charge levels (voltage) between runs. I see that the scripts control some sort of USB mux board - is that to take care of consistent charging?

The phones we use are always plugged in so should always be at full charge. The USB mux board we use is the https://www.yepkit.com/product/300115/YKUSHXS, which allows us to programmatically disconnect and reconnect the phones. As such, there shouldn't be large differences in initial charge levels, but to allow for better insight into if that does start causing problems I added the voltage as above and the phone battery level (percentage) as values we capture.

jandupej:
> The calculations were added here:

Nice! I think this is the best we can do right now. 

> The phones we use are always plugged in so should always be at full charge

Thanks for clearing that up. I agree, now that we use voltages to estimate energy, the process should be invariant to initial charge level/voltage.

src/scenarios/shared/devicepowerconsumption.py
```diff
@@ -0,0 +1,288 @@
...
+                
+                captureProcStats = RunCommand(captureBatteryStatsCmd, verbose=True)
+                captureProcStats.run()
+
+                capturedValues = {}
+                # Get the mAh estimated power use based on the Uid from the battery stats
+                # Explanation of the 4 groups: https://stackoverflow.com/questions/75390939/android-how-to-interpret-pwi-power-use-item-from-battery-stats-dumpsys
+                # Example output section and target:
+                # 9,11458,l,pwi,uid,0.0820,0,0.134,0.0342 captures total usage, is battery consumer (consult link below), screen usage, and proportional usage
+                totalPowerSearchString = r"[\d]*," + uid + r",l,pwi,uid,([\d]*.?[\d]*),([\d]*.?[\d]*),([\d]*.?[\d]*),([\d]*.?[\d]*)"
```
cincuranet, line 218: Is there a reason you use `[\d*]` instead of `\d*`. 

Also, I guess, the number matched is "required", thus maybe `\d+` would be a better.

cincuranet (approved):

LoopedBard3:
Errors are unrelated as no runs were added to use the new powerconsumption tooling, merging.

LoopedBard3:
Errors are unrelated as this PR doesn't add any new tests, just the tooling to run power consumption tests.

PR #3114, opened 2023-07-09
Add `BinaryReader` benchmarks et. al.

teo-tsirpanis:
Asked in dotnet/runtime#80331. I haven't run the benchmarks myself.

c.c. @adamsitnik

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -63,18 +63,14 @@
     <PackageReference Include="System.Drawing.Common" Version="$(SystemVersion)" Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '8.0'" />
     <PackageReference Include="System.Formats.Cbor" Version="$(SystemVersion)" Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &gt;= '5.0'" />
     <PackageReference Include="System.IO.Pipelines" Version="$(SystemVersion)" />
+    <PackageReference Include="System.Threading.Channels" Version="$(SystemVersion)" />
+    <PackageReference Include="Utf8Json" Version="1.3.7" />
+    <PackageReference Include="System.Collections.Immutable" Version="8.0.*-*" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '8.0')" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
     <PackageReference Include="System.Memory" Version="4.5.5" />
     <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
-    <PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
-    <PackageReference Include="System.Runtime.Serialization.Json" Version="4.3.0" />
-    <PackageReference Include="System.Threading.Channels" Version="$(SystemVersion)" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
-    <PackageReference Include="System.Xml.XmlSerializer" Version="4.3.0" />
-    <PackageReference Include="Utf8Json" Version="1.3.7" />
-    <!-- following two dependencies must point to version 4.3.0 or newer to make it possible to publish MicroBenchmarks.csproj -->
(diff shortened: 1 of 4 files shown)


src/benchmarks/micro/libraries/System.IO/BinaryReaderTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                var stream = new MemoryStream();
+                var writer = new BinaryWriter(stream, Encoding.UTF8);
+                writer.Write(value);
+                return stream;
+            }
+        }
+
+        [Benchmark]
+        public BinaryReader DefaultCtor() => new BinaryReader(Stream.Null);
```
adamsitnik, line 42: thank you for adding the benchmarks for all the APIs 👍

src/benchmarks/micro/libraries/System.IO/BinaryWriterTests.cs
```diff
@@ -57,6 +57,14 @@ public void WriteUInt64()
             _bw.Write((ulong)0xdeadbeef_aabbccdd);
         }
 
+#if NET5_0_OR_GREATER
+        [Benchmark]
+        public void WriteHalf()
```
adamsitnik, line 62: thank you for covering the benchmark gap! 👍

src/benchmarks/micro/libraries/System.IO/BinaryReaderTests.cs
```diff
@@ -0,0 +1,95 @@
...
+#if NET5_0_OR_GREATER
+        [Benchmark]
+        public Half ReadHalf() => _dummyReader.ReadHalf();
+#endif
+
+        [Benchmark]
+        public float ReadSingle() => _dummyReader.ReadSingle();
+
+        [Benchmark]
+        public double ReadDouble() => _dummyReader.ReadDouble();
```
adamsitnik, line 79: Is my understanding correct that the dummy reader is going to return `0` for all the benchmarks that return numeric types? If that is the case, it would be better to let it perform more work (to mimic more realistic scenario), by for example reading the values used for `BinaryWriter` benchmarks.

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -63,19 +63,15 @@
...
     <PackageReference Include="System.IO.Pipelines" Version="$(SystemVersion)" />
-    <PackageReference Include="System.Memory" Version="4.5.5" />
-    <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
-    <PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
-    <PackageReference Include="System.Runtime.Serialization.Json" Version="4.3.0" />
     <PackageReference Include="System.Threading.Channels" Version="$(SystemVersion)" />
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
-    <PackageReference Include="System.Xml.XmlSerializer" Version="4.3.0" />
     <PackageReference Include="Utf8Json" Version="1.3.7" />
-    <!-- following two dependencies must point to version 4.3.0 or newer to make it possible to publish MicroBenchmarks.csproj -->
```
adamsitnik, line 74: The NativeAOT CI leg is green, which means that the comment is no longer valid. Thank you for simplifying the project file 👍

adamsitnik (commented):
@teo-tsirpanis thank you for your contribution! PTAL at my comment regarding the benchmarks that read numeric values.

src/benchmarks/micro/libraries/System.IO/BinaryReaderTests.cs
teo-tsirpanis: Isn't the goal of this microbenchmark to measure only the performance of `BinaryReader`? `DummyReadStream` does the absolute minimum that makes reading succeed. The `BinaryWriter` benchmarks do something similar by discarding the data written.

teo-tsirpanis: > the reader reads first value, it turns out to be 0, so it returns zero and finishes

I don't understand. Only `ReadString` exhibits this behavior and we don't use `DummyReadStream` there. Reading primitive types from a `BinaryReader` will always read one value no matter the binary data.

adamsitnik: @teo-tsirpanis please excuse me, for some reason I forgot that it's binary deserialization, so it does not depend on the input size like it does for Utf8 strings etc:

https://github.com/dotnet/runtime/blob/e71faeeeccba29c1f06c7020929cfdadbd32401e/src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Utf8Parser/Utf8Parser.Integer.Unsigned.D.cs#L21-L27

adamsitnik (approved):
LGTM, thank you @teo-tsirpanis !

adamsitnik:
@cincuranet the build failures are unrelated (scenarios), could you please merge the PR?

cincuranet:
Merged.

PR #3118, opened 2023-07-11
[wasm] Run utf8 encoding tests on wasm now

lewing:
@LoopedBard3 looks like we can run these now

cc @kg @radical

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text.Encoding/Perf.Utf8Encoding.cs
+++ b/src/benchmarks/micro/libraries/System.Text.Encoding/Perf.Utf8Encoding.cs
@@ -8,7 +8,7 @@ using System.IO;
 
 namespace System.Text
 {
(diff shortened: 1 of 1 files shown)


radical (approved):

radical:
I'm unable to run the pipeline for any PRs because the helix jobs never get any agents assigned, even after >3hrs. I was trying earlier today for a different PR on build [20230710.5](https://dev.azure.com/dnceng/internal/_build/results?buildId=2218039).

lewing:
reverts https://github.com/dotnet/performance/pull/1482 to verify https://github.com/dotnet/performance/pull/1482

radical:
I meant that we'll have to depend on just local validation, because the pipeline can't reliably be run for the PR right now. Since it runs locally, it should be fine. I'll keep an eye on the daily perf runs.

PR #3120, opened 2023-07-11
Update BDN

cincuranet:
* Update BDN
* Update version of Microsoft.CodeAnalysis.CSharp in ILLink benchmarks
* Update ILLink package versions

### DIFF
--- a/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
+++ b/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.5.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
+++ b/src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.5.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.0" />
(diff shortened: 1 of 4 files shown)


cincuranet:
* Failures are on par with what's [currently happening in _main_](https://dev.azure.com/dnceng-public/public/_build/results?buildId=337077&view=results).
* The _micro+nativeaot8.0_ is going to be handled in separate task #3135.

LoopedBard3 (dismissed):
Changes look good 👍

LoopedBard3 (dismissed):

LoopedBard3 (dismissed):

cincuranet:
Currently blocked by https://github.com/dotnet/BenchmarkDotNet/issues/2369.

src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
```diff
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.5.0" />
```
adamsitnik, line 18: This one is simple, but tricky. IIRC the main point of the Roslyn benchmarks was to see how runtime changes affect compilation time with Roslyn. That is why we have been using a fixed version of Roslyn (BTW we always wanted to add another set of CI legs that would use a fixed runtime and latest Roslyn).

Is this bump mandatory due to the BDN changes? If so, it's acceptable, we just need to remember that if there is a regression reported for these benchmarks it might be caused by the Roslyn change, not runtime change.

adamsitnik (commented):
> Currently blocked by https://github.com/dotnet/BenchmarkDotNet/issues/2369.

I've just merged https://github.com/dotnet/BenchmarkDotNet/pull/2365, there should be a new preview version available in the BDN CI feed in +- 1h.

src/benchmarks/real-world/Roslyn/CompilerBenchmarks.csproj
adamsitnik: I just noticed that the Roslyn dependency got downgraded to 4.1.0 in https://github.com/dotnet/BenchmarkDotNet/pull/2359 so this change should not be required anymore

caaavik-msft (approved):

PR #3122, opened 2023-07-12
Fix mono aot build

radical:
Issue: https://github.com/dotnet/BenchmarkDotNet/issues/2311

- Latest BDN copies the `PackageReferences` from
  `MicroBenchmarks.csproj` to the generated project.
  - but it does not copy the version property, thus breaking the build.
  - Move the version property to a Common.props, so it can be used by
    the main MicroBenchmarks.csproj, and the autogenerated project.


radical:
This depends on https://github.com/dotnet/BenchmarkDotNet/pull/2367 .

src/benchmarks/micro/MicroBenchmarks.Mono.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <!-- Shared with the autogenerated project -->
+  <Import Project="*Common.props" />
+  <PropertyGroup>
+    <SelfContained>true</SelfContained>
```
kotlarmilos, line 5: The `SelfContained` parameter is required in  the `src/BenchmarkDotNet/Templates/MonoAOTLLVMCsProj.txt` template. To have clear separation of responsibility between the performance and BenchmarkDotNet repositories, I recommend avoiding any dependencies if possible.

We can add the property to the `MonoAOTLLVMCsProj.txt` template and import the `$([System.IO.Path]::ChangeExtension('$(OriginalCSProjPath)', '.Common.props'))`.

kotlarmilos (dismissed):
A minor suggestion, otherwise LGTM!

src/benchmarks/micro/MicroBenchmarks.Mono.props
radical: > The `SelfContained` parameter is required in the `src/BenchmarkDotNet/Templates/MonoAOTLLVMCsProj.txt` template. To have clear separation of responsibility between the performance and BenchmarkDotNet repositories, I recommend avoiding any dependencies if possible.

Done. I didn't want to assume how the project gets used.

> We can add the property to the `MonoAOTLLVMCsProj.txt` template and import the `$([System.IO.Path]::ChangeExtension('$(OriginalCSProjPath)', '.Common.props'))`.

Is the suggestion to name the props file as `Common.props` instead of `Mono.props`?
The reason I chose Mono.props is because it allows the benchmarks project to have the settings separate per project type. We could have a `Common.props` imported by template projects, but then we also need some property to indicate the type of project being built. And you would need to set this property in the template project itself. But I don't know enough about how these get used from bdn.

This would be better approach, and if you want to implement it:
1. change the name of Mono.props/targets, and the .Wasm.props/targets to Common.props/targets
2. Set some property to indicate the project type in the template project
3. You also might want to add this props/targets as a common pattern to the other template projects too.
4. in dotnet/performance, add the Common.props and use the project type property to differentiate the settings.
5. Make sure to test the whole thing from `dotnet/performance`, and `dotnet/runtime`.

kotlarmilos: Thanks!

> Is the suggestion to name the props file as Common.props instead of Mono.props?

If we set the SelfContained property in the template and remove the `MicroBenchmarks.Mono.props`, I assume that the `<Import Project="*Common.props" />` should be moved from `MicroBenchmarks.Mono.props`.

Everything else looks good to me and I wouldn't change the naming.

radical: We still need the versions set in `Common.props`, and imported into the generated project via `Mono.props` because:
```
Latest BDN copies the PackageReferences from
MicroBenchmarks.csproj to the generated project.

- but it does not copy the version property, thus breaking the build.
- Move the version property to a Common.props, so it can be used by
the main MicroBenchmarks.csproj, and the autogenerated project.
```

re:remove-selfcontained, done 👍

LoopedBard3 (approved):
LGTM, and the failures are the same as on main so good to merge.

radical:
We do need the bdn update though.

radical:
As for the failures, I *think* it might need a fix like https://github.com/dotnet/runtime/pull/88755/commits/2e873dd69a2640330bda888ab9cd6582d5541952 . Essentially set `DOTNET_ROOT`. I'm testing that hypothesis in https://github.com/dotnet/runtime/pull/88755 .

radical:
@LoopedBard3 I only see a nightly nuget pushed - https://github.com/dotnet/BenchmarkDotNet/actions/runs/5532878163/jobs/10095576828

LoopedBard3:
> @LoopedBard3 I only see a nightly nuget pushed - https://github.com/dotnet/BenchmarkDotNet/actions/runs/5532878163/jobs/10095576828

Do you know the earliest version that we need? @cincuranet is working on getting a BDN update PR in: https://github.com/dotnet/performance/pull/3120

radical:
I think updating to the latest bdn will also run into https://github.com/dotnet/BenchmarkDotNet/issues/2346 .

radical:
The failures on windows are same as https://github.com/dotnet/runtime/issues/88779 . I'm fixing it for wasm with https://github.com/dotnet/runtime/pull/88751/commits/35477aab96f35e316e998759fe1498c89654111e . I think a similar fix is needed here too - to set `DOTNET_ROOT`. And the reason for that is https://github.com/dotnet/roslyn/pull/68918 changed
to expect to build the dotnet path using `DOTNET_ROOT` instead of depending `PATH`.

LoopedBard3:
@radical It does seem that the BDN update is causing us to hit https://github.com/dotnet/BenchmarkDotNet/issues/2346. As we wait for that fix, do we want to get this PR merged so we are closer to being in a good state once the fix is ready?

radical:
> @radical It does seem that the BDN update is causing us to hit [dotnet/BenchmarkDotNet#2346](https://github.com/dotnet/BenchmarkDotNet/issues/2346). As we wait for that fix, do we want to get this PR merged so we are closer to being in a good state once the fix is ready?

Do you mean merge this(#3122), and then continue to wait for BDN update via #3120?
Yeah, we can merge this. The "fix" won't take effect yet, but no need to keep this one open till the bdn update.

LoopedBard3:
> > @radical It does seem that the BDN update is causing us to hit [dotnet/BenchmarkDotNet#2346](https://github.com/dotnet/BenchmarkDotNet/issues/2346). As we wait for that fix, do we want to get this PR merged so we are closer to being in a good state once the fix is ready?
> 
> Do you mean merge this(#3122), and then continue to wait for BDN update via #3120? Yeah, we can merge this. The "fix" won't take effect yet, but no need to keep this one open till the bdn update.

Pretty much. Merge this #3122 with #3120 coming shortly while we wait for the fix to https://github.com/dotnet/BenchmarkDotNet/issues/2346. 

Let me know if you don't have merge permissions and I will merge this in.

radical:
ah, looks like I can't merge on Red.`Required statuses must pass before merging`

adamsitnik:
> Latest BDN copies the PackageReferences from
MicroBenchmarks.csproj to the generated project.

It was a bad idea and just got reverted: https://github.com/dotnet/BenchmarkDotNet/pull/2365

IMO once https://github.com/dotnet/performance/pull/3120 gets merged we should revert this PR too, as it won't be needed anymore and adds some non-trivial complexity to our MSBuild files that are already quite complex (cc @cincuranet )

@radical kudos for finding a quick fix!

cincuranet:
I'll take care of the revert.

PR #3126, opened 2023-07-12
Use new TrimMode for HelloWorld ILLink template

cincuranet:
As per https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options and https://learn.microsoft.com/en-us/dotnet/core/compatibility/deployment/7.0/deprecated-trimmer-default-action.


cincuranet:
Failures are on par with what's [currently happening in _main_](https://dev.azure.com/dnceng-public/public/_build/results?buildId=337077&view=results).

LoopedBard3 (approved):
LGTM

DrewScoggins:
In the linked docs, I didn't see a comparison between copy and partial. Are those the same thing?

cincuranet:
@DrewScoggins Yes. Based on [this](https://learn.microsoft.com/en-us/dotnet/core/compatibility/deployment/7.0/trim-all-assemblies#recommended-action).

PR #3139, opened 2023-07-14
Add MemoryRandomizationAttribue

DrewScoggins:
Initial PR for adding the MemoryRandomizationAttribute. This is currently applied to all tests allocate at least 1000B.


src/benchmarks/micro/Serializers/Binary_FromStream.cs
```diff
@@ -37,6 +37,7 @@ public void SetupBinaryFormatter()
 
         [BenchmarkCategory(Categories.Libraries)]
         [Benchmark(Description = nameof(BinaryFormatter))]
+        [MemoryRandomizationAttribute]
```
kunalspathak, line 40: Why is there `Attribute` word in this? Why is it not just `MemoryRandomization`?

DrewScoggins: I just mis-read the documentation. Forgot that you don't include the Attribute part of the name for attributes. Also fixed up the extra `#if` errors.

adamsitnik (commented):
> This is currently applied to all tests allocate at least 1000B.

It's a good start, but we should consider filtering it to benchmarks that:
- are multimodal in our Reporting System
- consume allocated memory. Example: just allocating and returning an array via `list.ToArray()` is not interesting, as allocated memory is not used, so whatever alignment it has, it does not matter.

And the most important question: how the Reporting System and the tooling for detecting regressions is going to handle that? IIRC the main problem was that it was always using a single number: average of all iterations to represent given run. If we keep using the average, it won't help us much (the numbers will most likely stay the same).

PR #3140, opened 2023-07-17
Don't run PDN scenarios on x86 systems 

LoopedBard3:
Don't run PDN scenarios on x86 systems as there is no portable.zip available for x86, only arm64 and x64. Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2223768&view=results

The error is occuring on the sdk_scenarios runs because we download PDN for all the scenarios, not just scenarios.proj, and the sdk_scenarios is the only one to run for x86. This also means the architecture condition on the PaintDotNet is not strictly needed but was added to ensure clarity if x86 was added for scenarios.proj in the future.

### DIFF
--- a/eng/performance/scenarios.proj
+++ b/eng/performance/scenarios.proj
@@ -47,7 +47,7 @@
       <ScenarioDirectoryName>wpfsfc</ScenarioDirectoryName>
       <PayloadDirectory>$(ScenariosDir)%(ScenarioDirectoryName)</PayloadDirectory>
     </UIScenario>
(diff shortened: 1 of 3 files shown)


DrewScoggins:
LGTM. Did you do a run on internal, since we don't use PDN on public runs?

DrewScoggins (approved):

LoopedBard3:
> LGTM. Did you do a run on internal, since we don't use PDN on public runs?

Yes, the test run is internal testing the sdk_scenarios.proj and and scenarios.proj since those are the two projects targeted by these changes, and all the jobs succeeded.

Issue #3141, opened 2023-07-17
First attempts at a command line tool for running Mono Scenarios locally


LoopedBard3:
https://github.com/dotnet/performance/pull/3142

PR #3142, opened 2023-07-17
Add basic local testing script that builds and runs benchmarks

LoopedBard3:
Finishes: https://github.com/dotnet/performance/issues/3141

This V1 version adds support for CoreRun, Mono Interpreter, and Mono JIT with a placeholder for MonoAOTLLVM. CoreRun is consistently successful on both Windows and Linux while the Mono types are only consistently successful on Linux. 

The script assumes that you have the requirements for building the target run-type already setup.

### DIFF
--- a/scripts/benchmarks_local.py
+++ b/scripts/benchmarks_local.py
@@ -62,7 +62,7 @@ def check_for_runtype_specified(parsed_args: Namespace, run_types_to_check: list
     return False
 
 # Builds libs and corerun by default
-def build_runtime_dependency(parsed_args: Namespace, repo_path: str, subset: str = "clr+libs+libs.tests", configuration: str = "Release", additional_args: list = []):    
+def build_runtime_dependency(parsed_args: Namespace, repo_path: str, subset: str = "clr+libs", configuration: str = "Release", additional_args: list = []):    
     # Run the command
     if parsed_args.os == "windows":
         build_libs_and_corerun_command = [
@@ -125,7 +125,10 @@ def generate_all_runtype_dependencies(parsed_args: Namespace, repo_path: str, co
         dest_dir_mono_jit = os.path.join(get_run_artifact_path(parsed_args, RunType.MonoJIT, commitish_information), "dotnet_mono")
 
         if parsed_args.rebuild_artifacts or not os.path.exists(dest_dir_mono_interpreter) or not os.path.exists(dest_dir_mono_jit):
-            build_runtime_dependency(parsed_args, repo_path, "clr+mono+libs")
+            build_runtime_dependency(parsed_args, repo_path)
+            shutil.move(os.path.join(repo_path, "artifacts"), os.path.join(repo_path, "artifacts-coreclr"))
+            build_runtime_dependency(parsed_args, repo_path, "mono+libs")
+            shutil.copytree(os.path.join(repo_path, "artifacts-coreclr"), os.path.join(repo_path, "artifacts"))
             build_runtime_dependency(parsed_args, repo_path, "libs.pretest", additional_args=['-testscope', 'innerloop', '/p:RuntimeFlavor=mono', f"/p:RuntimeArtifactsPath={os.path.join(repo_path, 'artifacts', 'bin', 'mono', f'{parsed_args.os}.{parsed_args.architecture}.Release')}"])
             generate_layout(parsed_args, repo_path)
             # Create the mono-dotnet
--- a/scripts/benchmarks_local.py
+++ b/scripts/benchmarks_local.py
@@ -32,7 +32,7 @@ import platform
 import shutil
 import sys
 import os
-
+import ctypes
 
 # Assumptions: We are only testing this Performance repo, should allow single run or multiple runs
 # For dotnet_version based runs, use the benchmarks_monthly .py script instead
@@ -48,6 +48,12 @@ start_time = datetime.now()
 
 def is_windows(parsed_args: Namespace):
     return parsed_args.os == "windows"
+
+def is_running_as_admin(parsed_args: Namespace):
+    if is_windows(parsed_args):
+        return ctypes.windll.shell32.IsUserAnAdmin()
+    else:
+        return os.getuid() == 0
 
 def kill_dotnet_processes(parsed_args: Namespace):
     if is_windows(parsed_args):
@@ -319,6 +325,7 @@ def add_arguments(parser):
     parser.add_argument('--rebuild-artifacts', action='store_true', help='Whether to rebuild the artifacts for the specified commits before benchmarking.')
     parser.add_argument('--build-only', action='store_true', help='Whether to only build the artifacts for the specified commits and not run the benchmarks.')
     parser.add_argument('--skip-local-rebuild', action='store_true', help='Whether to skip rebuilding the local repo and use the already built version (if already built). Useful if you need to run against local changes again.')
+    parser.add_argument('--allow-non-admin-execution', action='store_true', help='Whether to allow non-admin execution of the script. Admin execution is highly recommended as it minimizes the chance of encountering errors, but may not be possible in all cases.')
     def __is_valid_run_type(value):
         try:
             RunType[value]
@@ -344,11 +351,18 @@ def __main(args: list):
     
     setup_loggers(verbose=parsed_args.verbose)
 
+    # Ensure we are running as admin
+    if not is_running_as_admin(parsed_args):
+        if parsed_args.allow_non_admin_execution:
+            getLogger().warning("This script is not running as an administrator. This may cause errors.")
+        else:
+            raise Exception("This script must be run as an administrator or --allow-non-admin-execution must be passed.")
+
     # If list cached builds is specified, list the cached builds and exit
     if parsed_args.list_cached_builds:
         for folder in os.listdir(parsed_args.artifact_storage_path):
             if any([run_type.name in folder for run_type in RunType]):
-                print(folder)
+                getLogger().info(folder)
         return
 
     # Check to make sure we have something specified to test
@@ -372,7 +386,6 @@ def __main(args: list):
     try:
         getLogger().info("Killing any running dotnet, vstest, or msbuild processes... (ignore system cannot find path specified)")
         kill_dotnet_processes(parsed_args)
-        getLogger().info("****** MAKE SURE TO RUN AS ADMINISTRATOR ******")
 
         # Generate the artifacts for each of the remote versions
         if parsed_args.commits:
--- a/scripts/benchmarks_local.py
+++ b/scripts/benchmarks_local.py
@@ -118,7 +118,7 @@ def generate_layout(parsed_args: Namespace, repo_path: str, additional_args: lis
                 "generatelayoutonly",
                 "/p:LibrariesConfiguration=Release"
             ] + additional_args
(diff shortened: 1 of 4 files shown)


scripts/benchmarks_local.py
```diff
@@ -0,0 +1,404 @@
...
+            os.makedirs(dest_dirpath)
+        for src_filename in src_filenames:
+            if os.path.exists(os.path.join(dest_dirpath, src_filename)) and os.path.samefile(os.path.join(src_dirpath, src_filename), os.path.join(dest_dirpath, src_filename)):
+                continue
+            shutil.copy2(os.path.join(src_dirpath, src_filename), dest_dirpath)
+        
+# Builds libs and corerun by default
+def build_runtime_dependency(parsed_args: Namespace, repo_path: str, subset: str = "clr+libs", configuration: str = "Release", additional_args: list = []):    
+    # Run the command
+    if parsed_args.os == "windows":
```
cincuranet, line 84: Maybe a helper function like `is_windows`? Just in case in the future we need to service this logic.

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,404 @@
...
+        build_libs_and_corerun_command = [
+                "pwsh",
+                "-File",
+                f"build.ps1"
```
cincuranet, line 88: `f` not needed here.

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,404 @@
...
+    else:
+        build_script = "./build.sh"
+    generate_layout_command = [
+                f"{build_script}",
+                "release",
+                parsed_args.architecture,
+                "generatelayoutonly",
+                "/p:LibrariesConfiguration=Release"
+            ] + additional_args
+    RunCommand(generate_layout_command, verbose=True).run(os.path.join(repo_path, "src/tests"))
```
cincuranet, line 118: Do we care that on Windows this should be `src\tests`?

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,404 @@
...
+        dest_dir_mono_interpreter = os.path.join(get_run_artifact_path(parsed_args, RunType.MonoInterpreter, commit), "dotnet_mono")
+        dest_dir_mono_jit = os.path.join(get_run_artifact_path(parsed_args, RunType.MonoJIT, commit), "dotnet_mono")
+
+        if force_regenerate or not os.path.exists(dest_dir_mono_interpreter) or not os.path.exists(dest_dir_mono_jit):
+            build_runtime_dependency(parsed_args, repo_path, "clr+mono+libs")
+            build_runtime_dependency(parsed_args, repo_path, "libs.pretest", additional_args=['-testscope', 'innerloop', '/p:RuntimeFlavor=mono', f"/p:RuntimeArtifactsPath={os.path.join(repo_path, 'artifacts', 'bin', 'mono', f'{parsed_args.os}.{parsed_args.architecture}.Release')}"])
+
+            # Create the mono-dotnet
+            src_dir = os.path.join(repo_path, "artifacts", "bin", "runtime", f"net8.0-{parsed_args.os}-Release-{parsed_args.architecture}")
+            dest_dir = os.path.join(repo_path, "artifacts", "bin", "testhost", f"net8.0-{parsed_args.os}-Release-{parsed_args.architecture}", "shared", "Microsoft.NETCore.App", "8.0.0")
```
cincuranet, line 151: I think for all the `net8.0`/`8.0.0`/... (here and and all the below) we should create constant(s) for easy future changes.

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,404 @@
...
+    add_arguments(parser)
+    parsed_args = parser.parse_args(args)
+    
+    setup_loggers(verbose=parsed_args.verbose)
+
+    # If list cached builds is specified, list the cached builds and exit
+    if parsed_args.list_cached_builds:
+        for folder in os.listdir(parsed_args.artifact_storage_path):
+            if any([run_type.name in folder for run_type in RunType]):
+                print(folder)
```
cincuranet, line 348: Shouldn't this use _logger_?

cincuranet (changes requested):

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,451 @@
...
+def enum_name_list_to_enum_list(EnumType, enum_name_list: list):
+    return [enum_name_to_enum(EnumType, enum_name) for enum_name in enum_name_list]
+
+def check_for_runtype_specified(parsed_args: Namespace, run_types_to_check: list) -> bool:
+    for run_type in run_types_to_check:
+        if run_type.name in parsed_args.run_type_names:
+            return True
+    return False
+
+# Uses python copy to copy the contents of a directory to another directory while overwriting any existing files
```
DrewScoggins, line 82: nit: `copy to copy` -> `copy, to copy`

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,451 @@
...
+        for commit in parsed_args.commits:
+            repo_dirs.append(f"runtime-{commit.replace('/', '-')}")
+
+    try:
+        getLogger().info("Killing any running dotnet, vstest, or msbuild processes... (ignore system cannot find path specified)")
+        kill_dotnet_processes(parsed_args)
+
+        # Generate the artifacts for each of the remote versions
+        if parsed_args.commits:
+            getLogger().info(f"Checking if references {parsed_args.commits} exist in {repo_url}.")
```
DrewScoggins, line 423: Should this go on line 413 when we are actually checking if these references exist?

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,451 @@
...
+        repo = Repo.clone_from(repo_url, repo_combined_path)
+    else:
+        repo = Repo(repo_combined_path)
+        repo.remotes.origin.fetch()
+
+    # Check if each reference exists in the repository
+    for reference in references:
+        try:
+            result = repo.git.branch('-r', '--contains', reference) # Use git branch -r --contains <commit> to check if a commit is in a branch
+            if "error: malformed object" in result:
```
DrewScoggins, line 342: What are the two different types of error that we are trying to handle here? I see we raise the same exception, is it just the case we can get in this state from two different paths?

LoopedBard3: The "error: malformed object" is a leftover from when I was first figuring out the check for references. It will be removed in the next feedback commit.

cincuranet (dismissed):

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,449 @@
...
+        if parsed_args.local_test_repo:
+            generate_artifacts_for_commit(parsed_args, "local", parsed_args.local_test_repo, "local", True)
+
+        if not parsed_args.build_only:
+            # Run the benchmarks
+            commitsToRun = []
+            if parsed_args.commits:
+                commitsToRun = parsed_args.commits
+            if parsed_args.local_test_repo:
+                commitsToRun.append("local")
```
DrewScoggins, line 439: Should we make `"local"` a constant where we are using it?

DrewScoggins (approved):
LGTM

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,450 @@
+# This is a script for testing the performance of the different dotnet/runtime build types locally. It is assumed that prereqs for the build type are already installed.
+# Example usage from the performance/scripts folder: 
+# python .\benchmarks_local.py --local-test-repo "<absolute path to runtime folder>/runtime" --run-type MonoJIT --filter *Span.IndexerBench.CoveredIndex2*
+# or if you want remotes:
+# python .\benchmarks_local.py --commits dd079f53b95519c8398d8b0c6e796aaf7686b99a --repo-storage-dir "<absolute path to where you want to store runtime clones>" --run-types MonoInterpreter MonoJIT --filter *Span.IndexerBench.CoveredIndex2*
```
kotlarmilos, line 5: Use --repo-storage-path instead of --repo-storage-dir

kotlarmilos, line 5: Quotation marks are missing for *Span.IndexerBench.CoveredIndex2*

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,450 @@
...
+        except KeyError:
+            raise ArgumentTypeError(f"Invalid run type: {value}.")
+        return value
+    parser.add_argument('--run-types', dest='run_type_names', nargs='+', type=__is_valid_run_type, choices=[run_type.name for run_type in RunType], help='The types of runs to perform.')
+    parser.add_argument('--quiet', dest='verbose', action='store_false', help='Whether to not print verbose output.')
+    
+    # Arguments specifically for dependency generation and BDN
+    parser.add_argument('--bdn-arguments', type=str, default="", help='Command line arguments to be passed to BenchmarkDotNet, wrapped in quotes')
+    parser.add_argument('--architecture', choices=['x64', 'x86', 'arm64', 'arm'], default=get_machine_architecture(), help='Specifies the SDK processor architecture')
+    parser.add_argument('--os', choices=['windows', 'linux'], default=platform.system().lower(), help='Specifies the operating system of the system')
```
kotlarmilos, line 371: It would be good to add support for osx. Currently, it selects darwin which is incorrect os for the runtime.

kotlarmilos, line 371:
```suggestion
    parser.add_argument('--os', choices=['windows', 'linux', 'osx'], default=get_default_os(), help='Specifies the operating system of the system')
```

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,450 @@
...
+    parser.add_argument('--filter', type=str, default='*', help='Specifies the benchmark filter to pass to BenchmarkDotNet')
+    parser.add_argument('-f', '--framework', choices=ChannelMap.get_supported_frameworks(), default='net8.0', help='The target framework to run the benchmarks against.') # Can and should this accept multiple frameworks?
+    parser.add_argument('--csproj', type=str, default=os.path.join("..", "src", "benchmarks", "micro", "MicroBenchmarks.csproj"), help='The path to the csproj file to run benchmarks against.')    
+
```
kotlarmilos, line 375:
```suggestion

def get_default_os():
    system = platform.system().lower()
    if system == 'darwin':
        return 'osx'
    elif system in ['windows', 'linux']:
        return system
    else:
        raise Exception("Unsupported operating system: {system}.")

```

kotlarmilos:
Otherwise it works fine! Thanks for such great effort!

scripts/benchmarks_local.py
```diff
@@ -0,0 +1,459 @@
...
+            if os.path.exists(os.path.join(dest_dirpath, src_filename)) and os.path.samefile(os.path.join(src_dirpath, src_filename), os.path.join(dest_dirpath, src_filename)):
+                continue
+            shutil.copy2(os.path.join(src_dirpath, src_filename), dest_dirpath)
+        
+# Builds libs and corerun by default
+def build_runtime_dependency(parsed_args: Namespace, repo_path: str, subset: str = "clr+libs", configuration: str = "Release", additional_args: list = []):    
+    # Run the command
+    if is_windows(parsed_args):
+        build_libs_and_corerun_command = [
+                "pwsh",
```
cincuranet, line 99: Is _pwsh_ needed. Maybe people don't have it installed. Maybe regular _powershell_ would do.

cincuranet (approved):

PR #3145, opened 2023-07-18
[perf][nativeaot] Add Native AOT performance measurements for iOS

kotlarmilos:
This PR introduces Native AOT startup and SOD measurements for the Maui apps. It also includes the Xamarin performance scenarios. Once the PR is merged, the performance dashboard should be expanded with Mono and Native AOT measurements.


cincuranet (approved):

kotlarmilos:
Xamarin startup job failed due to incorrect package name in https://dev.azure.com/dnceng/internal/_build/results?buildId=2224727&view=results.

cincuranet (dismissed):

LoopedBard3 (approved):
This looks great!

cincuranet (approved):

kotlarmilos:
The perf tests have passed and a failure tracked in https://github.com/dotnet/performance/issues/3148 shouldn't be related.

src/scenarios/xamarinios/pre.py
```diff
@@ -0,0 +1,41 @@
...
+precommands.execute(['/p:_RequireCodeSigning=false', '/p:ApplicationId=net.dot.xamarintesting'])
+
+# Remove the aab files as we don't need them, this saves space
+output_dir = const.PUBDIR
+if precommands.output:
+    output_dir = precommands.output
+remove_aab_files(output_dir)
+
+# Copy the XamarinVersion to a file so we have it on the machine
+xamarin_version = get_version_from_dll_powershell_ios(rf"./{const.APPDIR}/obj/Release/{precommands.framework}/ios-arm64/ipa/Payload/{EXENAME}.app/Microsoft.iOS.dll")
```
ivanpovazan, line 38: Is this done for NativeAOT as well? 
If yes, it could become a problem as we do not ship assemblies with NativeAOT any more.

kotlarmilos: Yes. Also, it is done for Maui scenarios in https://github.com/dotnet/performance/blob/89c09452ebcff0c483ea6d29e69ba12dccbe2b45/src/scenarios/mauiios/pre.py#L38C8-L38C8.

My assumption is that the version is presented on the dashboards. /cc: @LoopedBard3

LoopedBard3: I am not aware of any other places where the version may be grabbed from, but I also have not taken a deep look. This is also being grabbed from the obj folder just after the build, so if you mean that assemblies are not put into the publish results we use for testing but still show up in the intermediary steps we should be good.

kotlarmilos: Currently we retrieve the version from the dll in the app bundle. If the Native AOT doesn't ship assemblies anymore, as you suggested we could retrieve the version from an intermediate directory, such as `bin/Release/net8.0-ios/ios-arm64/Microsoft.iOS.dll`.

kotlarmilos: The code is updated to retrieve the version from the intermediate directory `/obj/Release/net8.0-ios/ios-arm64/linked/Microsoft.iOS.dll`. The scenarios are tested locally and the PR is ready for to be reviewed/merged.

cincuranet (approved):

kotlarmilos:
The tests have passed in https://dev.azure.com/dnceng/internal/_build/results?buildId=2226634&view=logs&j=db08f39a-0f85-5b5b-e0d1-f245bb65a1ac&t=886fade1-3a4b-56ee-b0da-81de8c32b899.

cincuranet:
After talking with @kotlarmilos, the failures should be unrelated, merging.

kotlarmilos:
Thanks. If it starts happening on the CI, we will act accordingly.

src/scenarios/mauiblazorios/pre.py
```diff
@@ -72,8 +72,8 @@
 remove_aab_files(output_dir)
 
 # Copy the MauiVersion to a file so we have it on the machine
-maui_version = get_version_from_dll_powershell_ios(rf"./{const.APPDIR}/obj/Release/{precommands.framework}/ios-arm64/ipa/Payload/{EXENAME}.app/Microsoft.Maui.dll")
+maui_version = get_version_from_dll_powershell_ios(rf"./{const.APPDIR}/obj/Release/net8.0-ios/ios-arm64/linked/Microsoft.Maui.dll")
```
ivanpovazan, line 75: nit: I know this already got merged in, so for future reference it might be better to use `{precommands.framework}` instead of explicit `net8.0-ios` tfm

kotlarmilos: Yes we missed that, sorry. Follow-up PR in https://github.com/dotnet/performance/pull/3154.

PR #3146, opened 2023-07-19
New TimeToMain2 measurement for LTTng on Ubuntu 22.04 workaround (future work).

cincuranet:
This switches _scenarios_ to use our own "OnMain" from using event from runtime to get "Time to Main". We can't use the runtime one, because currently we'll not be able to capture it on Ubuntu 22.04 and LTTng. 

The workaround for LTTng and 22.04 is future PR.

### DIFF
--- a/src/scenarios/shared/precommands.py
+++ b/src/scenarios/shared/precommands.py
@@ -243,7 +243,6 @@ class PreCommands:
             trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.OnMain()"
         elif language_file_extension == 'fs':
             trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.OnMain()"
-        #print(f"{' ' * 10}Hello world")
         else:
             raise Exception(f"{language_file_extension} not supported.")
         self.add_event_source(file, line, trace_statement, language_file_extension)
--- a/src/scenarios/shared/precommands.py
+++ b/src/scenarios/shared/precommands.py
@@ -225,10 +225,30 @@ class PreCommands:
             startup_args += self.crossgen_arguments.get_crossgen2_command_line()
             RunCommand(startup_args, verbose=True).run(self.crossgen_arguments.coreroot)
 
-    def add_startup_logging(self, file: str, line: str):
-        self.add_event_source(file, line, "PerfLabGenericEventSource.Log.Startup();")
+    def add_startup_logging(self, file: str, line: str, language_file_extension: str = 'cs', indent: int = 0):
+        if language_file_extension == 'cs':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.Startup();"
+        elif language_file_extension == 'vb':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.Startup()"
+        elif language_file_extension == 'fs':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.Startup()"
+        else:
+            raise Exception(f"{language_file_extension} not supported.")
+        self.add_event_source(file, line, trace_statement, language_file_extension)
 
-    def add_event_source(self, file: str, line: str, trace_statement: str):
+    def add_onmain_logging(self, file: str, line: str, language_file_extension: str = 'cs', indent: int = 0):
+        if language_file_extension == 'cs':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.OnMain();"
+        elif language_file_extension == 'vb':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.OnMain()"
+        elif language_file_extension == 'fs':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.OnMain()"
+        #print(f"{' ' * 10}Hello world")
+        else:
+            raise Exception(f"{language_file_extension} not supported.")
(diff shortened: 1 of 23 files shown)


cincuranet:
Failures (around RC1) are unrelated.

src/scenarios/shared/precommands.py
```diff
@@ -225,28 +225,48 @@ def execute(self, build_args: list = []):
...
+        self.add_event_source(file, line, trace_statement, language_file_extension)
+
+    def add_onmain_logging(self, file: str, line: str, language_file_extension: str = 'cs', indent: int = 0):
+        if language_file_extension == 'cs':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.OnMain();"
+        elif language_file_extension == 'vb':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.OnMain()"
+        elif language_file_extension == 'fs':
+            trace_statement = f"{' ' * indent}PerfLabGenericEventSource.Log.OnMain()"
+        #print(f"{' ' * 10}Hello world")
```
DrewScoggins, line 246: nit: Remove comment

src/tools/ScenarioMeasurement/Startup/TimeToMain2Parser.cs
```diff
@@ -0,0 +1,108 @@
...
+            source.Kernel.ProcessStart += evt =>
+            {
+                if (!pid.HasValue && ParserUtility.MatchProcessStart(evt, source, processName, pids, commandLine))
+                {
+                    pid = evt.ProcessID;
+                    start = evt.TimeStampRelativeMSec;
+                }
+            };
+
+            source.Source.Kernel.ThreadCSwitch += evt =>
```
DrewScoggins, line 45: Do we know get CSwitch events on Linux? Otherewise, shouldn't this go inside the `IsWindows` check?

src/tools/ScenarioMeasurement/Startup/TimeToMain2Parser.cs
```diff
@@ -0,0 +1,108 @@
...
+                        threadTime = 0;
+                        start = 0;
+                    }
+                });
+            }
+            else
+            {
+                source.Source.Clr.EventSourceEvent += evt =>
+                {
+                    if (pid.HasValue && evt.ProcessID == pid && evt.EventID == OnMainEventId)
```
DrewScoggins, line 85: Just for my own understanding. This `EventSourceEvent` is only activated if you register your own events with `EventSource`, and then enable them and they fire. In that way we know that only our events will show up here and there will not be any overlap of `EventID` with another event. Is that the correct understanding? If so should we add a comment that explains that all of our custom events need to have unique IDs across all events?

src/scenarios/staticfsconsoletemplate/pre.py
```diff
@@ -11,4 +11,5 @@
 setup_loggers(True)
 precommands = PreCommands()
 precommands.existing(os.path.join(sys.path[0], const.SRCDIR), ('%s.fsproj' % EXENAME))
+precommands.add_onmain_logging("Program.fs", "let main _ =", language_file_extension="fs", indent=4)
```
DrewScoggins, line 14: Do we need to update the test.py for this test to use the new parser as well?

src/scenarios/shared/precommands.py
cincuranet: Fixed. Thanks.

src/tools/ScenarioMeasurement/Startup/TimeToMain2Parser.cs
cincuranet: We don't. But it's also doing nothing on Linux, so I left it without check, just in case it starts to work in the future.

src/tools/ScenarioMeasurement/Startup/TimeToMain2Parser.cs
cincuranet: We're setting the `COMPlus_EventSourceFilter`.

DrewScoggins: OK, so we can only ever use one at a time anyone. I understand now.

src/tools/ScenarioMeasurement/Startup/TimeToMain2Parser.cs
```diff
@@ -0,0 +1,108 @@
...
+                };
+            }
+
+            source.Process();
+        }
+
+        var result = new List<Counter>();
+        result.Add(new Counter() { Name = "Time To Main", MetricName = "ms", DefaultCounter = true, TopCounter = true, Results = results.ToArray() });
+        // below is supported only on Windows, other platform will have 0s
+        if (!threadTimes.All(x => x == 0))
```
DrewScoggins, line 102: So this change moves it from, we can have no value be zero, to, we can't have all values be zero. Do we expect that some values being zero and others not a scenario we would expect and want to save the data from? It feels like if that happened than something went wrong.

src/tools/ScenarioMeasurement/Startup/TimeToMain2Parser.cs
cincuranet: I've added also `MagicConstant`, to "prevent" collisions.

LoopedBard3 (approved):

DrewScoggins (approved):

PR #3154, opened 2023-07-21
[perf][ios] Use precommands.framework var instead of explicit framework

kotlarmilos:
This PR updates iOS scenarios to use `{precommands.framework}` var instead of explicit framework.


ivanpovazan (approved):
Thanks!

cincuranet (approved):

Issue #3155, opened 2023-07-21
Only one PerfLab.cs across repository


cincuranet:

DrewScoggins:
I am a little confused. I only see one copy of Perflab.cs checked in?

cincuranet:
The content, though, is in multiple files. Difficult to maintain. I want to have one source of truth.

DrewScoggins:
Ok, so removing the definition of this type from each individual parser and moving it to this one location?

cincuranet:
Not parser(s), but scenario files.

PR #3156, opened 2023-07-21
Extract copying of PerfLab.cs into separate method, so it can be called directly without add_event_source. Preparation for #3155.

cincuranet:


LoopedBard3 (approved):
Looks good to me.

PR #3163, opened 2023-07-25
Fix Android bug and Maui builds

LoopedBard3:
Fix bug introduced when adding power consumption capabilities due to += and append a string on string function differently. We are also hitting signing errors when installing Maui nightly workload as they are not signed, added --skip-sign-check to workaround this.


LoopedBard3:
Test run with successful android run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2229310&view=results

cincuranet (approved):

LoopedBard3:
Both android tests succeeded and workload installed for ios. There was an unrelated ios startup failure but merging to get the fixes in.

PR #3165, opened 2023-07-26
[wasm] Replace hash with VERSION.

ilonatommy:
All dotnet versions used to be squashed into the same name, e.g. `dotnet.VERSION.js.br`. This PR changes it to be `dotnet.8.0.0-VERSION.js.br`.


cincuranet:
Are the match group for `path`, `jsExtension`, `dotnet`, etc. really needed?

Nit: `hash` isn't really a good name, given it's not matching only the hash, but the whole version part.

PR #3167, opened 2023-07-26
Add `-bl:` argument to the restore command line

radical:
This utilizes the `binlog` path passed in, and appends `-restore.binlog`
to the string, which could be a full path. So the filename can be like
`foo.binlog-restore.binlog`.

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -294,7 +294,8 @@ class CSharpProject:
     def restore(self, 
                 packages_path: str, 
                 verbose: bool,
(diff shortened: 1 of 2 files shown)


LoopedBard3 (approved):
LGTM, let's let the verification pipeline run and then merge if there are no failures due to this change.

radical:
The failure for `ubuntu 1804 x64 micro Open nativeaot8.0` is same as that on `main`.

radical:
I can't merge on red.

DrewScoggins (approved):
LGTM

Issue #3164, opened 2023-07-26
[perf] Dotnet restore unable to find 8.0.0-rc.* packages


kotlarmilos:
## Description

The perf jobs failed to restore `MicroBenchmarks.csproj`.

```
+ python3 /home/helixbot/work/AAE30989/w/C7D30A73/e/performance/scripts/benchmarks_ci.py --csproj /home/helixbot/work/AAE30989/w/C7D30A73/e/performance/src/benchmarks/micro/MicroBenchmarks.csproj --incremental no --architecture x64 -f net8.0 --upload-to-perflab-container --dotnet-versions 8.0.100-rc.1.23375.8 --cli-source-info args --cli-branch refs/heads/main --cli-commit-sha 764f77412ef4ec1a7e5dc521608247487a66cd7e --cli-repository https://github.com/dotnet/runtime --cli-source-timestamp 2023-07-25T19:04:35Z --bdn-artifacts /home/helixbot/work/AAE30989/w/C7D30A73/e/artifacts/BenchmarkDotNet.Artifacts --bdn-arguments=--anyCategories Libraries Runtime  --category-exclusion-filter NoAOT NoWASM --runtimes monoaotllvm --aotcompilerpath /home/helixbot/work/AAE30989/p/monoaot/sgen/mini/mono-sgen --customruntimepack /home/helixbot/work/AAE30989/p/monoaot/pack --aotcompilermode llvm --logBuildOutput --generateBinLog  --partition-count 30 --partition-index 1
----------------------------------------------
Initializing logger 2023-07-25 17:20:58.395098
----------------------------------------------
Installing tools.
----------------------
Downloading DotNet Cli
----------------------
DotNet Install Path: '/home/helixbot/work/AAE30989/w/C7D30A73/e/performance/tools/dotnet/x64'
Downloading https://dot.net/v1/dotnet-install.sh
Downloaded https://dot.net/v1/dotnet-install.sh OK
$ pushd "/home/helixbot/work/AAE30989/w/C7D30A73/e/performance"
$ /home/helixbot/work/AAE30989/w/C7D30A73/e/performance/tools/dotnet/x64/dotnet-install.sh -InstallDir /home/helixbot/work/AAE30989/w/C7D30A73/e/performance/tools/dotnet/x64 -Architecture x64 -Version 8.0.100-rc.1.23375.8
[...]

PR #3168, opened 2023-07-27
Perf: Workaround a nuget restore issue for MicroBenchmarks.csproj

radical:
```
error NU1102: Unable to find package System.Text.Json with version (>= 8.0.0-rc.1.23375.7)
error NU1102:   - Found 1290 version(s) in dotnet8 [ Nearest version: 8.0.0-rc.1.23375.4 ]
error NU1102: Unable to find package System.Security.Cryptography.Xml with version (>= 8.0.0-rc.1.23375.7)
error NU1102:   - Found 1290 version(s) in dotnet8 [ Nearest version: 8.0.0-rc.1.23375.4 ]
```

Note the difference between the referenced version, and the one
available in the feed.

These packages are getting added as transitive references.

The hypothesis is that the direct package references have version string
as `8.0.*-*` which select the latest available version.
    - but those packages transitively reference others for whom the
      latest version might not have been published to nuget yet.

To work around this, we add explicit references to such packages, so
they can get the versions selected from the ones in the feed.

Issue: https://github.com/dotnet/performance/issues/3164

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -87,7 +87,7 @@
   <!--
     Workaround for https://github.com/dotnet/performance/issues/3164
   -->
-  <ItemGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '8.0'))">
+  <ItemGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '8.0'))">
     <PackageReference Include="System.Text.Json" Version="$(SystemVersion)" />
     <PackageReference Include="System.Security.Cryptography.Xml" Version="$(SystemVersion)" />
     <PackageReference Include="System.Text.Encodings.Web" Version="$(SystemVersion)" />
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -84,6 +84,15 @@
     <PackageReference Include="System.Collections.Immutable" Version="8.0.*-*" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '8.0')" />
   </ItemGroup>
 
+  <!--
+    Workaround for https://github.com/dotnet/performance/issues/3164
+  -->
+  <ItemGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '8.0'))">
+    <PackageReference Include="System.Text.Json" Version="$(SystemVersion)" />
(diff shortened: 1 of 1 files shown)


lewing (approved):

lewing:
I'm in favor of seeing if this resolves the restore problem and gets things building again, it has some unfortunate issues around what will actually get benchmarked but those already exist with other wildcards currently being used

radical:
Testing this here - https://dev.azure.com/dnceng/internal/_build/results?buildId=2230266 .

radical:
Though something to keep in mind is that the issue doesn't show up on every run, so if we don't want to merge, then we might have to run this a few times.

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -84,6 +84,15 @@
     <PackageReference Include="System.Collections.Immutable" Version="8.0.*-*" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' &lt; '8.0')" />
   </ItemGroup>
 
+  <!--
+    Workaround for https://github.com/dotnet/performance/issues/3164
+  -->
+  <ItemGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '8.0'))">
```
cincuranet, line 90: This fails with `error MSB4184: The expression "[MSBuild]::VersionGreaterThanOrEquals('', 8.0)" cannot be evaluated. Version string was not in a correct format.`. Better to use `$(_TargetFrameworkVersionWithoutV)` (used couple times above).

cincuranet (changes requested):

src/benchmarks/micro/MicroBenchmarks.csproj
radical: thanks!

ericstj (changes requested):
What's the theory behind why this would even work?  If it fixes anything at all it means we'd be testing in a torn configuration.  I don't think we want to do that.  Instead we'd just want to retry and hope that more packages were pushed.

cincuranet:
I'm working on the Maestro version.

lewing:
> What's the theory behind why this would even work? If it fixes anything at all it means we'd be testing in a torn configuration. I don't think we want to do that. Instead we'd just want to retry and hope that more packages were pushed.

As I mentioned in my comment it, my view is we're already potentially testing a torn configuration for any of the wildcard packs that are listed explicitly (there are a lot).  I agree we need a better long term solution but I'm also extremely frustrated that I'm getting zero perf data as we approach rc1.

LoopedBard3:
> I agree we need a better long term solution but I'm also extremely frustrated that I'm getting zero perf data as we approach rc1.

Could you give some specifics for either what data is missing or where the data is missing? I want to make sure that there is not another issue occuring that is being overshadowed by this issue. Thanks!

lewing:
https://github.com/dotnet/performance/issues/3181
https://github.com/dotnet/performance/issues/3180

are a couple

PR #3170, opened 2023-07-27
 EventListener that listens for the events we care about and then emits them via own LTTng tracepoint provider that works against the updated ABI

cincuranet:


src/tools/ScenarioMeasurement/Startup/TimeToMain2Parser.cs
```diff
@@ -1,26 +1,29 @@
...
     public void EnableUserProviders(ITraceSession user)
     {
-        user.EnableUserProvider(EventSourceName, TraceEventLevel.Verbose);
+        user.EnableUserProvider(PerfLabValues.EventSourceName, TraceEventLevel.Verbose);
+        if (user is LinuxTraceSession)
+        {
+            user.AddRawEvent($"{PerfLabValues.LTTngProviderName}:{PerfLabValues.OnMainEventName}");
+        }
+        var baseDir = Environment.GetEnvironmentVariable("HELIX_CORRELATION_PAYLOAD") ?? Path.GetFullPath("..");
+        Startup.AddTestProcessEnvironmentVariable("DOTNET_STARTUP_HOOKS", Path.Join(baseDir, PerfLabValues.ForwarderName, $"{PerfLabValues.ForwarderName}.dll"));
```
brianrob, line 26: I think `DOTNET_STARTUP_HOOKS` should only be set on Linux.

src/tools/ScenarioMeasurement/Startup/TraceEventParsers/PerfLabGenericEventSourceLTTngProviderParser.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        protected override Delegate Target
+        {
+            get => _target;
+            set => _target = (Action<TSelf>)value;
+        }
+
+        public override string[] PayloadNames => Array.Empty<string>();
+
+        public override object PayloadValue(int index) => throw new NotSupportedException();
```
brianrob, line 91: Recommend that you return `null` here instead.  This will ensure that the behavior here matches what is expected, and since you don't have any event arguments, it should be safe to do so.

src/tools/PerfLabGenericEventSourceLTTngProvider/build.sh
```diff
@@ -0,0 +1,30 @@
...
+            ;;
+    esac
+done
+
+pushd $(dirname $(realpath $0)) > /dev/null
+
+# install liblttng-ust-dev
+
+gcc -c -I. -fpic PerfLabGenericEventSourceLTTngProvider.c && \
+gcc -o $OUTPUT/PerfLabGenericEventSourceLTTngProvider.so -shared PerfLabGenericEventSourceLTTngProvider.o -llttng-ust
```
brianrob, line 24: Do you know what the default optimization level is for this compilation?  I did a bit of searching and didn't find a clear answer.  Mostly just want to make sure that we're not trying to run "debug" codegen unintentionally.

brianrob (commented):
This looks really great!  A few small things.

src/tools/ScenarioMeasurement/Startup/TimeToMain2Parser.cs
cincuranet: As we discussed offline, I want to keep the Linux and Windows versions close to each other, so even on Windows the startup hook is used, but only on Linux it is emitting the LTTng events ([here](https://github.com/dotnet/performance/pull/3170/files#diff-87862d1da7576ce7a9dbc519aadcdfad0ec8626cc81e66c30c5bb5ada317441aR20)).

brianrob: Yup, makes total sense to me.

src/tools/ScenarioMeasurement/Startup/TraceEventParsers/PerfLabGenericEventSourceLTTngProviderParser.cs
cincuranet: Done.

src/tools/PerfLabGenericEventSourceLTTngProvider/build.sh
cincuranet: Good catch! I added `O2` optimization level.

brianrob (approved):
LGTM!

LoopedBard3 (approved):
Looks good to me! For my knowledge, are these changes backwards compatible with say Ubuntu 20.04?

cincuranet:
> are these changes backwards compatible with say Ubuntu 20.04

Not compatible. Ubuntu 20.04 has `liblttng-ust0`, which is the previous version (which would work out of the box).

PR #3202, opened 2023-07-31
Increase Sorting invocations per iteration

AndyAyersMS:
Increase by 10x, since the benchmark currently runs each iteration
in 10ms or so, and we'd prefer 100ms.


AndyAyersMS:
Per https://github.com/dotnet/runtime/issues/87194#issuecomment-1656186354

@adamsitnik PTAL

adamsitnik (approved):
LGTM, thank you for identifying and applying the fix @AndyAyersMS !

adamsitnik:
the build failure is unrelated (NativeAOT compilation failed with OOM), the PR can be merged, but I don't have the permissions to merge on red anymore

PR #3203, opened 2023-07-31
Avoid benchmarking cached IPAddress ToString

MihaZupan:
https://github.com/dotnet/runtime/issues/88671#issuecomment-1658081776

Alternatively, we could delete this one given that we already have benchmarks covering `ctor` and `TryFormat`, which are almost the same.


cincuranet (approved):

cincuranet:
LGTM. Thanks.

PR #3222, opened 2023-08-03
Revert https://github.com/dotnet/performance/commit/860d6a17d0c17592e2f4d803038d99969f3d8778

LoopedBard3:
Reverts: https://github.com/dotnet/performance/pull/3179 (https://github.com/dotnet/performance/commit/860d6a17d0c17592e2f4d803038d99969f3d8778) as the tool build is still broken on the iOS builds. 

Test of proper building of tools (each scenario made it past building tools): https://dev.azure.com/dnceng/internal/_build/results?buildId=2235822&view=results

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -277,8 +277,8 @@ jobs:
       buildMachines:
         - win-x64
         - ubuntu-x64
-        - win-arm64
-        - ubuntu-arm64-ampere
+        #- win-arm64
+        #- ubuntu-arm64-ampere
       isPublic: false
       jobParameters:
         kind: scenarios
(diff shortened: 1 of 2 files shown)


DrewScoggins:
How did we miss that this didn't fix it before?

LoopedBard3:
The runs I tested for this before was only on window and ubuntu missing the mac version. The same removal also went into the runtime repo without causing any issues, so there is something (likely a version somewhere) that still has the breaking change on OSX.

LoopedBard3:
We also don't run OSX steps on public-ci because we only have on-prem testing hardware.

DrewScoggins:
OK, that makes sense.

DrewScoggins (approved):

PR #3230, opened 2023-08-04
Crank runner for Helix

caaavik-msft:
This PR contains initial work around making it so that devs inside Microsoft can use crank to schedule performance runs on Helix machines locally, rather than having to do them through CI pipelines only.

Currently, a lot of the logic around running performance tests are stored in a mix of Azure Pipelines yaml files, shell scripts, powershell scripts, python scripts, MSBuild project files across both the dotnet/runtime and dotnet/performance repositories. This PR attempts to move as much of that logic as possible into python scripts so that they can be run outside the context of a CI pipeline, and eventually we can update our CI pipelines to run these python scripts instead.

I also took this opportunity to fix up our python scripts to have proper type hints set, as well as configuring some settings so that we get good autocompletion and linting inside VSCode.

I have set up a crank agent that runs inside corpnet, which those inside Microsoft on a corpnet-connected machine will be able to use. For Microsoft employees working remotely, they can either use a DevBox (which has access to corpnet), or they can talk to a member of the .NET Performance Team to get an Azure Relay connection string which will allow communication outside corpnet.

An example of using crank to schedule a performance run on a remote machine is:
```
crank --config .\helix.yml --scenario micro --profile win-x64 --variable bdnArgs="--filter *Linq*" --profile msft-internal --variable buildNumber="caaavik-20230804.2"
```
This will run all the Linq microbenchmarks on a win-x64 helix machine against a local pre-built corerun. The results will be returned back and displayed in the terminal, it may also be uploaded to the Azure Data Explorer to view there.


caaavik-msft:
Some things that still need fixing before this can be merged:
- I need to add some documentation on how to use this
- There are a few bugs left with running scenarios I need to fix
- I also don't have support just yet for getting the scenarios to send the results back to crank, they will be uploaded to ADX though.
- I had to add an exception for `.git` in the `.gitignore` otherwise it would not get uploaded to the crank agent where we run git commands

scripts/ci_setup.py
```diff
@@ -315,28 +347,37 @@ def __main(args: list) -> int:
...
-                source_timestamp = parsed_timestamp.strftime('%Y-%m-%dT%H:%M:%SZ')
-            except ValueError:
-                getLogger().warning('Invalid commit_time format. Please use YYYY-MM-DD HH:MM:SS +/-HHMM. Attempting to get commit time from api.github.com.')
-                source_timestamp = dotnet.get_commit_date(target_framework_moniker, commit_sha, repo_url)
-        else:
-            source_timestamp = dotnet.get_commit_date(target_framework_moniker, commit_sha, repo_url)
+        # TODO: Fix this
+        source_timestamp = datetime.datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
+
+        # if(args.commit_time is not None):
```
caaavik-msft, line 361: Reminder to uncomment this and figure out what to do for commit time for local runs.

eng/performance/benchmark_jobs.yml
```diff
@@ -123,7 +123,7 @@ jobs:
               ${{ channel }}:
                 _Channel: ${{ channel }}
                 _Configs: CompilationMode=Tiered RunKind="${{ parameters.kind }}"
-                _PerfLabArguments: $(PerfLabArguments)
+                _PerfLabArguments: $(PerfLabArguments) # not necessary?
```
LoopedBard3, line 126: It looks like we do still use this: https://github.com/dotnet/performance/blob/main/eng/performance/benchmark_jobs.yml#L99

eng/performance/helix_sender.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.10-bullseye
```
cincuranet, line 1: I would vote for `3.10-bookworm` (or maybe better `3.10-slim-bookworm`), to be on current Debian 12.

scripts/run_performance_job.py
```diff
@@ -0,0 +1,517 @@
...
+                "export PYTHONPATH=",
+                "python3 -m pip install -U pip",
+                "pip3 install --user urllib3==1.26.15 --force-reinstall",
+                "pip3 install --user azure.storage.blob==12.7.1 --force-reinstall",
+                "pip3 install --user azure.storage.queue==12.1.5 --force-reinstall",
+                f'export PERFLAB_UPLOAD_TOKEN="{args.perflab_upload_token}"'
+            ]
+        else:
+            if args.runtime_type == "wasm":
+                wasm_precommand = (
```
cincuranet, line 182: This needs to be synced with https://github.com/dotnet/runtime/pull/90093 and https://github.com/dotnet/runtime/pull/90113.

eng/performance/benchmark_jobs.yml
```diff
@@ -123,11 +123,20 @@ jobs:
...
```
LoopedBard3, line 126: I am not finding _PerfLabArguments anywhere else so I think the rename got all the updates. We can probably remove this.

eng/performance/benchmark_jobs.yml
```diff
@@ -123,11 +123,20 @@ jobs:
...
                 _BuildConfig: ${{ parameters.architecture }}_$(_Channel)_${{ parameters.kind }} # needs to be unique to avoid logs overwriting in mc.dot.net
         steps:
         - checkout: self
           clean: true
+        - task: UsePythonVersion@0
+          inputs:
+            versionSpec: '3.10'
```
LoopedBard3, line 133: Since we install python in at least 2 places (here and scnearios.yml), it may be beneficial to see if we can pull the version into a single variable or version in a file so we only have to update it in one place.

scripts/performance_setup.py
```diff
@@ -0,0 +1,362 @@
...
+    if args.branch is not None:
+        setup_arguments.branch = args.branch
+
+    if args.commit_sha is not None:
+        setup_arguments.commit_sha = args.commit_sha
+
+    if not args.internal:
+        setup_arguments.not_in_lab = True
+
+    # TODO: Figure out if this should be the runtime or performance commit time, or if we need to capture both
```
LoopedBard3, line 239: I think we should either be just the runtime commit time, or both the runtime and performance times. I prefer to have both, even if the performance repo commit time gets added as additional data.

eng/performance/benchmark_jobs.yml
caaavik-msft: I've changed it now so that we just use python 3.6 actually, but dataclasses are supported by installing the dataclasses pip package

scripts/performance_setup.py
caaavik-msft: At the moment it's only practical to get the performance repo commit time, as only the core root folder is uploaded to the crank agent. To get the runtime commit information we may need to look at sending up the `.git` directory somehow to crank. I'll leave this TODO in here since I don't think I will be able to resolve it just yet.

eng/performance/helix_sender.Dockerfile
caaavik-msft: I tested out slim-bookworm but it was causing too many issues with other dependencies such as wget so just using bookworm now.

helix.yml
```diff
@@ -0,0 +1,157 @@
+jobs:
+  run_performance_job:
+    isConsoleApp: true
+    waitForExit: true
+    framework: net7.0
```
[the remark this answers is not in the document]
caaavik-msft: I have removed this setting now, it's only used if crank builds a .NET program which we aren't doing inside crank.

scripts/run_performance_job.py
caaavik-msft: Done

eng/performance/benchmark_jobs.yml
caaavik-msft: Removed

scripts/performance_setup.py
LoopedBard3: Sounds good to me.

eng/performance/benchmark_jobs.yml
LoopedBard3: Is dataclasses something we should add to the requirements.txt file?

eng/performance/helix.proj
```diff
@@ -1,63 +1,145 @@
 <Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">
 
-  <PropertyGroup Condition="'$(TargetsWindows)' == 'true'">
+  <PropertyGroup Condition="'$(TargetsWindows)' == 'true' AND '$(UseCoreRun)' == 'true'">
+    <PerformanceDirectory>%HELIX_WORKITEM_ROOT%\performance</PerformanceDirectory>
+    <HelixPreCommands>$(HelixPreCommands) &amp;&amp; robocopy /np /nfl /e %HELIX_CORRELATION_PAYLOAD%\performance $(PerformanceDirectory) /XD %HELIX_CORRELATION_PAYLOAD%\performance\.git</HelixPreCommands>
```
DrewScoggins, line 5: It seems like we are setting the HelixPreCommands here and then resetting them a few lines down. Am I missing something here?

scripts/dotnet.py
```diff
@@ -919,25 +916,11 @@ def add_arguments(parser: ArgumentParser) -> ArgumentParser:
     '''
     Adds new arguments to the specified ArgumentParser object.
     '''
-
     parser = __add_arguments(parser)
-
-    # .NET Compilation modes.
```
DrewScoggins, line 925: Do we just no longer need this argument?

eng/performance/helix.proj
caaavik-msft: This is with UseCoreRun == true set which is true for runs against a local runtime build, the reason for this is that in the runtime repo we set the precommands in the helix.proj, but in the performance repo we set them in the yaml file. So this was a way to merge the two .proj files. In the future I think consolidating it so that more of this is done inside the python scripts would be good.

scripts/dotnet.py
caaavik-msft: Yep, I checked and this argument was not used anywhere

scripts/performance_setup.py
DrewScoggins: Sorry to unresolve, but I am confused here. Shouldn't we be getting the commit time from calling ci_setup.py? Specifically, we call `dotnet.get_commit_time` and this properly gets the commit based on what repo we are running against.

scripts/run_performance_job.py
```diff
@@ -0,0 +1,518 @@
+from dataclasses import dataclass, field
```
DrewScoggins, line 1: Not sure exactly where to put this comment, but do we actually support running against WASM as this currently stands? I am having a hard time understanding where we package up the various runtime packs and such. Is that something that just needs to be handled by the user?

scripts/performance_setup.py
DrewScoggins: Oh, I see I missed that we don't follow that code path here because we are running against something local. It does feel like this is something we could grab locally when we send the job off to crank, but that feels outside the scope for sure.

scripts/run_performance_job.py
caaavik-msft: no it doesn't support wasm right now, I have a TODO for it in the code but I need to understand more about what all the inputs are and where it downloads everything

caaavik-msft: I know there are some files in the runtime repo that are used in wasm runs, and so to support that we would need to add additional sources to the helix.yml which ensures those directories are sent to crank

DrewScoggins (approved):
LGTM. Obviously there is a lot here, and I would be lying if I said that I now fully understand all of the intricacies. Overall, the change looks good though and I am confident that this shouldn't cause any breakage in our existing data pipelines.

docs/crank-to-helix-workflow.md
```diff
@@ -0,0 +1,105 @@
...
+
+- [Helix](https://github.com/dotnet/arcade/blob/main/Documentation/Helix.md) is the work scheduler that we use in our CI pipelines to run performance tests. We have many Helix queues available to us which provide different hardware capabilities so that we are able to test a wide array of situations.
+
+- [Crank](https://github.com/dotnet/crank) is a tool that provides infrastructure for software performance measurement and is mainly used today to support our [TechEmpower Web Framework Benchmarks](https://github.com/aspnet/benchmarks).
+
+## Current limitations
+- This workflow is only available to Microsoft employees. If you are not a Microsoft employee you can continue to run benchmarks using our other [benchmarking workflow documentation](./benchmarking-workflow-dotnet-runtime.md).
+- Currently, only support for running the BenchmarkDotNet benchmarks has been thoroughly tested. In the future, we will work towards adding support for scenarios such as Startup and Size on Disk.
```
DrewScoggins, line 32: Maybe also include additional configurations here (Mono, WASM, iOS, Android, etc...)

docs/crank-to-helix-workflow.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Workflow
+
+### Building the runtime repository
+The crank configuration only supports builds that have been generated to the Core_Root folder. Please see [these docs](https://github.com/dotnet/runtime/blob/main/docs/workflow/testing/coreclr/testing.md#building-the-core_root) in the runtime repo for more information about how to build this folder. If you wish to compile for a different OS or architecture, please read the [Cross-Building](https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/cross-building.md#cross-building-for-different-architectures-and-operating-systems) documentation. Below are some examples of what to run for some common scenarios.
+
+#### Building for Windows on Windows
+Run the following in the cloned runtime repository
+```cmd
+.\build.cmd clr+libs -c Release && .\src\tests\build.cmd release generatelayoutonly
```
DrewScoggins, line 60: This command does not work in Powershell in Windows Terminal. I had to replace the `&&` with `;`

docs/crank-to-helix-workflow.md
```diff
@@ -0,0 +1,105 @@
...
+  - It is not required, but running crank will be simpler if the two repositories are cloned to the same parent directory such that doing `cd ../runtime` in the performance repository will navigate you to the runtime repository.
+- Crank must be installed to your machine.
+  - Only the crank controller is required. We are hosting a crank agent accessible to Microsoft employees which has all the required environment variables set up to schedule Helix jobs and upload performance results.
+  - Crank can be installed with `dotnet tool install -g Microsoft.Crank.Controller --version "0.2.0-*"`
+  - Please see the [crank](https://github.com/dotnet/crank) GitHub repository for further information and documentation.
+- Microsoft's corporate VPN is required to be active to connect to the crank agent.
+- Corpnet access
+  - If your machine is not connected to corpnet, then [DevBox](https://devbox.microsoft.com) is our recommended alternative.
+  - If you are unable to use DevBox and can't get corpnet access, then please email dotnetperf so that we can give you an alternative.
```
DrewScoggins, line 47: Do we currently have an alternative?

caaavik-msft: Yes, using Azure Relay, didn't want to go into it in too much detail in the docs.

docs/crank-to-helix-workflow.md
```diff
@@ -0,0 +1,134 @@
...
```
LoopedBard3, line 47:
```suggestion
  - If you are unable to use DevBox and can't get corpnet access, then please email [dotnetperf@microsoft.com](mailto:dotnetperf@microsoft.com) so that we can give you an alternative.
```

docs/crank-to-helix-workflow.md
DrewScoggins: OK, I wasn't sure.

docs/crank-to-helix-workflow.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Prerequisites
+- The [dotnet/runtime](https://github.com/dotnet/runtime) and [dotnet/performance](https://github.com/dotnet/perforamnce) repositories must be cloned to your machine. 
+  - It is not required, but running crank will be simpler if the two repositories are cloned to the same parent directory such that doing `cd ../runtime` in the performance repository will navigate you to the runtime repository.
+- Crank must be installed to your machine.
+  - Only the crank controller is required. We are hosting a crank agent accessible to Microsoft employees which has all the required environment variables set up to schedule Helix jobs and upload performance results.
+  - Crank can be installed with `dotnet tool install -g Microsoft.Crank.Controller --version "0.2.0-*"`
+  - Please see the [crank](https://github.com/dotnet/crank) GitHub repository for further information and documentation.
+- Microsoft's corporate VPN is required to be active to connect to the crank agent.
+- Corpnet access
```
DrewScoggins, line 45: Maybe we can make this a little clearer. It is not immediately obvious that your computer must both be connected to the ethernet in a MS building and connected to the VPN. Maybe we should also start with saying that people should use Devbox, just because I wonder how many people actually have a computer they use day-to-day that meets the requirements.

docs/crank-to-helix-workflow.md
```diff
@@ -0,0 +1,105 @@
...
+
+An explanation for each argument:
+- `--config .\helix.yml`: This tells crank what yaml file defines all the scenarios and jobs
+- `--scenario micro`: Runs the microbenchmarks scenario
+- `--profile win-x64`: Configures crank to a local Windows x64 build of the runtime, and sets the Helix Queue to a Windows x64 queue.
+- `--variable bdnArgs="--filter *Linq*"`: Sets arguments to pass to BenchmarkDotNet that will filter it to only Linq benchmarks
+- `--profile msft-internal`: Sets the crank agent endpoint to the internal hosted crank agent
+- `--variable buildNumber="myalias-20230811.1"`: Sets the build number which will be associated with the results when it gets uploaded to our storage accounts. You can use this to search for the run results in Azure Data Explorer. This build number does not have to follow any convention, the only recommendation would be to include something unique to yourself so that it doesn't conflict with other build numbers.
+
+#### Profiles
```
DrewScoggins, line 97: Maybe it is also worth adding here that using `--profiles` multiple times is additive and does not destroy what was inserted with a previous call. That is a fairly unintuitive way for a command line option to work.

docs/crank-to-helix-workflow.md
```diff
@@ -0,0 +1,105 @@
...
+- `--variable osGroup="windows"`: Set what type of OS the helix queue is, examples of valid values are `windows`, `osx`, and `linux`, `ios`, `freebsd`, etc. Set to `windows` by default.
+- `-variable architecture="x64"`: Sets the architecture to use e.g. `x64`, `x86`, `arm64`. Set to `x64` by default.
+- `--variable internal="true"`: Sets whether or not the Helix Queue is a public or internal queue. If the queue is public the results will not be uploaded. Defaults to `true`.
+- `--json results.json`: Will export all the raw benchmark results as a JSON with the given file name.
+
+### Accessing results
+#### Crank CLI output
+Once the helix jobs have completed, crank will output a simplfied benchmark results with a list of all the benchmarks and the average runtime.
+
+#### Azure Data Explorer
```
DrewScoggins, line 118: Might it be worth it to add a query here that allows people to easily compare their results against some build we already have in the data?

caaavik-msft: I was going to include the one you sent me, but I think for now I'm not sure if this is the right way to do comparisons. Not going to block the PR on this.

cincuranet (approved):

PR #3250, opened 2023-08-11
Add MicrosoftDotNetXHarnessCLIVersion to the maui iOS test proj.

LoopedBard3:
Add MicrosoftDotNetXHarnessCLIVersion to the maui iOS test proj as it is now required. This version is known working from the runtime repo for the startup tests. Same change was made to other branch I am testing getting a successful send to helix in this test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2241031&view=results.

### DIFF
--- a/eng/performance/maui_scenarios_ios.proj
+++ b/eng/performance/maui_scenarios_ios.proj
@@ -3,6 +3,7 @@
 
   <PropertyGroup>
     <IncludeXHarnessCli>true</IncludeXHarnessCli>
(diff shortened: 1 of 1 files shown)


kotlarmilos (approved):
Generally It would be good to retrieve the version from an arcade, like we do for the Maui workload.

cincuranet (approved):
I suppose we need to update this value manually whenever new version is available, right? Otherwise LGTM.

PR #3251, opened 2023-08-11
Remove dataclasses from ci_setup.py

caaavik-msft:
Dataclasses dependency in ci_setup.py is breaking the perf runs in the runtime repository, just going to remove the dependency altogether for now while we fix it in the future

### DIFF
--- a/eng/performance/benchmark_jobs.yml
+++ b/eng/performance/benchmark_jobs.yml
@@ -129,11 +129,6 @@ jobs:
         steps:
         - checkout: self
           clean: true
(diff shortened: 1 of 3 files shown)


caaavik-msft:
@radical this PR combined with [your PR](https://github.com/dotnet/runtime/pull/90424) will fix the issues

DrewScoggins:
Runs all have gotten through ci_setup step. And output looks good.

DrewScoggins (approved):

Issue #3148, opened 2023-07-20
[perf] iOS Maui blazor default startup didn't get the fully-drawn magic string event


kotlarmilos:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2225910
```
[2023/07/20 01:42:04][INFO] {"count":16,"finished":1}
[2023/07/20 01:42:04][INFO] ==========
[2023/07/20 01:42:04][INFO] /private/tmp/helix/working/A9A00928/w/B1E70900/e/mauiblazorios/tmp/iteration0.logarchive
[2023/07/20 01:42:04][INFO] ==========
[2023/07/20 01:42:04][INFO] $ log show --predicate "(processIdentifier == 1511) && (composedMessage contains \"__MAUI_Blazor_WebView_OnAfterRender__\")" --info --style ndjson tmp/iteration0.logarchive
[2023/07/20 01:42:04][INFO] {"count":0,"finished":1}
[2023/07/20 01:42:04][INFO] ==========
[2023/07/20 01:42:04][INFO] /private/tmp/helix/working/A9A00928/w/B1E70900/e/mauiblazorios/tmp/iteration0.logarchive
[2023/07/20 01:42:04][INFO] ==========
Versions: {'mauiVersion': '8.0.0-nightly.8767+sha.15061214f-azdo.8112471.15061214f153f9ed81de921c34a9777d8c8a0230'}
Traceback (most recent call last):
  File "test.py", line 16, in <module>
    Runner(traits).run()
  File "/tmp/helix/working/A9A00928/p/shared/runner.py", line 763, in run
    raise Exception("Didn't get the fully-drawn magic string event.")
Exception: Didn't get the fully-drawn magic string event.
Removing app,tmp,traces,pub,bin,crossgen.out,__pycache__,emsdk if exist ...
Removing empty log files:
ls: /tmp/helix/working/A9A00928/w/B1E70900/uploads/xunit-*.xml: No such file or directory
+ exit_code=1
+ '[' -f /tmp/helix/working/A9A00928/w/B1E70900/e/.retry ']'
+ '[' -f /tmp/helix/working/A9A00928/w/B1E70900/e/.reboot ']'
+ exit 1
+ export _commandExitCode=1
+ _commandExitCode=1
+ export PYTHONPATH=:/etc/helix/scripts
+ PYTHONPATH=:/etc/helix/scripts
[...]

PR #3263, opened 2023-08-15
[perf] Add hybrid globalization performance measurements for Maui scenarios

kotlarmilos:
This PR adds performance measurements of Maui scenarios with hybrid globalization. Once the PR is merged, the performance dashboard should be expanded with new results.

Native AOT
Metric                                                                   |Hybrid Globalization  (MB)          |Standard Globalization (MB)   
-------------------------------------------------------------------------|-------------------|--------------
SOD - Xamarin iOS Default Native AOT IPA Size         | 1,98 | 2,14
SOD - Maui iOS Default Native AOT IPA Size         | 10,06 | 10,22
SOD - Maui Blazor iOS Default Native AOT IPA Size         | 11,47 | 11,62

Mono AOT
Metric                                                                   |Hybrid Globalization  (MB)          |Standard Globalization (MB)   
-------------------------------------------------------------------------|-------------------|--------------
SOD - Xamarin iOS Default Mono AOT IPA Size         | 4,25 | 4,41
SOD - Maui iOS Default Mono AOT IPA Size         | 14,817 | 14,97
SOD - Maui Blazor iOS Default Mono AOT IPA Size         | 17,68 | 17,83

Dotnet version: 8.0.100-preview.7.23376.3


kotlarmilos:
Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2244041&view=results

The startup measurements for Native AOT with hybrid globalization are failing, possibly due to app crashes. /cc: @mkhamoyan

eng/performance/maui_scenarios_ios.proj
```diff
@@ -26,19 +27,19 @@
   </ItemDefinitionGroup>
 
   <ItemGroup>
-    <MAUIiOSScenario Include="Xamarin iOS Default $(RuntimeFlavor)">
+    <MAUIiOSScenario Include="Xamarin iOS Default $(RuntimeFlavor) HybridGlobalization $(HybridGlobalization)">
```
LoopedBard3, line 30: We should think about the best way to add these properties going forward to ensure the best compatibility with using ADX and PowerBI for the scenario names. Specifically, I think if we could have the properties be additive to the name rather than added to every name, it would keep any PowerBI Maintenance and ADX queries that use test name from breaking each time we add another true or false property.

LoopedBard3 (commented):
Thank you for making these changes! It looks really good. The primary comment is around figuring out a good way to deal with how we add properties to names without disrupting flows that may depend on the test names.

eng/performance/maui_scenarios_ios.proj
kotlarmilos: That is right. 

> I think if we could have the properties be additive to the name rather than added to every name

Could you provide an example?

Alternative approach would be matching them by a unique value, such as primary key or something similar which is unique to scenarios. When a new scenario is introduced, a key is added to the database, and a config is updated on the CI.

LoopedBard3: A good example of this making the naming additive is the name setup found here: https://github.com/dotnet/runtime/pull/89825/files#diff-cda44392aa226e93aee0bcef10f241925a7cb882ca107cef95a333e3ee1da8a6R35.

kotlarmilos:
Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2262843&view=results

@LoopedBard3 The startup measurements of Maui and Xamarin templates are disabled due to https://github.com/dotnet/performance/issues/3148.

LoopedBard3 (approved):
LGTM, verified that the configs are updated as expected.

PR #3275, opened 2023-08-17
Support processor groups in the affinity assertions

markples:


markples:
We won't be merging this as we are deleting the Python infrastructure in favor of the C# infrastructure. However, it might be useful as a stopgap for existing uses of the Python infrastructure.

PR #3277, opened 2023-08-17
Make the ILLink benchmark much more realistic

vitek-karas:
Previously we ran with copy default action (or the equivalence of that) - which is not the default for console/ASP.NET apps
* Didn't test the setup which most people will experience
* ILLink reported many warnings because the framework assumes certain code to be removed by it (due to feature switches and such), but that didn't always kick in due to the copy action - default should be no warnings
* One upside of the previous setting was that the tests stressed the linker much more (especially the MarkStep) and thus the resulting duration is much more precise measuring marking performance as compared to the rest of the linker - effectively it emulated a much larger application.

Fixed a wrong singlewarn- command line.
Fixed feature switched.
Updated feature switches to match the .NET 8 default for command line apps.

Marking this as a draft first since there's some discussion to be had:
* Should we rename the benchmark name? So that history data is not "polluted"? https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-immutable
* Should we keep the existing one and just introduce a new one which is more realistic?
* Should we change the msbuild variation as well? It currently uses TrimMode=partial.

Future considerations:
* Add another benchmark for ASP.NET AOT project - as that is a good sample of a more complex app
* Add another benchmark with mono, MAUI or Blazor project template - as that is where illink is used heavily and with different settings

/cc Youssef1313

### DIFF
--- a/src/benchmarks/real-world/ILLink/BasicBenchmark.cs
+++ b/src/benchmarks/real-world/ILLink/BasicBenchmark.cs
@@ -3,7 +3,6 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using BenchmarkDotNet.Attributes;
-using BenchmarkDotNet.Engines;
 
 namespace ILLinkBenchmarks;
 
@@ -16,9 +15,8 @@ public partial class BasicBenchmark
     string[] _linkerOutArgs => new string[] { "-out", Path.Combine(_publishOutputFolder, "linked") };
     string[] _extraArgs = new string[] {
         "--singlewarn",
-        "--trim-mode", "copy",
-        "--action", "copy",
-        "--singlewarn-", "link6",
+        "--trim-mode", "link",
+        "--action", "link",
         "--nowarn", "1701;1702;IL2121;1701;1702",
         "--warn", "5",
         "--warnaserror-", "--warnaserror", ";NU1605",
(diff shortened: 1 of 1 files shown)


jtschuster:
> Should we rename the benchmark name? So that history data is not "polluted"? 
> Should we keep the existing one and just introduce a new one which is more realistic?

I'm in favor of keeping both. While the previous benchmark wasn't a realistic situation, it did have to analyze all the framework libraries, which covered more analysis scenarios and makes it useful for overall performance benchmarking of the linker. Since this is just a HelloWorld app, we're rooting a lot less and analyzing much less if we use 'trim'. I think both are valid scenarios to test and offer different insights into the performance.

> Should we change the msbuild variation as well? It currently uses TrimMode=partial.

Do you think we could remove that? The only benefit I could see from that test is that the arguments to the linker will be updated as we update the SDK.

vitek-karas:
Another idea is to:
- In the benchmark setup, run `dotnet publish -p:PublishTrimmed=true /bl`, grab the binlog, and using structured log package parse it and find the command line used to run the `illink`. Then run that in the benchmark.
- This would have two advantages:
  - No need to maintain the command line compatible and representative of the current version
  - No need for code to grab list of framework assemblies, prepare a temp folder and so on - it would happen in the normal output directory
- Disadvantage is the dependency on msbuild's binlog and especially the package to parse it

jtschuster:
The LinkTask is a ToolTask, which looks like it has some functionality to create a response file. Maybe we could have an option to write it to disk permanently and use that in this benchmark? That would have been helpful for debugging some issues in the past, too.

vitek-karas:
Good point - let's solve the "repro" problem for the product first and then we can use it in the benchmarks.
Sven - how much do you think there's risk in switching the illink invocation from normal command line, to rsp file? It doesn't feel dangerous at all - especially if we do it for 9 only.

jtschuster:
If we don't want to wait for msbuild, we could also override ExecuteTool and have a switch to just output the rsp file somewhere and skip actually running the linker. Then we wouldn't need to run the whole linker before doing it again in the benchmark.

Or we could create a whole new task which just outputs the rsp file somewhere and shares code with the link task.

PR #3284, opened 2023-08-18
Added an override nuget feed for some packages that don't exist in our eng feeds

mrsharm:
A few users have been reporting issues with installing required packages such as XPlot needed by the infra. This is a workaround to add the nuget feed for the GC.Infrastructure to prevent install errors.


mrsharm:
Closing in favor of adding 2 missing packages to our dotnet-public feed: 

[NuGet Gallery | Spectre.Console 0.47.0](https://www.nuget.org/packages/spectre.console)
[NuGet Gallery | XPlot.Plotly 4.0.6](https://www.nuget.org/packages/XPlot.Plotly)

Adding these packages to the feed is a better way to do this than referencing the public nuget source.

CC: @markples

PR #3293, opened 2023-08-23
Added Improvements to the Features and Messaging to the GC Infra re: ASP.NET Benchmarks

mrsharm:
1. Fixed headers of commands and reporting results.
2. Improved the communication of failures and added documentation that delineates the reasons and troubleshooting steps.
3. Added the ability to specify additional arguments.
4. Added the ability to specify the framework version at the top and run level.
5. Early exit of analysis in case there is just one run.
6. Added the ability to upload either the entire folder or a specific file to crank.
7. Added the ability to specify the name of the run and benchmark if the log file is specified.
8. Added a notebook that's associated with analyzing results from the ASP.NET benchmarks.


src/benchmarks/gc/GC.Infrastructure/Configurations/ASPNetBenchmarks/ASPNetBenchmarks.yaml
```diff
@@ -3,6 +3,7 @@ runs:
     corerun: C:\CoreRuns\Empty\
     environment_variables:
       COMPlus_GCServer: 1
+    framework_version: net6.0
```
markples, line 6: I'm curious about the choice of 6.0 here. My guess would have been that samples include the latest version.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/CommandBuilders/ASPNetBenchmarks.CommandBuilder.cs
```diff
@@ -72,14 +72,45 @@ public static (string, string) Build(ASPNetBenchmarksConfiguration configuration
...
+                frameworkVersion = configuration.Environment.framework_version;
 
-            string corerunToSend = run.Value.corerun.EndsWith("\\") ? run.Value.corerun.Remove(run.Value.corerun.Length - 1) : run.Value.corerun;
-            commandStringBuilder.Append($" --application.options.outputFiles {Path.Combine(Path.GetDirectoryName(corerunToSend), "*.*" )}");
+                // If the framework version is set at the run level, use it.
+                if (!string.IsNullOrEmpty(run.Value.framework_version))
+                {
+                    frameworkVersion = run.Value.framework_version;
+                }
+            }
```
markples, line 93: nit/style comment:

If you have/define a helper method somewhere

``` c#
    public static string NullIfEmpty(this string value) => String.IsNullOrEmpty(value) ? null : value;
```

then this could be
``` c#
// Choose in preference order
string frameworkVersion = configuration.Environment.framework_version.NullIfEmpty()
    ?? run.Value.framework_version.NullIfEmpty()
    ?? "net8.0";
```
(also consider if you want to use IsNullOrWhiteSpace instead, though that seems far less common than somehow ending up with the empty string)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/CommandBuilders/ASPNetBenchmarks.CommandBuilder.cs
```diff
@@ -72,14 +72,45 @@ public static (string, string) Build(ASPNetBenchmarksConfiguration configuration
...
+            // If the corerun specified is a directory, upload the entire directory.
+            // Else, we upload just the file.
+            if (Directory.Exists(run.Value.corerun!))
+            {
+                artifactsToUpload = Path.Combine(artifactsToUpload, "*.*");
+            }
+            commandStringBuilder.Append($" --application.options.outputFiles {artifactsToUpload} ");
 
             // Get the log.
+            // TODO: Specify the path.
```
markples, line 107: just checking that this isn't completed by the use of `configuration.Output.Path` below

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/ASPNetBenchmark.Configuration.cs
```diff
@@ -11,17 +11,20 @@ public sealed class ASPNetBenchmarksConfiguration : ConfigurationBase
...
     {
         public string? corerun { get; set; }
+        public string? framework_version { get; set; }
     }
 
     public class Environment
     {
         public Dictionary<string, string> environment_variables { get; set; } = new();
         public uint default_max_seconds { get; set; } = 300;
+        public string framework_version { get; set; } = "net8.0";
```
markples, line 21: Is this default value going to always "win" the choice above?

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksAnalyzeCommand.cs
```diff
@@ -38,7 +41,17 @@ public override int Execute([NotNull] CommandContext context, [NotNull] AspNetBe
...
-            Dictionary<string, List<MetricResult>> results = ExecuteAnalysis(configuration, configurationToCommand, new());
+            Dictionary<string, List<MetricResult>> result = ExecuteAnalysis(configuration, configurationToCommand, new());
+            if (result.Count == 0)
+            {
+                AnsiConsole.MarkupLine($"[bold green] No report generated since there were no results to compare. [/]");
+            }
+
+            else
+            {
+                AnsiConsole.MarkupLine($"[bold green] Report generated at: {Path.Combine(configuration.Output.Path, "Results.md")} [/]");
```
markples, line 52: This report name seems like something that could be in a shared location (and allow some kind of `--report-name <val>` argument if desired)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksAnalyzeCommand.cs
```diff
@@ -64,10 +84,10 @@ public override int Execute([NotNull] CommandContext context, [NotNull] AspNetBe
                 }
             }
 
-            // Launch new process.
             Dictionary<string, string> benchmarkToComparisons = new();
             Dictionary<string, List<MetricResult>> metricResults = new();
 
+            // Compute the compare functionality if there is a single run.
```
markples, line 90: I don't understand this or the following added comment about a "single run" after the `return new()` above.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
```diff
@@ -132,27 +139,32 @@ public static AspNetBenchmarkResults RunASPNetBenchmarks(ASPNetBenchmarksConfigu
                         }
                     }
 
-                    string outputDetails = output.ToString();
-                    File.WriteAllText(Path.Combine(outputPath, $"{GetKey(c.Key, run.Key)}.log"), "Output: \n" + outputDetails + "\n Errors: \n" + error.ToString());
+                    // For the case where the output file doesn't exist implies that was an issue connecting to the asp.net machines or error number 1.
+                    // This case also applies for incorrect crank arguments or error number 2.
+                    // Move the standard out to the standard error as the process failed.
```
markples, line 144: style - move comment into the `else`?  I personally find this a bit confusing because it makes the `if` looks like it doesn't have an `else`

src/benchmarks/gc/GC.Infrastructure/docs/ASPNETBenchmarks.md
```diff
@@ -0,0 +1,30 @@
+# ASP.NET Benchmarks
```
markples, line 1: great, thanks!

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/ASPNetBenchmark.Configuration.cs
mrsharm: Yes, this was the value hardcoded before in the crank command.

src/benchmarks/gc/GC.Infrastructure/Configurations/ASPNetBenchmarks/ASPNetBenchmarks.yaml
mrsharm: This was just an example to show case that one can override the framework version at the level of the run.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/ASPNetBenchmark.Configuration.cs
markples: I don't mean to question the specific value.  Is line 84

``` c#
    if (!string.IsNullOrEmpty(configuration.Environment.framework_version))
```

referring to this value?  If so, then if it has a default hard-coded value, then won't the `if` always be `true`?  (and also the same string appears on line 82)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/CommandBuilders/ASPNetBenchmarks.CommandBuilder.cs
markples: I read the original code backwards.  The order is flipped and should be

``` c#
// Choose in preference order
string frameworkVersion = run.Value.framework_version.NullIfEmpty()
    ?? configuration.Environment.framework_version.NullIfEmpty()
    ?? "net8.0";
```

(but also if my other comment is correct, the "net8.0" will never be used here because the second value will be "net8.0" if not overridden to be something else - unless it is possible to override it with empty/null)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/ASPNetBenchmark.Configuration.cs
mrsharm: Ah - true. Will change so that we are only checking if there is an override.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksAnalyzeCommand.cs
mrsharm: Will consider this for all the test suites in a separate PR.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/CommandBuilders/ASPNetBenchmarks.CommandBuilder.cs
mrsharm: Not needed anymore since ``configuration.Environment.framework_version`` is never null.

Maoni0:
> {

let's not checkin the output of the notebooks.

---
Refers to: src/benchmarks/gc/GC.Infrastructure/Notebooks/ASPNetBenchmarkAnalysis.ipynb:1 in 8bf7c48. [](commit_id = 8bf7c48492ac9c2f1aba3ad506bc94036020fb8d, deletion_comment = False)

Maoni0 (dismissed):
aside from the output, rest LGTM

Maoni0 (approved):

mrsharm:
@cincuranet - can we please merge this? Thanks in advance!

PR #3298, opened 2023-08-25
ComWrappers RCW parallel access

AaronRobinsonMSFT:
Benchmark created as fallout of https://github.com/dotnet/runtime/pull/91120.

### DIFF
--- a/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
+++ b/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
@@ -43,14 +43,16 @@ namespace Interop
         [Benchmark]
         public async Task ParallelRCWLookUp()
         {
-            // Define a large number of iterations for parallel action.
-            var count = 3_000_000;
             await Task.WhenAll(
-                Enumerable.Repeat(_ptr, count)
-                    .Select(ptr =>
+                Enumerable.Range(0, Environment.ProcessorCount)
+                    .Select(_ =>
                         Task.Run(delegate
                         {
-                            s_instance.GetOrCreateObjectForComInstance(ptr, CreateObjectFlags.None);
+                            // Define a large number of iterations for parallel action.
+                            for (int i = 0; i < 3_000_000; i++)
+                            {
+                                s_instance.GetOrCreateObjectForComInstance(_ptr, CreateObjectFlags.None);
+                            }
                         })));
         }
 
--- a/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
+++ b/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
@@ -13,7 +13,7 @@ using System.Threading.Tasks;
 
 namespace Interop
 {
-    [BenchmarkCategory(Categories.Runtime, Categories.NoWASM)]
+    [BenchmarkCategory(Categories.Runtime, Categories.NoWASM, Categories.NoMono)]
     public class ComWrappersTests
     {
         static readonly ComWrappers s_instance = new DummyComWrappers();
--- a/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
+++ b/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
@@ -25,6 +25,7 @@ namespace Interop
             => s_instance.GetOrCreateObjectForComInstance(ptr, CreateObjectFlags.None);
 
         private List<nint> iterations;
+        private object targetRcw;
 
         public ComWrappersTests()
         {
@@ -45,7 +46,7 @@ namespace Interop
             nint rcwPtr = (nint)ptr;
 
             // Transfer ownership to RCW instance. This populates the RCW cache.
-            object rcw = s_instance.GetOrCreateObjectForComInstance((nint)ptr, CreateObjectFlags.None);
+            targetRcw = s_instance.GetOrCreateObjectForComInstance((nint)ptr, CreateObjectFlags.None);
 
             // Ownership was transferred.
             Marshal.Release((nint)ptr);
--- a/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
+++ b/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
@@ -7,63 +7,60 @@ using MicroBenchmarks;
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using System.Threading;
 using System.Threading.Tasks;
 
 namespace Interop
 {
-    [BenchmarkCategory(Categories.Runtime)]
-    public unsafe class ComWrappersTests
+    [BenchmarkCategory(Categories.Runtime, Categories.NoWASM)]
+    public class ComWrappersTests
     {
         static readonly ComWrappers s_instance = new DummyComWrappers();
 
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/runtime/Interop/ComWrappers.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                Marshal.AddRef(externalComObject);
+                return new object();
+            }
+
+            protected override void ReleaseObjects(IEnumerable objects)
+                => throw new NotImplementedException();
+        }
+
+        static Guid IID_IUnknown = new Guid(0x00000000, 0x0000, 0x0000, 0xc0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46);
```
danmoseley, line 78: I'm surprised there isn't a convenient place this is already exposed?
Maybe
https://apisof.net/catalog/37ae96ea-3761-766e-fc9d-0d46cba60387

AaronRobinsonMSFT: There isn't a lot of value in consuming this from a NuGet package. It is a well defined constant where duplicates don't breed confusion in practice and pointing at a non-BCL library seems unnecessary. I do agree though that there is an argument to be made to add it to `Marshal` or perhaps the `ComWrappers` API.

AaronRobinsonMSFT: I've filed https://github.com/dotnet/runtime/issues/91150 to see if we can find a good place for some other COM IIDs.

src/benchmarks/micro/runtime/Interop/ComWrappers.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        public ComWrappersTests()
+        {
+            // Allocate a simple IUnknown instance.
+            var ptr = (void**)NativeMemory.Alloc(5, (nuint)sizeof(void*));
+
+            ptr[1] = (void*)1; // Initial ref count.
+
+            // Define IUnknown vtable.
+            ptr[2] = (delegate* unmanaged<void*, Guid*, void**, int>)&QueryInterface;
```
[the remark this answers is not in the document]
jkotas: Or even better: Instead of creating your own bespoke COM implementation for the foreign COM object, you can use a second instance of ComWrappers to create the foreign COM object. I believe that it is going to test the same critical path - the two ComWrappers instances should not influence each other in any way.

Like:

```csharp
...
        static readonly ComWrappers s_foreignInstance = new DummyComWrappers();
...
            nint ptr = s_foreignInstance.GetOrCreateComInterfaceForObject(new object(), CreateComInterfaceFlags.None);

            // Transfer ownership to RCW instance. This populates the RCW cache.
            targetRcw = s_instance.GetOrCreateObjectForComInstance((nint)ptr, CreateObjectFlags.None);

            // Ownership was transferred.
            Marshal.Release((nint)ptr);
...
            protected override unsafe ComInterfaceEntry* ComputeVtables(object obj, CreateComInterfaceFlags flags, out int count)
            {
                count = 0;
                return null;
            }
...
          // delete IID_IUnknown, QueryInterface, AddRef2, AddRef, Release
```

src/benchmarks/micro/runtime/Interop/ComWrappers.cs
```diff
@@ -0,0 +1,118 @@
...
+            Marshal.Release((nint)ptr);
+
+            // Define a large number of iterations for parallel action.
+            var limit = 3_000_000;
+            iterations = Enumerable.Repeat(rcwPtr, limit).ToList();
+        }
+
+        [Benchmark]
+        public unsafe ParallelLoopResult ParallelRCWLookUp()
+            => Parallel.ForEach(iterations, static (nint ptr) => QueryRCW(ptr));
```
jkotas, line 61: I know that I have suggested `Parallel.ForEach` when we talked about this. I do not see this pattern used for testing concurrent anywhere in this repro. I am not sure how well it is going to work (e.g. how stable it is going to be).

I would stick to one of the existing concurrent execution patterns used in the repo. For example, this one:

https://github.com/dotnet/performance/blob/c30478de1d1e8903221d8b21920a9c980d65ee55/src/benchmarks/micro/libraries/System.Buffers/ArrayPoolTests.cs#L63-L81

danmoseley: Just curious, what does Parallel.For do that is meaningfully different to what is done there? In both cases it's presumably one threadpool thread per core.

jkotas: `Parallel.For` takes IEnumerable as the input. It needs to take lock when fetching and scheduling items from this enumerable. 

The work being tested in each iteration is very small: take a lock and do a lookup in a dictionary. The lock taken by `Parallel.For` scheduler can dominate or otherwise interfere with the work being tested.

src/benchmarks/micro/runtime/Interop/ComWrappers.cs
```diff
@@ -7,63 +7,60 @@
...
+        {
+            // Define a large number of iterations for parallel action.
+            var count = 3_000_000;
+            await Task.WhenAll(
+                Enumerable.Repeat(ptr, count)
+                    .Select(ptr =>
+                        Task.Run(delegate
+                        {
+                            s_instance.GetOrCreateObjectForComInstance(ptr, CreateObjectFlags.None);
+                        })));
```
jkotas:
```suggestion
            await Task.WhenAll(
                Enumerable
                .Range(0, Environment.ProcessorCount)
                .Select(_ =>
                Task.Run(async delegate
                {
                    // Define a large number of iterations for parallel action.
                    for (int i = 0; i < 3_000_000; i++)
                    {
                        s_instance.GetOrCreateObjectForComInstance(ptr, CreateObjectFlags.None);
                    }
                })));
```
Scheduling a task is more expensive operation than `s_instance.GetOrCreateObjectForComInstance`. (On my notebook, the perf tests spends 1.3 seconds in the task scheduler and 0.5 seconds in  `s_instance.GetOrCreateObjectForComInstance` on .NET 8 P7 - before the perf fix.)

The perf test needs to schedule a lot fewer tasks than `s_instance.GetOrCreateObjectForComInstance` calls. Otherwise, it would be testing the perf of the task scheduler and not the perf of the ComWrappers.

jkotas: My suggestion is copy&paste from https://github.com/dotnet/performance/blob/c30478de1d1e8903221d8b21920a9c980d65ee55/src/benchmarks/micro/libraries/System.Buffers/ArrayPoolTests.cs#L66C1-L81C1

src/benchmarks/micro/runtime/Interop/ComWrappers.cs
```diff
@@ -0,0 +1,77 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+
+namespace Interop
+{
+    [BenchmarkCategory(Categories.Runtime, Categories.NoWASM)]
```
jkotas, line 16: Do we need to disable it for all Mono targets?

jkotas (dismissed):
Thanks!

jkotas (approved):

cincuranet (approved):

PR #3309, opened 2023-08-31
SizeOnDisk: fix up version replacement for dotnet.native, and dotnet.runtime js files

radical:


lewing (approved):

cincuranet:
Small remark: The dots in regex for dots in name should be escaped (i.e. `dotnet\.(?:native...`). Now it is matching any character.

PR #3322, opened 2023-09-06
GCPerfSim Improvements

mrsharm:
1. Removed some dead code and added explanation for why some code is uncommented.
2. Fixed compilation issue due to the nonexistence of the POH before .NET 5.
3. Replaced the StreamWriter with simple Console Writes.


markples:
The diffs here are all straightforward, but I don't have enough experience here to understand if the commented code is old, pending, useful-to-uncomment, etc.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -2499,27 +2476,13 @@ bool GoToNextPhase()
     // Allocates object and survives it. Returns the object size.
     void MakeObjectAndMaybeSurvive()
     {
-        // if (isLarge && args.lohPauseMeasure)
```
markples, line 2502: For example, the top comment describes `-lohPauseMeasure`, which isn't actually parsed.  Would this be part of it?

Maoni0: we should not get rid of this code - it's useful for measuring LOH alloc pauses. see the comment at the beginning in "Other things worth noting" -

> 1) There's also a lohPauseMeasure var that you could make into a config - for now I just always measure pauses for LOH 
>    allocs (since it was used for measuring LOH alloc pause wrt BGC sweep).

if you want to improve this, you could make it into a config.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -2702,12 +2665,9 @@ static TestResult DoTest(in Args args, int currentPid)
         Debug.Assert(SimpleRefPayLoad.NumPinned == SimpleRefPayLoad.NumUnpinned);
 #endif
 
+        Console.WriteLine($"Time Taken: {tEnd - tStart}ms");
         sw.WriteLine("Took {0}ms", tEnd - tStart);
         sw.Flush();
-
-        // sw.WriteLine("after init: heap size {0}, press any key to continue", GC.GetTotalMemory(false));
-        // Console.ReadLine();
-
```
Maoni0, line 2710: again, we shouldn't get rid of this. this is useful when you want to have the process paused at the end instead of just exit. could you turn this into a config.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -2702,12 +2665,9 @@ static TestResult DoTest(in Args args, int currentPid)
         Debug.Assert(SimpleRefPayLoad.NumPinned == SimpleRefPayLoad.NumUnpinned);
 #endif
 
+        Console.WriteLine($"Time Taken: {tEnd - tStart}ms");
```
Maoni0, line 2668: you might as well add a config to say if you want it to print to console or print to a file. so you don't do duplications like this

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
```diff
@@ -805,11 +804,6 @@ struct SizeSlot
     // an object size distribution for SOH/LOH derived from a real server scenario
     public static SizeSlot[] sohLohSizeSlots = new SizeSlot[]
     {
-//        new SizeSlot(        24,  16229766), We are creating some small wrapper objects of 32 and 40 bytes around the
-//        new SizeSlot(        32,  48790439), payload objects - somewhat compensate for this by distorting the distribution
-//        new SizeSlot(        40,  21496250), at the low end
-//        new SizeSlot(        48,  15020039),
-//
```
Maoni0, line 812: I would not get rid of these - this is data from running real workloads - we should add the ability to create very small objects instead of getting rid of this.

src/benchmarks/gc/src/exec/GCPerfSim/GCPerfSim.cs
mrsharm: Sure, I'll add a comment with this code as a TODO and reference the comment.

PR #3331, opened 2023-09-08
Update BDN version

DrewScoggins:
This updates the BDN version to include this fix, https://github.com/dotnet/BenchmarkDotNet/pull/2375/files. This resolves an issue with filtering that was happening with the GC performance tests.


mrsharm (dismissed):
LGTM! Thanks!

PR #3338, opened 2023-09-12
Add MonoAOTLLVM to local script

LoopedBard3:
Adds MonoAOTLLVM scenario support to the local testing script. Because the MonoAOTLLVM toolchain is only setup to run one BDN test set per invocation, the tool runs multiple which then can be compared using the ResultsComparer tool.

PR #3363, opened 2023-09-22
Fix default repo_storage_path with os.getcwd

LoopedBard3:
Fix default repo_storage_path with os.getcwd rather than . to ensure full path usage. Fixes the error: https://github.com/dotnet/performance/pull/3338#issuecomment-1731548399.


matouskozak:
LGTM! I tried with this change and now the build finishes correctly.

cincuranet (approved):

PR #3381, opened 2023-10-04
Cleanup benchmarks_local.py script

LoopedBard3:
This cleans up the benchmarks_local.py script based on the Pylance recommendations. This mostly consisted of adding type specifications. It also included refactoring how the enum string methods interact with the enum types and updating the benchmarks_ci.py __main method to main so it could be properly called from the local script.

### DIFF
--- a/scripts/benchmarks_local.py
+++ b/scripts/benchmarks_local.py
@@ -200,7 +200,10 @@ def generate_all_runtype_dependencies(parsed_args: Namespace, repo_path: str, co
     if check_for_runtype_specified(parsed_args, [RunType.MonoAOTLLVM]):
         artifact_mono_aot_llvm = os.path.join(get_run_artifact_path(parsed_args, RunType.MonoAOTLLVM, commit), "monoaot")
         if force_regenerate or not os.path.exists(artifact_mono_aot_llvm):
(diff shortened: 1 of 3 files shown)


scripts/benchmarks_local.py
```diff
@@ -76,36 +75,38 @@ def get_os_short_name(os_name: str):
...
         raise ValueError(f"Unknown OS {os_name}")
 
-def is_running_as_admin(parsed_args: Namespace):
+def is_running_as_admin(parsed_args: Namespace) -> bool:
     if is_windows(parsed_args):
+        import ctypes
         return ctypes.windll.shell32.IsUserAnAdmin()
     else:
-        return os.getuid() == 0
+        return os.getuid() == 0 # type: ignore os.getuid() returns an int on Unix-like systems
```
cincuranet, line 83: It's fine that it returns int. You're comparing it with 0, which is also int. Or am I missing something?

cincuranet (dismissed):

scripts/benchmarks_local.py
LoopedBard3: Pylance (for me) displays `Type of "getuid" is unknown`. Clarified this in the latest push.

Issue #3399, opened 2023-10-10
Wasm warning 'unknown flag --experimental-wasm-eh'


LoopedBard3:
It appears that the `--experimental-wasm-eh` flag we are passing to v8 is not being recognized. This is affecting both the wasm aot and wasm wasm microbenchmarks. I have not looked back to see how far back this is affecting yet. The version of v8 that was used in the run I found this is 11.7.439. I tried searching the help for this flag on version 12.0.* of v8 and am not finding it as an option so it may have been removed.

@radical, do you know if this is the proper use of the --experimental-wasm-eh flag and if the flag still exists?

LoopedBard3:
After some digging it looks like the --experimental-wasm-eh flag may only be needed by older node versions and can be removed from our runs: https://github.com/dotnet/runtime/issues/91318

radical:
Yeah, the versions of v8 being used in perf pipeline, and library tests were not the same earlier. But they are controlled, and in sync now to be the latest stable which shouldn't be needed now.

LoopedBard3:
This appears to be working as expected.

PR #3402, opened 2023-10-11
Remove --experimental-wasm-eh argument from the WASM runtype args

LoopedBard3:
Remove --experimental-wasm-eh argument from the WASM runtype args for the benchmarks_local.py script. The versions of v8 should now be in sync to latest so --experimental-wasm-eh is no longer needed.

Per: https://github.com/dotnet/performance/issues/3399

### DIFF
--- a/scripts/benchmarks_local.py
+++ b/scripts/benchmarks_local.py
@@ -21,7 +21,7 @@
 # gitpython (pip install --global gitpython)
 # Ubuntu Version 22.04 if using Ubuntu
 # May need llvm+clang 16 for MonoAOTLLVM (https://apt.llvm.org/)
-# Wasm need jsvu installed and setup (No need to setup EMSDK, the tool does that automatically when building)
+# Wasm runs need jsvu/v8 installed and setup. Latest v8 preferred, must be post --experimental-wasm-eh removal/enabled by default. (No need to setup EMSDK, the tool does that automatically when building)
 
 
 import glob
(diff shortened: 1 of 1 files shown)


scripts/benchmarks_local.py
```diff
@@ -436,7 +436,7 @@ def generate_single_benchmark_ci_args(parsed_args: Namespace, specific_run_type:
             '--category-exclusion-filter', 'NoInterpreter', 'NoWASM', 'NoMono',
             '--wasmDataDir', os.path.join(get_run_artifact_path(parsed_args, RunType.WasmInterpreter, commit), "wasm_bundle", "wasm-data"),
             '--wasmEngine', parsed_args.wasm_engine_path,
-            '--wasmArgs', '\" --experimental-wasm-eh --expose_wasm --module\"',
```
radical, line 439: I guess we don't want to remove this from the local script, makes sense.

radical: lol!! This *is* the local script!!

Ok, so we are removing it for that so it assumes a minimum version of v8. Maybe we should check for that, or .. add somewhere that a minimum version is required?

LoopedBard3: Do you know if there is a minimum version that we don't need the argument for? If not, I will add a note to get latest.

radical: @radekdoulik @pavelsavara might know the version where this was removed.

cincuranet (approved):

scripts/benchmarks_local.py
LoopedBard3: If we find a version, we can add it in another PR. In place of that, I clarified the v8/jsvu requirement at the top.

PR #3410, opened 2023-10-12
Adding benchmarks from Akade.IndexedSet

akade:
I was asked in https://github.com/dotnet/runtime/issues/90101 by @AndyAyersMS if I would like to include the benchmarks from my nuget package in this repo and I finally found the time to so. Including a few less synthetic benchmarks.

Snapshot of benchmarks https://github.com/akade/Akade.IndexedSet/commit/eacb9a8a31e184e0790e5c0c254baff8f47f686c but only including the benchmarks concerning the library itself - leaving out any comparison benchmarks.

I followed the lines of the ImageSharp benchmarks, but I have the following remaining questions:
- How can I get the nuget package listed in the configured sources? I removed the clear for now...
- Do you have limits for the runtime for all the benchmark? Currently, running through all of them in .NET 8 takes about 12min on my machine. I might be able to tune that a bit on the amount of data the real-world-benchmarks are run against.

Anything else that I need to do to?


akade:
@dotnet-policy-service agree

NuGet.config
```diff
@@ -4,7 +4,7 @@
     <add key="disableSourceControlIntegration" value="true" />
   </solution>
   <packageSources>
-    <clear />
+    <!--<clear />-->
```
cincuranet, line 7: Let me know what package(s) needs to be available in the feeds below and I'll make sure it's there. Then this change can go away.

cincuranet (changes requested):

cincuranet:
@akade Also the `azure-pipelines.yml` needs to be modified to run these benchmarks. It will be basically the same as "ImageSharp" parts.

@AndyAyersMS Do you want to run on standard set of platforms we do for real-world benchmarks or something in particular?

AndyAyersMS:
> @AndyAyersMS Do you want to run on standard set of platforms we do for real-world benchmarks or something in particular?

The standard set.

NuGet.config
akade: it needs Akade.IndexedSet 1.0.1 and Bogus 34.0.2. How would it go btw, if I update the package? Which I will do soon for .NET 8...

cincuranet:
@akade Sorry for this taking so long. I was bit busy with other stuff. The packages are now mirrored in `dotnet-public` feed, I fixed the MD problem linter was complaining about and finally fixed the typo in `azure-pipelines.yml` file. If (fingers crossed) everything goes smoothly, I'll merge it and prepare all the reporting and data ingestion tomorrow.

cincuranet:

cincuranet:
Reports are being generated. Links:

- https://pvscmdupload.blob.core.windows.net/reports/allTestHistory/9.0_x64_ubuntu%2022.04_RunKind=akadeindexedset/AllTestindex.html
- https://pvscmdupload.blob.core.windows.net/reports/allTestHistory/9.0_arm64_ubuntu%2020.04_RunKind=akadeindexedset/AllTestindex.html
- https://pvscmdupload.blob.core.windows.net/reports/allTestHistory/9.0_x64_Windows%2010.0.18362_RunKind=akadeindexedset/AllTestindex.html
- https://pvscmdupload.blob.core.windows.net/reports/allTestHistory/9.0_arm64_Windows%2010.0.19041_RunKind=akadeindexedset/AllTestindex.html

Thanks for the contribution @akade.

Issue #3313, opened 2023-09-01
Creating emptyvbconsoletemplate scenario often fails because it cannot access the file


cincuranet:
```
  [2023/09/01 12:26:12][INFO] $ pushd "D:\a\1\s\src\scenarios\emptyvbconsoletemplate"
  [2023/09/01 12:26:12][INFO] $ dotnet new console --output app --no-restore --force --name emptyvbconsoletemplate --language vb
  [2023/09/01 12:26:12][INFO] Template "Console App" could not be created.
  [2023/09/01 12:26:12][INFO] Failed to create template.
  [2023/09/01 12:26:12][INFO] Details: Error while processing file /content/ConsoleApplication-VisualBasic/Company.ConsoleApplication1.vbproj
  [2023/09/01 12:26:12][INFO] The process cannot access the file 'D:\a\1\s\src\scenarios\emptyvbconsoletemplate\app\emptyvbconsoletemplate.vbproj' because it is being used by another process.
  [2023/09/01 12:26:12][INFO] 
  [2023/09/01 12:26:12][INFO] 
  [2023/09/01 12:26:12][INFO] For details on the exit code, refer to https://aka.ms/templating-exit-codes#100
  [2023/09/01 12:26:12][INFO] $ popd
```

JanKrivanek:
FYI @dotnet/templating-engine-maintainers

marcpopMSFT:
@MiYanni can you take a look and see if you can reproduce this (or theorize why this would happen)? I believe we own the empty console templates.

MiYanni:
So, just trying the basic scenario with the 8.0.100-rc.2.23477.19 build of the SDK:
```
PS C:\Users\miyanni> pushd C:\Workspace\TemplateTest\NewOne
PS C:\Workspace\TemplateTest\NewOne> dotnet new console --output app --no-restore --force --name emptyvbconsoletemplate --language vb
The template "Console App" was created successfully.

PS C:\Workspace\TemplateTest\NewOne>
```

Then, since the issue seemed to suggest it was intermittent, I ran this which runs the command 100 times and deletes the folder before running the command again:
```powershell
foreach ($i in 1..100){ dotnet new console --output app --no-restore --force --name emptyvbconsoletemplate --language vb; Remove-Item -Force -Recurse -Path App }
```

No issues there. It just printed `The template "Console App" was created successfully.` 100 times.

[...]

PR #3415, opened 2023-10-16
Fix grouping so post.py is properly executed for every item in scenario.

cincuranet:
Side effect of this is fixing of https://github.com/dotnet/performance/issues/3313 (at least I'm 99% sure).


cincuranet:
Correct.

PR #3420, opened 2023-10-17
Update PowerShell benchmarks to use net8.0

TIHan:
We are wanting to use SuperPMI collections for the PowerShell benchmarks but we have to update the targetframework to net8 for it to work.


cincuranet:
@TIHan Do you want to merge immediately?

TIHan:
@cincuranet yes, it would be good to merge now if we can

cincuranet:

@TIHan Done.

Issue #3408, opened 2023-10-12
benchmarks_local bdn-arguments join requires space


LoopedBard3:
When trying to run benchmarks_local.py with join in bdn-arguments, it only works when there is a space after the join. `--bdn-arguments "--join "` works, `--bdn-arguments "--join"` does not work.

LoopedBard3:
It looks like passing in argument values with starting with -- is not cleanly suppported: https://stackoverflow.com/questions/16174992/cant-get-argparse-to-read-quoted-string-with-dashes-in-it
and
https://bugs.python.org/issue9334

To get around this, add an equals sign after `--bdn-arguments`, like `--bdn-arguments="--join"`. This should fix the problem. I am also going to add a note about this to the help output.
FYI @matouskozak

PR #3434, opened 2023-10-20
Benchmarks_local script: Clarify that bdn-arguments must be passed using =.

LoopedBard3:
Fixes: https://github.com/dotnet/performance/issues/3408.

Clarifies that `--bdn-arguments` should be pass with ="args..". This is further explored in the above issue but is due to how pythons arg_parse parses arguments. Also removed unnecessary default.


matouskozak (approved):
Thanks for investigating. It works for me when passed with `=`.

PR #3435, opened 2023-10-21
Add microbenchmark for core UnmanagedMemoryStream scenario

jkotas:
Related to https://github.com/dotnet/runtime/pull/93766

### DIFF
--- a/src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
+++ b/src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
@@ -41,7 +41,7 @@ namespace System.IO.Tests
         [Benchmark]
         public void ReadAsArrays()
         {
-            var array = new byte[1];
+            var array = new byte[128];
             using (var ums = new UnmanagedMemoryStream(_buffer, Length))
             {
                 while (ums.Read(array, 0, array.Length) != 0)
--- a/src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
+++ b/src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
@@ -13,24 +13,24 @@ namespace System.IO.Tests
     public unsafe class UnmanagedMemoryStreamTests
     {
         private const int Length = 10000;
-        private byte* buffer;
+        private byte* _buffer;
 
         [GlobalSetup]
         public void Setup()
         {
-            buffer = (byte*)Marshal.AllocCoTaskMem(Length);
+            _buffer = (byte*)Marshal.AllocCoTaskMem(Length);
         }
 
         [GlobalCleanup]
         public void Cleanup()
         {
-            Marshal.FreeCoTaskMem((IntPtr)buffer);
+            Marshal.FreeCoTaskMem((IntPtr)_buffer);
         }
 
         [Benchmark]
         public void ReadAsBytes()
         {
-            using (var ums = new UnmanagedMemoryStream(buffer, Length))
+            using (var ums = new UnmanagedMemoryStream(_buffer, Length))
             {
                 while (ums.ReadByte() >= 0)
                 {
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        [GlobalCleanup]
+        public void Cleanup()
+        {
+            Marshal.FreeCoTaskMem((IntPtr)buffer);
+        }
+
+        [Benchmark]
+        public void ReadAsBytes()
+        {
+            using (var ums = new UnmanagedMemoryStream(buffer, Length))
```
jkotas, line 33: I am not sure whether this should better be in IterationSetup. The docs for [IterationSetup](https://github.com/dotnet/BenchmarkDotNet/blob/e170684208103ca5ba4212ad8dc7c2aad5cf02d4/src/BenchmarkDotNet.Annotations/Attributes/IterationSetupAttribute.cs#L7) say that it should not be used for microbenchmarks.

cincuranet: As long as the benchmark takes roughly more than 100ms the IterationSetup is fine.

jkotas: This microbenchmark takes a lot less than 100ms and I do not think it makes sense to artificially inflate it to take more than 100ms.

src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    [BenchmarkCategory(Categories.Libraries)]
+    public unsafe class UnmanagedMemoryStreamTests
+    {
+        private const int Length = 10000;
+        private byte* buffer;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            buffer = (byte*)Marshal.AllocCoTaskMem(Length);
```
Jozkee, line 21: Just wondering why you chose this over AllocHGlobal?

Jozkee (approved):

src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
```diff
@@ -0,0 +1,53 @@
...
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.IO.Tests
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public unsafe class UnmanagedMemoryStreamTests
+    {
+        private const int Length = 10000;
+        private byte* buffer;
```
Jozkee, line 16:
```suggestion
        private byte* _buffer;
```

src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
jkotas: No particular reason. Either of these two would work just fine for this test.

src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
```diff
@@ -0,0 +1,53 @@
...
+                while (ums.ReadByte() >= 0)
+                {
+                }
+            }
+        }
+
+        [Benchmark]
+        public void ReadAsArrays()
+        {
+            var array = new byte[1];
```
jkotas, line 44:
```suggestion
            var array = new byte[128];
```

cincuranet (approved):

src/benchmarks/micro/libraries/System.IO/UnmanagedMemoryStreamTests.cs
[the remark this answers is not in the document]
jkotas: This test is meant to test raw cost of UnmanagedMemoryStream. With larger array sizes, this test would be testing performance of memory copy and the raw cost of UnmanagedMemoryStream would be diminished.

I have changed it to 128. It is still small enough to make the raw cost of UnmanagedMemoryStream dominate.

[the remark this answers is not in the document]
cincuranet: Makes sense. Thanks.

cincuranet:
Failures are unrelated.

Issue #3448, opened 2023-10-26
ComWrappersTests.ParallelRCWLookUp long op time in specific queue


LoopedBard3:
The ComWrappersTests.ParallelRCWLookUp test is taking 10+ minutes per op when run in the Windows.Server.Arm64.Perf machine queue. This is much greater than the ~9s per op when running on other queues. This long op is also causing many of the Partition10 jobs to fail due to timing out, failing the jobs and taking out one of the few machines for a long period of time.  

Test added in: https://github.com/dotnet/performance/commit/36ab046503c47ea9a9f1abc62e61093456f9468e

FYI @AaronRobinsonMSFT, about the massive time difference.

LoopedBard3:
Run with the error can be found in the pipeline https://dev.azure.com/dnceng/internal/_build/results?buildId=2301502&view=logs&j=86ff836e-79c0-5c2b-d692-74a6138ae50d&t=67773ac8-bb27-54c6-5c6e-b22882c3b62f, under the job for windows arm64, perfampere. There is only one partition failure (Partition10), the partition this test is in.

AaronRobinsonMSFT:
@LoopedBard3 How can I view the other runs? I looked at Partition 10 for the other runs and didn't see `ParallelRCWLookUp`. Is there some algorithm for looking at this data?

/cc @dotnet/interop-contrib @jkotas

LoopedBard3:
It should be available in the windows arm64 perfsurf job above the failing perfampere one.

LoopedBard3:
Here is a link to the test itself: https://github.com/dotnet/performance/blob/4bfa9d4974654c9e04fc68914e925aacf8aaf69e/src/benchmarks/micro/runtime/Interop/ComWrappers.cs#L43-L56

jkotas:
cc @mangod9 @JulieLeeMSFT Microbenchmark takes 10+ minutes on Windows Arm64 server vs. 9 seconds in other queues.

AaronRobinsonMSFT:
I don't have a lot of bandwidth to investigate this now, but I am dubious this has anything to do with `ComWrappers` or is specific to the API. I say that because there is nothing that is architecture specific in `ComWrappers` - either in managed or unmanaged code.

mangod9:
Is this a regression or a new benchmark and/or config?

jkotas:
It is a new microbenchmark added two months ago in #3298. This microbenchmark is a regression test for a customer reported performance issue that we have fixed.

[...]

PR #3457, opened 2023-10-31
Update ComWrapper.ParallelRCWLookup loop count

LoopedBard3:
Update ComWrapper.ParallelRCWLookup loop count to be 3 mil per iteration, not per ProcessorCount in the iteration. Manual test using this setup resulted in a mean time of ~3s per op and total runtime of ~70s, a time we can manage. 

Mitigates: https://github.com/dotnet/performance/issues/3448 by setting a consistent loop count that will not cause issues on high core count machines. There may still be some investigation to be done on windows vs linux runtimes and Arm vs x64 runtimes.

From my understanding of https://github.com/dotnet/runtime/pull/91120, this seems like it should still catch any potential regressions similar to the previous setup while ensuring high-core machines don't get hit harder than expected. @AaronRobinsonMSFT does my understanding above make sense?

### DIFF
--- a/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
+++ b/src/benchmarks/micro/runtime/Interop/ComWrappers.cs
@@ -41,7 +41,6 @@ namespace Interop
         }
 
         [Benchmark]
-        [OperatingSystemsArchitectureFilter(allowed: false, Architecture.Arm64)]
         public async Task ParallelRCWLookUp()
         {
             var iterationsPerTask = 3_000_000/Environment.ProcessorCount;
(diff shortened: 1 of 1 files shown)


AaronRobinsonMSFT (approved):

mangod9:
Doesnt still feel like this test is good to measure across OS/platforms. Assuming this normalizes the test somewhat to get it to finish in reasonable time. On 80 core machines its still going to cause more contention than on smaller machines.

LoopedBard3:
I think that this format still works for OS comparisons on the same hardware since the some total work is being done even taking into account the Task overhead. Another option may be to make something like 6 tasks each doing 500,000 calls. Let me know if this seems like a more desirable solution and I can update the PR to match it.

mangod9:
> Another option may be to make something like 6 tasks each doing 500,000 calls. Let me know if this seems like a more desirable solution and I can update the PR to match it.

will defer to @AaronRobinsonMSFT to determine the right intent for this benchmark.

LoopedBard3:
Merging for now to get some form of fix in. We can still iterate on the test if we decide it is necessary.

PR #3462, opened 2023-11-02
add Int128 and CopySign benchmarks

adamsitnik:
Benchmarks for https://github.com/dotnet/runtime/issues/77579

### DIFF
--- a/src/benchmarks/micro/libraries/System.Runtime/Perf.Int128.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime/Perf.Int128.cs
@@ -1,0 +1,54 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Linq;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.Tests
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public class Perf_Int128
+    {
+        private char[] _destination = new char[Int128.MinValue.ToString().Length];
+
+        public static IEnumerable<object> StringValues => Values.Select(value => value.ToString()).ToArray();
+
+        public static IEnumerable<object> Values => new object[]
+        {
+            Int128.MinValue,
+            (Int128)12345, // same value used by other tests to compare the perf
+            Int128.MaxValue,
+        };
+
+        [Benchmark]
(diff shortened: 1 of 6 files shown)


cincuranet (approved):

src/benchmarks/micro/libraries/System.Runtime/Perf.Int16.cs
```diff
@@ -32,5 +32,11 @@ public class Perf_Int16
         [Benchmark]
         [ArgumentsSource(nameof(Values))]
         public string ToString(short value) => value.ToString();
+
+#if NET7_0_OR_GREATER
+        [Benchmark]
+        [Arguments(1, -1)]
```
tannergooding, line 38: Shouldn't this test copying both a positive and negative sign?

The actual code, due to it being two's complement, has different costs based on whether `value` is positive or negative and whether `sign` is positive or negative.

src/benchmarks/micro/libraries/System.Runtime/Perf.Int128.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.Tests
+{
+    [BenchmarkCategory(Categories.Libraries)]
+    public class Perf_Int128
```
tannergooding, line 13: Why only this subset of tests? Do we want to provide meaningful benchmarks for most of Int128/UInt128 since it is not a language primitive, but does provide the full set of functionality?

tannergooding (approved):

src/benchmarks/micro/libraries/System.Runtime/Perf.Int16.cs
adamsitnik: > Shouldn't this test copying both a positive and negative sign?

I tried that and the reported time was almost identical for all test cases (like 0.4 ns for Int16)

src/benchmarks/micro/libraries/System.Runtime/Perf.Int128.cs
adamsitnik: > Why only this subset of tests

My main goal was to add `CopySign` benchmarks. I've realized that `Int128` has zero benchmarks, so I've copy-pasted `Int64` benchmarks and adjusted them. Just to have some benchmarks as a starting point

adamsitnik:
@cincuranet could you please merge it (the failures are unrelated)

cincuranet:
Done.

PR #3463, opened 2023-11-02
Update benchmarks_local default framework from net8.0 -> net9.0.

LoopedBard3:
Update benchmarks_local default framework from net8.0 -> net9.0 now that 9.0 SDK is available and the default on main.


cincuranet (approved):

LoopedBard3:
This doesn't impact any pipeline testing, merging.

matouskozak (commented):
Thanks for fixing!

PR #3469, opened 2023-11-07
Added retries and more checks for the ASP.NET Benchmarks

mrsharm:
1. Added a corpnet connection check. 
2. Added a check for when the machines reboot - sleep until the reboot is complete.
3. Added the ability to retry _once_ after sleeping in case of hangs and unresponsiveness from the host. 
4. If there are failures in the run, those are automatically persisted as a new configuration file to make it easy to rerun after failures.

### DIFF
--- a/src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
+++ b/src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
@@ -42,9 +42,165 @@ namespace GC.Infrastructure.Commands.ASPNetBenchmarks
 
             ConfigurationChecker.VerifyFile(settings.ConfigurationPath, nameof(AspNetBenchmarksCommand));
             ASPNetBenchmarksConfiguration configuration = ASPNetBenchmarksConfigurationParser.Parse(settings.ConfigurationPath);
+
+            // Before running the ASP.NET benchmarks, execute a few checks:
+            // Check 1. If you are connected to corpnet. These tests only run if you are connected to corp-net and therefore, preemptively, check this.
+            // Check 2. The host machines reboot between 12:00 AM - 12:08 AM PST. Check if we are running during that time, and if so, delay running the infrastructure.
+
+            // Check 1.
+            if (!Directory.Exists(@"\\clrmain\tools\"))
+            {
+                // Not connected to corpnet if that folder is unavailable.
+                AnsiConsole.MarkupLine($"[red bold] Not connected to corpnet. Ensure you are connected before proceeding with the ASP.NET Benchmarks [/]");
+                return -1;
+            }
+
+            // Check 2.
+            SleepUntilHostsHaveRestarted();
+
             RunASPNetBenchmarks(configuration);
             AnsiConsole.MarkupLine($"[bold green] Report generated at: {configuration.Output.Path} [/]");
             return 0;
+        }
+
+        private static void SleepUntilHostsHaveRestarted()
+        {
+            DateTime now = DateTime.UtcNow;
+
+            // Get the Pacific Standard Time zone (considering daylight saving).
+            TimeZoneInfo pstZone = TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time");
+            DateTime pstNow = TimeZoneInfo.ConvertTimeFromUtc(now, pstZone);
+
+            // Check if the current time is between 12:00 AM and 12:08 AM.
+            DateTime start = pstNow.Date; // 12:00 AM today.
+            DateTime end = start.AddMinutes(8); // 12:08 AM today.
+
+            if (pstNow >= start && pstNow < end)
+            {
+                TimeSpan timeUntilEnd = end - pstNow;
+                int secondsLeft = (int)timeUntilEnd.TotalSeconds + 60;
+
+                // If we are between 12:00 AM and 12:08 AM PST, sleep for 
+                AnsiConsole.MarkupLine($"[yellow bold] ({DateTime.Now}) ASP.NET Benchmarks Sleeping for {secondsLeft} seconds since the host machines are rebooting. [/]");
+                Thread.Sleep((secondsLeft) * 1000);
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
```diff
@@ -42,11 +42,167 @@ public override int Execute([NotNull] CommandContext context, [NotNull] AspNetBe
...
+            // Check if the current time is between 12:00 AM and 12:08 AM.
+            DateTime start = pstNow.Date; // 12:00 AM today.
+            DateTime end = start.AddMinutes(8); // 12:08 AM today.
+
+            if (pstNow >= start && pstNow < end)
+            {
+                TimeSpan timeUntilEnd = end - pstNow;
+                int secondsLeft = (int)timeUntilEnd.TotalSeconds + 60;
+
+                // If we are between 12:00 AM and 12:08 AM PST, sleep for
```
markples, line 83:
```suggestion
            // Check if the current time is between 12:00 AM and 12:09 AM.
            DateTime start = pstNow.Date; // 12:00 AM today.
            DateTime end = start.AddMinutes(9); // 12:09 AM today.

            if (pstNow >= start && pstNow < end)
            {
                TimeSpan timeUntilEnd = end - pstNow;
                int secondsLeft = (int)timeUntilEnd.TotalSeconds;

                // If we are between 12:00 AM and 12:09 AM PST, sleep for 
```

markples: (just being consistent about 12:07:59 and 12:08:01)

markples: also a question - should this sleep at 11:59PM as well, or will the retry logic handle that?  Starting a long run at something like 11:30PM is useful :)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
```diff
@@ -42,11 +42,167 @@ public override int Execute([NotNull] CommandContext context, [NotNull] AspNetBe
...
+            // Check 1.
+            if (!Directory.Exists(@"\\clrmain\tools\"))
+            {
+                // Not connected to corpnet if that folder is unavailable.
+                AnsiConsole.MarkupLine($"[red bold] Not connected to corpnet. Ensure you are connected before proceeding with the ASP.NET Benchmarks [/]");
+                return -1;
+            }
+
+            // Check 2.
+            SleepUntilHostsHaveRestarted();
```
markples, line 59: Is it correct that this check is at the beginning of the overall run?  This check is probably best at the beginning of each benchmark (in fact, each attempt at each benchmark in the case of a retry)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
markples: Retry waits 2 minutes, so I think attempt #2 could hit the 12:00-12:08 window.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
```diff
@@ -110,116 +266,31 @@ public static AspNetBenchmarkResults RunASPNetBenchmarks(ASPNetBenchmarksConfigu
...
-                    // 3. Test fails because of a test error.
-
-                    // Launch new crank process.
-                    int exitCode = -1;
-                    string logfileOutput = Path.Combine(outputPath, $"{GetKey(c.Key, run.Key)}.log");
-                    StringBuilder output = new();
-                    StringBuilder error = new();
-
-                    using (Process crankProcess = new())
+                    // Wait 2 minutes and then retry if the run timed out or the host was non-responsive (post corp-net connection check).
```
markples, line 280: Another case is the VPN dropping in the middle of a run, which unfortunately seems to be common.

Of course, the run is toast at that point, so it doesn't matter much.  This will cause each test post-drop to run twice, but I don't think that makes anything worse than it was.  I guess a script running ASP.NET benchmarks and then something else would have to wait a bit longer.  I think this PR should go in without addressing this.  (And the only way seems to be to reconnect the VPN, if that is even possible to do programmatically?)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
mrsharm: 100% - adding now.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
mrsharm: Yeah, agreed - I think we are out of luck if we run into the aforementioned situation. The recommendation here is to run the benchmarks on the internal GC machines where this state will never arise.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/ASPNetBenchmarks/AspNetBenchmarksCommand.cs
mrsharm: As a heads up, added this check at the beginning of each benchmark run - this will take care of the issue with hitting the restart period without retrying twice.

mrsharm:
@cincuranet, this PR is ready to go - could this be merged? Thanks in advance!

cincuranet:
Done.

PR #3474, opened 2023-11-08
remove BinaryFormatter benchmarks, as it's soon going to be deleted

adamsitnik:
https://github.com/dotnet/announcements/issues/284


cincuranet:
@LoopedBard3 @DrewScoggins @caaavik-msft @sblom Given we want to run everything from `main`, I think `#ifdef`s (or similar) approach is going to be required here. Opinions?

GrabYourPitchforks (approved):

DrewScoggins:
> @LoopedBard3 @DrewScoggins @caaavik-msft @sblom Given we want to run everything from `main`, I think `#ifdef`s (or similar) approach is going to be required here. Opinions?

Yeah, given that desire to not maintain a bunch of branches for servicing branches we will need to do it that way if we want to continue to run these tests downlevel. I suppose there is another discussion we could have on the possibility of not running these as all of our currently supported frameworks had the feature deprecated, but I am not sure how we could be confidant we are not setting ourselves up for a nasty surprise.

adamsitnik:
From my point of view, we just want everyone to stop using `BinaryFormatter` ASAP. I doubt we are going to backport any changes to it (so we should not see any new regressions) and I doubt even more that we would fix perf regressions if there were any introduced by other backports. So IMO we should remove these benchmarks everywhere.

@GrabYourPitchforks you are leading the effort of removal of `BinaryFormatter`, what is your opinion on that?

DrewScoggins:
That point makes sense @adamsitnik.

GrabYourPitchforks:
> From my point of view, we just want everyone to stop using `BinaryFormatter` ASAP. I doubt we are going to backport any changes to it (so we should not see any new regressions) and I doubt even more that we would fix perf regressions if there were any introduced by other backports. So IMO we should remove these benchmarks everywhere.
> 
> @GrabYourPitchforks Levi Broderick FTE you are leading the effort of removal of `BinaryFormatter`, what is your opinion on that?

Correct. No further changes to the code are planned. We're not even fixing bugs (see https://github.com/dotnet/runtime/issues/90387). The code base is essentially dead.

PR #3479, opened 2023-11-09
Fix Blazor iOS Maui Package Failure

LoopedBard3:
Fix Blazor iOS Maui package failure that is part of what is preventing the dotnet-performance pipeline from green.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2311336&view=results


cincuranet (approved):
🎉

PR #3485, opened 2023-11-10
Fix startup affinity check bug and set bug

LoopedBard3:
Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2314387&view=results

This fixes a bug occuring when the ProcessorAffinity returns -1 (64+ processors), though there will need to be investigation into this is we every want to pass an affinity using more than 64 processors. This also fixes a bug where the affinity was not being properly set and displayed for startup scenario testing. 

FYI @kunalspathak, I am not sure how closely the PDN scenario using this is watched, but this may cause a change in the results.

### DIFF
--- a/src/tools/ScenarioMeasurement/Startup/Startup.cs
+++ b/src/tools/ScenarioMeasurement/Startup/Startup.cs
@@ -144,7 +144,7 @@ class Startup
             var currentProcessAffinity = Process.GetCurrentProcess().ProcessorAffinity;
             if (affinity > currentProcessAffinity.ToInt64())
             {
-                throw new ArgumentException(nameof(affinity) + " cannot be greater than the number of processors available to this process!");
+                throw new ArgumentException(nameof(affinity) + " cannot be greater than the number of processors available to this process! (Current process affinity: " + currentProcessAffinity.ToInt64() + " Target affinity: " + affinity + ")");
             }
             currentProcessAffinity = (IntPtr)affinity;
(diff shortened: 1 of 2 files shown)


kunalspathak:
Wondering what scenario was affected that led us discover this bug?

LoopedBard3:
> Wondering what scenario was affected that led us discover this bug?

It was found during investigation into why ampere machines were failing the affinity comparison.

src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -142,12 +142,13 @@ static void checkArg(string arg, string name)
         if (affinity > 0 && (OperatingSystem.IsWindows() || OperatingSystem.IsLinux()))
         {
             var currentProcessAffinity = Process.GetCurrentProcess().ProcessorAffinity;
-            if (affinity > currentProcessAffinity.ToInt64())
+            if (affinity > currentProcessAffinity && currentProcessAffinity != -1) // -1 means all processors TODO: Check if there is a more proper way to deal with affinity for systems with more than 64 processors
             {
-                throw new ArgumentException(nameof(affinity) + " cannot be greater than the number of processors available to this process!");
+                throw new ArgumentException(nameof(affinity) + " cannot be greater than the number of processors available to this process! (Current process affinity: " + currentProcessAffinity + " Target affinity: " + affinity + ")");
```
danmoseley, line 147: Interpolated string?

cincuranet (approved):

DrewScoggins (approved):

PR #3488, opened 2023-11-13
Expunged all references to GoodLinq

mrsharm:
Renamed GoodLinq to EagerLinq and added it as a part of the System.Linq namespace as an extension method.

### DIFF
--- a/src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/CPUAnalysis/MarkPhaseAnalysis.cs
+++ b/src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/CPUAnalysis/MarkPhaseAnalysis.cs
@@ -8,15 +8,15 @@ namespace GC.Analysis.API
         public static PlotlyChart[] ChartStatisticsOfMarkPhaseByType(this GCProcessData processData, int generation, MarkRootType type)
         {
             int heapCount = processData.Stats.HeapCount;
-            var generationGCs = processData.GCs.Where(gc => gc.Generation == generation);
+            var generationGCs = processData.GCs.EagerWhere(gc => gc.Generation == generation);
 
-            var maxPromoted = generationGCs.Select(gc => (gc.Number, gc.PerHeapMarkTimes.Max(gc => gc.Value.MarkPromoted[(long)type])));
-            var minPromoted = generationGCs.Select(gc => (gc.Number, gc.PerHeapMarkTimes.Min(gc => gc.Value.MarkPromoted[(long)type])));
-            var avgPromoted = generationGCs.Select(gc => (gc.Number, gc.PerHeapMarkTimes.Sum(gc => gc.Value.MarkPromoted[(long)type]) / heapCount));
+            var maxPromoted = generationGCs.EagerSelect(gc => (gc.Number, gc.PerHeapMarkTimes.Max(gc => gc.Value.MarkPromoted[(long)type])));
+            var minPromoted = generationGCs.EagerSelect(gc => (gc.Number, gc.PerHeapMarkTimes.Min(gc => gc.Value.MarkPromoted[(long)type])));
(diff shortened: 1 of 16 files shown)


src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/CPUAnalysis/MarkPhaseAnalysis.cs
```diff
@@ -98,11 +98,11 @@ public static PlotlyChart[] ChartStatisticsOfMarkPhaseByType(this GCProcessData
         public static PlotlyChart ChartAverageMarkPhaseTimeByMarkType(this GCProcessData processData, int generation, IEnumerable<MarkRootType> types)
         {
             int heapCount = processData.Stats.HeapCount;
-            var generationGCs = processData.GCs.Where(gc => gc.Generation == generation);
+            var generationGCs = processData.GCs.EagerWhere(gc => gc.Generation == generation);
 
             var layout = new Layout.Layout
             {
-                title = $"Average Mark Promoted Bytes for Gen{generation}",
+                title = $"EagerAverage Mark Promoted Bytes for Gen{generation}",
```
markples, line 105:
```suggestion
                title = $"Average Mark Promoted Bytes for Gen{generation}",
```

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/Common.cs
```diff
@@ -4,6 +4,70 @@
...
+            double sum = 0;
+            foreach(var sourceItem in data)
+            {
+                sum += (double)map(sourceItem);
+            }
+
+            return sum / count;
+        }
+    }
+}
```
markples, line 69: The standard Linq methods are already eager for these two.  These differ from Linq
- in their handling of null (no exception)
- Average(<empty>) returning NaN instead of throwing.  (though there are Linq overloads that takes enumerables of nullables, and if the enumerable has no nonnull values (empty enumerable is a degenerate case of this) returns `null`)
- calling `Count()` rather than counting during the `foreach` - if the enumerable is a List this won't matter, but for others it could enumerate twice
- Linq uses checked arithmetic
-

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/GCAnalysis/GCSummarization.cs
```diff
@@ -85,44 +85,44 @@ void AddStr(string c, object val)
...
+            AddStr("Gen2 Total Promoted MB - Blocking", processData.Gen2Blocking.EagerSum(gc => gc.PromotedMB));
 
             // Allocations
-            AddStr("Mean Size Before MB", processData.GCs.Average(gc => gc.HeapSizeBeforeMB));
+            AddStr("Mean Size Before MB", processData.GCs.EagerAverage(gc => gc.HeapSizeBeforeMB));
             AddStr("Mean Size After MB", processData.Stats.MeanSizeAfterMB);
 
             // Speeds
-            AddStr("Ephemeral Average Speed (MB/MSec)", (processData.Generations[0].TotalPromotedMB + processData.Generations[1].TotalPromotedMB) / (processData.Generations[0].TotalPauseTimeMSec + processData.Generations[1].TotalPauseTimeMSec));
+            AddStr("Ephemeral EagerAverage Speed (MB/MSec)", (processData.Generations[0].TotalPromotedMB + processData.Generations[1].TotalPromotedMB) / (processData.Generations[0].TotalPauseTimeMSec + processData.Generations[1].TotalPauseTimeMSec));
```
markples, line 105:
```suggestion
            AddStr("Ephemeral Average Speed (MB/MSec)", (processData.Generations[0].TotalPromotedMB + processData.Generations[1].TotalPromotedMB) / (processData.Generations[0].TotalPauseTimeMSec + processData.Generations[1].TotalPauseTimeMSec));
```

PR #3491, opened 2023-11-14
Add 8.0 to pipeline testing channels

LoopedBard3:
This adds 8.0 channel runs to all of the performance pipeline jobs. This is a first pass to start discussion on if there are specific ones we do/don't want running 8.0.


cincuranet (approved):

DrewScoggins (dismissed):
I think that we should turn them all on. If we end up seeing that certain ones are taking up too much time we can always back that out if we need. But my guess is that it shouldn't matter.

cincuranet (approved):

PR #3566, opened 2023-12-06
Fix typo in metric name for BDN.Extensions Duration.

LoopedBard3:
Fix typo in metric name for BDN.Extensions Duration. The MetricName was denoting that the value is in ms while we are actually returning ns. This was found while investigating why Duration was showing up as ms in PowerBI, made a note of this discrepancy in the PowerBI for now.

Starting as a draft so we can talk about this change as this seems like it may have impacts in our perfcommand code. Specifically CompareAggregate equals and potentially hashcode generation. Although I have not found any other obvious places were this change may cause breaks.


cincuranet:
There's a `TestCounterMetricName` in our ADX, but I think nobody is relying on that value.

caaavik-msft (approved):

LoopedBard3:
Took another look and I think this should be good to go in. I plan on merging it 12/18 (Monday) so we have the whole week to catch any potential side effects.

PR #3586, opened 2023-12-08
Fix Blazor Inner Loop Client SDK_Scenario

LoopedBard3:
Updated SurveyPrompt.razor file path to Pages as the new blazor wasm template does not have a Shared folder. The template we are using was updated in https://github.com/dotnet/aspnetcore/pull/50554 and had the Shared folder removed resulting in the file copy failing.


LoopedBard3:
Made an issue to track this negative value showing up in time on thread diff, but it appears to be rare, especially on dedicated hardware. https://github.com/dotnet/performance/issues/3601. Opening for approval.

cincuranet (approved):

PR #3630, opened 2023-12-15
Update from the OSX.1015 queues to the OSX.13 queues.

LoopedBard3:
Updating from OSX.1015 to OSX.13 includes updating the queue pointed to by the build matrix, updating the XHarnessCLIVersion, and updating the runner to support new changes. The specific runner updates include fixing the log time collection to be based on the time in the previous log + 1, starting with the current time; retrieving and using the device ID in the launch and kill commands, updating the app_pid regex if on iOS 17 or newer, and updating log validation to be more general for new log format.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2335913&view=results
This should be backwards compatible with the version check, but the OSX.1015 device is not currently being found in the queue. Both devices will be upgraded soon regardless.

### DIFF
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -2,15 +2,15 @@
 Module for running scenario tasks
 '''
 
-from genericpath import exists
 import sys
 import os
 import glob
 import re
 import time
-from datetime import datetime
 import json
 
+from genericpath import exists
+from datetime import datetime, timedelta
 from logging import exception, getLogger
 from collections import namedtuple
 from argparse import ArgumentParser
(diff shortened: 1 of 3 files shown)


src/scenarios/shared/runner.py
```diff
@@ -607,9 +607,22 @@ def run(self):
...
+            #  Connected Devices:
+            #    PERFIOS-01 00008101-001A09223E08001E    17.0.2        iPhone iOS
+            deviceInfoMatch = re.search(r'Connected Devices:\s+(?P<deviceName>\S+)\s+\S+\s+(?P<deviceVersion>\S+)', apple_state.stdout)
+            if deviceInfoMatch:
+                deviceName = deviceInfoMatch.group('deviceName')
+                deviceVersion = deviceInfoMatch.group('deviceVersion')
+                getLogger().info(f"Device Name: {deviceName}")
+                getLogger().info(f"Device Version: {deviceVersion}")
+            else:
+                raise Exception("Device name or version not found in the output of apple_state command.")
```
kotlarmilos, line 624: Should the script continue even when the device details are unknown? I wonder if this is a valid reason to stop the execution.

kotlarmilos (approved):
LGTM!

cincuranet (approved):
Do we maybe want to plug in `MicrosoftDotNetXHarnessCLIVersion` into Maestro/DARC?

akoeplinger (approved):

src/scenarios/shared/runner.py
LoopedBard3: The device name at least needs to be available as it is now required for the launch and kill commands. The version is also used for now to determine the app launch line to look for as they changed slightly sometime between the previous iOS version and iOS 17 or between the XHarness versions.

PR #3653, opened 2023-12-18
Backport iOS17/OSX13 updates to release/6.0.

LoopedBard3:
Backport of https://github.com/dotnet/performance/pull/3630. Specifically to continue support for the runtime release/6.0 branch as we no longer run iOS in the performance runs.


LoopedBard3:
Closing as further investigation turned up that we only do SOD testing on runtime release/6.0 and don't use the iphone queues for that testing. As such, there is no need to backport the update.

PR #3768, opened 2024-01-16
Include the tzinfo in the start time argument when getting the logs from iOS devices

LoopedBard3:
Include the tzinfo in the start time argument when getting the logs from iOS devices. This will fix an error where not having the timezone info is causing more logs then expected to be grabbed.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2354941&view=results


DrewScoggins (approved):
LGTM

LoopedBard3:
Errors are unrelated as the change only impacts iOS device startup testing. Merging.

PR #3778, opened 2024-01-18
Update maui runs to net9.0

LoopedBard3:
This adds the Maui net9.0 runs to the internal performance runs as Maui now supports net9.0 runs, disables net8.0 runs as we work through fixing the net8.0 builds, and removes HybridGlobalization for iOS testing as it is now enabled by default.

Test run here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2356259&view=results

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -368,7 +368,7 @@ jobs:
         projectFile: maui_scenarios_android.proj
         dotnetVersionsLinks:
           9.0: https://maui.blob.core.windows.net/metadata/sdks/net9.0.json
-          # 8.0: https://aka.ms/dotnet/sdk/maui/net8.0.json # Use 8.0 key and net8.0 link until Maui is ready for net9.0
+          # 8.0: https://aka.ms/dotnet/sdk/maui/net8.0.json # 8.0 build flow is failing, disable temporarily https://github.com/dotnet/performance/issues/3655
 
   # Maui iOS Mono scenario benchmarks
   - template: /eng/performance/build_machine_matrix.yml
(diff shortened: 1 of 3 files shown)


cincuranet (approved):

azure-pipelines.yml
```diff
@@ -367,7 +367,8 @@ jobs:
         kind: maui_scenarios_android
         projectFile: maui_scenarios_android.proj
         dotnetVersionsLinks:
-          8.0: https://aka.ms/dotnet/sdk/maui/net8.0.json # Use 8.0 key and net8.0 link until Maui is ready for net9.0
+          9.0: https://maui.blob.core.windows.net/metadata/sdks/net9.0.json
```
jonathanpeppers, line 370: @mattleibow can someone create the aka.ms links for these? I don't think I can create them without "owning" the URL.

eng/performance/maui_scenarios_ios.proj
```diff
@@ -151,16 +149,17 @@
           sign MauiBlazoriOSDefault.app
 
           # Testing commands
-          $(Python) test.py devicestartup --device-type ios --package-path MauiBlazoriOSDefault.app --package-name net.dot.mauiblazortesting --scenario-name "%(Identity)" --use-fully-drawn-time --fully-drawn-magic-string __MAUI_Blazor_WebView_OnAfterRender__ --startup-iterations 5
+          $(Python) test.py devicestartup -1-device-type ios -1-package-path MauiBlazoriOSDefault.app -1-package-name net.dot.mauiblazortesting -1-scenario-name "%(Identity)" -1-use-fully-drawn-time -1-fully-drawn-magic-string __MAUI_Blazor_WebView_OnAfterRender__ -1-startup-iterations 5
```
jonathanpeppers, line 152: I know nothing about python, what does this change do?
```diff
-- --device-type
++ -1-device-type
```

jonathanpeppers (approved):
Until we get aka.ms links, the URLs you used should work for now. 👍

eng/performance/maui_scenarios_ios.proj
LoopedBard3: It is not clear in the compare, but this is part of a larger comment out block and the xml gets unhappy when there is --'s inside a larger block comment. So, this is just something to stop the comment from causing errors/warnings.

PR #3779, opened 2024-01-18
For the AoT scenarios, always invoke with allowing to be used with a standalone gc

mrsharm:
Note: In addition to passing this build arg, the user must specify the GCName via ``DOTNET_GCName`` and not ``COMPlus_GCName`` i.e., ``DOTNET_GCName: clrgc.dll``.


markples:
Is there a way to fail if `complus_` is used instead of `dotnet_`?  (or maybe lack of `dotnet_` already fails?)  (I'm not sure what behavior I'm actually asking for other than "I don't want to mess up a run"!)

mrsharm:
> Are multiple `--application.buildArguments` supported? At
> 
> https://github.com/dotnet/performance/blob/f2e017e804f13d303a24f6147e998dfeb94f72ca/src/benchmarks/gc/GC.Infrastructure/Configurations/ASPNetBenchmarks/ASPNetBenchmarks-All.csv#L17
> 
> , multiple arguments are combined into one `buildArguments`.

Yes, definitely supported: 

  --[JOB].buildArguments <argument>                                        An argument to pass to msbuild. Can be used multiple times to define multiple values.

Source: https://github.com/dotnet/crank/blob/main/src/Microsoft.Crank.Controller/README.md#usage 



> Is there a way to fail if complus_ is used instead of dotnet_? (or maybe lack of dotnet_ already fails?) (I'm not sure what behavior I'm actually asking for other than "I don't want to mess up a run"!)

If the environment variable isn't specified, there are no seeming implications i.e., failures. Unfortunately, no easy way to do this unless we look out for the name of the benchmark and check if it contains "AOT" in it and then fail if the environment variables contain any COMPlus vars. Can work on this independent to the PR.

Maoni0:
another option - are folks ok with just to making it a simple policy that we always specify env vars with DOTNET_, instead of COMPlus_? so the infra can just check for it and tell you to always use DOTNET_ if you are not. I'd be ok with that, just so that I don't need to care about this distinction between nativeaot and not.

mrsharm:
> another option - are folks ok with just to making it a simple policy that we always specify env vars with DOTNET_, instead of COMPlus_? so the infra can just check for it and tell you to always use DOTNET_ if you are not. I'd be ok with that, just so that I don't need to care about this distinction between nativeaot and not.

This is the best option imo - considering we're probably not doing much testing before .NET 6. I can work on this if you all are fine with this - please let me know if you have any objections: @markples @cshung.

cshung (approved):

mrsharm:
Hi @cincuranet, can we merge this PR? Seems like the CI runs are still red.

cincuranet:

PR #3795, opened 2024-01-22
Fix secret sync

DrewScoggins:
Updating the secret sync to grab the SDK from the global.json instead of a fixed version. We had to make this update to fix this run after https://github.com/dotnet/performance/pull/3737


cincuranet (dismissed):

cincuranet (approved):

cincuranet (approved):

DrewScoggins:
This is a stop gap until we can complete https://github.com/dotnet/performance/issues/3824

PR #3798, opened 2024-01-22
process compute as a value on cmdline; get rid of unsafe code

Maoni0:
compute is already processed as an int when you specify it in a file; should be the same for cmdline.


mrsharm (approved):
LGTM!

Maoni0:
@cincuranet could you merge please? also, any news on when this broken CI would be fixed? if it keeps on going, can you change the policy?

cincuranet:
@Maoni0 We're actively looking at the failures.

PR #3806, opened 2024-01-23
Consolidate yaml/script/proj logic into a single python script that is reusable across all contexts

caaavik-msft:
This PR is the first part of two PRs (the other one will follow in the runtime repository after this gets merged), in which we consolidate all our performance testing infrastructure code into python scripts that can be called locally and from our CI pipelines. This PR does not have everything needed to be able to run against local versions of the runtime repository, but it has support for every performance test defined in the performance repository which runs against .NET installed by the installer.

In particular of note, you can now see that the scenarios.yml and benchmark_jobs.yml are now much simpler, with the only logic now being to check out the repository, build the arguments for the python script and run it, then send the job to helix. All the logic that used to be in these files is now located in `run_performance_job.py`.

Another key aspect worth mentioning is that we have decided to keep the `send_to_helix.yml` step in for now in the pipeline to have distinct CI steps, however the python script does have support to send to helix itself which is useful when running locally. 

Helix projects take all their arguments in via environment variables, and until now it was not easy to see all the different environment variables that are necessary for it to run. `send_to_helix.yml` provided a list of some of these environment variables and took them in from yaml parameters to just pass into `env`, but the list was not exhaustive and we relied upon environment variables being defined ambiently from the environment or from other pipeline steps and ci_setup.py. With this change, I have an exhaustive list of all the needed environment variables defined in `send_to_helix.py` defined as a dataclass. I then also provide a helper function `set_environment_variables` which will lets you either save these variables to the calling CI pipeline, or inside the python script processes environment variables.

Additionally, having scenarios and microbenchmarks defined inside the same python script means that we are able to combine the pre-command logic which until now has been duplicated across different yaml files, scripts, and proj files into a single function which is commented to explain what each precommand is for.

Lastly, I modified the `helix.proj` file to simplify it significantly so that it's logic is moved to the python script as well where it is much easier to debug as a developer. However, it was not possible to move the logic out of the scenarios project files at this time due to them creating multiple `HelixWorkItem`s of different types, and that was too complicated to encode inside an environment variable.


cincuranet (dismissed):
Looks good. And we can finally debug/run the whole "pipeline" locally! 💪

I would maybe split some of the long methods (i.e. `get_pre_commands`) into multiple smaller one for better orientation. But I understand (I guess) that for now we want to have some kind of correlation between old yamls/scripts to check if something is not behaving as expected. It can be done as a next step later, when all the dust settles.

scripts/run_performance_job.py
```diff
@@ -336,72 +683,94 @@ def run_performance_job(args: RunPerformanceJobArgs):
...
+            dotnet_executable_path, "publish", 
+            "-c", "Release", 
+            "-o", os.path.join(payload_dir, "PerfLabGenericEventSourceForwarder"),
+            "-f", framework,
+            "-r", runtime_id,
+            os.path.join(args.performance_repo_dir, "src", "tools", "PerfLabGenericEventSourceForwarder", "PerfLabGenericEventSourceForwarder", "PerfLabGenericEventSourceForwarder.csproj"),
+            "-p:DisableTransitiveFrameworkReferenceDownloads=true"]).run()
+        
+        # build PerfLabGenericEventSourceLTTngProvider
+        if args.os_group != "windows" and args.os_group != "osx" and args.os_version == "2204":
```
cincuranet, line 706: Is this check correct? Because even in the latest run I don't see `PerfLabGenericEventSourceLTTngProvider` compiled.

caaavik-msft: I believe so, this is what the current yaml looks like that calls this: https://github.com/dotnet/performance/blob/1d11344caab47e35dcf38d5f17dacb924b35fcba/eng/performance/scenarios.yml#L258

This line has two conditions on it in the yaml:
`- ${{ if and(ne(parameters.osName, 'windows'), ne(parameters.osName, 'osx')) }}:`
`condition: and(succeeded(), contains(variables['System.JobDisplayName'], '2204'))`

So this is the combination of those two. Is there a specific pipeline run you were expecting to see it in that it isn't being run in?

caaavik-msft: Ah, my bad, there is an error there. The check for os_version is not correct.

cincuranet: Is the `os_version` really `2204` and not like `22.04`?

The most recent run here is still failing on _ubuntu 2204 x64 scenarios_ and I don't see `PerfLabGenericEventSourceLTTngProvider` compilation in output...

caaavik-msft:
The CI failures are caused by the global.json changes made recently. They are only modifying the global.json defined in the original version of the performance repository, rather than inside the copy in the `CorrelationStaging` directory. I'll see if I can find a nice way to have this all work.

caaavik-msft:
> Looks good. And we can finally debug/run the whole "pipeline" locally! 💪
> 
> I would maybe split some of the long methods (i.e. `get_pre_commands`) into multiple smaller one for better orientation. But I understand (I guess) that for now we want to have some kind of correlation between old yamls/scripts to check if something is not behaving as expected. It can be done as a next step later, when all the dust settles.

To add on to this, once the runtime repo starts using this script as well, we can also start refactoring ci_setup.py as the runtime repo won't be calling it directly.

LoopedBard3:
All the failures aside from the sdk_scenarios appears to be failing due to being unable to find package(s): 
`error NU1102: Unable to find package <package> with version (>= 9.0.0-preview.2.24101.5)`

The sdk_scenarios seems to be because LTTNG failed during install:
```
[2024/02/01 21:36:15][INFO] Installing perf_event packages.
[2024/02/01 21:36:15][INFO] Installing LTTng packages.
[2024/02/01 21:36:15][INFO] 
[2024/02/01 21:36:15][INFO] tput: unknown terminal "unknown"
[2024/02/01 21:36:15][INFO] tput: unknown terminal "unknown"
[2024/02/01 21:36:15][INFO] E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
[2024/02/01 21:36:15][INFO] E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
[2024/02/01 21:36:15][INFO] tput: unknown terminal "unknown"
[2024/02/01 21:36:15][INFO] tput: unknown terminal "unknown"
[2024/02/01 21:36:15][INFO] E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
[2024/02/01 21:36:15][INFO] E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
[2024/02/01 21:36:15][INFO] 
[2024/02/01 21:36:16][INFO] LTTng not installed.
[2024/02/01 21:36:16][INFO] Run ./perfcollect install
[2024/02/01 21:36:16][INFO] or install LTTng manually.
```

LoopedBard3:
Looking at the AkadeIndexResults and comparing it to a normal run in performance-ci, it looks like the csproj is not being properly setup/selected. The test run here used mircobenchmarks resulting in 15 tests while the normal run only had 1. Here are the two pipeline runs I compared: Test https://dev.azure.com/dnceng-public/public/_build/results?buildId=549996&view=results, and normal: https://dev.azure.com/dnceng-public/public/_build/results?buildId=550003&view=results.

caaavik-msft:
> All the failures aside from the sdk_scenarios appears to be failing due to being unable to find package(s): `error NU1102: Unable to find package <package> with version (>= 9.0.0-preview.2.24101.5)`
> 
> The sdk_scenarios seems to be because LTTNG failed during install:
> 
> ```
> [2024/02/01 21:36:15][INFO] Installing perf_event packages.
> [2024/02/01 21:36:15][INFO] Installing LTTng packages.
> [2024/02/01 21:36:15][INFO] 
> [2024/02/01 21:36:15][INFO] tput: unknown terminal "unknown"
> [2024/02/01 21:36:15][INFO] tput: unknown terminal "unknown"
> [2024/02/01 21:36:15][INFO] E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
> [2024/02/01 21:36:15][INFO] E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
> [2024/02/01 21:36:15][INFO] tput: unknown terminal "unknown"
> [2024/02/01 21:36:15][INFO] tput: unknown terminal "unknown"
> [2024/02/01 21:36:15][INFO] E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
> [2024/02/01 21:36:15][INFO] E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
> [2024/02/01 21:36:15][INFO] 
> [2024/02/01 21:36:16][INFO] LTTng not installed.
> [2024/02/01 21:36:16][INFO] Run ./perfcollect install
> [2024/02/01 21:36:16][INFO] or install LTTng manually.
> ```

This issue seems to be occurring on other runs too (e.g. build 20240201.23), so I think it is a separate issue unrelated to my PR.

cincuranet:
>This issue seems to be occurring on other runs too (e.g. build 20240201.23), so I think it is a separate issue unrelated to my PR.

Very likely. It will solve itself.

scripts/run_performance_job.py
```diff
@@ -421,56 +802,132 @@ def run_performance_job(args: RunPerformanceJobArgs):
...
-def main(argv: list[str]):
+def main(argv: List[str]):
     args: dict[str, Any] = {}
 
     i = 1
     while i < len(argv):
         key = argv[i]
         bool_args = {
             "--internal": "internal",
+            "--physical-promotion": "physical_promotion_run_type",
```
DrewScoggins, line 921: Is there a reason we are doing our own arg parsing here instead of using the arg parsing library?

scripts/run_performance_job.py
```diff
@@ -50,258 +50,613 @@ def output_counters_for_crank(reports: list[Any]):
...
         helix_pre_commands += ["export MSBUILDDEBUGCOMM=1", 'export "MSBUILDDEBUGPATH=$HELIX_WORKITEM_UPLOAD_ROOT"']
+
+    # Copy the performance repo and root directory to the work item directory
+    if args.os_group == "windows":
+        helix_pre_commands += [ 
+            "robocopy /np /nfl /e %HELIX_CORRELATION_PAYLOAD%\\performance %HELIX_WORKITEM_ROOT%\\performance",
+            "robocopy /np /nfl /e %HELIX_CORRELATION_PAYLOAD%\\root %HELIX_WORKITEM_ROOT%" ]
+    else:
+        helix_pre_commands += [ 
+            "cp -R $HELIX_CORRELATION_PAYLOAD/performance/* $HELIX_WORKITEM_ROOT/performance",
```
LoopedBard3, line 241: This line seems to be failing at least in the ubuntu main akadeindex runs:
```
+ cp -R /datadisks/disk1/work/AFFE096D/p/performance/CODE-OF-CONDUCT.md /datadisks/disk1/work/AFFE096D/p/performance/Directory.Build.props /datadisks/disk1/work/AFFE096D/p/performance/Directory.Build.targets /datadisks/disk1/work/AFFE096D/p/performance/LICENSE.TXT /datadisks/disk1/work/AFFE096D/p/performance/NuGet.config /datadisks/disk1/work/AFFE096D/p/performance/README.md /datadisks/disk1/work/AFFE096D/p/performance/azure-pipelines.yml /datadisks/disk1/work/AFFE096D/p/performance/docs /datadisks/disk1/work/AFFE096D/p/performance/eng /datadisks/disk1/work/AFFE096D/p/performance/global.json /datadisks/disk1/work/AFFE096D/p/performance/global.net8.json /datadisks/disk1/work/AFFE096D/p/performance/helix.yml /datadisks/disk1/work/AFFE096D/p/performance/requirements.txt /datadisks/disk1/work/AFFE096D/p/performance/scripts /datadisks/disk1/work/AFFE096D/p/performance/src /datadisks/disk1/work/AFFE096D/p/performance/tools /datadisks/disk1/work/AFFE096D/w/B65309F1/e/performance
cp: target '/datadisks/disk1/work/AFFE096D/w/B65309F1/e/performance' is not a directory
```

The runs still seems to succeed though. Looking at the logs and the directories for everything, is there a reason we try to copy this stuff to the HELIX_WORKITEM_ROOT? We seem to use the HELIX_WORKITEM_ROOT in the python path and it seems to be the default working directory, but all our other calls use the HELIX_CORRELATION_PAYLOAD for running.

LoopedBard3: This is not something we have to decide on for this PR, but we should double check this logic and can potentially simplify part of our pipeline in the near future.

scripts/run_performance_job.py
```diff
@@ -50,258 +50,613 @@ def output_counters_for_crank(reports: list[Any]):
...
+    else:
+        helix_pre_commands += ["export ORIGPYPATH=$PYTHONPATH"]
 
-def run_performance_job(args: RunPerformanceJobArgs):
-    if args.project_file is None:
-        args.project_file = os.path.join(args.performance_repo_dir, "eng", "performance", "helix.proj")
+    # Allow using OpenSSL 1.0.2: https://github.com/dotnet/runtime/pull/60728 
+    # TODO: Is this still needed? 
+    if args.os_group != "windows":
+        helix_pre_commands += ["export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true"]
```
LoopedBard3, line 123: At least in current performance micro testing (akadeindexset) on ubuntu, this is not being exported based on the run logs.

LoopedBard3:
A few minor differences I noticed between build using this and one using the old pipeline (performance-ci/public):
Lines missing from new runs:
* `sudo apt-get -y install python3-venv`, We only used the venv's internally and it looks like an internal run would install and use the venv as expected 👍.
* `$HELIX_PYTHONPATH $HELIX_CORRELATION_PAYLOAD/reporter/run.py ...` at the end of the run. I wasn't able to find where this is added from in the perf repo. It is part of the execute.sh rather than benchmarks_ci.py. May be worth looking into.

Lines added in the new runs:
* `export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true`. Appears to originally be internal only, though has a TODO to check if we need it and I don't forsee any issues with having it in public runs. 👍
* `export MSBUILDDEBUGCOMM=1; export "MSBUILDDEBUGPATH=$HELIX_WORKITEM_UPLOAD_ROOT"`. Don't see this at all in original flow, was it added for testing? I don't think it will cause any issues but it would be good to know why it was added. 
* `cp -R $HELIX_CORRELATION_PAYLOAD/performance/* $HELIX_WORKITEM_ROOT/performance; cp -R $HELIX_CORRELATION_PAYLOAD/root/* $HELIX_WORKITEM_ROOT`. /performance already brought up here: https://github.com/dotnet/performance/pull/3806#discussion_r1476567908. /root/* seems to only copy machine setup to the place we do call machine-setup from. May be worth investigating separately if we can remove this copy and use the machine-setup already downloaded.

scripts/run_performance_job.py
LoopedBard3: Digging deeper, it looks like this originally comes from helix.proj and was only added when UseCoreRun was true. It may be worth moving or wrapping these file copies with a check for UseCoreRun for consistency, although I don't think it will directly affect the performance runs if we keep it aside from the failure to copy.

It also seems that the additional /* is causing the command to expand rather than using `/performance/*`, which combined with not having a / at the end of the target directory is causing the error. An example from the dotnet-runtime-perf repo of the call is `+ cp -R <PAYLOAD>/performance <WORKITEM>/performance`.

scripts/run_performance_job.py
caaavik-msft: I went with this approach as argparse is not compatible with python type hints/dataclasses, which I think is more valuable here. With this your python editor will be able to give you type errors if you are not using your args correctly.

scripts/run_performance_job.py
caaavik-msft: This change is actually something new I am introducing as part of this script, and is now required for all runs, not just corerun runs. I had forgotten to mention it in the original post. 

Every helix job has one work item directory (which Helix will upload and then copy into the working directory on the host machine), and N correlation payload directories. My understanding is that when we first started out, we always just made the work item directory equal to the performance directory. Then eventually we found that this caused issues on Windows because we would be logging inside the performance directory at the same time as we were uploading it, which caused the helix publish to hang. As a result, we made it so that the performance directory would be copied to another folder, "notLocked" and then once it gets into Helix it would be copied back to the work item root again. Similarly, in the runtime repository it also does something similar except it has to upload some additional correlation payloads for coreroot. With this change, I've made it so that the performance repository will always be copied into `CorrelationPayload/performance` on the local machine, which will then get uploaded to a `performance` directory in Helix, and then this line will copy it back into the work item root.

As for the issue you raised, it turns out that even though I attempted to port the code that would copy the correlation payload to the work item root, I had left the code that referenced the performance repository still pointing to the helix correlation payload. I have just pushed a fix now that ensures that the target directory in the work item root exists and that it actually gets called from there. I think that even though it did work without this copy happening, it still probably needs to happen so that the custom nuget config that we copy into the work item root gets referenced when the project is built as part of the benchmark.

scripts/run_performance_job.py
caaavik-msft: I think I will leave it in for now, but in a future refactor of this file we can remove it.

caaavik-msft:
> * `sudo apt-get -y install python3-venv`, We only used the venv's internally and it looks like an internal run would install and use the venv as expected 👍.

This was an intentional change I made in this PR. My understanding is the only reason we didn't use a venv before on non-internal runs was just because they weren't installing any python packages. I think just to make the experience of a PR build as similar as possible to an internal run, having everything run in a venv seems like a safer choice.

> * `$HELIX_PYTHONPATH $HELIX_CORRELATION_PAYLOAD/reporter/run.py ...` at the end of the run. I wasn't able to find where this is added from in the perf repo. It is part of the execute.sh rather than benchmarks_ci.py. May be worth looking into.

Ah, thanks for catching this. This is caused by SYSTEM_ACCESSTOKEN not being propagated to send-to-helix.yml which causes arcade not to launch the azure pipelies reported

> * `export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true`. Appears to originally be internal only, though has a TODO to check if we need it and I don't forsee any issues with having it in public runs. 👍
> * `export MSBUILDDEBUGCOMM=1; export "MSBUILDDEBUGPATH=$HELIX_WORKITEM_UPLOAD_ROOT"`. ~Don't see this at all in original flow, was it added for testing?~ I don't think it will cause any issues but it would be good to know why it was added. (update) Looks like this is present in the dotnet-runtime-perf runs so it is probably from somewhere in there. I still don't think having it will cause any issues so 👍.

These two lines were added just to consolidate the two repositories. I looked into what these environment variables do and really it's just additional logging so thought it couldn't hurt to enable it by default.

> * `cp -R $HELIX_CORRELATION_PAYLOAD/performance/* $HELIX_WORKITEM_ROOT/performance; cp -R $HELIX_CORRELATION_PAYLOAD/root/* $HELIX_WORKITEM_ROOT`. /performance already brought up here: [Consolidate yaml/script/proj logic into a single python script that is reusable across all contexts #3806 (comment)](https://github.com/dotnet/performance/pull/3806#discussion_r1476567908). /root/* seems to only copy machine setup to the place we do call machine-setup from. May be worth investigating separately if we can remove this copy and use the machine-setup already downloaded, especially since we also use different pythonpath paths ( vs ), at least in scenarios.

I've responded to this one here: https://github.com/dotnet/performance/pull/3806#discussion_r1476941879

scripts/run_performance_job.py
```diff
@@ -312,11 +667,15 @@ def run_performance_job(args: RunPerformanceJobArgs):
         else:
             runtime_id = f"linux-{args.architecture}"
 
+        dotnet_executable_path = os.path.join(ci_setup_arguments.install_dir, "dotnet")
+
+        os.environ["MSBUILDDISABLENODEREUSE"] = "1" # without this, MSbuild will be kept alive
+
         # build Startup
         RunCommand([
```
LoopedBard3, line 675: May be worth adding a binlog output here, something like from copied below: `f"/bl:{os.path.join(args.performance_repo_dir, 'artifacts', 'log', build_config, 'StartupBuild.binlog')}"]`. This would help with debugging why any build failures are occuring if we see them.

LoopedBard3: This expands to all of the build steps at this part of the script. This feedback comes from the first run here: https://dev.azure.com/dnceng-public/public/_build/results?buildId=552028&view=results having all net8.0 runs fail but not having any useful information about why they failed. Although rerunning the tests seems to have fixed that issue.

LoopedBard3:
I think the performance-ci testing is showing great results. @caaavik-msft, do we have any internal test runs yet?

caaavik-msft:
> Do we have any runs of this on the internal CI system yet?

I have run a few, here is the latest, it was having some issues because of the helix queue backlog though https://dev.azure.com/dnceng/internal/_build/results?buildId=2370506&view=results

LoopedBard3 (approved):
LGTM

LoopedBard3 (commented):
Currently looks good, will take another look at internal results once we have them.

LoopedBard3:
On an internal scenarios run (Device startup), it seems that the `reporter\run.py` call is missing. Not sure if the run I am using should have that fix (20240205.10).

PR #3815, opened 2024-01-24
Disallowing using COMPlus_ has the prefix for environment variables.

mrsharm:
Disallowing using COMPlus_ has the prefix for environment variables and therefore, this PR entails:

1. Updating all instances in the configurations. 
2. Adding checks that throw an exception in case COMPlus env vars are used. 
3. Updated readme.


src/benchmarks/gc/GC.Infrastructure/Configurations/ASPNetBenchmarks/ASPNetBenchmarks.yaml
```diff
@@ -2,12 +2,12 @@ runs:
   baseline:
     corerun: C:\CoreRuns\Empty\
     environment_variables:
-      COMPlus_GCServer: 1
+      DOTNET_GCServer: 1
```
Maoni0, line 5: actually this is 'gcServer' instead of 'GCServer`

src/benchmarks/gc/GC.Infrastructure/README.md
```diff
@@ -172,10 +172,10 @@ runs:
   run:
     corerun: C:\ASPNetUpload\clrgc.dll
     environment_variables:
-      COMPlus_GCName: clrgc.dll # This environment variable was set.
+      DOTNET_GCName: clrgc.dll # This environment variable was set.
 ```
 
-NOTE: For this case, ensure the environment variable ``COMPlus_GCName`` or ``DOTNET_GCName`` is set to clrgc.dll.
+NOTE: For this case, ensure the environment variable ``DOTNET_GCName`` or ``DOTNET_GCName`` is set to clrgc.dll.
```
Maoni0, line 178:
```suggestion
NOTE: For this case, ensure the environment variable ``DOTNET_GCName`` is set to clrgc.dll.
```

Maoni0 (approved):
otherwise LGTM

mangod9:
what is the motivation for this change? Just to keep things consistent?

mrsharm:
> what is the motivation for this change? Just to keep things consistent?

NativeAOT doesn't allow for COMPlus variables and we didn't want to add additional logic in the infra to differentiate. considering our testing is now .NET 6 and onwards, we decided to exclusively make use of DOTNET_*.

mangod9:
makes sense

mrsharm (commented):
Fixed all instances of GCServer -> gcServer

PR #3818, opened 2024-01-25
Setup flow to install net9.0 with older requested dotnet SDKs to support global.json

LoopedBard3:
Setup flow to install net9.0 with older requested dotnet SDKs to support global.json


LoopedBard3:
Closed in favor of: https://github.com/dotnet/performance/pull/3853

PR #3821, opened 2024-01-25
Exclude GC related commits from default CI

cincuranet:


mrsharm (approved):
LGTM - we'll use a similar pattern to come up with a few sanity checks on the GC Infra.

PR #3853, opened 2024-01-30
Fix 8.0 runs by replacing global.json with net8.0 compatible global.json

LoopedBard3:
Adds global.net8.json file that replaces the normal global.json file when ci_setup is ran and setting up for a net8.0/nativeaot8.0 framework run. This will fix our pipeline and make 8.0 testing work again (to enable in a different PR).

Test runs:
Performance (pre-nativeaot fix but with multiple 8.0 runs): https://dev.azure.com/dnceng/internal/_build/results?buildId=2364393&view=results 
Release/8.0 runtime: https://dev.azure.com/dnceng/internal/_build/results?buildId=2364454&view=results
Main runtime: https://dev.azure.com/dnceng/internal/_build/results?buildId=2365843&view=results

### DIFF
--- a/global.net8.json
+++ b/global.net8.json
@@ -1,0 +1,12 @@
+{
+  "tools": {
+    "dotnet": "8.0.100"
+  },
+  "msbuild-sdks": {
+    "Microsoft.DotNet.Arcade.Sdk": "9.0.0-beta.23580.7",
+    "Microsoft.DotNet.Helix.Sdk": "9.0.0-beta.23580.7"
+  },
+  "native-tools": {
+    "python3": "3.7.1"
+  }
+}
--- a/global.net8.json
+++ b/global.net8.json
@@ -1,10 +1,15 @@
 {
+  "sdk": {
+    "version": "8.0.101",
+    "allowPrerelease": true,
+    "rollForward": "latestMinor"
+  },
   "tools": {
-    "dotnet": "8.0.100"
+    "dotnet": "8.0.101"
   },
   "msbuild-sdks": {
-    "Microsoft.DotNet.Arcade.Sdk": "9.0.0-beta.23580.7",
-    "Microsoft.DotNet.Helix.Sdk": "9.0.0-beta.23580.7"
+    "Microsoft.DotNet.Arcade.Sdk": "8.0.0-beta.24059.4",
+    "Microsoft.DotNet.Helix.Sdk": "8.0.0-beta.24059.4"
   },
   "native-tools": {
     "python3": "3.7.1"
--- a/scripts/ci_setup.py
+++ b/scripts/ci_setup.py
@@ -349,7 +349,7 @@ def main(args: Any):
     # if repository is set, user needs to supply the commit_sha
     if not ((args.commit_sha is None) == (args.repository is None)):
         raise ValueError('Either both commit_sha and repository should be set or neither')
-    
+   
     # for CI pipelines, use the agent OS
     if not args.local_build:
         args.target_windows = sys.platform == 'win32'
@@ -368,7 +368,7 @@ def main(args: Any):
                 channel=args.channel,
                 verbose=verbose
             )
-            
+           
         init_tools(
             architecture=architecture,
             dotnet_versions=args.dotnet_versions,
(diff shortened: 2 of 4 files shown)


scripts/ci_setup.py
```diff
@@ -378,6 +379,15 @@ def main(args: Any):
     else:
         dotnet.setup_dotnet(args.dotnet_path)
 
+    framework = ChannelMap.get_target_framework_moniker(args.channel)
+    if framework == 'net8.0' or framework == 'nativeaot8.0':
+        # Copy the global.json file to global.bak.json
```
cincuranet, line 384: This comment and the copy below do not match - _global.bak.json_ vs _global.net9.json_.

scripts/ci_setup.py
```diff
@@ -378,6 +379,15 @@ def main(args: Any):
...
+        global_json_path = os.path.join(get_repo_root_path(), 'global.json')
+        shutil.copy(global_json_path, os.path.join(get_repo_root_path(), 'global.net9.json'))
```
cincuranet, line 386: Do we need to keep the "original" file? We could just blindly overwrite it with _net8_ file, right?

global.net8.json
```diff
@@ -0,0 +1,12 @@
+{
+  "tools": {
+    "dotnet": "8.0.100"
+  },
+  "msbuild-sdks": {
+    "Microsoft.DotNet.Arcade.Sdk": "9.0.0-beta.23580.7",
```
DrewScoggins, line 6: Shouldn't we be targeting the 8.0 version of Arcade here? Otherwise we might end up in the same boat where we need the 9.0 SDK to build some part of Arcade.

global.net8.json
```diff
@@ -0,0 +1,12 @@
+{
```
DrewScoggins, line 1: As currently put together this will not enforce the 8.0 SDK. This should be fine as we can only end up using this global.json when we have already downloaded the 8.0 SDK, but just want to make sure that we are being explicit about what we are doing.

scripts/ci_setup.py
```diff
@@ -378,6 +379,15 @@ def main(args: Any):
     else:
         dotnet.setup_dotnet(args.dotnet_path)
 
+    framework = ChannelMap.get_target_framework_moniker(args.channel)
+    if framework == 'net8.0' or framework == 'nativeaot8.0':
```
DrewScoggins, line 383: We don't need to add net6.0, and net7.0 because they are still in their own branches, right?

global.net8.json
cincuranet: We could probably add `rollForward` and explicitly disallow rolling forward on major and minor version.

scripts/ci_setup.py
LoopedBard3: Yes, we will need to keep this in mind when running locally, but we already have to keep the global.json in mind when running locally as most people probably don't run ci_setup.py when running locally.

scripts/ci_setup.py
LoopedBard3: Great catch, updated.

global.net8.json
LoopedBard3: Updated to match the versions specified in the dotnet/runtime release/8.0-staging global.json.

global.net8.json
LoopedBard3: I copied the sdk setup used in dotnet/runtime release/8.0-staging global.json. It uses version 8.0.101, allowPrerelease: true, and rollForward: major. I think this is the correct setup if we have the sdk set as it will use the latest feature found for 8.0.101, but will continue rolling forward to the closest newer sdk. If we want to block against accidentally updating to 9.*, I think using rollForward: "minor" would be correct.

cincuranet: With `rollForward: major` we need to make sure there's no .NET 9 SDK around (i.e. globally installed).

LoopedBard3: We could probably also change rollForward to latestMinor since we should be always getting the latest net8.0 version. Let me know what y'all think and I will test it out. (Going to test latestMinor)

cincuranet: SGTM.

LoopedBard3: This seems to have worked as expected.

cincuranet (approved):

PR #3863, opened 2024-01-31
Add additional benchmarks for reflection's FieldInfo get\set

steveharter:
For measuring changes from https://github.com/dotnet/runtime/pull/97784

### DIFF
--- a/src/benchmarks/micro/runtime/System.Reflection/Invoke.cs
+++ b/src/benchmarks/micro/runtime/System.Reflection/Invoke.cs
@@ -259,7 +259,7 @@ namespace System.Reflection
         {
             for (int i = 0; i < Iterations; i++)
             {
-                s_int = (int)s_staticField_int.GetValue(s_MyClass);
+                s_int = (int)s_staticField_int.GetValue(null);
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/runtime/System.Reflection/Invoke.cs
```diff
@@ -265,6 +327,33 @@ public void Field_Set_class()
             }
         }
 
+        [Benchmark(OperationsPerInvoke = Iterations)]
+        public void Field_SetStatic_class()
+        {
+            for (int i = 0; i < Iterations; i++)
+            {
+                s_staticField_class.SetValue(s_MyClass, 42);
```
buyaa-n, line 335: Just curious, for static fields if setting the value with null object make any difference for perf

buyaa-n (approved):
LGTM

src/benchmarks/micro/runtime/System.Reflection/Invoke.cs
steveharter: The perf is the same; it is not used or validated. Thanks for noticing that - will just change to null.

buyaa-n (approved):

PR #3870, opened 2024-02-01
Check for Empty ASP.NET Benchmark Filters + Documentation

mrsharm:


Maoni0 (dismissed):

mrsharm:
@cincuranet, could we please merge this? It's ready as the MD linting is fixed.

src/benchmarks/gc/GC.Infrastructure/README.md
```diff
@@ -217,6 +217,19 @@ It's worth noting that if you have specified Linux based binaries in the corerun
    1. Remove the ``crank`` prefix from the command line.
    2. Remove the ``--application.aspNetCoreVersion``, ``--application.runtimeVersion`` and ``--application.sdkVersion`` command args from the command line that you paste in the CSV as the versions are set by the infrastructure itself.
 
+###### How To Filter Benchmarks
+
+You can filter benchmarks of interest from the entire set of benchmarks specified by the referenced `benchmarks_file` using a list of regex patterns such as the following in the `benchmark_settings` section:
```
markples, line 222: wording nit - The use of `*` is a wildcard pattern rather than a regex.  I assume that `Foo*` is such a common pattern that you don't want to only support regexes (`Foo.*`), but it makes explanation difficult.  (see also the exception message and that the implementation doesn't handle something like `Foo*Bar*` as might be expected or `?` - I assume that you don't want to dive into more implementation on this, so maybe something like "regex or pattern with trailing * wildcard" is good enough?)

mrsharm: I agree - will update this in the next round of doc changes. Thanks!

PR #3878, opened 2024-02-02
Add benchmark to verify performance of dependency injection container validation on build

christiaanderidder:
This benchmark supports the work done in https://github.com/dotnet/runtime/pull/96254 based on the issue raised in https://github.com/dotnet/runtime/issues/96253


cincuranet:
Thanks for the contribution!

PR #3880, opened 2024-02-02
Type.GetType benchmarks

adamsitnik:
I would like to get it merged soon, so we start gathering historical data so once https://github.com/dotnet/runtime/pull/97864 gets merged we can know for sure how the performance has changed.

### DIFF
--- a/src/benchmarks/micro/libraries/System.Runtime/Perf.Type.cs
+++ b/src/benchmarks/micro/libraries/System.Runtime/Perf.Type.cs
@@ -56,6 +56,18 @@ namespace System.Tests
         [ArgumentsSource(nameof(FullyQualifiedNamesArguments))]
         public Type GetType_FullyQualifiedNames(Type input) => Type.GetType(input.FullName);
 
+        public IEnumerable<object> ResolverArguments()
+        {
+            yield return typeof(int); // elemental type
+            yield return typeof(int[]); // SZArray
+            yield return typeof(Nested); // nested type
+        }
+
+        [Benchmark]
+        [ArgumentsSource(nameof(ResolverArguments))]
+        public Type GetType_Resolvers(Type input)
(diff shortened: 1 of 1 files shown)


jkotas:
The time spent in `Type.GetType(string)` is dominated by stack walk required to get the caller (`Type.GetType(string)` is one of those poorly designed APIs that look at its caller) and by the time spent in assembly loader and type loader to find the type.  The raw type parsing does not contribute much. It tends to be noisy due to number of moving parts involved.

If you would like to measure the performance of raw type name parsing, it would be better to use the `Type.GetType` overload that takes `assemblyResolver` and `typeResolver` instead or in addition to `Type.GetType(string)`. You can provide cheap trivial implementation of these resolvers (e.g. the resolvers can always just return the same singleton instance).

jkotas:
The actual time spent time on type name parsing in `Type.GetType(string)` overload is less than 20% of the time. Benchmarking this overload is measuring a lot of unrelated stuff. 

Benchmarking the overload that takes callbacks would give you more accurate answer for the work that you are doing. The overload that takes callbacks has less noise and it is a thin wrapper over the parser.

Benchmarking the `Type.GetType(string)` is certainly better than nothing. I will leave up to you to decide whether to benchmark the other overload too.

jkotas (dismissed):

adamsitnik:
> Benchmarking the overload that takes callbacks would give you more accurate answer for the work that you are doing. 

@jkotas I've added benchmark coverage for that, PTAL.

jkotas (approved):
LGTM. Thank you!

Do you happen to know how much faster are the `GetType_Resolvers` numbers compared to full `GetType` numbers?

adamsitnik:
@DrewScoggins @sblom @caaavik-msft @LoopedBard3 could you please merge the PR (the failures are unrelated)? or give me the rights for merging on red if you don't mind ;)

PR #3934, opened 2024-02-14
Benchmarks_local disable terminating of all dotnet processes

LoopedBard3:
Initially, due to msbuild node reuse, the artifact build steps would occasionally fail due to files still being "in use" when building more than 1 commit for testing. To get around this the benchmarks local script terminated any potential dotnet file holders with the side affect of terminating other unrelated dotnet tasks on the machine. This commit updates the default of the script to disable msbuild Node reuse (found via https://stackoverflow.com/questions/13510465/the-mystery-of-stuck-inactive-msbuild-exe-processes-locked-stylecop-dll-nuget/13510466#13510466, thanks @caaavik-msft) which should keep this problem from occurring. A side effect of this change is that builds may be slower, so the option to enable node reuse is available.

The other primary update includes the adding of a force terminate option if we find edge cases where we still hit problems due to running processes. Minor updates include consolidating get_mono_corerun and forced cleaning of dotnet_mono folder to avoid a state with intermixed dotnet_mono folders from between runs.


cincuranet (approved):

LoopedBard3:
Failures are unrelated, merging.

PR #3946, opened 2024-02-15
Minor benchmarks_local.py cleanup

LoopedBard3:
Minor benchmarks_local.py cleanup based on lint warnings and updating to match latest python practices we have in files like run_performance_job.py. There should be no changes to the general functionality of the script.

### DIFF
--- a/scripts/benchmarks_local.py
+++ b/scripts/benchmarks_local.py
@@ -3,7 +3,7 @@ import os
 import platform
 import shutil
 import sys
-from typing import List
+from typing import List, Optional
 import xml.etree.ElementTree as xmlTree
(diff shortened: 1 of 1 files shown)


cincuranet:
For my own education... Why the removal of f-strings usage?

LoopedBard3:
> For my own education... Why the removal of f-strings usage?

Ultimately, it has to do with deferring the string interpolation until the logger decides it actually needs the string. The specific rule is: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html (W1203) and this blog post explains in better detail the reason with some performance reference: https://blogs.rahultgeorge.com/w1203-logging-fstring-interpolation-solved

PR #3967, opened 2024-02-20
Don't always write MAUI_VERSION and ensure version env vars have value

LoopedBard3:
Don't always write MAUI_VERSION and ensure version env vars have value. Only add entries to Additional data if they have values and don't write MAUI_VERSION env var in ci_setup if maui_version is not passed.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2388245&view=results
Will also need to update the ADX MauiVersion column generation.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -323,37 +323,37 @@ jobs:
       buildMachines:
         - win-x64
         - ubuntu-x64
-        # - win-arm64
-        # - ubuntu-arm64-ampere
+        - win-arm64
+        - ubuntu-arm64-ampere
       isPublic: false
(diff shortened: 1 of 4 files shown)


LoopedBard3:
Turns out that python always writes envvars as UPPERCASE on Windows which was causing some other wierd behavior: https://docs.python.org/3/library/os.html#os.environ. I added in a normalization to uppercase key values so it does it on all platforms.

LoopedBard3:
Also will need to update the ADX table update query with the matching MAUIVERSION casing.

cincuranet (approved):

Issue #3662, opened 2023-12-19
Temporarily disable AndroidBDN runs in runtime main


LoopedBard3:
AndroidBDN builds in runtime main are currently failing due to the UpdateXlf issue mentioned here: https://github.com/dotnet/performance/issues/3655. This is not affecting other apps as the AndroidBDN app is the only run in runtime that is dependent on Maui. This can be reenabled once we have the correct urls to point to for building against net9.0 for the AndroidBDN app/Maui.

The disabling PR is: https://github.com/dotnet/runtime/pull/96222

PR #3971, opened 2024-02-21
Add device restart in pre.py for android instrumentation runs

LoopedBard3:
We recently had to disable PerfBDN runs in runtime testing due to the app failing to install: https://github.com/dotnet/performance/issues/3662. A single device reboot would only temporarily fix the issue, quickly getting back into a failing state. By restarting the device each time before trying to run the test we ensure that we are able to download the app. This must be merged before the usage is added to the runtime runs.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2383951&view=results

### DIFF
--- a/src/scenarios/bdnandroid/pre.py
+++ b/src/scenarios/bdnandroid/pre.py
@@ -1,14 +1,16 @@
 '''
 pre-command
 '''
-import sys
 import os
-from zipfile import ZipFile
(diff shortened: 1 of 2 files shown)


cincuranet (approved):

LoopedBard3:
Failing tests are unrelated, merging.

PR #3973, opened 2024-02-21
Add `MemoryRandomization` attribute

DrewScoggins:
This checkin adds the `MemoryRandomization` attribute to a select set of benchmarks in the performance repo.

In order to determine which benchmarks should be selected the following procedure was used.

1. All tests were run in a seperate config with the `MemoryRandomization` attribute turned on.
2. 5 weeks of data from these runs was used to calculate an average coefficient fo variance (CoV) for the runs with the attribute
3. This was then compared against the CoV of the regular runs that happened on the same builds and machines.
4. Any test that had all 5 weeks of CoV improve by more than 30% with the attribute were included.


src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
```diff
@@ -44,6 +44,8 @@ public char GetPinnableReference()
         
         [Benchmark(OperationsPerInvoke = 16)]
         [ArgumentsSource(nameof(IndexOfStringArguments))]
+
```
LoopedBard3, line 47: This space seems unintentional?

src/benchmarks/micro/libraries/System.Runtime/Perf.String.cs
```diff
@@ -110,6 +115,8 @@ public string Substring_IntInt(string s, int i1, int i2)
         [Arguments("A B C D E F G H I J K L M N O P Q R S T U V W X Y Z", new char[] { ' ' }, StringSplitOptions.RemoveEmptyEntries)]
         [Arguments("ABCDEFGHIJKLMNOPQRSTUVWXYZ", new char[]{' '}, StringSplitOptions.None)]
         [Arguments("ABCDEFGHIJKLMNOPQRSTUVWXYZ", new char[]{' '}, StringSplitOptions.RemoveEmptyEntries)]
+
```
LoopedBard3, line 118: Space

src/benchmarks/micro/libraries/System.Text.Encodings.Web/Perf.Encoders.cs
```diff
@@ -21,12 +21,13 @@ public IEnumerable<object> GetEncoderArguments()
                 yield return new EncoderArguments("no <escaping /> required", size, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
                 yield return new EncoderArguments("hello \"there\"", size, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
                 yield return new EncoderArguments("&lorem ipsum=dolor sit amet", size, UrlEncoder.Default);
-                yield return new EncoderArguments("�2020", size, UrlEncoder.Default);
+                yield return new EncoderArguments("�2020", size, UrlEncoder.Default);
```
LoopedBard3, line 24: We should make sure this was not actually changed probably.

src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Common.cs
```diff
@@ -38,14 +38,19 @@ public void Setup()
...
-        [Benchmark] public void Email_IsNotMatch() => _email.IsMatch("yay.performance@dot.net#");
+        [Benchmark] [MemoryRandomization]
+public void OneNodeBacktracking() => _oneNodeBacktracking.Match("This regex has the potential to be optimized further");
+        [Benchmark] [MemoryRandomization]
+public void Email_IsMatch() => _email.IsMatch("yay.performance@dot.net");
+        [Benchmark] [MemoryRandomization]
+public void Email_IsNotMatch() => _email.IsMatch("yay.performance@dot.net#");
 
-        [Benchmark] public void Date_IsMatch() => _date.IsMatch("Today is 11/18/2019");
+        [Benchmark]         [MemoryRandomization]
```
LoopedBard3, line 48: Extra space

src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Common.cs
```diff
@@ -38,14 +38,19 @@ public void Setup()
...
+public void Date_IsMatch() => _date.IsMatch("Today is 11/18/2019");
         [Benchmark] public void Date_IsNotMatch() => _date.IsMatch("Today is 11/18/201A");
 
-        [Benchmark] public void IP_IsMatch() => _ip.IsMatch("012.345.678.910");
+        [Benchmark]         [MemoryRandomization]
```
LoopedBard3, line 52: Extra space

src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Common.cs
```diff
@@ -56,12 +61,15 @@ public void Setup()
         [Benchmark] public int MatchesWord() => _searchWord.Matches(_loremIpsum).Count;
         [Benchmark] public int MatchesWords() => _searchWords.Matches(_loremIpsum).Count;
 
-        [Benchmark] public Match MatchWord() => _searchWords.Match(_loremIpsum);
+        [Benchmark]         [MemoryRandomization]
```
LoopedBard3, line 64: Extra space

src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Common.cs
```diff
@@ -56,12 +61,15 @@ public void Setup()
...
+public Match MatchWord() => _searchWords.Match(_loremIpsum);
         [Benchmark] public string ReplaceWords() => _searchWords.Replace(_loremIpsum, "amoveatur");
         [Benchmark] public string[] SplitWords() => _searchWords.Split(_loremIpsum);
 
-        [Benchmark] public void Ctor() => new Regex(WarningPattern, Options);
-        [Benchmark] public void CtorInvoke() => new Regex(WarningPattern, Options).IsMatch(@"(1");
+        [Benchmark]         [MemoryRandomization]
```
LoopedBard3, line 69: Extra space

src/benchmarks/micro/libraries/System.Threading/Perf.Volatile.cs
```diff
@@ -14,6 +14,8 @@ public class Perf_Volatile
         private double _newValue = 1;
         
         [Benchmark]
+
```
LoopedBard3, line 17: Space

src/benchmarks/micro/runtime/Devirtualization/GuardedThreeClassInterface.cs
```diff
@@ -38,6 +38,8 @@ public class E : B, I
   
         [Benchmark(OperationsPerInvoke=TestInput.N)]
         [ArgumentsSource(nameof(GetInput))]
+
```
LoopedBard3, line 41: Space

src/benchmarks/micro/runtime/Linq/Linq.cs
```diff
@@ -193,6 +193,11 @@ public bool Where00ForX()
     #region Where01
 
     [Benchmark]
+#endregion
```
LoopedBard3, line 196: The endregion and region seem unnecessary + space

src/benchmarks/micro/runtime/System.Reflection/Attributes.cs
```diff
@@ -67,11 +67,13 @@ public class MyAttribute : Attribute
         [Benchmark(Description = "GetCustomAttributes - Method Base: Hit (inherit)")]
         public object[] GetCustomAttributesMethodBaseHitInherit() => AttributedBase.GetCustomAttributes(typeof(MyAttribute), true);
         [Benchmark(Description = "GetCustomAttributes - Method Base: Miss (inherit)")]
+[MemoryRandomization]
```
LoopedBard3, line 70: Double check formatting, I am not sure of correct formatting here.

LoopedBard3: Same below

cincuranet (approved):
~Do we have some process planned of how to keep the list of benchmarks that need randomization up-to-date (for example when new benchmarks are added)?~ I see @AndyAyersMS already asked basically the same question.

DrewScoggins:
> Happy to see this at long last.
> 
> Any thoughts on how this will get applied to new benchmarks? Will we need to redo the experiment every so often?

I have all of the tooling, and queries that I used to generate this data. My guess is that we should check that in, but I think that it should probably not go in this repo. We don't really have tooling here unless it is directly for measuring performance.

I will make a PR in our infra repo.

src/benchmarks/micro/libraries/System.Text.RegularExpressions/Perf.Regex.Common.cs
DrewScoggins: I am going to reformat these so they are not all on the same line. It will look cleaner, and match up with what we have everywhere else.

src/benchmarks/micro/runtime/System.Reflection/Attributes.cs
DrewScoggins: I also added additional line breaks here for readability.

LoopedBard3 (approved):

DrewScoggins:
Failures are all on 8.0 and unrelated to this change.

Issue #3984, opened 2024-02-22
Wasm runs failing with empty WasmAssembliesToBundle


LoopedBard3:
Starting on [02/19](https://dev.azure.com/dnceng/internal/_build/results?buildId=2383193&view=results), Wasm runs switched from hitting one failure (now seemingly fixed: https://github.com/dotnet/runtime/issues/98334) to hitting a new failure: `WasmApp.Common.targets(350,5): error : WasmAssembliesToBundle item is empty. No assemblies to process [/home/helixbot/work/9CCA0950/w/B41609A7/e/performance/artifacts/bin/for-running/MicroBenchmarks/Job-HFTUUL/BenchmarkDotNet.Autogenerated.csproj]` when building the BenchmarkDotNet.Autogenerated.csproj (WasmNestedPublishApp) -> PrepareInputsForWasmBuild Target.

This failure seems to correspond with when we [updated our BDN version](https://github.com/dotnet/performance/pull/3940) from '0.13.11-nightly.20231126.107' -> '0.13.13-nightly.20240213.132'. Digging into a binlog from a previous working run and a recent failing run, the TargetDir property now has /AppBundle at the end and the configuration in the path, the TargetDir is being set to the same path for all of the projects, and the Target PrepareForWasmBuild is not adding any items (.dlls). The path TargetDir path update seems to be from https://github.com/dotnet/BenchmarkDotNet/commit/e6fdc6b3c4d540983641f5aeca9c93308dc0da7a. Was able to repro with updated perf repo using command line: `sudo python3 ./scripths/benchmarks_local.py --commits cc95b08e0060d388c5a48268f85d454fbe8e64fe --run-types WasmInterpreter --filter *Span.IndexerBench.CoveredIndex2* *WriteReadAsync* --wasm-engine-path /<path>/.jsvu/engines/v8/v8 --rebuild-artifacts --reinstall-dotnet`

FYI: @adamsitnik, @ilonatommy

LoopedBard3:
[...]

PR #3991, opened 2024-02-23
Revert BDN version to one that works with wasm based on local testing.

LoopedBard3:
Revert BDN version to one that works with wasm based on local testing. This version still has the VS test adapter.
This will give us more time to work through: https://github.com/dotnet/performance/issues/3984


caaavik-msft (dismissed):

DrewScoggins (approved):
LGTM

PR #4031, opened 2024-03-06
Don't set branch on performance CI runs

caaavik-msft:
With the python pipelines changes, a bug was introduced that would set the branch for dotnet/performance CI runs when instead it should not be passed and then ci_setup will fill the branch value with data from the ChannelMap


DrewScoggins (approved):
LGTM

PR #4033, opened 2024-03-06
Add and fix up DependencyInjection CreateInstance() benchmarks

steveharter:
- Fixed up the benchmarks that said "DI picks this over the one below and calls". This will do a slight reset on these, but that is desired since they can't be compared against each other very well.
- Added benchmarks that use `[ActivatorUtilitiesConstructor]`. There is current work in DI that will improve perf here.
- Removed some unnecessary benchmarks that are either never used (0 args) or can be inferred by others (removed 2,4 args since they can be calculated by 1,3,5 args).


buyaa-n (approved):
LGTM

PR #4063, opened 2024-03-18
Add OpenTelemetry capabilities to python scripts

LoopedBard3:
Add OpenTelemetry capabilities to python logging system and tracing to some of our python scripts.


LoopedBard3:
Currently looking for feedback on how we want to enable/disable the console log and trace exporters. I think we will want to have the option to enable/disable them, or maybe have it be based on something like the IN_LAB env var, to keep the output clean.

scripts/performance/logger.py
```diff
@@ -14,9 +14,27 @@
...
+        self.logger_opentelemetry_imported = False
+
+    def set_initialized(self, value: bool): self.logger_initialized = value
+    def set_opentelemetry_imported(self, value: bool): self.logger_opentelemetry_imported = value
+    def get_initialized(self) -> bool: return self.logger_initialized
+    def get_opentelemetry_imported(self) -> bool: return self.logger_opentelemetry_imported
+
+logger_state_manager = LoggerStateManager()
+
+try:
```
DrewScoggins, line 29: When is this code executed? I am unfamiliar with loose code just sitting in a class definition.

LoopedBard3: This code gets executed when the file is imported, though the import will be cached so it is only run once when imported in the same way.

scripts/benchmarks_ci.py
```diff
@@ -228,6 +232,24 @@ def __is_valid_datetime(dt: str) -> str:
...
+    parser.add_argument(
+        '--enable-open-telemetry-logger',
+        dest='enable_open_telemetry_logger',
+        required=False,
+        default=False,
+        action='store_true',
+        help='Enables OpenTelemetry logging',
+    )
+
+    parser.add_argument(
```
DrewScoggins, line 244: It looks like we never actually use this argument any where. Am I missing something?

scripts/performance/tracer.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    A OpenTelemetry aware tracer implementation that is used as a wrapper for OpenTelemetry calls where OpenTelemetry is not guaranteed be installed.
+    When not installed, the tracer is a no-op and the decorated functions are executed as if the decorator was not there..
+    """
+    tracer = None
+    def __init__(self, name="dotnet.performance"):
+        if is_opentelemetry_imported():
+            self.tracer = trace.get_tracer(name)
+
+    def start_as_current_span(self, *top_args, **top_kwargs):
```
DrewScoggins, line 72: I would expect this to need to have the decorator syntax.

scripts/performance/tracer.py
```diff
@@ -0,0 +1,93 @@
...
+    from opentelemetry.sdk.trace import TracerProvider
+    from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
+    tracing_state_manager.set_opentelemetry_imported(True)
+
+except ImportError:
+    pass
+
+def setup_tracing():
+    '''Set up the OpenTelemetry trace provider.'''
+    if tracing_state_manager.get_initialized() or not tracing_state_manager.get_opentelemetry_imported():
```
DrewScoggins, line 29: We are checking here to see if the `tracing_state_manager` has been initialized, but from this PR, I don't see us ever setting that value. So this will always be false. Is that intended?

DrewScoggins:
Do we have a plan to take calls to the current logger and also use them as part of the OTel information we capture?

DrewScoggins (approved):
LGTM. On the question of how we want to enable this, checking IN_LAB is probably a good way for the automation to decide what command line args to pass to `benchmark_ci.py`

LoopedBard3:
> LGTM. On the question of how we want to enable this, checking IN_LAB is probably a good way for the automation to decide what command line args to pass to `benchmark_ci.py`

That makes sense, I fully expect some changes to be made once we find weak points or features that are missing in this PR as we start using the OTel logging more.

DrewScoggins (approved):

PR #4079, opened 2024-03-22
Add benchmark for WinHttpHandler

brporter:
This adds a new benchmark for WinHttpHandler. It's essentially a copy of the existing SocketsHttpHandler, but for WinHttpHandler.

I've just copied the existing SocketsHttpHandler benchmark and modified it ever so slightly to use WinHttpHandler instead. I've also added a NuGet package reference for WinHttpHandler.

Fix for issue: #4080

### DIFF
--- a/src/benchmarks/micro/libraries/System.Net.Http/Perf.WinHttpHandler.cs
+++ b/src/benchmarks/micro/libraries/System.Net.Http/Perf.WinHttpHandler.cs
@@ -37,6 +37,10 @@ namespace System.Net.Http.Tests
         [GlobalSetup]
         public void Setup()
         {
+#if DEBUG
+            System.Diagnostics.Debugger.Launch();
+#endif
+
             _serverCert = Test.Common.Configuration.Certificates.GetServerCertificate();
             _listener = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
 
@@ -62,7 +66,7 @@ namespace System.Net.Http.Tests
                                 if (ssl)
                                 {
                                     var sslStream = new SslStream(stream, false, delegate { return true; });
-                                    await sslStream.AuthenticateAsServerAsync(_serverCert, false, SslProtocols.None, false);
+                                    await sslStream.AuthenticateAsServerAsync(_serverCert, false, SslProtocols.Tls12, false);
                                     stream = sslStream;
                                 }
 
--- a/src/benchmarks/micro/libraries/System.Net.Http/Perf.WinHttpHandler.cs
+++ b/src/benchmarks/micro/libraries/System.Net.Http/Perf.WinHttpHandler.cs
@@ -106,46 +106,6 @@ namespace System.Net.Http.Tests
             }
         }
 
-        [Benchmark]
-        [BenchmarkCategory(Categories.NoAOT)]
-        public async Task Get_EnumerateHeaders_Validated()
-        {
-            HttpMessageInvoker invoker = _invoker;
-            HttpRequestMessage req = _request;
-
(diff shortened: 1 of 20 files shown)


src/benchmarks/micro/libraries/System.Net.Http/Perf.WinHttpHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+            using (HttpResponseMessage resp = await invoker.SendAsync(req, CancellationToken.None))
+            using (Stream respStream = await resp.Content.ReadAsStreamAsync())
+            {
+                await respStream.CopyToAsync(Stream.Null);
+            }
+        }
+
+        [Benchmark]
+        [BenchmarkCategory(Categories.NoAOT)]
+        public async Task Get_EnumerateHeaders_Validated()
```
MihaZupan, line 111: I would drop the header benchmarks. The implementation that's being tested is completely separate from `WinHttpHandler` (it'll be the same with `SocketsHttpHandler`).

brporter: Sounds good, resolved.

src/benchmarks/micro/libraries/System.Net.Http/Perf.WinHttpHandler.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using System.Runtime.Versioning;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.Net.Http.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.NoWASM)]
+    [SupportedOSPlatform("windows")]
```
MihaZupan, line 19:
```suggestion
    [SupportedOSPlatform("windows")]
    [OperatingSystemsFilter(allowed: true, platforms: OS.Windows)]
```

The net462 builds are failing because this attribute and the `WriteAsync` overload used below don't exist there.

You could disable this benchmark on Framework like we do for SocketsHttpHandler:
https://github.com/dotnet/performance/blob/4983ccd9730d24bcffb036b26f0f4f05ea94c636/src/benchmarks/micro/MicroBenchmarks.csproj#L123-L125

MihaZupan (approved):
Other than the framework build issue LGTM

brporter:
> Other than the framework build issue LGTM

Thanks. I think I've taken care of that, but there a huge number of failing checks and I am unfamiliar with how to debug these. I'm not entirely clear if they are caused by my change or not. Any pointers would be appreciated here.

brporter:
> > Other than the framework build issue LGTM
> 
> Thanks. I think I've taken care of that, but there a huge number of failing checks and I am unfamiliar with how to debug these. I'm not entirely clear if they are caused by my change or not. Any pointers would be appreciated here.

It occurred to me that I don't have a conditional on the WinHttpHandler package reference. I don't see a lot to go off of from the logs, so I'm speculatively attempting to see if that's causing the test failures. :\

MihaZupan (approved):
The changes LGTM.

@LoopedBard3 I see you were involved in a bunch of recent PRs - are you the right person to ask about CI failures?
I opened a dummy PR #4157 with no changes and it hit similar CI issues as this PR - is the CI generally broken and we should be merging on red?

LoopedBard3:
@'ing me or @'caaavik-msft or @'DrewScoggins. 

Our goal is of course all green, but it is not unusual to have 1 or 2 specific runs failing, or 8.0 runs due to inconsistencies with dotnet versions trying to be used and their availability. I went ahead and reran the tests as the 8.0 runs should be working at the moment. I will take a look at any still failing jobs to see if we are seeing the same failures on the main branch testing.

PR #4103, opened 2024-03-29
Fix iOS device startup log starttime bug

LoopedBard3:
Setup the first iOS startup iteration to start 10 minutes in the past and use the latest of all events found to determine the time that should be used for the testing going forward.

Test run:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2418115&view=results

### DIFF
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -637,7 +637,7 @@ ex: C:\repos\performance;C:\repos\runtime
             getLogger().info("Completed install.")
 
             allResults = []
-            timeToFirstDrawEventEndDateTime = datetime.now() # This is used to keep track of the latest time to draw end event, we use this to calculate time to draw and also as a reference point for the next iteration log time.
+            timeToFirstDrawEventEndDateTime = datetime.now() + timedelta(minutes=-10) # This is used to keep track of the latest time to draw end event, we use this to calculate time to draw and also as a reference point for the next iteration log time.
             for i in range(self.startupiterations + 1): # adding one iteration to account for the warmup iteration
                 getLogger().info("Waiting 10 secs to ensure we're not getting confused with previous app run.")
                 time.sleep(10)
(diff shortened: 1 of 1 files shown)


DrewScoggins:
What was the behavior that we are seeing that this is fixing?

src/scenarios/shared/runner.py
```diff
@@ -637,7 +637,7 @@ def run(self):
             getLogger().info("Completed install.")
 
             allResults = []
-            timeToFirstDrawEventEndDateTime = datetime.now() # This is used to keep track of the latest time to draw end event, we use this to calculate time to draw and also as a reference point for the next iteration log time.
+            timeToFirstDrawEventEndDateTime = datetime.now() + timedelta(minutes=-10) # This is used to keep track of the latest time to draw end event, we use this to calculate time to draw and also as a reference point for the next iteration log time.
```
DrewScoggins, line 640: Do we ever intend to use the value that is ten minutes in the past, or do we always have a warmup iteration?

LoopedBard3: We should never have a reason to use a value in the actual past, just potentially in the past compared to the hosting computer if the iPhone and computer times get out of sync. For the warmup iteration, I could see a style of device startup where we uninstall and reinstall between every run, but until we have that I am pretty sure we will have a warmup run, and even if we add it we could use a different flow.

LoopedBard3:
> What was the behavior that we are seeing that this is fixing?

The phone and computer times would get out of sync causing the log command to get logs from previous runs while doing the warm-up resulting in more than 4 events and an error from the 4 event check. This changes it so that we take into account degree of potential time difference when doing the warm-up run/time to use check, and don't check for exactly 4 events when getting the device time from logs for the warm-up.

DrewScoggins (approved):
LGTM

LoopedBard3:
Failures are unrelated to the change, having to do with dotnet 8 sdk version not being found, merging.

PR #4108, opened 2024-04-01
Update ImageSharp from version 2.1.3 -> 2.1.7.

LoopedBard3:


DrewScoggins (approved):
LGTM

danmoseley:
incidentally I can throw up a dependabot.yml for this repo if it would be helpful (that would create update  PR's like this for all updates not just security)

LoopedBard3:
That would be great! Funnily enough, we did get a dependabot notification for this but the PR generation or something seemed to have been failing so I just manually made the change.

LoopedBard3:
There are successful windows ImageSharp runs after the package was mirrored to our NuGet feeds and the Linux runs are likely backed up/won't be finished for a while, so merging.

danmoseley:
> That would be great! Funnily enough, we did get a dependabot notification for this but the PR generation or something seemed to have been failing so I just manually made the change.

I had a quick try of a few things but (1) you only want certain projects updated, or at least automatically - maybe not benchmarks and (2) dependabot has some difficulties. https://github.com/dependabot/dependabot-core/issues/9297#issuecomment-2032739774

So I guess leave as is for now.

PR #4140, opened 2024-04-11
One file upload in gcperfsim server call

VincentBu:
Enable uploading one file in server call case and reduce constraint when parsing the configuration file.


mrsharm (approved):
LGTM!

PR #4188, opened 2024-04-30
enable iterations for microbenchmarks 

VincentBu:
- add iterations parameter to environment section of yaml file(Microbenchmarks_Server.yaml, Microbenchmarks_Workstation.yaml and ASPNetBenchmarks.yaml);
- add `Iterations` property to `Environment` class for Microbenchmarks and ASPNetBenchmarks;
- add loop for Microbenchmarks and naming trace and log with iteration index;


src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/Microbenchmarks.Configuration.cs
```diff
@@ -24,6 +24,7 @@ public sealed class Run : RunBase
     public class Environment
     {
         public uint default_max_seconds { get; set; } = 300;
+        public uint Iterations { get; set; } = 1;
```
mrsharm, line 27: I don't see a corresponding implementation of the iteration mechanism in the ASP.NET command. Should we get rid of this until we have one?

mrsharm:
Right now we are overwriting the other iteration results -- we'll need to figure out a way to preserve them and give us averages for the values in the results MD file.

PR #4193, opened 2024-04-30
GCPerfSim updates

markples:
- Implement finalizable `Item`s
- Implement `finishWithFullCollect` when a file is used
- Display whether the run is using SVR GC

### DIFF
--- a/src/benchmarks/gc/GCPerfSim/GCPerfSim.cs
+++ b/src/benchmarks/gc/GCPerfSim/GCPerfSim.cs
@@ -542,7 +542,7 @@ abstract class Item : ITypeWithPayload
     public byte[] GetPayload() =>
         Util.NonNull(payload);
 
-    class ItemFinalizable : Item
+    sealed class ItemFinalizable : Item
     {
         public ItemFinalizable(uint size, bool isPinned, bool isWeakLong, bool isPoh)
             : base(size, isPinned, isWeakLong, isPoh)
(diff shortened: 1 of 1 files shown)


Maoni0 (approved):

src/benchmarks/gc/GCPerfSim/GCPerfSim.cs
```diff
@@ -1840,6 +1865,18 @@ private static Args ParseFromCommandLine(string[] args)
...
+            double pohWeight;
+
+            if ((sohAllocRatioArg == 0) && (pohAllocRatioArg == 0))
+            {
+               lohWeight = 1;
+               sohWeight = 0;
+               pohWeight = 0;
+            }
+            else
+            {
```
markples, line 1879: undo this - #4255 has a more comprehensive fix (handles other soh==0 cases)

src/benchmarks/gc/GCPerfSim/GCPerfSim.cs
```diff
@@ -533,6 +541,26 @@ public void Free()
 
     public byte[] GetPayload() =>
         Util.NonNull(payload);
+
+    class ItemFinalizable : Item
```
mrsharm, line 545: Can we make this class and the class below internal sealed?

markples: The whole file is using default internal accessibility.  For sealed, I should change the ReferenceItem analogue as well.  I'm ambivalent on this since it's a single file project and they don't necessarily need to be sealed, but also they don't need to be unsealed and this documents that there are no further subclasses, so sure.

mrsharm (approved):
LGTM - probably be worthwhile to run the latest bits using the LowVolatilityScenario.yaml and possibly add a new Functional Test for this case.

PR #4208, opened 2024-05-09
fix rpolayout config setting

AndyAyersMS:
Config value needs to be set to `1`, not `*`.

### DIFF
--- a/scripts/ci_setup.py
+++ b/scripts/ci_setup.py
@@ -419,7 +419,7 @@ def main(args: Any):
     elif args.experiment_name == "jitoptrepeat":
         experiment_config = variable_format % ('DOTNET_JitOptRepeat', '*')
     elif args.experiment_name == "rpolayout":
-        experiment_config = variable_format % ('DOTNET_JitDoReversePostOrderLayout', '*')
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@DrewScoggins FYI

DrewScoggins (approved):
LGTM

amanasifkhalid:
Sorry I'm late to this -- @AndyAyersMS were you able to verify that the layout experiment wasn't running, or is this a proactive fix, just in case it isn't running?

AndyAyersMS:
Pretty sure it was not running properly with `*` as the setting, the parsing is really simplistic. You might want to check this locally; I think anything that's not a digit sequence gets parsed as 0.

The value of `*` for `optRepeat` is correct, though, because it is a `RELEASE_CONFIG_METHODSET` and not a `RELEASE_CONFIG_INTEGER`.

amanasifkhalid:
> You might want to check this locally; I think anything that's not a digit sequence gets parsed as 0.

You're right; SuperPMI even warns about passing a string for an int config value. Thanks for the fix!

PR #4213, opened 2024-05-12
Fix link in `benchmarking-workflow-dotnet-runtime.md`

ranma42:


LoopedBard3 (approved):
Looks good, thanks for the update!

PR #4240, opened 2024-05-29
TryAddWithoutValidation for multiple values could be more efficient

pedrobsaila:
Contributes to dotnet/runtime#64049

### DIFF
--- a/src/benchmarks/micro/libraries/System.Net.Http/TryAddWithoutValidation.cs
+++ b/src/benchmarks/micro/libraries/System.Net.Http/TryAddWithoutValidation.cs
@@ -1,0 +1,44 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Engines;
+using MicroBenchmarks;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
+namespace System.Net.Http.Tests
+{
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.Net.Http/TryAddWithoutValidation.cs
```diff
@@ -0,0 +1,44 @@
...
+using BenchmarkDotNet.Engines;
+using MicroBenchmarks;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
+namespace System.Net.Http.Tests
+{
+    [BenchmarkCategory(Categories.Libraries, Categories.LINQ)]
```
MihaZupan, line 15:
```suggestion
    [BenchmarkCategory(Categories.Libraries)]
```

src/benchmarks/micro/libraries/System.Net.Http/TryAddWithoutValidation.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class TryAddWithoutValidationTests
+    {
+        [Benchmark]
+        [ArgumentsSource(nameof(IEnumerableArgument))]
+        public bool AddHeaders(AddWithoutValidationTestData data) => new HttpResponseMessage().Headers.TryAddWithoutValidation("headerName", data.Values);
```
MihaZupan, line 20: We could make the test more focused on just the adding part, e.g.
```c#
private readonly HttpResponseHeaders _responseHeaders = new HttpResponseMessage().Headers;

public bool SomeTest()
{
    _responseHeaders.Add(...);
    _responseHeaders.Clear();
}

```

MihaZupan (commented):
Thank you for the improving coverage here

Issue #4027, opened 2024-03-05
8.0 Perf CI Runs Failing


DrewScoggins:
As seen [here](https://dev.azure.com/dnceng-public/public/_build/results?buildId=588533&view=results), 8.0 runs have started failing. It looks like the `dotnet-install` script has started returning a different version of the SDK that does not seem to exist, as runs from a couple of days ago were getting different builds with the same input.

PR #4251, opened 2024-06-05
Use full dotnet version instead of the version in file path

LoopedBard3:
Fixes 8.0 runs where the file path for the dotnet sdk contains the short version of the dotnet sdk rather than the full version, this switches to using the full version located in the sdks .version file instead. This will allow the dotnet-install script to redownload the correct dotnet version for running the tests.

Fixes: https://github.com/dotnet/performance/issues/4027

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -555,7 +555,7 @@ def get_dotnet_path() -> str:
     return dotnet_path
 
 
-def get_dotnet_version(
+def get_dotnet_version_path(
         framework: str,
         dotnet_path: Optional[str] = None,
         sdk_path: Optional[str] = None) -> str:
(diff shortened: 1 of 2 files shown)


scripts/dotnet.py
```diff
@@ -555,7 +555,7 @@ def get_dotnet_path() -> str:
     return dotnet_path
 
 
-def get_dotnet_version(
+def get_dotnet_version_path(
```
caaavik-msft, line 558: looking at this method, it doesn't seem to return a path, it seems to return a version number. Is this the right name?

scripts/dotnet.py
```diff
@@ -586,19 +586,28 @@ def get_dotnet_version(
 
     return sdk
 
+def get_dotnet_version_precise(
+        framework: str,
+        dotnet_path: Optional[str] = None,
+        sdk: Optional[str] = None) -> str:
+    sdk_path = get_sdk_path(dotnet_path)
+    sdk = get_dotnet_version_path(framework, dotnet_path,
```
caaavik-msft, line 594: I think I would prefer if this is an if statement rather than a ternary since it's only doing something if the value is not None.

caaavik-msft (approved):

scripts/dotnet.py
LoopedBard3: Makes sense, updated.

scripts/dotnet.py
LoopedBard3: Good point. I think based on how it is used it would make sense, but I switched the name to match what it does. I updated the name to get_dotnet_version_from_path as it is getting the dotnet version based on the folder name in the sdk_path. I am definitely open to suggestions for a better name.

caaavik-msft: `get_dotnet_version_from_path` is perfect, I misunderstood it before as thinking it meant getting a path to a dotnet version

caaavik-msft (approved):

PR #4255, opened 2024-06-06
Various GC infra improvements

markples:
GCPerfSim:
- Fix overhead coefficient in weight derivation (add the `1000 *`)
- Handle zero and very low sohar values

GC Infra for GCPerfSim:
- Fix a few null references
- Also fix a few other nullable warnings (before giving up for now, there are a lot)
- Store the gc log

General:
- .gitignore the build-generated xlf folder
- Update Spectre.Console version to what the build is finding anyway (but without a warning)
- Fix some "code is Windows-specific" warnings (by doing nothing if not on Windows)


src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/Common.cs
```diff
@@ -29,7 +29,7 @@ public static void VerifyFile(string configurationPath, string prefix)
                 throw new ArgumentNullException($"{prefix}: The provided path to yaml file {nameof(configurationPath)} doesn't exist or is empty - please ensure you are passing in a valid .yaml file.");
             }
 
-            if (Path.GetExtension(configurationPath) != ".yaml")
+            if ((Path.GetExtension(configurationPath) != ".yaml") && (Path.GetExtension(Path.GetFileNameWithoutExtension(configurationPath)) != ".yaml"))
```
markples, line 32: Undo this (local workaround for my scripting)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/GCPerfSim/GCPerfSimCommand.cs
```diff
@@ -157,6 +157,15 @@ public static GCPerfSimResults RunGCPerfSim(GCPerfSimConfiguration configuration
 
                         foreach (var environVar in environmentVariables)
                         {
+                            // Check if the log file is specified, also store it in a run-specific location.
+                            // This log file should be named in concordance with the name of the run and the benchmark.
+                            if (string.CompareOrdinal(environVar.Key, "DOTNET_GCLogFile") == 0)
+                            {
+                                string gcLogDirectory = Path.Combine(outputPath, runInfo.RunDetails.Key + "_GCLog");
+                                Core.Utilities.TryCreateDirectory(gcLogDirectory);
+                                gcperfsimProcess.StartInfo.WorkingDirectory = gcLogDirectory;
```
markples, line 166: I don't particularly like this fix.  Currently the infra will just run gcperfsim from the current working directory and leave the gc log there, so this is probably an improvement.  It doesn't handle the server case.

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/AnalyzerManager.cs
```diff
@@ -22,7 +22,7 @@ public static class AnalyzerManager
         {
             Dictionary<string, Analyzer> analyzers = new();
 
-            IEnumerable<string> allFiles = null;
+            IEnumerable<string> allFiles;
```
Maoni0, line 26: I presume this is to fix a warning?

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/GCPerfSim/GCPerfSimCommand.cs
Maoni0: I thought this was by design. do you need to be able to specify a different dir? @mrsharm could take a look when he gets back. for server I don't think we ever needed to specify anything other than the current directory (because we just get files in the current dir back).

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/AnalyzerManager.cs
markples: Yes, it's not `IEnumerable<string>?`, so `null` isn't appropriate.  The original value isn't needed since the `if/else` takes care of it.  This is from diving into the nullable warnings, which is a much larger project but I kept the few things that I did.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/GCPerfSim/GCPerfSimCommand.cs
markples: The infra has logic for capturing things like stdout and putting the results in `outputPath`, so it seems inconsistent for the gc log to not end up there.  I did the `_GCLog` thing to match the ASP.NET testing, though with the right filename it probably doesn't need to be in a subdirectory.  The infra is generally creating `outputPath`, so the directory doesn't exist when launching the infra.

The ASP.NET case has some explicit logic asking for the GCLog directory to be retrieved, so I was assuming that something similar was needed (and missing) for gcperfsim.  I haven't actually tried it though.

mrsharm: I think a cleaner solution to this would be to update the value of the environment variable of ``DOTNET_GCLogFile`` path to point to the output path combined with the new folder (*_GCLog) and then the name of the file specified with the environment variable; this way we don't need to change anything with the way the process will be run and will have made a decision for the user that's consistent with the ASP.NET functionality.

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/CPUAnalysis/CPUAnalysisExt.cs
```diff
@@ -246,10 +246,14 @@ public static List<CPUInfo> GetPerGCMethodCost(this CPUProcessData processData,
...
         public static void SetSourcePath(this CPUProcessData processData, string sourcePath)
             => processData.StackView.SymbolReader.SourcePath = sourcePath;
-        public static string Annotate(this CPUProcessData processData, string methodName)
+        public static string? Annotate(this CPUProcessData processData, string methodName)
         {
-            SourceLocation location = processData.StackView.GetSourceLocation(methodName, out var @out);
-            SortedDictionary<int, float> metricsOnLine = @out;
+            if (!processData.StackView.TryGetSourceLocation(methodName, out SourceLocation? location, out SortedDictionary<int, float>? metricsOnLine))
+            {
+                // TODO: Log an error.
```
mrsharm, line 253: Would throwing an exception make more sense?

src/benchmarks/gc/GCPerfSim/GCPerfSim.cs
```diff
@@ -1854,8 +1861,8 @@ private static Args ParseFromCommandLine(string[] args)
             double a31 = 1;
             double a32 = 1;
             double a33 = 1;
-            double b1 = lohAllocRatioArg * overhead;
-            double b2 = pohAllocRatioArg * overhead;
+            double b1 = 1000 * lohAllocRatioArg * overhead;
```
mrsharm, line 1864: Since this is an esoteric calc, we should probably add a comment explaining why the 1000x is needed.

markples: The documentation for all of this code is in https://cshung.github.io/posts/poh-tuning-2/.  @cshung any chance you could update your blog post to include this fix?  (the two `+ O` under "Since we assumed the weights..." should be `+ 1000 O` and the two `R_{xoh}O` under "Now we think of..." should be `1000 R_{xoh}O`)

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/CPUAnalysis/CPUAnalysisExt.cs
markples: I copied this from `GetSourceLocation`.  I'm happy to add a throw, but I wasn't sure what the model was here (which exception, etc.).  Any thoughts on that?

(also this is just to remove one of 100s of null warnings, so if this is complicated, I should just revert and leave it for another time)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/GCPerfSim/GCPerfSimCommand.cs
markples: I pushed a change to do this (still need to test it though).

Note that this requires a casing match.  I could enumerate the variables and do case-insensitive matches if you'd prefer.  (I think it would be foreach-to-identify, update-outside-of-foreach, foreach-to-set-env-vars.  Technically I could just do it when setting the envvars, but I didn't like leaving the envvar dictionary in the "wrong" state as it seemed like it could invite a mistake in the future in code later in the method.)

Also, if the variable has an absolute path, then this code won't do anything.  I think I'm ok with that.

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/Analyzer.cs
```diff
@@ -38,7 +38,7 @@ public Analyzer(string tracePath, HashSet<string> processNames)
                     AllGCProcessData[p.Name] = values = new();
                 }
 
-                values.Add(new GCProcessData(p, managedProcess, processIdToGCThreads[p.ProcessID], this, p.EndTimeRelativeMsec - p.StartTimeRelativeMsec));
+                values.Add(new GCProcessData(p, managedProcess, processIdToGCThreads.GetValueOrDefault(p.ProcessID) ?? new(), this, p.EndTimeRelativeMsec - p.StartTimeRelativeMsec));
```
markples, line 41: @mrsharm This (and the similar change below) are new.  I was trying to analyze a trace that didn't have this mapping.  (If I recall correctly, I have a notebook side change to not filter these out, but this access will throw without this change.)

mrsharm: Looks good!

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/CPUAnalysis/CPUAnalysisExt.cs
mrsharm: I see - yes, I filched this from the PerfView source. I think it's fine, as you have it. We can revisit later.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/GCPerfSim/GCPerfSimCommand.cs
mrsharm: >> do case-insensitive matches if you'd prefer

No preference. 

>> Also, if the variable has an absolute path, then this code won't do anything. I think I'm ok with that. 

Yes, my rationale here was to main consistency with the output_path that the user has expressed intention where the outputs should live vs. an arbitrary path. I'd rather have everything in one place (enforced by the Infra).

mrsharm (dismissed):
LGTM! Let's have the vendors test this out for further validation.

mrsharm (approved):
LGTM

src/benchmarks/gc/GCPerfSim/GCPerfSim.cs
cshung: > @cshung any chance you could update your blog post to include this fix? 

The blog is updated.

PR #4277, opened 2024-06-19
Add DefaultMutualHandshakeContext* benchmarks

rzikm:
This PR adds microbenchmarks for mutually authenticated SslStream pair which uses SslStreamCertificateContext on the server side. This allows benchmarking TLS resume with client certificates (supported on Windows, newly supported on Linux in .NET 9 via https://github.com/dotnet/runtime/pull/102656)


wfurt (approved):

rzikm:
CI Failures seem to be unrelated.

src/benchmarks/micro/libraries/System.Net.Security/SslStreamTests.Context.cs
```diff
@@ -29,26 +28,34 @@ public partial class SslStreamTests
...
-                ServerCertificateContext  = _context,
+                ServerCertificateContext = _context,
             };
 
             using (var sslClient = new SslStream(client, leaveInnerStreamOpen: true, delegate { return true; }))
             using (var sslServer = new SslStream(server, leaveInnerStreamOpen: true, delegate { return true; }))
             {
                 await Task.WhenAll(
-                    sslClient.AuthenticateAsClientAsync("localhost", null, SslProtocols.None, checkCertificateRevocation: false),
+                    sslClient.AuthenticateAsClientAsync("localhost", requireClientCert ? new X509CertificateCollection() { _clientCert } : null, SslProtocols.None, checkCertificateRevocation: false),
```
adamsitnik, line 58: just for my education: Why do you want to create a new instance of `X509CertificateCollection` every time the benchmark is called? Does this reflect typical real-life scenario?

adamsitnik (approved):
LGTM, thank you @rzikm !

src/benchmarks/micro/libraries/System.Net.Security/SslStreamTests.Context.cs
rzikm: I wanted to mirror what we do in non-context benchmarks for SslStream.

https://github.com/dotnet/performance/blob/b69d1bdd939f8ccbebf62cd276d3f3d95149b1da/src/benchmarks/micro/libraries/System.Net.Security/SslStreamTests.cs#L130

Typical real-world scenarios would probably always use the overload accepting Ssl(Client|Server)AuthenticationOptions and reuse a single instance of it for multiple connections, or always create new instance entirely from scratch (as we do in linked source code).

rzikm:
@adamsitnik I can't merge this PR because of a failed GH check on `performance-ci`, I don't think the failures are related to my changes. Can you override it and merge this?

adamsitnik:
> Can you override it and merge this?

I don't have such permissions anymore.

But @caaavik-msft or @LoopedBard3  or @DrewScoggins should be able to do that.

LoopedBard3 (approved):

PR #4287, opened 2024-06-26
Download PDN with Managed Identity

caaavik-msft:
Will need to make these changes in the 6.0 and 7.0 branches too after this goes in. This is done by downloading PDN in a separate Azure Pipelines build step and the copying it into the helix upload directories during the performance setup scripts.

### DIFF
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -99,7 +99,7 @@ class RunPerformanceJobArgs:
     use_local_commit_time: bool = False
     javascript_engine_path: Optional[str] = None
     maui_version: Optional[str] = None
-    download_pdn: bool = False
+    pdn_path: Optional[str] = None
     os_version: Optional[str] = None
     dotnet_version_link: Optional[str] = None
     target_csproj: Optional[str] = None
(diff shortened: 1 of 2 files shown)


LoopedBard3 (approved):
Looks good presuming a successful test run.

scripts/run_performance_job.py
```diff
@@ -729,18 +729,14 @@ def run_performance_job(args: RunPerformanceJobArgs):
                 os.path.join(args.performance_repo_dir, "src", "tools", "PerfLabGenericEventSourceLTTngProvider", "build.sh"),
                 "-o", os.path.join(payload_dir, "PerfLabGenericEventSourceForwarder")]).run()
         
-        # download PDN
-        if args.os_group == "windows" and args.architecture != "x86" and args.download_pdn:
-            print("Downloading PDN")
-            escaped_upload_token = str(os.environ.get("PerfCommandUploadTokenLinux")).replace("%25", "%")
-            pdn_url = f"https://pvscmdupload.blob.core.windows.net/assets/paint.net.5.0.3.portable.{args.architecture}.zip{escaped_upload_token}"
+        # copy PDN
+        if args.os_group == "windows" and args.architecture != "x86" and args.pdn_path is not None:
```
DrewScoggins, line 733: Is the reasoning behind this for when people do local runs? We shouldn't need this check as the yaml ensures that pdn_path is empty unless we are in the right OS/Arch combination.

DrewScoggins:
One comment, but LGTM.

DrewScoggins (approved):

scripts/run_performance_job.py
caaavik-msft: Yes, the check is just for local runs

PR #4295, opened 2024-07-02
Don't include AzureCLI for PDN if running a public or PR build

LoopedBard3:
Fix public runs not running due to attempting to instantiate a service connection that does not exist on our public runs.

### DIFF
--- a/eng/performance/scenarios.yml
+++ b/eng/performance/scenarios.yml
@@ -82,15 +82,16 @@ jobs:
         steps:
         - checkout: self
           clean: true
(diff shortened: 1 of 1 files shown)


DrewScoggins:
I am confused. The title talks about secret synchronization, but the change seems to be about limiting when we download PDN.

LoopedBard3:
> I am confused. The title talks about secret synchronization, but the change seems to be about limiting when we download PDN.

Updated. The initial draft PR I made included the assumption that the cause of the error was related to secret-synchronization, this was not the case.

PR #4303, opened 2024-07-09
Make sure the DynamicEvent unit tests are compiled only when using custom trace event

cshung:
This will prevent the code to fail the build in case we are not using the latest TraceEvent.


mrsharm (approved):
LGTM

PR #4315, opened 2024-07-16
comment out output.path check logic for aspnetbenchmarks

VincentBu:
In an end-to-end run, "output.path" can be null when creating suites for aspnetbenchmarks. We can comment out the logic for this case.

### DIFF
--- a/src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/ASPNetBenchmark.Configuration.cs
+++ b/src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/ASPNetBenchmark.Configuration.cs
@@ -62,10 +62,11 @@
                 throw new ArgumentNullException($"{nameof(ASPNetBenchmarksConfigurationParser)}: {nameof(configuration.Output)} is null. Check the syntax of the configuration.");
             }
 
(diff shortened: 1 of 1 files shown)


src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Configurations/ASPNetBenchmark.Configuration.cs
```diff
@@ -62,10 +62,11 @@ public static ASPNetBenchmarksConfiguration Parse(string path)
                 throw new ArgumentNullException($"{nameof(ASPNetBenchmarksConfigurationParser)}: {nameof(configuration.Output)} is null. Check the syntax of the configuration.");
             }
 
-            if (string.IsNullOrEmpty(configuration.Output.Path))
-            {
-                throw new ArgumentNullException($"{nameof(ASPNetBenchmarksConfigurationParser)}: {nameof(configuration.Output.Path)} is null or empty. Please specify an output path.");
-            }
+            // configuration.Output.Path can be null when creating suites for ene to end run.
```
mrsharm, line 65: Let's get rid of these lines.

mrsharm (approved):
LGTM!

PR #4321, opened 2024-07-18
Update Maui version flows

LoopedBard3:
Update how we determine dotnet version to use and rollback versions for Performance Maui.

Version.Details.XML update from: https://github.com/dotnet/performance/pull/4327. 
Darc subscription still needs to be setup.

Full test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2506223&view=results

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -27,5 +27,42 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>731d793be2d0a66bafc96b1a79dc96b4d1f0301b</Sha>
     </Dependency>
+     <!-- Dependencies for .NET MAUI workload -->
+    <Dependency Name="Microsoft.Maui.Controls" Version="9.0.0-preview.7.24368.8">
+      <Sha>cf76a629023c99557239464a278daec280d1f448</Sha>
+      <Uri>https://github.com/dotnet/maui</Uri>
+    </Dependency>
+    <Dependency Name="VS.Tools.Net.Core.SDK.Resolver" Version="9.0.100-preview.7.24360.5" CoherentParentDependency="Microsoft.Maui.Controls">
+      <Sha>9a028e12b90e6a583b09ccb3008fdfaf85761f19</Sha>
+      <Uri>https://github.com/dotnet/sdk</Uri>
+    </Dependency>
+    <Dependency Name="Microsoft.NETCore.App.Ref" Version="9.0.0-preview.7.24357.2" CoherentParentDependency="VS.Tools.Net.Core.SDK.Resolver">
+      <Sha>4e278fe17f69ea31fbdcbab74ac47ec6fa84914b</Sha>
+      <Uri>https://github.com/dotnet/runtime</Uri>
+    </Dependency>
+    <Dependency Name="Microsoft.NET.Workload.Emscripten.Current.Manifest-9.0.100.Transport" Version="9.0.0-preview.7.24319.4" CoherentParentDependency="Microsoft.NETCore.App.Ref">
+      <Sha>ffe9afdc046cf7a6f82cc7c5796aade54047af64</Sha>
+      <Uri>https://github.com/dotnet/emsdk</Uri>
+    </Dependency>
+    <Dependency Name="Microsoft.Android.Sdk.Windows" Version="34.99.0-preview.7.346" CoherentParentDependency="Microsoft.Maui.Controls">
+      <Sha>06bb1dc6a292ef5618a3bb6ecca3ca869253ff2e</Sha>
+      <Uri>https://github.com/dotnet/android</Uri>
+    </Dependency>
+    <Dependency Name="Microsoft.MacCatalyst.Sdk.net9.0_17.2" Version="17.2.9714-net9-p6" CoherentParentDependency="Microsoft.Maui.Controls">
+      <Sha>4741d540eee2390fa075d0bdb49083cb58e43048</Sha>
+      <Uri>https://github.com/xamarin/xamarin-macios</Uri>
+    </Dependency>
+    <Dependency Name="Microsoft.macOS.Sdk.net9.0_14.2" Version="14.2.9714-net9-p6" CoherentParentDependency="Microsoft.Maui.Controls">
+      <Sha>4741d540eee2390fa075d0bdb49083cb58e43048</Sha>
+      <Uri>https://github.com/xamarin/xamarin-macios</Uri>
+    </Dependency>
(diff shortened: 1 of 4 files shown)


eng/Version.Details.xml
```diff
@@ -27,5 +27,53 @@
...
+    <!--
+      Maui Rollback Version mappings, default means generated from sdk version. Allows for manual override of version similar to https://github.com/dotnet/maui/blob/df8cfcf635a590955a8cc3d6cf7ae6280449dd3f/eng/Versions.props#L138-L146, where the logic comes from:
+      Mapping_Microsoft.Maui.Controls:default
+      Mapping_Microsoft.NETCore.App.Ref:default
+      Mapping_Microsoft.NET.Workload.Emscripten.Current:default
+      Mapping_Microsoft.Android.Sdk:default
+      Mapping_Microsoft.MacCatalyst.Sdk:9.0.100-preview.6
+      Mapping_Microsoft.macOS.Sdk:9.0.100-preview.6
+      Mapping_Microsoft.iOS.Sdk:9.0.100-preview.6
+      Mapping_Microsoft.tvOS.Sdk:9.0.100-preview.6
```
DrewScoggins, line 39: How is this information generated? Is this done by Maestro or is this something we will need to manually update?

src/scenarios/shared/mauisharedpython.py
```diff
@@ -1,24 +1,95 @@
...
+    rollback_dict: dict[str, str] = {}
+    rollback_name_to_xml_name_mappings: dict[str, str] = {
+        "microsoft.net.sdk.android" : "Microsoft.Android.Sdk",
+        "microsoft.net.sdk.ios" : "Microsoft.iOS.Sdk",
+        "microsoft.net.sdk.maccatalyst" : "Microsoft.MacCatalyst.Sdk",
+        "microsoft.net.sdk.macos" : "Microsoft.macOS.Sdk",
+        "microsoft.net.sdk.maui" : "Microsoft.Maui.Controls",
+        "microsoft.net.sdk.tvos" : "Microsoft.tvOS.Sdk",
+        "microsoft.net.sdk.mono.toolchain.current" : "Microsoft.NETCore.App.Ref",
+        "microsoft.net.sdk.mono.emscripten.current" : "Microsoft.NET.Workload.Emscripten.Current"
```
DrewScoggins, line 29: Where does this information come from? We should have a comment here that says where we get this information from and when/why we would need to update it.

src/scenarios/shared/mauisharedpython.py
```diff
@@ -1,24 +1,95 @@
...
-    with open ("MauiNuGet.config", "wb") as f:
-        f.write(requests.get(f'https://raw.githubusercontent.com/dotnet/maui/{target_framework_wo_platform}/NuGet.config', allow_redirects=True).content)
+    with open("MauiNuGet.config", "wb") as f:
+        with urllib.request.urlopen(f'https://raw.githubusercontent.com/dotnet/maui/{target_framework_wo_platform}/NuGet.config') as response:
+            f.write(response.read())
 
     workload_install_args = ['--configfile', 'MauiNuGet.config', '--skip-sign-check']
-    if int(target_framework_wo_platform.split('.')[0][3:]) > 7: # Use the rollback file for versions greater than 7
-        workload_install_args += ['--from-rollback-file', f'https://maui.blob.core.windows.net/metadata/rollbacks/{target_framework_wo_platform}.json']
+    if int(target_framework_wo_platform.split('.')[0][3:]) > 8: # Use the rollback file for versions greater than 7
```
DrewScoggins, line 89: Doesn't this use the rollback file for versions greater than 8?

DrewScoggins:
In general, the logic looks good getting the values. Just have some questions on some of the additional data that we are using as part of the processing.

eng/Version.Details.xml
LoopedBard3: Yes, this is something we will have to manually update when Maui has to hardcode the data themselves. These values will have to match whatever is being generated in place of https://github.com/dotnet/maui/blob/net9.0/eng/Versions.props#L141-L155. Though this should be rare throughout the release cycle. The comment for these mappings does link to a specific snapshot of the values which can be used to figure out what the current values should be.

src/scenarios/shared/mauisharedpython.py
LoopedBard3: Yes, updated the number and left some information on what the number is for.

scripts/run_performance_job.py
```diff
@@ -546,12 +547,26 @@ def run_performance_job(args: RunPerformanceJobArgs):
         ci_setup_arguments.dotnet_path = f"{wasm_bundle_dir}/dotnet"
 
     if args.dotnet_version_link is not None:
-        with urllib.request.urlopen(args.dotnet_version_link) as response:
-            values = json.loads(response.read().decode('utf-8'))
-            if "dotnet_version" in values:
-                ci_setup_arguments.dotnet_versions = [values["dotnet_version"]]
+        if args.dotnet_version_link.startswith("https"): # Version link is a proper url
+            if args.dotnet_version_link.endswith(".json"):
```
caaavik-msft, line 551: We should raise an error for the case that this is false

src/scenarios/shared/mauisharedpython.py
LoopedBard3: Added a comment explaining that these are mappings to the Maui workloads in ./eng/Version.Details.xml file.

src/scenarios/shared/mauisharedpython.py
```diff
@@ -1,24 +1,98 @@
...
+
+    # Load in the Version.Details.xml file
+    with open(os.path.join(get_repo_root_path(), "eng", "Version.Details.xml"), encoding="utf-8") as f:
+        version_details_xml = f.read()
+        root = ET.fromstring(version_details_xml)
+
+    # Get the General Band version from the Version.Details.xml file sdk version
+    general_version_obj = root.find(".//Dependency[@Name='VS.Tools.Net.Core.SDK.Resolver']")
+    if general_version_obj is not None:
+        full_band_version_holder = general_version_obj.get("Version", "ERROR: Failed to get version")
```
caaavik-msft, line 43: I think it might be better to just do `get("Version")` then you can do `if full_band_version_holder is None`, since this string doesn't get rendered anywhere and it's just used for an equality check. Same for the other places that this is done.

caaavik-msft (approved):

src/scenarios/shared/mauisharedpython.py
LoopedBard3: Good catch, didn't notice that default return value for get was None.

scripts/run_performance_job.py
LoopedBard3: Good catch, updated.

DrewScoggins (approved):
LGTM

PR #4325, opened 2024-07-20
Add zip file of BenchmarkDotNet.Artifacts directory to Helix Upload Root

caaavik-msft:
This feature allows for us to access the binlogs and full report jsons that are generated when running BDN benchmarks in Helix jobs by zipping up the entire BenchmarkDotNet.Artifacts directory at the end of the run and putting it in the Helix Upload Root.


LoopedBard3 (approved):
LGTM given we have a run with this working.

PR #4330, opened 2024-07-23
Fixed notebooks for Dynamic Events and included them in the API

mrsharm:


src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/DynamicEvents.cs
```diff
@@ -43,69 +43,85 @@ public sealed class DynamicEventSchema
...
+            // One time add.
+            if (DynamicEventSchemas.Count == 0)
+            {
+                // Add the adaptation schemas.
+                Add(GCDynamicEvents.SizeAdaptationSampleSchema);
+                Add(GCDynamicEvents.SizeAdaptationTuningSchema);
+            }
+        }
+
+        public static void Add(DynamicEventSchema dynamicEventSchema, bool allowPartialSchema = true)
```
markples, line 57: [existing] The use of a static method/field for this means that the GC infra dll can only be used for one dynamic event schema at a time.

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/DynamicEvents.cs
```diff
@@ -43,69 +43,85 @@ public sealed class DynamicEventSchema
...
         public int MaxOccurrence { get; init; } = 1;
 
-        public static void Set(List<DynamicEventSchema> dynamicEventSchemas, bool allowPartialSchema = true)
+        static DynamicEventSchema()
         {
-            DynamicEventSchemas.Clear();
-            DynamicEventSchema.allowPartialSchema = allowPartialSchema;
-            foreach (DynamicEventSchema dynamicEventSchema in dynamicEventSchemas)
+            // One time add.
+            if (DynamicEventSchemas.Count == 0)
```
markples, line 49: A type initializer will only run once.  Is there some way that this check can fail?

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/DynamicEvents.cs
markples, line 57: This method is both compiling the schema and adding it to the list.  I think this might get simpler if you separated those.  If you add a `CompiledSchema Compile(DynamicEventSchema)`, then I think the explicit type initializer could go and `DynamicEventSchemas` could just be initialized to an expression.

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/DynamicEvents.cs
```diff
@@ -43,69 +43,85 @@ public sealed class DynamicEventSchema
...
+        }
+
+        public static void Add(DynamicEventSchema dynamicEventSchema, bool allowPartialSchema = true)
+        {
+            if (DynamicEventSchemas.ContainsKey(dynamicEventSchema.DynamicEventName))
+            {
+                throw new Exception($"Provided schema has a duplicated event named {dynamicEventSchema.DynamicEventName}");
+            }
+            CompiledSchema schema = new CompiledSchema();
+            if (dynamicEventSchema.MinOccurrence < 0)
```
markples, line 64: Since you have an abstract type for `dynamicEventSchema`, it seems like it should be the place to restrict `MinOccurrence >= 0`.

markples: (and max >= min)

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/DynamicEvents.cs
```diff
@@ -43,69 +43,85 @@ public sealed class DynamicEventSchema
...
+            }
+            if (dynamicEventSchema.MaxOccurrence < dynamicEventSchema.MinOccurrence)
             {
-                if (DynamicEventSchemas.ContainsKey(dynamicEventSchema.DynamicEventName))
+                throw new Exception($"Provided event named {dynamicEventSchema.DynamicEventName} has a MaxOccurrence smaller than MinOccurrence");
+            }
+            schema.MinOccurrence = dynamicEventSchema.MinOccurrence;
+            schema.MaxOccurrence = dynamicEventSchema.MaxOccurrence;
+            int offset = 0;
+            foreach (KeyValuePair<string, Type> field in dynamicEventSchema.Fields)
```
markples, line 75: just verifying - these fields aren't aligned, correct?

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/GCDynamicEvents.cs
```diff
@@ -0,0 +1,49 @@
+﻿namespace GC.Analysis.API.DynamicEvents
```
markples, line 1: As part of our notebook-samples-todo, we should include the workflow for adding a new dynamic event.  I think it would be something patterned after this code and calling the `Add` method, but seeing it would help evalulate it.  (Andrew has talked about how it's a lot of work to use dynamic events.  This would make it explicit.  This PR contains some of the infra (e.g., the size of a uint) which isn't needed for new events.)

markples: To then evaluate this fairly, we'd want the "best" version possible of the initialization code.  This is subjective of course...
- `KeyValuePair.Create("version", typeof(ushort)` seems better
- C# initialization syntax for the list?
- consider a "data" format for the schema rather than C# code?
- consider having the build of the GC itself (which defines how to write the dynamic event) write a schema or maybe embed it in the etl file so that the info doesn't need to be manually written twice

src/benchmarks/gc/GC.Infrastructure/Notebooks/Reports.dib
```diff
@@ -588,21 +588,169 @@ public static class Metrics
         public static Metric<TraceGC> HeapCount = new(gc => gc.HeapCount, "Heap count", "#");
 
         // HeapCountSample
-        public static Metric<TraceGC> HcsElapsedTimeBetweenGCs = new(gc => gc.HeapCountSample.ElapsedTimeBetweenGCsMSec, "HCSampleElapsed", "ms");
-        public static Metric<TraceGC> HcsGCIndex = new(gc => gc.HeapCountSample.GCIndex, "HCSampleGCIndex", "#");
-        public static Metric<TraceGC> HcsGCPauseTime = new(gc => gc.HeapCountSample.GCPauseTimeMSec, "HCSampleGCPause", "ms");
-        public static Metric<TraceGC> HcsMslWaitTime = new(gc => gc.HeapCountSample.MslWaitTimeMSec, "HCSampleGCMslWait", "ms");
+        public static Metric<TraceGC> TCPToConsider = new(gc =>
```
markples, line 591: `gc => { return E; }` can be simplified as `gc => E`

In looking at the usage side, consider making `DynamicEvents` a property.  It "stands out" compared to the non-dynamic events here.

How does a dynamic event have a built-in property name (like `SizeAdapatationTuning`)?  I guess that's from an earlier PR or I missed it in this one.  So would a non-built-in one need to do a dictionary lookup of some sort?  Or use an extension method to wrap that and make it look "nice"?

src/benchmarks/gc/GC.Infrastructure/Notebooks/Reports.dib
```diff
@@ -588,21 +588,169 @@ public static class Metrics
...
+        },
+        "AdjMetric", 
+        "#");
+
+        public static Metric<TraceGC> ElapsedTimeBetweenGCs = new(gc => 
+        {
+            return gc.DynamicEvents().SizeAdaptationSample?.ElapsedTimeBetweenGCs;
+        },
+        "ElapsedTimeBetweenGCs", 
+        "msec");
```
markples, line 715: I think `ms` is already used for this unit.  They need to match for the axis to be shared.

All of the names/titles should be checked against the others for consistency and that each has enough context.  Possibly these should be prefixed/described in order to scope them to datas.  What is `AdjMetric`? etc.

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/DynamicEvents.cs
mrsharm: Yes - this is the intended behavior. We currently only want one set of dynamic event schemas. In what circumstance should we want multiple dynamic event schemas?

markples: Admittedly this was initially just "statics are bad" commentary.

Initial thing that I can think of is iterating on a dynamic event in a local build and wanting to keep a notebook running with an old run and a new one at the same time?

Or maybe if the infra library ended up embedded in some tool (or server for a tool) like the insights report where it might have multiple traces open concurrently.  Perhaps such a scenario wouldn't overlap with a local build changing the dynamic events.  Though what is a versioning story/rules for dynamic events?  Part of the point is that they can be changed, but these datas ones are getting special treatment - what if .net 9 and .net 10 have different forms for the datas events?

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/DynamicEvents.cs
mrsharm: Yes, these aren't aligned - see: GCDynamicEvents.cs (the file I added) as reference.

src/benchmarks/gc/GC.Infrastructure/Notebooks/Reports.dib
cshung: We can't make `DynamicEvents` a property - we have C# extension method, but not C# extension property, that's a language constraint.

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/DynamicEvents.cs
cshung: I think a design like this might be appropriate.

- From the `GC.Analysis.API.dll`, we can expose a few public property that describe the dynamic event schema for each version, then
- When we open the trace, we can pick one from the list and associate it with the trace. That object from the public API is obviously just a template, we can modify it when we see fit (e.g. hacks on existing runtime with extra instrumentation)
- When `DynamicEvent()` is called, we know which trace it is, so we can figure out which schema to use for the decoding.

Is that something that will make us happy?

src/benchmarks/gc/GC.Infrastructure/Notebooks/Reports.dib
cshung: The high level goal of the dynamic event work is to make it really dynamic (i.e. as fluid as possible to introduce new event or change them) without having to involve changes in PerfView. Exposing the `byte[]` at the PerfView level abstraction and have the parsing done completely on the GC infrastructure side should future proof it.

We certainly could put all these parsing logic into PerfView, but then what are the trade offs? 

If we put the generic parsing into TraceEvent:

**Pros:**

1. We can use DynamicEvent as a property
2. We can store things on the TraceGC object (e.g. the parsed event objects)

**Cons:**

1. Next time when we want to do things in a new way again, we need to change PerfView again.
2. Any bug in TraceEvent needs some time to get in because of Nuget package publishing.

In fact, in the not too long future, I envision we will have to deal with variable-length array somehow (e.g. exposing some per heap data), so this will change again.

IMO, we have made the right trade-off. The small advantages doesn't worth the risk.

Alternatively, I could have made a `DynamicEvent` property on `TraceGC` that is a trivial read/write field of type `dynamic`. The parser on the API side can populate it and client can access it without going through extension method. This can be an option.

mrsharm:
@markples, @cshung - Implemented some of the feedback mentioned; rest, was out of scope for getting a fix for our broken notebooks. Please let me know if you have any questions / concerns.

cshung (approved):
LGTM

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/DynamicEvents/GCDynamicEvents.cs
mrsharm: Addressed the comments that are in scope. Resolving as the discussion here is out of scope - but this should be captured in an issue.

PR #4342, opened 2024-07-26
Use .NET 8.0 for scenarios libraries

caaavik-msft:
This change makes it so that Utils.csproj will get compiled with .NET 8 instead of .NET 6. The version of this library isn't too important, however there is a bug where .NET 6 projects are failing to build on the latest version of the .NET 8 SDK (https://github.com/dotnet/sdk/issues/42379), so this is good workaround to get our scenarios that run on .NET 8 working again. I also fixed an incorrect exclusion rule in the vscode settings that was causing the /src/tools directory to not be shown in VS and VS Code.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,7 +8,7 @@
         "**/logs/": true,
         "**/obj/": true,
         "**/packages/": true,
-        "tools/": true,
+        "/tools/": true,
     },
     "python.analysis.typeCheckingMode": "strict",
     "python.analysis.diagnosticSeverityOverrides": {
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -8,7 +8,7 @@
         "**/logs/": true,
         "**/obj/": true,
         "**/packages/": true,
-        "tools/": true,
+        "/tools/": true,
```
LoopedBard3, line 11: Would this make sense as **/tools/?

LoopedBard3 (approved):

.vscode/settings.json
caaavik-msft: No, since I think we only want to ignore the root tools directory. This was causing /src/tools to get ignored.

PR #4347, opened 2024-07-29
Enable log verbosity on setup build commands

caaavik-msft:
We have a few commands that get run in run_performance_job.py and this PR enables log verbosity on them so if they fail we can see why they failed. Currently the `PerfLabGenericEventSourceLTTngProvider` build is failing on ubuntu arm64 so with this in we can see the actual error message for that failure now.


LoopedBard3 (approved):
Looks good

PR #4348, opened 2024-07-29
Ensure ubuntu arm64 pipeline runs use the ubuntu dotnet prereqs container

caaavik-msft:
Runs on ubuntu arm64 were failing as lttng was not installed on the agent machine. This PR ensures that the prereqs container is used so these runs should start passing again.

### DIFF
--- a/eng/performance/build_machine_matrix.yml
+++ b/eng/performance/build_machine_matrix.yml
@@ -117,7 +117,7 @@ jobs:
       osVersion: 2204
       architecture: arm64
       pool:
-        vmImage: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


LoopedBard3 (approved):

LoopedBard3:
Looks good, though does adding this container and vmimage make a difference when we are running on baremetal?

caaavik-msft:
> Looks good, though does adding this container and vmimage make a difference when we are running on baremetal, or is it just the container for the pipeline/build steps?

This just affects the Azure Pipelines agent, it has no effect on the helix job

PR #4349, opened 2024-07-29
Update TraceEvent to 3.1.13

cshung:


src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/GC.Analysis.API.csproj
```diff
@@ -14,7 +14,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Html.Abstractions" Version="2.2.0" />
     <PackageReference Include="Microsoft.Data.Analysis" Version="0.19.1" />
-    <PackageReference Condition="'$(CustomTraceEvent)' != 'true'" Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.1.9" />
+    <PackageReference Condition="'$(CustomTraceEvent)' != 'true'" Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.1.13" />
```
mrsharm, line 17: Do we still need the "CustomTraceEvent" condition?

mrsharm:
Please update the .dib files as well with the appropriate versions of the TraceEvent library.

src/benchmarks/gc/GC.Infrastructure/GC.Analysis.API/GC.Analysis.API.csproj
cshung: Yes, it is intended to be kept.

Right now, we no longer need a custom trace event because 3.1.13 has all we want, but you never know if we will need anything new there, and by then we will need this support again.

Better just leave it here so we know what to do next time we needed the same support.

PR #4388, opened 2024-08-09
Add support for the GC reports notebook to read GC Infra gcperfsim output

markples:


mrsharm:
Curious - has this been tested with the crank runs for GCPerfSim?

markples:
> Curious - has this been tested with the crank runs for GCPerfSim?

I've been running locally.  Are they different?

mrsharm:
> > Curious - has this been tested with the crank runs for GCPerfSim?
> 
> I've been running locally. Are they different?

Yes, there is a mode that can supplied by passing in ``--server <Name Of Server>`` to run these via Crank. Worthwhile just testing if that mode works with your latest changes. 

Documentation: https://github.com/dotnet/performance/tree/main/src/benchmarks/gc/GC.Infrastructure#running-gcperfsim-scenarios-on-the-aspnet-machines

PR #4396, opened 2024-08-14
Update Maui Versions and name for runtime version

LoopedBard3:
Update Maui Versions and name for runtime version to use. The rename of the runtime version should be very rare and the Maui version updates going forward should mostly be done via maestro.

### DIFF
--- a/src/scenarios/shared/mauisharedpython.py
+++ b/src/scenarios/shared/mauisharedpython.py
@@ -17,7 +17,7 @@ def generate_maui_rollback_dict():
     # Generate and use rollback based on Version.Details.xml
     # Generate the list of versions starts to get and the names to save them as in the rollback.
     # These mapping values were taken from the previously generated rollback files for the maui workload. There should be at least one entry for each
-    # of the Maui Workload dependencies in the /eng/Version.Details.xml file, aside from VS.Tools.Net.Core.SDK.Resolver.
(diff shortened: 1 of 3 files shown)


src/scenarios/shared/mauisharedpython.py
```diff
@@ -37,11 +37,11 @@ def generate_maui_rollback_dict():
         root = ET.fromstring(version_details_xml)
 
     # Get the General Band version from the Version.Details.xml file sdk version
-    general_version_obj = root.find(".//Dependency[@Name='VS.Tools.Net.Core.SDK.Resolver']")
+    general_version_obj = root.find(".//Dependency[@Name='Microsoft.NET.Sdk']")
```
DrewScoggins, line 40: Why did this change?

DrewScoggins (approved):

src/scenarios/shared/mauisharedpython.py
LoopedBard3: I am not able to find the high level PR/issue, but it was due to VS.Tools.Net.Core.SDK.Resolver no longer being supported with Microsoft.NET.Sdk being the replacement: https://github.com/dotnet/maui/pull/23967#issuecomment-2263510484.

DrewScoggins: SGTM

PR #4440, opened 2024-09-07
[release/6.0] Disable Startup and SDK scenarios on Ubuntu

caaavik-msft:
This is backporting two changes as startup and SDK scenarios are currently broken due to LTTNG installation issues. This should make the release/6.0 scheduled runs show as green.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -538,21 +538,22 @@ jobs:
       channels:
         - release/6.0
 
+  # reenable under new machine on new ubuntu once lttng/events are available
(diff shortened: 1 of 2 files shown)


LoopedBard3 (commented):
Looks like the public Ubuntu sdk_scenario was not commented out, we should probably comment it out as well so PR runs are green.

LoopedBard3 (approved):

LoopedBard3:
@caaavik-msft Do we still want to merge this?

caaavik-msft:
Yes, thanks for the reminder

PR #4448, opened 2024-09-10
Added the Project to Facilitate Notebook Based Testing

mrsharm:
Added a Unit Test based library that makes use of dotnet-repl to run our notebooks to check to see if there are any errors in executing the cells.


mrsharm:
Hi @LoopedBard3, this is the unit test work I was mentioning earlier to add to the CI pipeline. We'd ideally want to run all the discovered tests from this project only when something is changed in the `src\benchmarks\gc\GC.Infrastructure` dir. What would be some of the next steps here to try to integrate this into the performance-ci or a possibly a new check?

cshung:
This is a great start, thanks!

I wonder if we can:

1. Do a directory listing and make sure we test all notebooks,
2. Check if any of the checked in notebook accidentally contains any execution results, and
3. Run `dotnet-format`, just to make sure the code has consistent formatting.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.NotebookTests/FunctionalTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        public void Setup()
+        {
+            if (!Utils.CheckDotnetReplInstalled())
+            {
+                Utils.InstallDotnetRepl();
+            }
+        }
+
+        [Test]
+        public void FunctionalTest_RunGCAnalysisExamples_Success()
```
markples, line 17: If you're going to list filenames, consider
``` c#
[TestCase("GCAnalysisExamples.ipynb")]
public void FunctionalTest_RunGCAnalysisExample_Success(string filename)
```

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.NotebookTests/Utils.cs
```diff
@@ -0,0 +1,134 @@
...
+            string failedNotebookText = File.ReadAllText(notebookPath);
+            NotebookRoot? deserializedNotebook = JsonConvert.DeserializeObject<NotebookRoot>(failedNotebookText);
+            deserializedNotebook.Should().NotBeNull();
+            return deserializedNotebook;
+        }
+
+        public static void ParseNotebookAndFindErrors(string outputNotebookPath, string baseNotebookPath)
+        {
+            NotebookRoot? deserializedNotebook = ParseNotebook(outputNotebookPath);
+            File.Delete(outputNotebookPath);
```
markples, line 111: `Delete` doesn't belong in a `Parse` method.  The caller is also already deleting the file.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.NotebookTests/Utils.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        public static void RunNotebookThatsExpectedToFail(string notebookPath)
+        {
+            string tempPathForOutputNotebook = Path.GetTempFileName();
+            using (Process dotnetReplProcess = SetupNotebookProcessRun(notebookPath, tempPathForOutputNotebook))
+            {
+                dotnetReplProcess.Start();
+                dotnetReplProcess.WaitForExit(TIMEOUT);
+                dotnetReplProcess.ExitCode.Should().NotBe(0);
+            }
```
markples, line 97: This path doesn't seem to delete the temp notebook.  I wonder if it's worth expanding that helper function to also execute/wait for the process and then delete the file.  I guess it would need a processing lambda to do stuff before deleting the file.

If not, then I think you want the `Delete` in a `finally` so that it always executes.

markples: But first, should the output notebooks be deleted?  When run under automation, typically files are left behind as artifacts for later inspection, either for all tests or for failing tests.  They also might be better stored out of the source tree.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.NotebookTests/Utils.cs
```diff
@@ -0,0 +1,134 @@
...
+        public static void ParseNotebookAndFindErrors(string outputNotebookPath, string baseNotebookPath)
+        {
+            NotebookRoot? deserializedNotebook = ParseNotebook(outputNotebookPath);
+            File.Delete(outputNotebookPath);
+
+            foreach (var cell in deserializedNotebook!.Cells)
+            {
+                foreach (var output in cell.Outputs)
+                {
+                    if (string.CompareOrdinal(output.Ename, "Error") == 0)
```
markples, line 117:
```suggestion
                    if (output.OutputType == "error")
```
I think this is a bit more robust - see https://nbformat.readthedocs.io/en/latest/format_description.html#error

(also for an equality check, I think this reads a bit better than Compare == 0)

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.NotebookTests/Utils.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                // If notebook execution fails, parse the notebook and find the error.
+                if (dotnetReplProcess.ExitCode != 0)
+                {
+                    ParseNotebookAndFindErrors(tempPathForOutputNotebook, notebookPath);
+                }
+
+                else
+                {
+                    dotnetReplProcess.HasExited.Should().BeTrue();
```
markples, line 81: I don't think this does anything.  If the process hasn't exited, `ExitCode` will throw.  I think you want to assert that the return value of `WaitForExit` is `true`.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.NotebookTests/Utils.cs
```diff
@@ -0,0 +1,134 @@
...
+            string tempPathForOutputNotebook = Path.GetTempFileName();
+            using (Process dotnetReplProcess = SetupNotebookProcessRun(notebookPath, tempPathForOutputNotebook))
+            {
+                dotnetReplProcess.Start();
+                dotnetReplProcess.WaitForExit(TIMEOUT);
+
+                // If notebook execution fails, parse the notebook and find the error.
+                if (dotnetReplProcess.ExitCode != 0)
+                {
+                    ParseNotebookAndFindErrors(tempPathForOutputNotebook, notebookPath);
```
markples, line 76: If this somehow doesn't find the error line (and therefore throw), I think this is going to report nunit test success for a test with a non-zero exit code.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.NotebookTests/FunctionalTests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        [Test]
+        public void FunctionalTest_IntentionallyFailedNotebook_Failure()
+        {
+            string? executionPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            Assert.IsNotNull(executionPath);
+            string failureNotebookPath = Path.Combine(executionPath, "TestNotebooks");
+            string[] failedNotebookPaths = Directory.GetFiles(failureNotebookPath);
+
+            foreach (var failedNotebookPath in failedNotebookPaths)
```
markples, line 40: This makes all of the "failed notebooks" a combined single test in test harness, so they can't be reported separately to the user.  Also if any of the individual notebooks throw an exception (not sure if a fluent assertion failure throws an exception or somehow calls into the nunit infra, but this would apply to any other uncaught exception), then the rest of the notebooks won't be tested.

Solutions could be to list the files (like `TestCase` above) or enumerate them in a `TestCaseSource`.  (Using more test features like this might be a reason to move over to xunit to match the runtime repo, but of course nunit is standard too.)

LoopedBard3:
> Hi @LoopedBard3, this is the unit test work I was mentioning earlier to add to the CI pipeline. We'd ideally want to run all the discovered tests from this project only when something is changed in the `src\benchmarks\gc\GC.Infrastructure` dir. What would be some of the next steps here to try to integrate this into the performance-ci or a possibly a new check?

After talking with the team, we think it would be best to make a new azure pipeline and associated yml file. The first step would be to verify the build and test steps locally that will eventually be followed in the yml. In parallel, we will need to create and merge a basic .yml file to create the pipeline. Once the pipeline is created, we will be able to iterate on the yml file as needed. I can handle getting a basic .yml file created and merged, and the pipeline created 👍. Is there a specific name you have in mind for the yml file, I am thinking something along the lines of gc-pr-verification-pipeline.yml?

mrsharm:
> > Hi @LoopedBard3, this is the unit test work I was mentioning earlier to add to the CI pipeline. We'd ideally want to run all the discovered tests from this project only when something is changed in the `src\benchmarks\gc\GC.Infrastructure` dir. What would be some of the next steps here to try to integrate this into the performance-ci or a possibly a new check?
> 
> After talking with the team, we think it would be best to make a new azure pipeline and associated yml file. The first step would be to verify the build and test steps locally that will eventually be followed in the yml. In parallel, we will need to create and merge a basic .yml file to create the pipeline. Once the pipeline is created, we will be able to iterate on the yml file as needed. I can handle getting a basic .yml file created and merged, and the pipeline created 👍. Is there a specific name and path you have in mind for the yml file, I am thinking something along the lines of gc-pr-verification-pipeline.yml at the top level of the repo, alongside azure-pipelines.yml?

Thanks for the response, @LoopedBard3! I think ``gc-pr-verification-pipeline.yml`` is an apt name. Just to make sure all the details are out there; a dependency of this project is that it assumes [dotnet-repl](https://www.nuget.org/packages/dotnet-repl) can be installed on the machine. Apart from that, I'll be happy to make any changes to the unit tests to facilitate they can be run using the pipeline. 

Please let me know if you have any further questions about the project; in terms of steps, I think the following are the only ones required: 

1. Download dotnet-repl using: `dotnet tool install --local dotnet-repl --version 0.1.216`.
2. Running the tests using: `dotnet test`.

mrsharm:
@cshung, @markples: Just incorporated the feedback. We are now:

1. Testing all the example notebooks with corresponding traces.
2. Have a unit test to test if there are any outputs saved to any notebooks. 
3. Updated the abstraction of running a notebook to make it clear when to relinquish resources. 
4. Added a `TestCase` based mechanism to test different notebooks. 

Please let me know if you any further feedback! Thanks!

src/benchmarks/gc/GC.Infrastructure/Notebooks/Examples/CustomTraceEventExample.ipynb
```diff
@@ -1,304 +0,0 @@
...
-    },
-    "vscode": {
-     "languageId": "polyglot-notebook"
-    }
-   },
-   "outputs": [],
-   "source": [
-    "// In order to use a custom TraceEvent library, we need to reference it directly here\n",
-    "\n",
-    "#r \"C:\\Dev\\perfview\\src\\TraceEvent\\bin\\Release\\netstandard2.0\\Microsoft.Diagnostics.Tracing.TraceEvent.dll\""
```
cshung, line 29: Do we have to delete the CustomTraceEvent example?
I think we will have times where we wanted to perform our analysis with a custom trace event (e.g. to fix a bug in TraceEvent or otherwise)

mrsharm: We didn't delete but rename the notebook as "CustomDynamicEvents".

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.NotebookTests/Utils.cs
mrsharm: Will address saving and retrieving notebook outputs in the future.

cshung (approved):
LGTM

PR #4449, opened 2024-09-10
Make GCPerfSim NativeAOT friendly

cshung:
This change fixes these warnings when published using NativeAOT by simply avoiding reflection:

```
GCPerfSim.cs: warning IL2075: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'System.Type.GetProperty(String)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

GCPerfSim.cs: Trim analysis warning IL2075: MemoryAlloc.GetProperty<T>(Object,String): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'System.Type.GetProperty(String)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```

WIth this change, the publish completed without any more warnings. The reflection is necessary only for .NET Core 3.0/3.1. Earlier versions just simply don't have that method and we will simply skip printing the numbers. I kept the old code just for this case. If we don't think we will ever want this for 3.0/3.1 (out of support anyway), we can simply delete them.

This change also make it easier to embed GCPerfSim as part of some other benchmarks:

We often want GCPerfSim to be a part of something else like request processing or background processing, this change make it easy so that we can simply drop in `GCPerfSim.cs` and call `MemoryAlloc.Test` instead of having to change the code.

### DIFF
--- a/src/benchmarks/gc/GCPerfSim/GCPerfSim.cs
+++ b/src/benchmarks/gc/GCPerfSim/GCPerfSim.cs
@@ -2713,7 +2713,7 @@ class MemoryAlloc
         return testResult;
     }
 
-    public static int Main(string[] argsStrs)
+    public static int Test(string[] argsStrs)
     {
         try
         {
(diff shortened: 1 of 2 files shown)


Maoni0 (dismissed):
:shipit:

mrsharm (approved):
LGTM!

PR #4463, opened 2024-09-17
Add JsonNode construction benchmarks

eiriktsarpalis:
Contributes to https://github.com/dotnet/runtime/issues/107869

### DIFF
--- a/src/benchmarks/micro/libraries/System.Text.Json/Node/Perf_Create.cs
+++ b/src/benchmarks/micro/libraries/System.Text.Json/Node/Perf_Create.cs
@@ -7,28 +7,54 @@ namespace System.Text.Json.Node.Tests
     [BenchmarkCategory(Categories.Libraries, Categories.JSON)]
     public class Perf_Create
     {
+        private readonly JsonNode[] _results = new JsonNode[50];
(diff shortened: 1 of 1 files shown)


LoopedBard3 (approved):
I think this is a good start, although we will probably want to put the super fast benchmarks (<100ns per iteration) into a loop or something to get their mean time closer to 200-300ns in order to help minimize noisiness. If you have an idea of how to do this while keeping the tests inline with their goal and can make the change, that would be great, otherwise doing this exercise is one of the items we want to do.

LoopedBard3:
@eiriktsarpalis Left a comment and approved the PR, whether you include the change mentioned in the approval, ping me when you are ready for this to be merged and I can merge it.

eiriktsarpalis:
Something like 26e65c2eb6c01d2b76af1f884ec07cfc0e3f62b8?

LoopedBard3 (approved):

LoopedBard3:
Yes, exactly like that 👍. Thanks! Just to double check, are you good having this merged once the tests run?

eiriktsarpalis:
Yep, go for it!

PR #4469, opened 2024-09-19
Fix configs for wasm aot and mono aot

caaavik-msft:
This bit of the performance-setup.sh was not ported properly in the move to python pipelines. This will have caused missing data, but should not have caused any bad data

### DIFF
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -227,7 +227,7 @@ def get_pre_commands(args: RunPerformanceJobArgs, v8_version: str):
             ]
 
     # Set MONO_ENV_OPTIONS with for Mono Interpreter runs
-    if args.codegen_type == "Interpreter" and args.runtime_type == "mono":
+    if args.codegen_type.lower() == "interpreter" and args.runtime_type == "mono":
         if args.os_group == "windows":
             helix_pre_commands += ['set MONO_ENV_OPTIONS="--interpreter"']
         else:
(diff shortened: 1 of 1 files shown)


scripts/run_performance_job.py
```diff
@@ -474,7 +474,7 @@ def run_performance_job(args: RunPerformanceJobArgs):
     configurations = { "CompilationMode": "Tiered", "RunKind": args.run_kind }
 
     using_mono = False
-    if mono_dotnet is not None:
+    if mono_dotnet is not None or mono_aot:
```
LoopedBard3, line 477: It seems this will pass exclusions for NoMono and NoAOT for mono_aot runs, it seems that the previous setup would have instead had NoAOT and NoWASM (https://github.com/dotnet/runtime/blob/00cb272da591b6c5466a294260a5f063f2960fac/eng/testing/performance/performance-setup.sh#L356-L364). I am not sure if NoWASM was correct in the first place, but is something we should double check/watch for in the test run.

scripts/run_performance_job.py
```diff
@@ -568,7 +568,7 @@ def run_performance_job(args: RunPerformanceJobArgs):
     if not args.internal and not args.performance_repo_ci:
         ci_setup_arguments.not_in_lab = True
 
-    if mono_dotnet is not None:
+    if mono_dotnet is not None and not mono_aot:
```
LoopedBard3, line 571: Not sure if there is a case where we could have both mono_dotnet and mono_aot so this may not strictly be necessary, but it is probably good to still verify that it is not mono_aot.

caaavik-msft: Yeah it's not possible, but just porting the exact same code from the previous script

LoopedBard3 (approved):
Looks good.

LoopedBard3 (approved):
LGTM

PR #4474, opened 2024-09-23
Update BenchmarkDotNet to 0.14.1-nightly.20240924.187

caaavik-msft:
This change is being done to pull the changes from https://github.com/dotnet/BenchmarkDotNet/pull/2645 which will allow for us to run the wasm performance tests targeting .NET 9 while the .NET 10 SDK is installed. This should unblock https://github.com/dotnet/runtime/pull/106599.

I have started a test run of the pipeline internally with this change and the .NET 10 change which I want to wait for before we merge this in.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,21 +1,21 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.NET.ILLink.Tasks" Version="9.0.0-rtm.24468.5">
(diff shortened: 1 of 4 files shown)


LoopedBard3 (dismissed):

lewing (approved):

caaavik-msft:
Running the pipeline again, just on wasm and mono aot so should finish a bit faster. Once that is all good, will merge in.

lewing:
I moved the BDN ref from the middle to the bottom of the other references so it should cause fewer update conflicts

lewing (approved):

PR #4485, opened 2024-09-26
Pass argument into scenarios to enable uploading results

caaavik-msft:
This PR does two things:
1. Adds a command line argument to our scenarios jobs that controls if we upload the results to Azure Storage
2. Adds an environment variable called ScenarioArgs which is passed into the command line of every scenario command.


LoopedBard3 (approved):
This looks good to me 👍.

caaavik-msft:
I realise this will cause the release/9.0 runs to fail so we should merge in https://github.com/dotnet/runtime/pull/108412 first before merging this one.

PR #4489, opened 2024-09-27
Fixing Syntax of the GC Pipeline Yaml

mrsharm:
- Fixing Syntax of the GC Pipeline Yaml. 
- Temporarily disable long running examples to get the pipeline running. 
- Excluding running the performance pipeline in case no changes are made in the GC Infra.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,7 +23,7 @@ trigger:
     - docs/*
     - '**/*.md'
     #- 'gc-azure-pipelines.yml'
-    - src/benchmarks/gc/*
+    #- src/benchmarks/gc/*
 
 pr:
   branches:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,7 +22,7 @@ trigger:
     exclude: # don't trigger if only docs and similar files changed
     - docs/*
     - '**/*.md'
-    #- 'gc-azure-pipelines.yml'
+    - 'gc-azure-pipelines.yml'
     - src/benchmarks/gc/*
 
 pr:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,8 +22,8 @@ trigger:
     exclude: # don't trigger if only docs and similar files changed
     - docs/*
     - '**/*.md'
-    #- 'gc-azure-pipelines.yml'
-    #- src/benchmarks/gc/*
+    - 'gc-azure-pipelines.yml'
+    - src/benchmarks/gc/*
 
 pr:
   branches:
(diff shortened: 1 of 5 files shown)


LoopedBard3:
@mrsharm it would probably be worth adding the gc-azure-pipelines.yml file to the normal azure-pipelines.yml exclude list, and maybe adding it to the include list of the gc-azure-pipelines.yml file (I am not sure if a yml considers itself as always in the include list or not 🤷‍♂️).

azure-pipelines.yml
```diff
@@ -22,6 +22,8 @@ trigger:
     exclude: # don't trigger if only docs and similar files changed
     - docs/*
     - '**/*.md'
+    - 'gc-azure-pipelines.yml'
```
LoopedBard3, line 25: It would probably also be worth adding the gc-azure-pipelines.yml file to the exclude list for pr's just below here.

LoopedBard3 (dismissed):
Left one comment. Looks good overall as well.

LoopedBard3 (dismissed):
This looks good to me 👍.

mrsharm:
Thanks for all the help, @LoopedBard3 and @DrewScoggins! This PR has got us to test a large part of the notebook functionality and I observe that the tests are all green from the pipeline:

image

And successfully running the tests: https://dev.azure.com/dnceng-public/public/_build/results?buildId=822709&view=logs&j=6f2f97bc-132f-5f9d-ae17-411e38c19357

image

Will follow up with re-enabling the example notebook tests as these are timing out on the lab machines even though successfully running locally.

gc-azure-pipelines.yml
```diff
@@ -33,17 +33,24 @@ jobs:
...
+      vmImage: windows-2019
  
     steps:
     # Install dotnet.
     - task: UseDotNet@2
       displayName: Install .NET 8.0 
       inputs:
-        version: 8.x
+        version: 8.0.x
+    - task: UseDotNet@2
```
cshung, line 45: Why do we install both 8.0 and 9.0 just to run the tests targetting 8.0?

mrsharm: We need the NET 9.0 one because that's what's expected from the global.json to install the dependency of dotnet-repl. The projects are all 8.0 - we'll move these to .NET 9 once it's out but needed to get these to run unless anyone has better suggestions to fix this.

cshung: ok - it is just unclear why from the context. A comment would be nice.

cshung (approved):

DrewScoggins (approved):
LGTM

PR #4520, opened 2024-10-15
[M] Cleanup Reporting.Reporter tool

e-kharion:
Reporting.Reporter tool polished

### DIFF
--- a/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
@@ -18,6 +18,8 @@ namespace BenchmarkDotNet.Extensions
 {
     public static class RecommendedConfig
     {
+        private static IEnvironment Environment = new EnvironmentProvider();
+
         public static IConfig Create(
             DirectoryInfo artifactsPath,
             ImmutableHashSet<string> mandatoryCategories,
@@ -58,7 +60,7 @@ namespace BenchmarkDotNet.Extensions
                 .AddValidator(new UniqueArgumentsValidator()) // don't allow for duplicated arguments #404
                 .WithSummaryStyle(SummaryStyle.Default.WithMaxParameterColumnWidth(36)); // the default is 20 and trims too aggressively some benchmark results
 
-            if (new EnvironmentProvider().IsLabEnvironment())
+            if (Environment.IsLabEnvironment())
             {
                 config = config.AddExporter(new PerfLabExporter());
             }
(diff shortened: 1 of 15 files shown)


src/tools/Reporting/Reporting/EnvironmentProvider.cs
```diff
@@ -3,13 +3,13 @@
...
+using System.Collections;
 
 namespace Reporting;
 
 public class EnvironmentProvider : IEnvironment
 {
     public string GetEnvironmentVariable(string variable) => Environment.GetEnvironmentVariable(variable);
-    public System.Collections.IDictionary GetEnvironmentVariables() => Environment.GetEnvironmentVariables();
+
+    public IDictionary GetEnvironmentVariables() => Environment.GetEnvironmentVariables();
```
DrewScoggins, line 14: Why move this onto one line, but not the property above?

src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
```diff
@@ -58,7 +58,7 @@ public static IConfig Create(
                 .AddValidator(new UniqueArgumentsValidator()) // don't allow for duplicated arguments #404
                 .WithSummaryStyle(SummaryStyle.Default.WithMaxParameterColumnWidth(36)); // the default is 20 and trims too aggressively some benchmark results
 
-            if (Reporter.CreateReporter().InLab)
+            if (new EnvironmentProvider().IsLabEnvironment())
```
DrewScoggins, line 61: Could we just make this a static method we call on EnvironmentProvider? Save the allocation.

DrewScoggins (dismissed):
I left a few comments, but overall this LGTM.

src/tools/Reporting/Reporting/EnvironmentProvider.cs
e-kharion: I wanted to insert an empty line between methods.

src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs
e-kharion: I wanted to have the possibility to mock environment for unit testing - thus the interface and it's implementation. 
At the same time I wanted to refactor only a limited portion code - thus the least possible changes outside the Reporting solution (like here in BenchmarkDotNet.Extensions). However good point with the allocation. I introduced a new static variable of the IEnvironment so the EnvironmentProvider is created only once.

LoopedBard3 (approved):

DrewScoggins (approved):

PR #4522, opened 2024-10-16
Add some benchmarks for array deabstraction

AndyAyersMS:
Various "equivalent" tests for enumerating arrays, illustrating the abstraction penalty.

See dotnet/runtime#108913

### DIFF
--- a/src/benchmarks/micro/runtime/DeAbstraction/Array.cs
+++ b/src/benchmarks/micro/runtime/DeAbstraction/Array.cs
@@ -1,0 +1,224 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@DrewScoggins PTAL
cc @dotnet/jit-contrib

LoopedBard3 (approved):
Looks good presuming there are no configs we need to filter out (does this work on WASM, Mono, and MonoAOT?).

AndyAyersMS:
> Looks good presuming there are no configs we need to filter out (does this work on WASM, Mono, and MonoAOT?).

It is fairly vanilla C# so I'd hope it works everywhere.

AndyAyersMS:
Looks like CI is unhappy though ?

LoopedBard3:
Yea, the ci is unhappy but not due to your changes. The error we are seeing is `Unable to find package Microsoft.NETCore.App.Ref with version (= 8.0.11)`. Not sure exactly why the main runs are hitting this, but it probably it is occurring when we build some tools so they may just need to be updated to target 9 instead of 8.

AndyAyersMS:
So can you merge? I can't....

PR #4523, opened 2024-10-17
Fix targetting for 9.0.0 packages.

LoopedBard3:
Update Util.csproj to use the same TargetFramework assignment logic as other projects in ScenarioMeasurement. Also removed the no longer used LibrariesTargetFramework property from the Scenario Measurement Directory.Build.props.

Aside from Util.csproj changes, added temporary feeds with references for 9.0.0 packages to fix the build failure for now.

Fixes: https://github.com/dotnet/performance/issues/4524


LoopedBard3:
The fsharpmicro runs seem to still hit issues with FSharp.Core missing for version >= 9.0.100, though I am not finding the proper dependency in darc to get this to work. This is currently being hit so I think we should get this merged to fix most of the other failing cases and come back to FSharp.Core in a different PR.

DrewScoggins:
Do we understand all of the other failures that we are seeing here? From a quick glance it looks like blazor_scenarios, sdk_scenarios, and Ubuntu micros.

DrewScoggins (approved):
LGTM. Failures are unrelated and will be fixed in later PRs.

LoopedBard3:
Partially, blazor_scenarios main seem to be failing to install wasm-tools, linux micro runs has 4 partitions failing due to a security exception or something when trying to run some SSL stream microbenchmarks, and fsharpmicro main are failing due to not being able to find the dependency for FSharp.Core (similar to what this is fixing and mentioned above).

PR #4570, opened 2024-11-12
Supress SYSLIB0057

lilinus:
I had trouble compiling since `X509Certificate2` ctor is marked as obsolete as of .NET 9.


lilinus:
My bad, seems like I forgot to pull the latest main before compiling and filing the PR.

PR #4581, opened 2024-11-21
Enable iteration for microbenchmarks

VincentBu:
Enable iteration for microbenchmarks. Now we can repeat bdnProcess several times based on `Environment.iteration` and each iteration has its own result folder and trace files.


mrsharm (approved):
LGTM - looks like this was tested and confirmed there are separate iteration folders, as expected. And the `run` command was successfully run.

PR #4617, opened 2025-01-03
Cert rotator

DrewScoggins:
Adds certificate handling to auth code for upload


caaavik-msft (dismissed):
LGTM, just two comments to optionally address

scripts/upload.py
```diff
@@ -32,14 +34,25 @@ def upload(globpath: str, container: str, queue: str, sas_token_env: str, storag
...
             dac = DefaultAzureCredential()
-            credential = ClientAssertionCredential(TENANT_ID, CLIENT_ID, lambda: dac.get_token("api://AzureADTokenExchange/.default").token)
+            credential = ClientAssertionCredential(TENANT_ID, ARC_CLIENT_ID, lambda: dac.get_token("api://AzureADTokenExchange/.default").token)
             credential.get_token("https://storage.azure.com/.default")
         except ClientAuthenticationError as ex:
-            getLogger().info("Unable to use managed identity. Falling back to environment variable.")
-            credential = os.getenv(sas_token_env)
+            credential = None
+            getLogger().info("Unable to use managed identity. Falling back to certificate.")
+            cmd_line = list((os.path.join(str(helixpayload()), 'certhelper', 'CertHelper.exe')))
```
e-kharion, line 42: is this meant to run on windows only?

e-kharion (dismissed):

scripts/upload.py
DrewScoggins: It is supposed to run on all platforms. Good catch, I will make this part platform agnostic.

scripts/upload.py
```diff
@@ -32,14 +34,25 @@ def upload(globpath: str, container: str, queue: str, sas_token_env: str, storag
...
+                    credential.get_token("https://storage.azure.com/.default")
+                except ClientAuthenticationError as ex:
+                    credential = None
+                    continue
         if credential is None:
-            getLogger().error("Sas token environment variable {} was not defined.".format(sas_token_env))
-            return 1
+            getLogger().error("Unable to authenticate with managed identity or certificates.")
+            getLogger().info("Falling back to environment variable.")
+            credential = os.getenv(sas_token_env)
```
LoopedBard3, line 55: If we are going to keep the envvar credential, I think we should also keep the check to see if it is None that it had before.

LoopedBard3 (approved):

scripts/upload.py
LoopedBard3: Although I could see us removing the SAS token functionality entirely once we have everything setup.

DrewScoggins: That was my plan. Is that we would be getting rid of this fallback in a couple of days.

DrewScoggins:
I was thinking, maybe we should move the behavior of CertRotator into CertHelper. That way we always bring along what we need on every run, and we don't have to worry about pushing updates to CertRotator out to machines.

LoopedBard3:
Would the idea be to then run the CertRotator functionality every run as well? I think that makes sense, especially from the standpoint of having a clear update flow.

DrewScoggins:
Yeah, that would be the idea. We would check on the certs before we started the upload.

src/tools/CertHelper/CertHelper.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
```
LoopedBard3, line 5: Looking at the test run results, it appears that you will need to use the same TargetFramework setup that we have for our other tools. Something along these lines:
```
    <TargetFramework>$(PERFLAB_TARGET_FRAMEWORKS)</TargetFramework>
    <!-- Supported target frameworks -->
    <TargetFramework Condition="'$(TargetFramework)' == ''">$(FallbackTargetFramework)</TargetFramework>
```

e-kharion (approved):

azure-pipelines.yml
```diff
@@ -459,135 +459,135 @@ jobs:
...
-        - win-x86
-        #- ubuntu-x64-1804 reenable under new machine on new ubuntu once lttng/events are available
-      isPublic: false
-      jobParameters:
-        kind: sdk_scenarios
-        projectFile: sdk_scenarios.proj
-        channels:
-          - main
-          - 8.0
+  # # SDK scenario benchmarks
```
LoopedBard3, line 462: We will want to make sure we uncomment these tests before merging.

src/tools/CertHelperTests/KeyVaultCertTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        mockCertClient = new Mock<CertificateClient>(new Uri("https://dotnetperfkeyvault.vault.azure.net/"), mockTokenCred.Object);
+        mockSecretClient = new Mock<SecretClient>(new Uri("https://dotnetperfkeyvault.vault.azure.net/"), mockTokenCred.Object);
+        KeyVaultCertificateWithPolicy? mockCert1, mockCert2;
+        X509Certificate2 cert1, cert2;
+        MakeCerts(out mockCert1, out mockCert2, out cert1, out cert2, localAndKeyVaultDifferent);
+
+        var certCollection = new X509Certificate2Collection { cert1, cert2 };
+
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
```
LoopedBard3, line 52: I think if we are going to have multiple disables in a very short section, it would be fine to have a single one with a comment that it is for all GetCertificateAsync or whatever calls.

src/tools/CertHelper/Constants.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CertHelper;
+public class Constants
+{
+    public static readonly string Cert1Name = "LabCert1";
+    public static readonly string Cert2Name = "LabCert2";
+    public static readonly Uri Cert1Id = new Uri("https://test.vault.azure.net/certificates/LabCert1/07a7d98bf4884e5c40e690e02b96b3b4");
```
LoopedBard3, line 12: Are these used for anything other than testing?

DrewScoggins: Nope, we just need a uri that looks like one from a key vault.

src/tools/CertHelper/KeyVaultCert.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            throw new Exception("Certificate not found in Key Vault");
+        }
+        var secret = await SecretClient.GetSecretAsync(keyVaultCert.Value.Name, keyVaultCert.Value.SecretId.Segments.Last());
+        if(secret.Value == null)
+        {
+            throw new Exception("Certificate secret not found in Key Vault");
+        }
+        var certBytes = Convert.FromBase64String(secret.Value.Value);
+#if NET9_0_OR_GREATER
```
DrewScoggins, line 86: This is the only change in the code, I believe.

DrewScoggins:
Test runs for Arm64 are here, https://dev.azure.com/dnceng/internal/_build/results?buildId=2622437&view=results.

Alpine and iOS are here, https://dev.azure.com/dnceng/internal/_build/results?buildId=2622440&view=results.

src/tools/CertHelper/KeyVaultCert.cs
```diff
@@ -0,0 +1,110 @@
...
+        SecretClient = secretClient ?? new SecretClient(new Uri(KeyVaultUrl), Credential);
+        KeyVaultCertificates = new X509Certificate2Collection();
+    }
+
+    public async Task GetKeyVaultCerts()
+    {
+        KeyVaultCertificates.Add(await FindCertificateInKeyVault(Constants.Cert1Name));
+        KeyVaultCertificates.Add(await FindCertificateInKeyVault(Constants.Cert2Name));
+
+        if (KeyVaultCertificates.Count != 2 || KeyVaultCertificates.Where(c => c == null).Count() > 0)
```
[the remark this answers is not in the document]
DrewScoggins: I think I will just get rid of that condition. I think we should keep the check if any of the entries are null. While I am not sure we could end up in that state, I would like to protect against it.

sblom (approved):

PR #4622, opened 2025-01-07
Update BDN version to support more net10.0 testing.

LoopedBard3:
Update BDN version to support more net10.0 testing. This should fix WASM runs.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -11,7 +11,7 @@
     <MicrosoftNETILLinkPackageVersion>10.0.0-alpha.1.24617.3</MicrosoftNETILLinkPackageVersion>
     <SystemThreadingChannelsPackageVersion>10.0.0-alpha.1.24617.3</SystemThreadingChannelsPackageVersion>
     <MicrosoftExtensionsLoggingPackageVersion>10.0.0-alpha.1.24617.3</MicrosoftExtensionsLoggingPackageVersion>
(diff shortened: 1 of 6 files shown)


e-kharion (dismissed):
LGTM

caaavik-msft:
It seems in the BDN update that the identifier for wasm on .NET 10 is wasmnet10_0, and Mono is Mono10_0. I think we need to update the code to use the underscore for it to work?

LoopedBard3:
> It seems in the BDN update that the identifier for wasm on .NET 10 is wasmnet10_0, and Mono is Mono10_0. I think we need to update the code to use the underscore for it to work?

Probably, at least for wasmnet10_0. I will check and see for mono but we may be using the generic version there.

LoopedBard3:
I verified that we only use the versioned runtime argument for wasm runs. MonoAotLLVM uses monoaotllvm and mono JIT does not appear to pass anything. So I updated the argument we pass to for wasm to be wasmnet10_0.

LoopedBard3:
Test run here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2616919&view=results running against runtime main and this performance branch shows wasm wasm run fixed and wasm aot mostly successful, though with some `Segmentation fault (core dumped)` failures. I think we should merge this and continue to investigate the new errors in a new thread. 

Other failures were either already known, i.e. Mono AOT Regex and crossgen failures, or were a handful of dead-lettered jobs with others succeeding.

DrewScoggins (approved):
LGTM

PR #4641, opened 2025-01-15
Add benchmarking for whitespace detection in spans

AlexRadch:
This commit introduces a new method `IsWhiteSpace` to benchmark the detection of whitespace in spans.

### DIFF
--- a/src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
+++ b/src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
@@ -95,6 +95,26 @@ namespace System.Memory
             yield return "abcdefg";
         }
 
+        [Benchmark]
+        [ArgumentsSource(nameof(IsWhiteSpaceArguments))]
+        [MemoryRandomization]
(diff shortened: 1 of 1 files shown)


src/benchmarks/micro/libraries/System.Memory/ReadOnlySpan.cs
```diff
@@ -95,6 +95,26 @@ public static IEnumerable<object> TrimArguments()
...
+            yield return "0abcdefg";
+            yield return " 1abcdefg";
+            yield return "  2abcdefg";
+            yield return "    4abcdefg";
+            yield return "      6abcdefg";
+            yield return "       7abcdefg";
+            yield return "        8abcdefg";
+            yield return "         9abcdefg";
+            yield return "                16abcdefg";
+            yield return "                                32abcdefg";
```
e-kharion, line 115: I'd consider adding other whitespace characters than just a space.
See https://learn.microsoft.com/en-us/dotnet/api/system.char.iswhitespace?view=net-9.0 for more info.

PR #4651, opened 2025-01-21
Write gcperfsim result in json

VincentBu:
Add a functionality to GC.Infrastructure which can write comparison result of LowVolatilityRun to a json file.
For `gcperfsim` command: add `- json` to `output.formats`(config file)
For `gcperfsim-compare` command: specify output file with "json" extension


mrsharm (approved):
LGTM!

PR #4657, opened 2025-01-22
Revert global.json to match dotnet/runtime

LoopedBard3:
Revert global.json to match what dotnet/runtime has as wasm runs are failing due to using the runtimes installed dotnet and not the latest daily dotnet like our other runs, making it fail due to too new of a minimum in global.json.

There is currently a blocking issue to merge the latest dotnet/arcade update and we will have to wait for it to go in before we can update it to the same in our repo.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,15 +1,15 @@
 {
   "sdk": {
-    "version": "10.0.100-alpha.1.25064.3",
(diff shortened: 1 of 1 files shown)


caaavik-msft (approved):

LoopedBard3:
Failures are the same as being seen elsewhere, merging.

PR #4664, opened 2025-01-27
Add doc for android startup scenario local testing.

LoopedBard3:
Add first iteration of a doc for android startup scenario local testing.


LoopedBard3:
Lint failures are unrelated to this PR, or markdown we manage. The failures are in eng/common/template-guidance.md.

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
+
+# Android Startup Test
+
+## Prereqs
+
+- Ensure `dotnet` is installed and available with the `dotnet` command. [Dotnet Download](https://dotnet.microsoft.com/en-us/download) or [Daily Dotnet Download](https://github.com/dotnet/sdk/blob/main/documentation/package-table.md)
```
matouskozak, line 6: Do we have any requirement regarding minimal .net version?

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
...
+- Ensure `xharness` is installed and available with the `xharness` command. The current version in use can be found in the `eng/performance/maui_scenarios_android.proj` file at line 7 (under the tag `MicrosoftDotNetXHarnessCLIVersion`), although any recent version should work. [XHarness Install Instructions](https://github.com/dotnet/xharness?tab=readme-ov-file#installation-and-usage).
+- Have an Android app APK available for testing.
+
+## Steps
+
+1. Initialize the environment (note the . for bash):
+
+    ```sh
+    cd src/scripts
```
matouskozak, line 15:
```suggestion
    cd src/scenarios
```

I think this should point to `scr/scenarios` folder instead (https://github.com/dotnet/performance/blob/main/docs/scenarios-workflow.md#linux).

matouskozak (commented):
Thanks a lot for submitting the doc!

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
...
+- Have an Android app APK available for testing.
+
+## Steps
+
+1. Initialize the environment (note the . for bash):
+
+    ```sh
+    cd src/scenarios
+    . ./init.sh  # or .\init.ps1 on Windows
+    cd ../
```
matouskozak, line 17:
```suggestion

```

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
...
+    ```
+
+2. Navigate to the `helloandroid` scenario directory:
+
+    ```sh
+    cd ./scenarios/helloandroid
```
matouskozak, line 23:
```suggestion
    cd ./helloandroid
```

matouskozak (commented):
These are no longer necessary I believe

docs/android-startup-scenarios.md
LoopedBard3: Good catches, updated.

docs/android-startup-scenarios.md
LoopedBard3: Added a comment on the .NET Core version to have. Which should be any supported version. Although this may not be true depending on if and how supporting startup testing tools are generated.

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
...
+    cd helloandroid
+    ```
+
+3. Copy the APK into the `helloandroid` directory.
+4. Run the test:
+
+    ```sh
+    python test.py devicestartup --device-type android --package-path <path to apk (e.g. .)>/<apkname>.apk --package-name <apk package name> [--disable-animations]
+    ```
+
```
caaavik-msft, line 32: Maybe add information about where the results of the performance test can be found or an example of what the output should look like.

caaavik-msft (dismissed):

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
...
+- Have an Android app APK available for testing.
+- Have an Android Device (with developer mode enabled) or emulator connected to computer, and viewable with `xharness android device` or `xharness android adb -- devices -l`.
+
+## Steps
+
+1. Initialize the environment (note the . for bash):
+
+    ```sh
+    cd src/scenarios
+    . ./init.sh  # or `.\init.ps1` on Windows
```
ivanpovazan, line 17: What about custom dotnet locations? 
Afaik there is/was `-dotnetdir` parameter which could specify custom dotnet installation.

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
...
+    ```
+
+2. Navigate to the `helloandroid` scenario directory:
+
+    ```sh
+    cd helloandroid
+    ```
+
+3. Copy the APK into the `helloandroid` directory.
```
ivanpovazan, line 26: Is there any correlation between dotnet version used for building the .apk and the one used as part of the environment?

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
...
+4. Run the test:
+
+    ```sh
+    python test.py devicestartup --device-type android --package-path <path to apk (e.g. .)>/<apkname>.apk --package-name <apk package name> [--disable-animations]
```
ivanpovazan, line 30: Are there any requirements on python version?

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,35 @@
...
+    ```
+
+## Notes
+
+- Example commands and additional logic can be found in the `maui_scenarios_android.proj` and `runner.py` files in the `performance` repository.
```
ivanpovazan, line 35: I would prefer to have here the exact command which measures the device startup.
The examples in `maui_scenarios_android.proj` are decorated with different MSBuild properties and cover different scenarios.

docs/android-startup-scenarios.md
matouskozak: This script should work with any .apk, cc: @LoopedBard3

docs/android-startup-scenarios.md
matouskozak: We could add an example something like: `python test.py devicestartup --device-type android --package-path HelloAndroid.apk --package-name net.dot.HelloAndroid` for the dotnet/runtime sample app

docs/android-startup-scenarios.md
LoopedBard3: There should not be any correlation if using a separately built APK. There are scenarios that this does not currently cover where the dotnet version would matter, e.g. when building Maui apps with pre.py the dotnet version will matter.

docs/android-startup-scenarios.md
LoopedBard3: Good callout, added to the prereqs.

docs/android-startup-scenarios.md
LoopedBard3: Added comment on this.

docs/android-startup-scenarios.md
LoopedBard3: Added a specific example.

docs/android-startup-scenarios.md
LoopedBard3: Added section on this.

matouskozak (dismissed):
LGTM!

docs/android-startup-scenarios.md
ivanpovazan: I am a bit confused, could you please clarify:
- Is this doc specific to HelloAndroid (the `dotnet/runtime`) sample, or it refers to any Android app - any .apk?
    - If it is the former - only HelloAndroid sample, there are couple of issues I can see with it:
        - we should have a more general solution and a doc for it, not just HelloAndroid sample, since the sample is not a real representative of an .NET Android application
        - it should be pointed that .apks of HelloAndroid sample can produce different results before and after: https://github.com/dotnet/runtime/pull/111742 gets merged to main and with it, it will matter which dotnet version is used
    - If it is the latter, then we should rename the scenario to avoid confusion (or create a new one which would be just for this purpose). Maybe `androidstartup` or something like that, the point is that it shouldn't refer to HelloAndroid in any way for the above mentioned reasons

docs/android-startup-scenarios.md
ivanpovazan: Thanks!

docs/android-startup-scenarios.md
LoopedBard3: The doc is more general for any APK, so I went ahead and added a new scenario folder named genericandroidstartup to make it more clear that it can be used for any APK. The reason we have the specific scenarios for some of the other apps is some of the pre.py files can build the apps (like the mauiandroid), but most of the test.py's for the android testing are interchangeable.

docs/android-startup-scenarios.md
```diff
@@ -0,0 +1,62 @@
...
+    ```sh
+    cd genericandroidstartup
+    ```
+
+3. Copy the APK into the `genericandroidstartup` directory.
+4. Run the test:
+
+    ```sh
+    python/(py -3 on Linux) test.py devicestartup --device-type android --package-path <path to apk (e.g. .)>/<apkname>.apk --package-name <apk package name> [--disable-animations] [--use-fully-drawn-time --fully-drawn-extra-delay <delay in sec> (**see below for note**)]
+    ```
```
ivanpovazan, line 32: This looks cryptic when observed as a shell command. I suggest splitting it into multiple commands for example:

- non-Linux

```sh
python test.py devicestartup --device-type android --package-path <path-to-apk> --package-name <apk-package name> [--disable-animations] [--use-fully-drawn-time --fully-drawn-extra-delay <delay-in-sec>]
``` 

- Linux

```sh
py -3 test.py devicestartup --device-type android --package-path <path-to-apk> --package-name <apk-package name> [--disable-animations] [--use-fully-drawn-time --fully-drawn-extra-delay <delay-in-sec>]
``` 

- Please refer to the [Notes](./android-startup-scenarios#notes) below about specifying `--use-fully-drawn-time --fully-drawn-extra-delay` parameters

src/scenarios/genericandroidstartup/pre.py
```diff
@@ -0,0 +1,49 @@
...
+from performance.logger import setup_loggers, getLogger
+from shutil import copyfile
+from shared.precommands import PreCommands
+from shared.const import PUBDIR
+from argparse import ArgumentParser
+
+setup_loggers(True)
+
+parser = ArgumentParser()
+parser.add_argument('--unzip', help='Unzip APK and report extracted tree', action='store_true', default=False)
```
ivanpovazan, line 16: Do we need the unzip logic if this scenario is only about startup measurements ?

matouskozak (commented):
I'm unsure about the `py -3` suggestion for Linux... Isn't `py -3` specific to Windows Python Launcher? In my experience, I always use `python` or `python3` on unix systems...

I think we should have example for:
- Windows
- Unix-based (that should work for both Linux and macOS)

LoopedBard3:
> I'm unsure about the `py -3` suggestion for Linux... Isn't `py -3` specific to Windows Python Launcher? In my experience, I always use `python` or `python3` on Ubuntu or macOS.
> 
> I think we should have example for:
> 
> * Windows
> * Unix-based (that should work for both Linux and macOS)

Looked again. It seems we should be able to get away with using just python on both Linux and Windows. So I switched to just using python in the doc.

src/scenarios/genericandroidstartup/pre.py
LoopedBard3: Updated as we shouldn't need it.

docs/android-startup-scenarios.md
LoopedBard3: Seems we can use python for both windows and Linux, so I kept just 1 reference. Also added the reference to the notes separately.

matouskozak (approved):

ivanpovazan (approved):
LGTM, thanks!

PR #4672, opened 2025-01-29
Add net9.0 runs into the performance repo channel runs.

LoopedBard3:
Add net9.0 runs into the performance repo channel runs.


LoopedBard3:
All failures are the same as what we are seeing on main. Merging.

PR #4680, opened 2025-02-04
Fix for Mac Certs

DrewScoggins:
From my work it does not appear that there is any programmatic way to recover the public/private key pair in a Key Chain without having to enter a password. As a result, we are going to grab the certificates directly off the disk, like we do on Linux machines.

### DIFF
--- a/scripts/performance/common.py
+++ b/scripts/performance/common.py
@@ -38,6 +38,9 @@ def get_machine_architecture():
 def iswin():
     return sys.platform == 'win32'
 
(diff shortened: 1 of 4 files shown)


LoopedBard3 (approved):
Looks good to merge, especially if we have a test run.

DrewScoggins:
Yeah, I have test runs on both Mac and non-Mac queues.

caaavik-msft (approved):

PR #4699, opened 2025-02-10
System.Reflection.Metadata.TypeName benchmarks

adamsitnik:
So far, we were benchmarking the `TypeName.Parse` implementation only in an indirect way, when benchmarking `Type.GetType` (which internally uses the same parser). We are now considering changing some parts of the implementation (https://github.com/dotnet/runtime/pull/112242), so it would be good to understand the perf and make sure we don't introduce any regressions.


jkotas (dismissed):
Thanks

adamsitnik:
@caaavik-msft could you please merge this PR? Thanks!

caaavik-msft (approved):

PR #4710, opened 2025-02-12
Update the runtime jobs we are running

LoopedBard3:
Copied current tiger runs to also run on viper queue so we have some time of overlapping data running as well as removed the nodynamicpgo and nor2r runs as they are not currently being looked at. The viper runs are for preparation to downsize our Viper machine queue, and the nodynamicpgo and nor2r runs are no longer being looked at so this should help our compute (talked with @AndyAyersMS about this offline).

Runtime pipeline test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2642451&view=results

### DIFF
--- a/eng/pipelines/runtime-perf-jobs.yml
+++ b/eng/pipelines/runtime-perf-jobs.yml
@@ -272,27 +272,7 @@ jobs:
         runtimeRepoAlias: ${{ parameters.runtimeRepoAlias }}
         performanceRepoAlias: ${{ parameters.performanceRepoAlias }}
         ${{ each parameter in parameters.jobParameters }}:
-          ${{ parameter.key }}: ${{ parameter.value }}
-
-  # run coreclr perfviper microbenchmarks perf jitoptrepeat jobs
-  - template: /eng/pipelines/common/platform-matrix.yml@${{ parameters.runtimeRepoAlias }}
-    parameters:
-      jobTemplate: /eng/pipelines/templates/runtime-perf-job.yml@${{ parameters.performanceRepoAlias }}
-      buildConfig: release
-      runtimeFlavor: coreclr
-      platforms:
-      - linux_x64
-      - windows_x64
-      jobParameters:
-        liveLibrariesBuildConfig: Release
-        runKind: micro
-        logicalMachine: 'perfviper'
-        experimentName: 'jitoptrepeat'
-        additionalJobIdentifier: jitoptrepeat
-        runtimeRepoAlias: ${{ parameters.runtimeRepoAlias }}
-        performanceRepoAlias: ${{ parameters.performanceRepoAlias }}
-        ${{ each parameter in parameters.jobParameters }}:
-          ${{ parameter.key }}: ${{ parameter.value }}
+          ${{ parameter.key }}: ${{ parameter.value }}ex
 
   # run coreclr crossgen perf job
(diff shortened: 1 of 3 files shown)


eng/pipelines/runtime-perf-jobs.yml
```diff
@@ -148,122 +134,123 @@ jobs:
...
 
+  # run coreclr perfviper microbenchmarks perf job
   - template: /eng/pipelines/common/platform-matrix.yml@${{ parameters.runtimeRepoAlias }}
     parameters:
       jobTemplate: /eng/pipelines/templates/runtime-perf-job.yml@${{ parameters.performanceRepoAlias }}
       buildConfig: release
       runtimeFlavor: coreclr
       platforms:
       - linux_x64
+      - windows_x64
```
DrewScoggins, line 248: On line 277 it seems like we already have a definition for this job. We probably only need one.

eng/pipelines/runtime-perf-jobs.yml
```diff
@@ -148,122 +134,104 @@ jobs:
...
         runtimeRepoAlias: ${{ parameters.runtimeRepoAlias }}
         performanceRepoAlias: ${{ parameters.performanceRepoAlias }}
         ${{ each parameter in parameters.jobParameters }}:
           ${{ parameter.key }}: ${{ parameter.value }}
 
+  # run mono aot microbenchmarks perf job
   - template: /eng/pipelines/common/platform-matrix.yml@${{ parameters.runtimeRepoAlias }}
     parameters:
-      jobTemplate: /eng/pipelines/templates/runtime-perf-job.yml@${{ parameters.performanceRepoAlias }}
+      jobTemplate: /eng/pipelines/templates/runtime-perf-job.yml@${{ parameters.performanceRepoAlias }} # NOTE: should we move this file out of coreclr tempelates because it contains mono jobs?
```
caaavik-msft, line 185: Can you remove this comment as well while you're here about the coreclr templates folder? This was something I had missed in my PR as it is no longer in that folder.

caaavik-msft (dismissed):

eng/pipelines/runtime-perf-jobs.yml
LoopedBard3: 👍

eng/pipelines/runtime-perf-jobs.yml
```diff
@@ -283,34 +264,15 @@ jobs:
...
-        liveLibrariesBuildConfig: Release
-        runKind: micro
-        logicalMachine: 'perfviper'
-        experimentName: 'jitoptrepeat'
-        additionalJobIdentifier: jitoptrepeat
-        runtimeRepoAlias: ${{ parameters.runtimeRepoAlias }}
-        performanceRepoAlias: ${{ parameters.performanceRepoAlias }}
-        ${{ each parameter in parameters.jobParameters }}:
-          ${{ parameter.key }}: ${{ parameter.value }}
+          ${{ parameter.key }}: ${{ parameter.value }}ex
```
DrewScoggins, line 275: Probably this ex should be removed?

LoopedBard3: Yes, removed. Thanks for the catch.

DrewScoggins (approved):

PR #4742, opened 2025-02-28
Fix FSharp and inner loop SDK scenarios.

LoopedBard3:
Fix FSharp and inner loop SDK scenarios. This consists of updating an accessibility reference in the FSharp test's source and updating the tests that explicitly take in a framework to use the PERFLAB_Framework as we no longer need to support netcoreapp and don't know if the framework has 3 or 4 characters for the version number. This also does stop running older targets using newer SDKs as the old packages are not available on with this setup.

### DIFF
--- a/eng/performance/sdk_scenarios.proj
+++ b/eng/performance/sdk_scenarios.proj
@@ -2,6 +2,18 @@
 
   <Import Project="Scenarios.Common.props" />
 
+  <!-- <ItemGroup>
+      <Framework Include="netcoreapp2.1" FrameworkName="%(Identity)"/>
+      <Framework Include="netcoreapp3.0" FrameworkName="%(Identity)" Condition="'$(FrameworkVersion)' &gt; '2.1'"/>
+      <Framework Include="netcoreapp3.1" FrameworkName="%(Identity)" Condition="'$(FrameworkVersion)' &gt; '3.0'"/>
+      <Framework Include="netcoreapp5.0" FrameworkName="%(Identity)" Condition="'$(FrameworkVersion)' &gt; '3.1'"/>
+      <Framework Include="net6.0" FrameworkName="%(Identity)" Condition="'$(FrameworkVersion)' &gt; '5.0'"/>
+      <Framework Include="net7.0" FrameworkName="%(Identity)" Condition="'$(FrameworkVersion)' &gt; '6.0'"/>
+      <Framework Include="net8.0" FrameworkName="%(Identity)" Condition="'$(FrameworkVersion)' &gt; '7.0'"/>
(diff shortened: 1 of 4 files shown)


LoopedBard3:
The only sdk_scenarios error is a flaky access failure: "PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'app'", the consistently broken sdk_scenarios are fixed.

eng/performance/sdk_scenarios.proj
```diff
@@ -22,12 +11,6 @@
   </ItemDefinitionGroup>
   
   <ItemGroup>
-    <SDKWorkItem Include="@(Framework -> 'SDK Console Template %(Identity)')" Exclude="*2.1;*3.0;*3.1;*5.0;*6.0;*7.0;*8.0">
```
e-kharion, line 25: Is this removal only temporary? If it is, we might want to keep a comment there referring to some task/dependency state so we do not forget to revert it.

LoopedBard3: My plan was to have it be permanent as it hits this error when targeting framework net8.0 with a net9.0 sdk: `Unable to find package Microsoft.NETCore.App.Ref with version (= 8.0.14)`, but commenting it out instead for easy reenablement probably makes more sense.

LoopedBard3: Found an alternative and pushed in the latest PR.

LoopedBard3:
The final sdk_scenario failure was a flaky failure due to the app directory being "in use": `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'app'`. This should be good to merge.

eng/performance/sdk_scenarios.proj
```diff
@@ -100,7 +102,8 @@
       <PreCommands>%(PreCommands) -f %(FrameworkName)</PreCommands>
     </SDKWorkItem> -->
 
-    <SDKWorkItem Include="@(Framework -> 'SDK Windows Forms Large %(Identity)')" Exclude="*2.1" Condition="'$(TargetsWindows)' == 'true' and ('$(FrameworkVersion)' == '3.0' or '$(FrameworkVersion)' == '3.1')">
+    <!-- We never run against Framework 3.0 or 3.1 -->
+    <!-- <SDKWorkItem Include="@(Framework -> 'SDK Windows Forms Large %(Identity)')" Exclude="*2.1" Condition="'$(TargetsWindows)' == 'true' and ('$(FrameworkVersion)' == '3.0' or '$(FrameworkVersion)' == '3.1')">
```
DrewScoggins, line 106: We should do it as another PR, but we might just want to move these tests forward to run against the current target frameworks.

DrewScoggins (approved):

PR #4750, opened 2025-03-06
new perf builds pipeline

e-kharion:
This PR adds new templates to be used by the new perf-builds pipeline and relates to the https://github.com/dotnet/runtime/pull/113221

### DIFF
--- a/eng/pipelines/register-build-jobs.yml
+++ b/eng/pipelines/register-build-jobs.yml
@@ -1,12 +1,12 @@
 parameters:
   runtimeRepoAlias: runtime
   performanceRepoAlias: self
-  buildConfiguration: []
+  buildType: []
   mauiFramework: ''
 
 jobs:
-- ${{ each configuration in parameters.buildConfiguration }}:
+- ${{ each type in parameters.buildType }}:
   - template: /eng/pipelines/templates/register-build-job.yml@${{ parameters.performanceRepoAlias }}
     parameters:
-      buildConfiguration: ${{ configuration }}
+      buildType: ${{ type }}
       buildId: $(Build.BuildId)
--- a/eng/pipelines/register-build-jobs.yml
+++ b/eng/pipelines/register-build-jobs.yml
@@ -1,0 +1,12 @@
+parameters:
+  runtimeRepoAlias: runtime
+  performanceRepoAlias: self
+  buildConfiguration: []
+  mauiFramework: ''
+
+jobs:
+- ${{ each configuration in parameters.buildConfiguration }}:
+  - template: /eng/pipelines/templates/register-build-job.yml@${{ parameters.performanceRepoAlias }}
+    parameters:
+      buildConfiguration: ${{ configuration }}
+      buildId: $(Build.BuildId)
--- a/eng/pipelines/runtime-perf-build-jobs.yml
+++ b/eng/pipelines/runtime-perf-build-jobs.yml
@@ -1,0 +1,84 @@
+parameters:
+  runtimeRepoAlias: runtime
+  performanceRepoAlias: self
+  buildConfiguration: []
+  mauiFramework: ''
+
+jobs:
+  - ${{ if containsValue(parameters.buildConfiguration, 'coreclr_arm64_linux') }}:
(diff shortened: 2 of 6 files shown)


eng/pipelines/upload-build-artifacts-jobs.yml
```diff
@@ -0,0 +1,167 @@
...
+  buildConfiguration: []
+  mauiFramework: ''
+
+jobs:
+  - ${{ if containsValue(parameters.buildConfiguration, 'coreclr_arm64_linux') }}:
+    - template: /eng/pipelines/templates/upload-build-artifacts-job.yml@${{ parameters.performanceRepoAlias }}
+      parameters:
+        buildConfiguration: 'coreclr_arm64_linux'
+        dependencyJobName: build_linux_arm64_release_coreclr
+        artifacts:
```
DrewScoggins, line 13: Is there any way that we can build these strings from the configuration, and not rely on hard coding the value?

eng/pipelines/templates/upload-build-artifacts-job.yml
```diff
@@ -0,0 +1,26 @@
...
+      - script: echo "pvscmdupload/builds/buildArtifacts/${{ variables['Build.SourceVersion'] }}/${{ parameters.buildConfiguration }}/${{ fileName }}"
+        displayName: 'Artifact upload path'
+      - task: AzureCLI@2
+        displayName: 'Upload ${{ fileName }} file'
+        inputs:
+          azureSubscription: '.NET Performance (790c4451-dad9-4fda-af8b-10bd9ca328fa)'
+          scriptType: 'pscore'
+          scriptLocation: 'inlineScript'
+          inlineScript: |
+            az storage blob upload --auth-mode login --account-name pvscmdupload --container-name builds --file "$(Pipeline.Workspace)/${{ artifact.artifactName }}/${{ fileName }}" --name "buildArtifacts/${{ variables['Build.SourceVersion'] }}/${{ parameters.buildConfiguration }}/${{ fileName }}" --overwrite true
```
DrewScoggins, line 26: Will this file ever want to be publicly available? If so we should move this into $web.

DrewScoggins: Same question holds for the build register output.

DrewScoggins (approved):
LGTM

eng/pipelines/templates/upload-build-artifacts-job.yml
e-kharion: That's a very good question and frankly speaking I could see pros and cons for both alternatives.
$web container - can be simply downloaded without any auth
new container - we can easily define quotas, policies, see statistics etc.

Since I expected we would be the only user accessing the bits and we would do so from pipeline/.net client, accessing the blob storage would be no issue, thus I ended up with going with the new container. But if anybody have a stronger preference on the $web container, we can definitely change it.

eng/pipelines/register-build-jobs.yml
```diff
@@ -0,0 +1,12 @@
...
+  runtimeRepoAlias: runtime
+  performanceRepoAlias: self
+  buildConfiguration: []
+  mauiFramework: ''
+
+jobs:
+- ${{ each configuration in parameters.buildConfiguration }}:
+  - template: /eng/pipelines/templates/register-build-job.yml@${{ parameters.performanceRepoAlias }}
+    parameters:
+      buildConfiguration: ${{ configuration }}
```
caaavik-msft, line 11: buildConfiguration is already a well known term that refers to whether it is compiled in Release or Debug so I think this would be confusing. Maybe `buildType` would be better?

eng/pipelines/runtime-perf-build-jobs.yml
```diff
@@ -0,0 +1,84 @@
+parameters:
+  runtimeRepoAlias: runtime
+  performanceRepoAlias: self
+  buildConfiguration: []
+  mauiFramework: ''
+
+jobs:
+  - ${{ if containsValue(parameters.buildConfiguration, 'coreclr_arm64_linux') }}:
```
caaavik-msft, line 8: This should be simplifiable to an `- ${{ each config in buildConfiguration }}:` where we check the `config` value inside the `parameters` section.

caaavik-msft (changes requested):

PR #4764, opened 2025-03-12
Add benchmarks for `ZipFile.CreateFromDirectory` and `ZipFile.ExtractToDirectory`

carlossanlop:
Also added a test zip file containing the files inside the existing TestData folder.

Will soon use this class for the async versions of these methods.


src/benchmarks/micro/libraries/System.IO.Compression/ZipFileTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        Directory.Delete(_testOutputDirectory, recursive: true);
+    }
+
+    [Benchmark]
+    public void ZipFileCreateFromDirectory()
+    {
+        ZipFile.CreateFromDirectory(_testInputDirectory, _testOutputZipFile);
+    }
+
+    public void ZipFileExtractToDirectory()
```
adamsitnik, line 38: Was that supposed to be a benchmark too?

```suggestion
    [Benchmark]
    public void ZipFileExtractToDirectory()
```

src/benchmarks/micro/libraries/System.IO.Compression/ZipFileTests.cs
```diff
@@ -0,0 +1,42 @@
...
+public class ZipFileTests
+{
+    private static readonly string _testOutputDirectory = FileUtils.GetTestFilePath();
+    private static readonly string _testOutputZipFile = Path.Combine(_testOutputDirectory, "TestOutputZipFile.zip");
+    private static readonly string _testInputDirectory = Path.Combine(
+        AppContext.BaseDirectory, "libraries", "System.IO.Compression", "TestData");
+    private static readonly string _testInputZipFile = Path.Combine(
+        AppContext.BaseDirectory, "libraries", "System.IO.Compression", "TestInputZipFile.zip");
+
+    [IterationSetup(Target = nameof(ZipFileCreateFromDirectory))]
```
adamsitnik, line 20: How long does it take to run this benchmark? We usually avoid `[IterationSetup]` for the reasons described here:

https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#iterationsetup

Does `ZipFile.CreateFromDirectory` overwrite the file?

src/benchmarks/micro/MicroBenchmarks.csproj
```diff
@@ -118,6 +118,9 @@
     <None Update="libraries\System.IO.Compression\TestData\**">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="libraries\System.IO.Compression\TestZipFile.zip">
```
adamsitnik, line 121: I was thinking about suggesting to just zip `TestData` during `[GlobalSetup]` but then I've realized that if somebody adds new file to `TestData`, the time reported by the benchmark is going to report a regression.

So please keep the zip file. 👍

adamsitnik (commented):
@carlossanlop thank you for providing more benchmarks! PTAL at my comments.

Issue #4754, opened 2025-03-07
[Android] Setup `dotnet new android` performance runs


matouskozak:
The goal is to setup a continuous lab measurement of startup and size for `dotnet new android` sample app. Requirements are to support Mono and CoreCLR (when available) runtimes. Additionally, to have a tight dev loop, we need appropriate cadence of runtime (SDK) and Android workload builds.

# Proposal
Using [VMR SDK builds](https://github.com/dotnet/dotnet/blob/main/docs/builds-table.md) (±daily cadence) and nightly Android workload from dotnet10 feed (https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet10/nuget/v3/index.json).

## Stage 1 (daily SDK + latest Android workload release)
In the first stage we will setup a measurement for daily runtime builds obtained via VMR builds (https://github.com/dotnet/dotnet/blob/main/docs/builds-table.md) together with the latest released Android workload (not incremental builds). This can be achieved by the following steps.

1. Download dotnet-install script (https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script)
2. Install the latest VMR SDK build: `./dotnet-install.sh --channel "10.0.1xx-ub" --quality daily --install-dir sdk`
3. Create a nuget.config with dotnet10 feed
    a. `/sdk/dotnet new nugetconfig`
    b. Add: `<add key="dotnet10" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet10/nuget/v3/index.json" />`
    c. Add "dotnet-android" feeds from https://github.com/dotnet/android/blob/main/NuGet.config
4. Install the Android workload: `./sdk/dotnet workload install android`
5. Create the Android sample app: `./sdk/dotnet new android`

The output of these steps as of 2025/03/07 is:
- Runtime version: `10.0.0-preview.3.25155.15` (commit: `666e9188f5e11f4a59708b5f18ff611aa950d965`)
- Android workload version: `35.99.0-preview.1.140/10.0.100-preview.1` (commit: `6ec531d9418573d56a66192a59411f548f17844d`)

## Stage 2 (daily SDK + nightly Android workload builds)
In the second stage we improve the setup by adding nightly Android workload builds. This can be achieved by the following steps:

### Option 1
1. Repeat steps 1.-3. from Stage 1
2. Setup workload to grap the latest manifests: `./sdk/dotnet workload config --update-mode manifests`
3. Install the Android workload: `./sdk/dotnet workload install android`
[...]

PR #4770, opened 2025-03-13
Setup new dotnet new android measurements using Mono or CoreCLR runtimes

matouskozak:
Switching MAUI-Android (and .NET Android) scenarios to use latest .NET SDK + dotnet10 feed, the steps are:
1. Install the latest .NET SDK: `.dotnet-install.sh --channel 10.0 --quality daily`
2. Using the latest feed ([dotnet10](https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet10/nuget/v3/index.json)), find the latest packs using `dotnet package search`.
3. Put the latest packs to a rollback file and install maui workload.

Adding new jobs:
- `dotnet new android` running CoreCLR.
- `dotnet new maui` (Android) running CoreCLR.
- `dotnet new maui-blazor` (Android) running CoreCLR.

Internal pipeline run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2684265

TODO:
- [x] Remove added android feeds from `Nuget.config` (should be added by maestro)

Out of scope
- Fix flaky memory consumption measurements for .NET Android and .NET MAUI Android https://github.com/dotnet/performance/issues/4826

---

https://github.com/dotnet/performance/issues/4754


src/scenarios/shared/precommands.py
```diff
@@ -288,6 +288,37 @@ def uninstall_workload(self, workloadid: str):
         if self.has_workload and not self.readonly_dotnet:
             subprocess.run(["dotnet", "workload", "uninstall", workloadid])
 
+    def create_nuget_config(self, feeds: dict[str, str]={"nuget.org": "https://api.nuget.org/v3/index.json"}):
```
LoopedBard3, line 291: We will likely get workitems to pop up if we use nuget.org, would it instead be possible to instead use the one at the top level of the performance repo? If there are some packages that only occasionally update but they are not being found in our feeds, we can have them added.

Also, this does not appear to currently be called anywhere.

scripts/run_performance_job.py
```diff
@@ -541,6 +541,17 @@ def run_performance_job(args: RunPerformanceJobArgs):
         configurations["CompilationMode"] = "JIT"
         configurations["RuntimeType"] = str(runtime_type)
 
+    if args.run_kind == "maui_scenarios_android":
+        if args.runtime_flavor == "mono":
+            runtime_type = "Mono"
+            configurations["CompilationMode"] = "ProfiledAOT"
```
LoopedBard3, line 547: We should make sure other similar android app testing matches this for Mono.

src/scenarios/shared/precommands.py
matouskozak: removed

scripts/run_performance_job.py
matouskozak: I added a comment explaining which scenario is which. We run different configuration for dotnet/runtime android sample than the default configuration used in .NET Android / .NET MAUI Android samples.

caaavik-msft (dismissed):
Looks good to me. One question I have, are any of these changes (e.g. the _MSBuildArgs) being made because they are coupled to the runtime repository itself or would you want them set even if we decided to run against an older version of the runtime? 

As part of the recent changes we made to move the yaml from the runtime to the performance repository, I'd like to know if any of the things we moved over were better living in the runtime repository. If so, we can look at restructuring where these live to better couple related changes.

e-kharion (approved):

eng/performance/maui_scenarios_android.proj
```diff
@@ -11,6 +11,10 @@
     <PreparePayloadOutDirectoryName>scenarios_out</PreparePayloadOutDirectoryName>
     <PreparePayloadWorkItemBaseDirectory Condition="'$(TargetsWindows)' == 'true'">$(CorrelationPayloadDirectory)$(PreparePayloadOutDirectoryName)\</PreparePayloadWorkItemBaseDirectory>
     <PreparePayloadWorkItemBaseDirectory Condition="'$(TargetsWindows)' != 'true'">$(CorrelationPayloadDirectory)$(PreparePayloadOutDirectoryName)/</PreparePayloadWorkItemBaseDirectory>
+
+    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'mono'">/p:UseMonoRuntime=true</_MSBuildArgs>
+    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'coreclr'">/p:UseMonoRuntime=false;/p:AndroidEnableMarshalMethods=false</_MSBuildArgs>
```
ivanpovazan, line 16:
```suggestion
    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'coreclr' and '$(PublishAot)' != 'true'">/p:UseMonoRuntime=false</_MSBuildArgs>
```
`AndroidEnableMarshalMethods ` is not needed anymore as of: https://github.com/dotnet/android/pull/9912

eng/performance/maui_scenarios_android.proj
```diff
@@ -25,19 +29,19 @@
...
 
   <ItemGroup>
-    <MAUIAndroidScenario Include="Maui Android Default">
-      <ScenarioDirectoryName>mauiandroid</ScenarioDirectoryName>
-      <PayloadDirectory>$(ScenariosDir)%(ScenarioDirectoryName)</PayloadDirectory>
-      <ApkName>com.companyname.mauiandroiddefault-Signed</ApkName>
-      <PackageName>com.companyname.mauiandroiddefault</PackageName>
-    </MAUIAndroidScenario>
-    <MAUIAndroidScenario Include=".NET Android Default">
+    <MAUIAndroidScenario Include=".NET Android Default $(RunConfigsString)">
```
ivanpovazan, line 32: I think we should revisit the naming here and/or introduce new cases.
The `Default` is different between Mono, CoreCLR and NativeAOT, Mono's defaults to profiled AOT, while CoreCLR defaults to JIT only *at the moment* etc, which can lead to confusion when analysing the measurements.
I suggest we add new more descriptive scenarios (JIT, AOT, ..) conditioned on the runtime type.
Lastly, marking which one of the scenarios is the default one is still valueable.

eng/performance/maui_scenarios_android.proj
```diff
@@ -25,19 +29,19 @@
...
       <PackageName>com.companyname.NetAndroidDefault</PackageName>
     </MAUIAndroidScenario>
-    <MAUIAndroidScenario Include="Maui Blazor Android Default">
+    <MAUIAndroidScenario Include="Maui Android Default $(RunConfigsString)" Condition="'$(RuntimeFlavor)' == 'mono'">
+      <ScenarioDirectoryName>mauiandroid</ScenarioDirectoryName>
+      <PayloadDirectory>$(ScenariosDir)%(ScenarioDirectoryName)</PayloadDirectory>
+      <ApkName>com.companyname.mauiandroiddefault-Signed</ApkName>
+      <PackageName>com.companyname.mauiandroiddefault</PackageName>
+    </MAUIAndroidScenario>
+    <MAUIAndroidScenario Include="Maui Blazor Android Default (TTFD) $(RunConfigsString)" Condition="'$(RuntimeFlavor)' == 'mono'">
```
ivanpovazan, line 44: Why do we only care about TTFD measurement here? Or why don't we measure both TTID and TTFD?

scripts/run_performance_job.py
```diff
@@ -531,16 +531,30 @@ def run_performance_job(args: RunPerformanceJobArgs):
...
     if android_coreclr:
         runtime_type = "CoreCLR"
         configurations["CompilationMode"] = "JIT"
         configurations["RuntimeType"] = str(runtime_type)
 
+    # .NET Android and .NET MAUI Android sample app scenarios
+    if args.run_kind == "maui_scenarios_android":
+        if args.runtime_flavor == "mono":
+            runtime_type = "Mono"
+            configurations["CompilationMode"] = "ProfiledAOT"
```
ivanpovazan, line 550: Can we use `CompilationMode` in the `MAUIAndroidScenario` naming above?

src/scenarios/mauiandroid/pre.py
```diff
@@ -11,9 +11,14 @@
 from test import EXENAME
 
 setup_loggers(True)
+logger = getLogger(__name__)
+logger.info("Starting pre-command for Maui Android Default App")
 
 precommands = PreCommands()
-install_versioned_maui(precommands)
+
+precommands.setup_workload_update_mode("manifests")
```
ivanpovazan, line 19: Should this be the default behavior? 
My understanding is that we have to set the update mode only if we want to use the nightly builds

eng/performance/maui_scenarios_android.proj
matouskozak: I don't know. I just saw that it is passing the `--use-fully-drawn-time` so I added it to the scenario name to make it clear. I can probably add TTID as well if this is something we want.

src/scenarios/mauiandroid/pre.py
matouskozak: > My understanding is that we have to set the update mode only if we want to use the nightly builds

This is true. 

> Should this be the default behavior?

The change is intended to be the new default, and I wasn't planning for use to have 2 separate pipelines, one of released and another of nightly builds. These measurements should serve as a tool for internal monitoring and not public benchmarks. Do you see benefits in having both scenarios (released/nightly) running in the lab?

ivanpovazan: If we checkout the repo and want to run the scenario locally with a specific release how would we do that?

matouskozak: I see, I have not considered this as a user scenario when designing this change. 

We can specify local dotnet install by `. ./init.sh -dotnetdir <custom dotnet directory>` so I think it could be enough to pass an argument to the `pre.py` command to skip installing the additional workloads and rely on the user that the `<custom dotnet directory>` contains the specific release installed. Would that be sufficient for the use case you have in mind?

matouskozak: Update on that, we actually already support this scenario. If you use:
1. `. ./init.sh -dotnetdir <custom dotnet directory>`
2. `pre.py <other_args> --readonly-dotnet --has-workload`

This should ensure that the test app is build using the locally installed dotnet that has the specific configuration of SDKs.

ivanpovazan:
@matouskozak 

There might be some issues with the approach of `dotnet workload config --update-mode manifests` especially around servicing releases, because the manifests will include a reference to a new servicing release (like atm `9.0.4`) which is not publicly available so the workload installation would fail with:
```
Workload installation failed: One or more errors occurred. (Version 9.0.4 of package microsoft.net.runtime.monoaotcompiler.task is not found in NuGet feeds ...
```
Will experiment a bit with this but I think we might need to the the `Option 2` from your proposal in: https://github.com/dotnet/performance/issues/4754

eng/performance/maui_scenarios_android.proj
matouskozak: @LoopedBard3 do you remember why for blazor we are using `--use-fully-drawn-time` and not the default startup measurement config?

eng/performance/maui_scenarios_android.proj
matouskozak: Added

matouskozak:
Refactored the PR to address some issues and changed the maui installation to use rollback file instead. The PR contains changes to Nuget.config which should be done maestro so should be reverted before merging.

eng/performance/maui_scenarios_android.proj
matouskozak: Actually, I'm thinking about removing the `$(RunConfigsString)` all together from the scenario name as per 
https://github.com/dotnet/runtime/pull/114117#discussion_r2023567496 and offline discussion with @LoopedBard3 . 

We can already see the configuration used in the job name 
image
+ everything we place in `configurations` https://github.com/dotnet/performance/blob/2659b9d72210c725f20e7a10499ff88d40bcc29f/scripts/run_performance_job.py#L550-L551 is going to flow to database so we can retrieve it for visualizations...

eng/performance/maui_scenarios_android.proj
LoopedBard3: I don't recall anything specifically. Maybe TTID doesn't include some of the blazor startup. Regardless, I don't see any reason not to include both.

eng/performance/maui_scenarios_android.proj
```diff
@@ -11,6 +11,10 @@
     <PreparePayloadOutDirectoryName>scenarios_out</PreparePayloadOutDirectoryName>
     <PreparePayloadWorkItemBaseDirectory Condition="'$(TargetsWindows)' == 'true'">$(CorrelationPayloadDirectory)$(PreparePayloadOutDirectoryName)\</PreparePayloadWorkItemBaseDirectory>
     <PreparePayloadWorkItemBaseDirectory Condition="'$(TargetsWindows)' != 'true'">$(CorrelationPayloadDirectory)$(PreparePayloadOutDirectoryName)/</PreparePayloadWorkItemBaseDirectory>
+
+    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'mono'">/p:UseMonoRuntime=true</_MSBuildArgs>
+    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'coreclr'">/p:UseMonoRuntime=false;</_MSBuildArgs>
```
ivanpovazan, line 16:
```suggestion
    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'coreclr'">/p:UseMonoRuntime=false</_MSBuildArgs>
```

Do we need to concat more than this property?

scripts/run_performance_job.py
```diff
@@ -531,16 +531,25 @@ def run_performance_job(args: RunPerformanceJobArgs):
...
     if android_coreclr:
         runtime_type = "CoreCLR"
         configurations["CompilationMode"] = "JIT"
         configurations["RuntimeType"] = str(runtime_type)
 
+    # .NET Android and .NET MAUI Android sample app scenarios
+    if args.run_kind == "maui_scenarios_android":
+        if not args.runtime_flavor in ("mono", "coreclr"):
+            raise Exception("Runtime flavor must be specified for maui_scenarios_android")
+        configurations["CodegenType"] = str(args.codegen_type)
```
ivanpovazan, line 550: Just for my understanding how is `CodegenType` different from `CompilationMode` used above?

src/scenarios/shared/mauisharedpython.py
```diff
@@ -93,4 +94,80 @@ def install_versioned_maui(precommands: PreCommands):
         dump_dict_to_json_file(rollback_dict, f"rollback_{target_framework_wo_platform}.json")
         workload_install_args += ['--from-rollback-file', f'rollback_{target_framework_wo_platform}.json']
 
-    precommands.install_workload('maui', workload_install_args) 
+    precommands.install_workload('maui', workload_install_args)
+
+# TODO: the feed should be updated to the latest
```
ivanpovazan, line 99: Once we have darc setup to update NuGet feeds, this won't be needed anymore, correct?

ivanpovazan (approved):
LGTM, only left few minor notes.
Thank you!

scripts/run_performance_job.py
ivanpovazan: Ok, I was referring to lines:
https://github.com/dotnet/performance/pull/4770/files#diff-75e8f68d6caa7704e8c72407bcfc108e368c9b3ea7e71ca1532033c8905bad46R543
https://github.com/dotnet/performance/pull/4770/files#diff-75e8f68d6caa7704e8c72407bcfc108e368c9b3ea7e71ca1532033c8905bad46R537

matouskozak: I think these are old version, `CompilationMode` is no longer being used there.

src/scenarios/shared/mauisharedpython.py
matouskozak: I wrote a new logic which extracts the latest feed from the repo's Nuget.config so we don't need this anymore.

src/scenarios/shared/versionmanager.py
```diff
@@ -34,4 +34,63 @@ def get_version_from_dll_powershell(dll_path: str):
...
+        print(f"Getting version from {dll_path}")
+        result = powershell_cmd(dll_path)
+        
+        if "Cannot find path" in result:
+            getLogger().warning(f"Cannot find {dll_name} in {dll_folder_path}. Skipping version extraction.")
+            continue
+
+        version, commit = parse_version_output(result)
+        results[f"{sdk}_version"] = version
+        results[f"{sdk}_commit_hash"] = commit
```
LoopedBard3, line 94: Rather than adding a new string to look for in the Reporter.cs, can this instead be `PERFLAB_DATA_{sdk}_commit_hash`? The PERFLAB_DATA_ is not included in the saved additional data so it should show up the same, but then we don't need to add the additional search string `_commit_hash`.

LoopedBard3 (commented):
This looks really great! I am excited to have what seems to be a more robust solution for building the apps. I left one comment but am also wondering if this supports 9.0 and 8.0 mobile app runs. It seems that it currently sets up the runs to only run main.

matouskozak:
>  I left one comment but am also wondering if this supports 9.0 and 8.0 mobile app runs.

Yes, currently works just on main. I think it could be extended to other branches. We would just search a different dotnet feed I think and make it more parametrized. However, I don't see much value now in having measurements on other branches but we could extend it in the future if needed.

LoopedBard3 (approved):

ivanpovazan (approved):
This looks neat, thank you!

The only thoughts left (not for this PR) are:
1. It seems like the rollback file could be provided by some other service rather than us manually creating it as part of performance runs 
2. Think about a way of doing performance runs on specific releases, versions i.e., answering a question: what is the performance of our preview.X release - and with that freeze/tag such measurements for reference

Other than that, really nice work!

caaavik-msft (dismissed):
LGTM, just the one minor comment on the yaml condition

LoopedBard3 (approved):

PR #4776, opened 2025-03-18
Add Android Startup Perfetto trace support

LoopedBard3:
Add support for getting perfetto traces for the android runs. This will start out as a primarily manual thing to help users of our scenario testing when doing manual runs, or manually setup pipeline runs.

### DIFF
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -581,7 +581,7 @@ ex: C:\repos\performance;C:\repos\runtime
                     time.sleep(self.closeToStartDelay) # Delay in seconds for ensuring a cold start
 
                 if self.traceperfetto:
-                    perfetto_device_save_file = f'/data/misc/perfetto-traces/perfetto_trace_{time.time()}' # Try a different trace location from #f'/sdcard/Android/data/{self.packagename}/files/perfetto_startup_trace'
+                    perfetto_device_save_file = f'/data/misc/perfetto-traces/perfetto_trace_{time.time()}'
 
                     # Setup the phone props to allow perfetto to run properly
                     getLogger().info("Setting up the device for Perfetto")
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -583,8 +583,17 @@ ex: C:\repos\performance;C:\repos\runtime
                 if self.traceperfetto:
                     perfetto_device_save_file = f'/data/misc/perfetto-traces/perfetto_trace_{time.time()}'
                     original_traced_enable = None
+                    perfetto_terminated = False
+
+                    stop_perfetto_cmd = xharness_adb() + [ # Stop perfetto now that the app. Sending a Terminate signal should be enough per the longer trace capturing guidance here: https://perfetto.dev/docs/concepts/config#android.
+                        'shell',
+                        'pkill -TERM perfetto'
+                    ]
 
                     try:
+                        getLogger().info("Clearing potential previous running perfetto traces")
+                        RunCommand(stop_perfetto_cmd, verbose=True).run()
+
                         # Get the current value of persist.traced.enable
                         getLogger().info("Getting current persist.traced.enable value")
                         get_traced_cmd = xharness_adb() + [
@@ -622,13 +631,9 @@ ex: C:\repos\performance;C:\repos\runtime
                         traced_start = RunCommand(androidHelper.startappcommand, verbose=True)
                         traced_start.run()
 
-                        # Stop perfetto now that the app has started. Sending a Terminate signal should be enough per the longer trace capturing guidance here: https://perfetto.dev/docs/concepts/config#android.
                         getLogger().info("Stopping perfetto trace capture")
-                        stop_perfetto_cmd = xharness_adb() + [
-                            'shell',
-                            'pkill -TERM perfetto'
-                        ]
                         RunCommand(stop_perfetto_cmd, verbose=True).run()
+                        perfetto_terminated = True
 
                         # Pull the trace from the device and store in the traceperfetto directory
                         pull_trace_cmd = xharness_adb() + [
@@ -647,6 +652,11 @@ ex: C:\repos\performance;C:\repos\runtime
                         RunCommand(delete_trace_cmd, verbose=True).run()
 
                     finally:
+                        if not perfetto_terminated:
+                            # If we didn't terminate perfetto, we need to stop it now, although it will time out as well if this manages to fail.
+                            getLogger().info("Perfetto not terminated, stopping it now")
+                            RunCommand(stop_perfetto_cmd, verbose=True).run()
+
                         # Restore original persist.traced.enable value if we saved one
                         if original_traced_enable is not None and original_traced_enable != "1":
                             getLogger().info(f"Restoring persist.traced.enable to original value: {original_traced_enable}")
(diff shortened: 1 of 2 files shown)


LoopedBard3:
Tested successfully locally and initial use is just in manual runs.

src/scenarios/shared/runner.py
```diff
@@ -577,7 +579,81 @@ def run(self):
                         raise Exception("Android Time Capture Failed! Incorrect number of captures found.")
                     allResults.append(formattedTime) # append TotalTime: (TIME)
                     time.sleep(self.closeToStartDelay) # Delay in seconds for ensuring a cold start
-                
+
+                if self.traceperfetto:
+                    perfetto_device_save_file = f'/data/misc/perfetto-traces/perfetto_trace_{time.time()}'
```
e-kharion, line 584: nit: Would it make sense to use formatted time instead of a number here? Would somebody be checking the files manually?

src/scenarios/shared/runner.py
```diff
@@ -577,7 +579,81 @@ def run(self):
...
+                        getLogger().info("Ensuring the trace capture has been completed.")
+                        while time.time() - perfetto_start_time_sec < max_startup_time_sec + 3:
+                            time.sleep(1)
+                            getLogger().info("Waiting for trace capture to complete. Time elapsed (Sec): %s, Max time (sec): %s", (time.time() - perfetto_start_time_sec), (max_startup_time_sec + 3))
+
+                        # Pull the trace from the device and store in the traceperfetto directory
+                        pull_trace_cmd = xharness_adb() + [
+                            'pull',
+                            perfetto_device_save_file,
+                            os.path.join(os.getcwd(), const.TRACEDIR, f'perfetto_startup_trace_{self.packagename}_{time.time()}.trace')
```
e-kharion, line 635: nit: Would it make sense to use formatted time instead of a number here? Would somebody be checking the files manually?

src/scenarios/shared/runner.py
```diff
@@ -577,7 +579,81 @@ def run(self):
...
+                        ]
+                        RunCommand(setup_perfetto_cmd, verbose=True).run()
+
+                        getLogger().info("Tracing with Perfetto")
+                        # Get the max TotalTime from the allResults list in seconds
+                        max_startup_time_sec = int(max(int(re.search(r"TotalTime: (\d+)", str(result)).group(1)) for result in allResults) / 1000)
+
+                        perfetto_cmd = xharness_adb() + [
+                            'shell',
+                            f'perfetto --background --txt -o {perfetto_device_save_file} --time {max_startup_time_sec + 3}s -b 64mb sched freq idle am wm gfx view binder_driver hal dalvik camera input res memory'
```
e-kharion, line 612: where the +3 comes from?

I have no idea about the usual "magnitude" of the max_startup_time_sec - but if it's just some kind of reserve, it might be better to add some percentage to the original value (in case the value might vary significantly)?

e-kharion (dismissed):
LGTM in general, left few comments/questions

src/scenarios/shared/runner.py
```diff
@@ -577,7 +579,81 @@ def run(self):
...
                     time.sleep(self.closeToStartDelay) # Delay in seconds for ensuring a cold start
-                
+
+                if self.traceperfetto:
+                    perfetto_device_save_file = f'/data/misc/perfetto-traces/perfetto_trace_{time.time()}'
+                    original_traced_enable = None
+
+                    try:
+                        # Get the current value of persist.traced.enable
+                        getLogger().info("Getting current persist.traced.enable value")
```
e-kharion, line 589: super nit: I know you follow the pattern here, but I feel that decomposing the code a bit would improve the readability/maintainability significantly. For example separating all the commands into some helper commands library class and only call those here would make the code much nicer in my opinion.

src/scenarios/shared/runner.py
matouskozak: > Would somebody be checking the files manually?

The idea is that users would upload the captured trace to https://ui.perfetto.dev and analyzed the results there.

e-kharion: Thanks. Does the filename format matter anyhow for the tool? If not (meaning the filename would not get processed by a machine) than some more human friendly time format make more sense to me. Still a nit, though.

src/scenarios/shared/runner.py
matouskozak: We generally want to run the tracing for the startup duration, that is what is calculated in:
`max_startup_time_sec = int(max(int(re.search(r"TotalTime: (\d+)", str(result)).group(1)) for result in allResults) / 1000)`.

+3s is probably just to introduce some slack if the startup would take unexpectedly longer.

I would maybe add the `max_startup_time_sec + 3` as a separate variable value with the comment from below why we use + 3 to make it a bit clearer.

> I have no idea about the usual "magnitude" of the max_startup_time_sec

In scenarios we are measuring the startup ranges from hundreds of milliseconds to low seconds.  Making it, for example, 1.5x of the max startup should be good enough as well.

matouskozak (dismissed):
Thanks a lot Parker!

src/scenarios/shared/runner.py
LoopedBard3: I agree with this assessment, so I made a task here: https://github.com/dotnet/performance/issues/4784.

src/scenarios/shared/runner.py
```diff
@@ -577,7 +579,81 @@ def run(self):
...
+
+                        # Run the startup test with the trace running (only once)
+                        getLogger().info("Running startup test with Perfetto trace running")
+                        traced_start = RunCommand(androidHelper.startappcommand, verbose=True)
+                        traced_start.run()
+
+                        # Wait until the total time taken to start the app is greater than the max startup time + 3
+                        # This is to ensure that the trace has captured the entire startup process
+                        getLogger().info("Ensuring the trace capture has been completed.")
+                        while time.time() - perfetto_start_time_sec < max_startup_time_sec + 3:
```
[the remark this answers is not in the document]
LoopedBard3: Completely moved away from using the seconds and instead switched to killing the tracing after we finish the startup command. This background start and then kill seems to be the recommended approach after further digging in the docs: https://perfetto.dev/docs/concepts/config#android.

src/scenarios/shared/runner.py
LoopedBard3: Completely moved away from using the seconds and instead switched to killing the tracing after we finish the startup command. This background start and then kill seems to be the recommended approach after further digging in the docs: https://perfetto.dev/docs/concepts/config#android. Still kept a max time for the tracing at 2x the max found startup, up to 60 seconds. At 60 seconds it throws an error as an arbitrary but, I think reasonable, cut off.

src/scenarios/shared/runner.py
```diff
@@ -573,11 +575,87 @@ def run(self):
...
+                        RunCommand(perfetto_cmd, verbose=True).run()
+
+                        # Run the startup test with the trace running (only once)
+                        getLogger().info("Running startup test with Perfetto trace running")
+                        traced_start = RunCommand(androidHelper.startappcommand, verbose=True)
+                        traced_start.run()
+
+                        # Stop perfetto now that the app has started. Sending a Terminate signal should be enough per the longer trace capturing guidance here: https://perfetto.dev/docs/concepts/config#android.
+                        getLogger().info("Stopping perfetto trace capture")
+                        stop_perfetto_cmd = xharness_adb() + [
```
caaavik-msft, line 627: Should the perfetto killall also be in the finally block?

caaavik-msft: Additionally, it might make sense to run a precautionary perfetto killall before running the kill command at the top. I know a couple times we have been bitten by processes not being killed.

caaavik-msft (changes requested):

src/scenarios/shared/runner.py
LoopedBard3: Having a kill before the execution is a great idea. I have updated the code per these comments. I included a perfetto kill in the finally block, though am not super worried about it as the time parameter in the call starting the trace should ensure it ends eventually.

matouskozak (approved):

caaavik-msft (approved):

Issue #4779, opened 2025-03-19
[android] Investigate reliance on `ActivityManager` vs `ActivityTaskManager` in adb logs


ivanpovazan:
## Description

When parsing adb logs for Android startup measurements: https://github.com/dotnet/performance/blob/9a4c4192117366d1d8fa14ba2bedc055a0fa5151/src/scenarios/shared/runner.py#L536
we expect that the TTID event gets reported by: `ActivityTaskManager`

However, official docs suggests that TTID gets reported by `ActivityManager`: https://developer.android.com/topic/performance/vitals/launch-time#time-initial

The issue does not seem to happen on "newer" Androids ie:
- Pixel7  pro Android 15 generates: 
```
03-19 15:10:25.296  1547  1658 I ActivityTaskManager: Displayed com.companyname.dotnetnewmauisamplecontent/crc646280ffa8b0b9e474.MainActivity for user 0: +1s435ms
```  
- Oneplus3T Android 9 generates:
```
03-19 09:52:20.543  1315  1424 I ActivityManager: Displayed com.companyname.dotnet_new_android/crc6456e79b14b408fe9d.MainActivity: +478ms
```

## Proposal

Support both variants, ie add a fallback option to cover both variants.

ivanpovazan:
@LoopedBard3  @matouskozak

matouskozak:
We could possibly change the logcat grep https://github.com/dotnet/performance/blob/eae9f655898a77d96e485fa974fdf7fe53110ec3/src/scenarios/shared/runner.py#L541 to  `f"logcat -d | grep -E 'ActivityManager|ActivityTaskManager: Displayed {androidHelper.activityname}'"` 

+ the same for TTFD I suppose

LoopedBard3:
Do we happen to know or have a solid grasp on whether this should be only one or the other and that both won't show up in the same log? Based on my understanding, that seems to be the case as it will be dependent on some version of Android or some other system version.

ivanpovazan:
> Do we happen to know or have a solid grasp on whether this should be only one or the other and that both won't show up in the same log? Based on my understanding, that seems to be the case as it will be dependent on some version of Android or some other system version.

I am not aware of any potential problems regarding this. I think if we support both variants it would be "good-enough" for now.

PR #4780, opened 2025-03-19
[android] Accept both `ActivityManager` and `ActivityTaskManager` for reporting TTID and TTFD

ivanpovazan:
## Description

As the title says, this PR changes how we parse logcat for TTID and TTFD events, accepting both `ActivityManager` and `ActivityTaskManager` as sources.

Fixes: https://github.com/dotnet/performance/issues/4779

### DIFF
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -533,12 +533,12 @@ ex: C:\repos\performance;C:\repos\runtime
                 # Create the fullydrawn command
                 fullyDrawnRetrieveCmd = xharness_adb() + [ 
                     'shell',
-                    f"logcat -d | grep 'ActivityTaskManager: Fully drawn {self.packagename}'"
+                    f"logcat -d | grep -E 'ActivityManager|ActivityTaskManager: Fully drawn {self.packagename}'"
                 ]
 
                 basicStartupRetrieveCmd = xharness_adb() + [ 
                     'shell',
-                    f"logcat -d | grep 'ActivityTaskManager: Displayed {androidHelper.activityname}'"
+                    f"logcat -d | grep -E 'ActivityManager|ActivityTaskManager: Displayed {androidHelper.activityname}'"
                 ]
 
                 clearLogsCmd = xharness_adb() + [


matouskozak (approved):
Happy that the fix worked.

src/scenarios/shared/runner.py
```diff
@@ -533,12 +533,12 @@ def run(self):
...
                 fullyDrawnRetrieveCmd = xharness_adb() + [ 
                     'shell',
-                    f"logcat -d | grep 'ActivityTaskManager: Fully drawn {self.packagename}'"
+                    f"logcat -d | grep -E 'ActivityManager|ActivityTaskManager: Fully drawn {self.packagename}'"
                 ]
 
                 basicStartupRetrieveCmd = xharness_adb() + [ 
                     'shell',
-                    f"logcat -d | grep 'ActivityTaskManager: Displayed {androidHelper.activityname}'"
+                    f"logcat -d | grep -E 'ActivityManager|ActivityTaskManager: Displayed {androidHelper.activityname}'"
```
matouskozak, line 541:
```suggestion
                    f"logcat -d | grep -E 'ActivityManager|ActivityTaskManager' | grep ': Displayed {androidHelper.activityname}'"
```

I realized it might be safer to do the regex grep for "ActivityManager" or "ActivityTaskManager" separately from the `androidHelper.activityname` as it might happen that some character inside the activity name is a special regex char.

matouskozak (changes requested):

src/scenarios/shared/runner.py
LoopedBard3: I think this second update looks good 👍.

LoopedBard3 (approved):

ivanpovazan:
I haven't had a chance to retest this with my old phone. 
Do we have some unit tests which we could use for cases like this i.e., take a static logcat output and run it through the parser?

src/scenarios/shared/runner.py
```diff
@@ -533,12 +533,12 @@ def run(self):
                 # Create the fullydrawn command
                 fullyDrawnRetrieveCmd = xharness_adb() + [ 
                     'shell',
-                    f"logcat -d | grep 'ActivityTaskManager: Fully drawn {self.packagename}'"
+                    f"logcat -d | grep -E 'ActivityManager|ActivityTaskManager: Fully drawn {self.packagename}'"
```
matouskozak, line 536:
```suggestion
                    f"logcat -d | grep -E 'ActivityManager|ActivityTaskManager' | grep ': Fully drawn {self.packagename}'"
```

matouskozak (dismissed):

src/scenarios/shared/runner.py
LoopedBard3: Sorry, missed that spot.

matouskozak (approved):

e-kharion (approved):

PR #4795, opened 2025-03-25
Ensure binlogs are copied to helix root and don't upload combined report to perf lab

caaavik-msft:
It seems that the --artifacts argument we pass in when running the benchmarks does not affect where the binlogs and other build artifacts are stored. Rather, it is always being put in the performance/artifacts directory. I'm not sure what the best solution is right now for what folder should be the true artifacts folder or which artifacts go where, but this PR just keeps the current situation as-is and ensures that we copy all binlogs to the helix upload root so we can use it to debug issues.

I also fixed a bug here where we are uploading the combined report json to the perf lab, and this was causing every performance measurement to be doubled in Kusto. This file should only be stored in the helix upload root.

### DIFF
--- a/scripts/benchmarks_ci.py
+++ b/scripts/benchmarks_ci.py
@@ -30,7 +30,7 @@ import shutil
 import sys
 from typing import Any, List, Optional
 
-from performance.common import validate_supported_runtime, get_artifacts_directory, helixuploadroot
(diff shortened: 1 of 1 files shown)


LoopedBard3 (approved):
Looks good to me.

Issue #4819, opened 2025-04-08
Crossgen runtime runs are completely failing


LoopedBard3:
Currently, all crossgen runs in the [dotnet-runtime-perf](https://dev.azure.com/dnceng/internal/_build?definitionId=702) pipeline are failing. One such failure being seen is the following:

```
(.venv) <DIR>\w\B2DF099F\e>python <DIR>\w\B2DF099F\e\performance\src\scenarios\crossgen2\pre.py crossgen2 --core-root <DIR>\p\Core_Root --single System.Net.WebProxy.dll 
----------------------------------------------
Initializing logger 2025-04-08 04:23:43.559211
----------------------------------------------
$ pushd "<DIR>\p\Core_Root"
$ <DIR>\p\Core_Root\corerun.exe <DIR>\p\Core_Root\crossgen2\crossgen2.dll <DIR>\p\Core_Root\System.Net.WebProxy.dll -o <DIR>\w\B2DF099F\e\crossgen.out\System.Net.WebProxy.ni.dll -O -r <DIR>\p\Core_Root\System.*.dll -r <DIR>\p\Core_Root\Microsoft.*.dll -r <DIR>\p\Core_Root\netstandard.dll -r <DIR>\p\Core_Root\mscorlib.dll
Activation Context creation failed. Error: 0x00000002
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly '<DIR>\p\Core_Root\crossgen2\crossgen2.dll'. The system cannot find the file specified.
File name: '<DIR>\p\Core_Root\crossgen2\crossgen2.dll'
$ popd
Process exited with status 3762504530
```

PR #4820, opened 2025-04-08
Disable Performance Crossgen tests

LoopedBard3:
Crossgen tests are currently failing per https://github.com/dotnet/performance/issues/4819. They have been this way for a while, and we haven't had a chance to look into them. Disabling them will allow us to more quickly act on other issues that pop up in our testing.


kunalspathak:
There is no runtime issue to my knowledge that reports such failures so this seems to be some kind of setup/infra issue.

caaavik-msft (approved):

e-kharion (approved):

Issue #4794, opened 2025-03-24
Introduce measuring MAUI template app with Syncfusion Toolkit Integration


ivanpovazan:
## Description

To collect data and measure performance of a "real-world" MAUI application we should introduce a new MAUI scenario for measuring size and startup performance of the `dotnet new maui -sc` aka `Sample Content` template application. This template was introduced in: https://github.com/dotnet/maui/pull/25419 and includes integration of: 

- Syncfusion Toolkit for .NET MAUI
- .NET MAUI Community Toolkit
- CommunityToolkit.Mvvm
- SQLite

and demonstrates basic usage patterns. 

By measuring size and startup performance of this template app we would collect and track data which is much closer to the "real-world" application, than the standard `dotnet new maui` sample. These measurements will also allow us to catch early any potential regression or issues with performance our customers are facing as it should match a feature-rich application.

For similar reasons, previously, we measured performance of the [Podcasts](https://github.com/microsoft/dotnet-podcasts) app. However, due to challenges of maintaining the app and mobile workload dependencies, eventually we stopped measuring this scenario. The new `Sample Content` template is regularly maintained an should not have this problem.

ivanpovazan:
@jonathanpeppers @matouskozak @LoopedBard3

matouskozak:
Completed in https://github.com/dotnet/performance/pull/4837

PR #4837, opened 2025-04-16
[Android] Add MAUI Android Sample Content template app + refactor TTFD naming

matouskozak:
- Add support for maui sample content android scenario by passing an extra argument to the dotnet new command
- Refactor MAUI Blazor Android naming so that TTFD tag is only included in startup measurements.

Internal run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2688185&view=results

---

https://github.com/dotnet/performance/issues/4794

### DIFF
--- a/eng/performance/maui_scenarios_android.proj
+++ b/eng/performance/maui_scenarios_android.proj
@@ -76,16 +76,16 @@
       <Command>$(Python) test.py sod --scenario-name &quot;%(Identity)&quot; $(ScenarioArgs)</Command>
     </HelixWorkItem>
     <HelixWorkItem Include="@(MAUIAndroidScenario -> 'Device Startup - %(Identity)')">
-      <_StartupAdditionalArguments>%(HelixWorkItem.StartupAdditionalArguments)</_StartupAdditionalArguments>
-      <_StartupTTFD Condition="$(_StartupAdditionalArguments.Contains('--use-fully-drawn-time'))"> (TTFD)</_StartupTTFD>
+      <_ScenarioName Condition="%(HelixWorkItem.StartupAdditionalArguments) == ''">%(Identity)</_ScenarioName>
+      <_ScenarioName Condition="%(HelixWorkItem.StartupAdditionalArguments) != ''">%(Identity) [%(HelixWorkItem.StartupAdditionalArguments)]</_ScenarioName>
       <PreCommands>echo on; xcopy %HELIX_CORRELATION_PAYLOAD%\$(PreparePayloadOutDirectoryName)\%(HelixWorkItem.ScenarioDirectoryName) %HELIX_WORKITEM_ROOT%\pub\ /E /I /Y</PreCommands>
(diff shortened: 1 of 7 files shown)


eng/performance/maui_scenarios_android.proj
```diff
@@ -70,12 +76,16 @@
...
       <PreCommands>echo on; xcopy %HELIX_CORRELATION_PAYLOAD%\$(PreparePayloadOutDirectoryName)\%(HelixWorkItem.ScenarioDirectoryName) %HELIX_WORKITEM_ROOT%\pub\ /E /I /Y</PreCommands>
-      <Command>$(Python) test.py devicestartup --device-type android --package-path pub\%(HelixWorkItem.ApkName).apk --package-name %(HelixWorkItem.PackageName) --scenario-name &quot;%(Identity)&quot; %(HelixWorkItem.StartupAdditionalArguments) $(ScenarioArgs)</Command>
+      <Command>$(Python) test.py devicestartup --device-type android --package-path pub\%(HelixWorkItem.ApkName).apk --package-name %(HelixWorkItem.PackageName) --scenario-name &quot;%(_ScenarioName)&quot; $(_StartupAdditionalArguments) $(ScenarioArgs)</Command>
     </HelixWorkItem>
     <HelixWorkItem Include="@(MAUIAndroidScenario -> 'Device Startup - %(Identity) NoAnimation')">
+      <_ScenarioName Condition="%(HelixWorkItem.StartupAdditionalArguments) == ''">%(Identity)</_ScenarioName>
+      <_ScenarioName Condition="%(HelixWorkItem.StartupAdditionalArguments) != ''">%(Identity) [%(HelixWorkItem.StartupAdditionalArguments)]</_ScenarioName>
```
e-kharion, line 86: I'm definitely missing a wider context, but if the HelixWorkItem.StartupAdditionalArguments is used as scenario name, rename might make sense.

matouskozak: I made a change before which included `(TTFD)` in the `%(Identity)` string which caused that also size and memory tests had `(TTFD)` in their name even though this is startup specific. What I tried to do here is to pass this information only to the startup test name.

LoopedBard3: Does this mean that the SOD tests are being duplicated, or that the change you had tried to make just added the TTFD incidentally and we are not doing SOD tests for both TTFD and non-TTFD?

LoopedBard3 (approved):
Left one comment but this looks good to me 👍.

eng/performance/maui_scenarios_android.proj
matouskozak: I think they could become duplicated if we would run both TTFD and non-TTFD measurements. Since we currently don't have that case we don't see the issue. This change just makes sure that the startup measurement test name reflects the additional startup arguments and SOD/Memory are unaffected.

PR #4842, opened 2025-04-23
Add build time measurement to Startup tool

kotlarmilos:
## Description

This PR adds support for measuring total build time in the `Startup` tool using binlog files.

## Changes

 - Add `DownloadBuildArtifacts` task to retrieve binlog files from the build machines
 - Update `run_performance_job.py` to copy binlog files into the helix payload directory
 - Update `Startup` tool to invoke `BuildTimeParser`
 - Implement BuildTimeParser to retrieve overall publish duration, plus `ILLink` and Mono-specific task times

## Validation

Run the internal dotnet-runtime-perf tool to verify that `BuildTimeParser` collects build times.

## Out-of-scope

These measurements are collected from the build machines. With lack of any build time measurements atm, we want to standup these now and iterate later if needed. As suggested in https://github.com/dotnet/runtime/issues/113073, apps build may be moved to Helix in the future.

Contributes to https://github.com/dotnet/runtime/issues/113073

### DIFF
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -742,11 +742,11 @@ def run_performance_job(args: RunPerformanceJobArgs):
         android_binlog_dir = os.path.join(root_payload_dir, "androidHelloWorldBinlog")
         shutil.copytree(os.path.join(args.built_app_dir, "androidHelloWorldBinlog"), android_binlog_dir)
 
-        for file in glob(os.path.join(android_binlog_dir, "**", "*.binlog"), recursive=True):
+        binlog_files = glob(os.path.join(android_binlog_dir, "**", "*.binlog"))
+        if binlog_files:
             dest = os.path.join(android_binlog_dir, "msbuild.binlog")
-            getLogger().info(f"Moving {file} to {dest}")
-            shutil.move(file, dest)
-            break
+            getLogger().info(f"Moving {binlog_files[0]} to {dest}")
+            shutil.move(binlog_files[0], dest)
         # Disabled due to not successfully building at the moment. https://github.com/dotnet/performance/issues/4729
         # if android_mono:
             # shutil.copy(os.path.join(args.built_app_dir, "MonoBenchmarksDroid.apk"), os.path.join(root_payload_dir, "MonoBenchmarksDroid.apk"))
@@ -765,20 +765,20 @@ def run_performance_job(args: RunPerformanceJobArgs):
         shutil.copytree(os.path.join(args.built_app_dir, "iosHelloWorldZip"), ios_hello_world_zip_dir)
 
         # Find the zip file in the directory and move it to iOSSampleApp.zip
-        for file in glob(os.path.join(ios_hello_world_zip_dir, "**", "*.zip")):
+        zip_files = glob(os.path.join(ios_hello_world_zip_dir, "**", "*.zip"))
+        if zip_files:
             dest = os.path.join(ios_hello_world_zip_dir, "iOSSampleApp.zip")
-            getLogger().info(f"Moving {file} to {dest}")
-            shutil.move(file, dest)
-            break
+            getLogger().info(f"Moving {zip_files[0]} to {dest}")
+            shutil.move(zip_files[0], dest)
 
         ios_hello_world_binlog_dir = os.path.join(payload_dir, "iosHelloWorldBinlog")
         shutil.copytree(os.path.join(args.built_app_dir, "iosHelloWorldBinlog"), ios_hello_world_binlog_dir)
 
-        for file in glob(os.path.join(ios_hello_world_binlog_dir, "**", "*.binlog")):
+        binlog_files = glob(os.path.join(ios_hello_world_binlog_dir, "**", "*.binlog"))
+        if binlog_files:
             dest = os.path.join(ios_hello_world_binlog_dir, "msbuild.binlog")
-            getLogger().info(f"Moving {file} to {dest}")
-            shutil.move(file, dest)
-            break
+            getLogger().info(f"Moving {binlog_files[0]} to {dest}")
+            shutil.move(binlog_files[0], dest)
 
     # ensure work item directory is not empty
     getLogger().info("Copying docs to work item directory so it isn't empty")
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -738,6 +738,7 @@ def run_performance_job(args: RunPerformanceJobArgs):
             raise Exception("Built apps directory must be present for Android benchmarks")
         getLogger().info("Copying Android apps to payload directory")
         shutil.copy(os.path.join(args.built_app_dir, "androidHelloWorld", "HelloAndroid.apk"), os.path.join(root_payload_dir, "HelloAndroid.apk"))
+        shutil.copy(os.path.join(args.built_app_dir, "androidHelloWorldBinlog", "msbuild.binlog"), os.path.join(root_payload_dir, "msbuild.binlog"))
         # Disabled due to not successfully building at the moment. https://github.com/dotnet/performance/issues/4729
         # if android_mono:
             # shutil.copy(os.path.join(args.built_app_dir, "MonoBenchmarksDroid.apk"), os.path.join(root_payload_dir, "MonoBenchmarksDroid.apk"))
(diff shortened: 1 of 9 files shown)


src/tools/ScenarioMeasurement/Startup/Startup.cs
```diff
@@ -25,6 +25,7 @@ enum MetricType
     WinUI,
     WinUIBlazor,
     TimeToMain2,
+    BuildTime,
```
matouskozak, line 28: Unsure whether `BuildTime` qualifies as part of "Startup.cs" but seems like it is sharing a lot of the infrastructure here.

Could we add test for it to "StartupTests.cs"?

scripts/run_performance_job.py
```diff
@@ -738,6 +738,15 @@ def run_performance_job(args: RunPerformanceJobArgs):
             raise Exception("Built apps directory must be present for Android benchmarks")
         getLogger().info("Copying Android apps to payload directory")
         shutil.copy(os.path.join(args.built_app_dir, "androidHelloWorld", "HelloAndroid.apk"), os.path.join(root_payload_dir, "HelloAndroid.apk"))
+
+        android_binlog_dir = os.path.join(root_payload_dir, "androidHelloWorldBinlog")
+        shutil.copytree(os.path.join(args.built_app_dir, "androidHelloWorldBinlog"),android_binlog_dir)
```
matouskozak, line 743:
```suggestion
        shutil.copytree(os.path.join(args.built_app_dir, "androidHelloWorldBinlog"), android_binlog_dir)
```

src/scenarios/shared/const.py
```diff
@@ -18,14 +18,16 @@
 DEVICEMEMORYCONSUMPTION = "devicememoryconsumption"
 ANDROIDINSTRUMENTATION = "androidinstrumentation"
 DEVICEPOWERCONSUMPTION = "devicepowerconsumption"
+DEVICEBUILD = "devicebuild"
```
matouskozak, line 21: https://github.com/dotnet/runtime/pull/113469#discussion_r2055921250

matouskozak (commented):
Thanks a lot for introducing these. As a follow-up we should extend this to .NET sample app scenarios.

Is there an internal CI run where we can check the output of this measurements?

src/scenarios/shared/const.py
kotlarmilos: Updated, thanks.

kotlarmilos:
> Is there an internal CI run where we can check the output of this measurements?

Yes, invoked the internal CI. Sharing it offline.

src/tools/ScenarioMeasurement/Startup/Startup.cs
kotlarmilos: Yes, we are able to reuse code and just implement a new parser.

I reviewed the tests and noticed that all artifact-related tests are currently disabled. We could add a binlog and write a test, but it would only serve as a one-off check -- since the input is artificial and parser is unlikely to change.

Let me know what you think.

matouskozak: > parser is unlikely to change
Not sure how likely it is that we extend the parser to measure additional build part. However, I agree that we might not need the test, especially since we don't test some other parsers either (e.g., parsing ADB log to extract startup time).

matouskozak (approved):
LGTM! thank you.

src/scenarios/shared/runner.py
```diff
@@ -957,3 +966,10 @@ def run(self):
             if not (self.dirs or builtdir):
                 raise Exception("Dirs was not passed in and neither %s nor %s exist" % (const.PUBDIR, const.BINDIR))
             sod.runtests(scenarioname=self.scenarioname, dirs=self.dirs or builtdir, upload_to_perflab_container=self.upload_to_perflab_container, artifact=self.traits.artifact)
+
+        elif self.testtype == const.BUILDTIME:
+            startup = StartupWrapper()
+            if not (self.binlogpath):
+                raise Exception("Binlog path was not passed provided.")
```
ivanpovazan, line 973: Should we check if the path exists? 
Additionally, the exception message sounds wrong, maybe something like:
```py
raise Exception("For build time measurements a valid binlog path must be provided")
```

src/tools/ScenarioMeasurement/Util/Parsers/BuildTimeParser.cs
```diff
@@ -0,0 +1,78 @@
...
+                    androidappbuilderTimes.Add(ms);
+                }
+            }
+
+            publishTimes.Add(build.Duration.TotalMilliseconds);
+        }
+
+
+        if (illinkTimes.Count > 0)
+            yield return new Counter { Name = "ILLink Time", MetricName = "ms", DefaultCounter = false, TopCounter = true, Results = illinkTimes.ToArray() };
```
ivanpovazan, line 68: How do we process the `Counter.Results` later on in the code, are we calculating the average?
I am asking this because in Android builds we will have multiple ILLink task executions (for each rid/abi).

src/tools/ScenarioMeasurement/Util/Util.csproj
```diff
@@ -14,6 +14,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" />
     <PackageReference Include="System.Security.Principal.Windows" />
+    <PackageReference Include="MSBuild.StructuredLogger" VersionOverride="2.2.386" />
```
ivanpovazan, line 17: Why do we need to override the version?

ivanpovazan (approved):
Nice work! I only left few minor comments, thank you!

src/tools/ScenarioMeasurement/Util/Util.csproj
kotlarmilos: The project is using central package version management, so version numbers can either be overridden or defined in Versions.props. I moved it to Versions.props.

src/tools/ScenarioMeasurement/Util/Parsers/BuildTimeParser.cs
kotlarmilos: I think it stores the configuration along with the measurements. @matouskozak please correct me if I am wrong.

matouskozak: Yes, the configuration is stored along the measurements. E.g., for the testing run, the stored data looks like: 

| Series                                      | Test Name                  | Avg          |
|---------------------------------------------|----------------------------|--------------|
| coreclr_JIT-Build Time - Android HelloWorld | ILLink Time                | 5,090.2713   |
| coreclr_JIT-Build Time - Android HelloWorld | AndroidAppBuilderTask Time | 5,660.4876   |
| coreclr_JIT-Build Time - Android HelloWorld | Publish Time               | 25,808.5949  |
| coreclr_R2R-Build Time - Android HelloWorld | AndroidAppBuilderTask Time | 5,114.4365   |
| coreclr_R2R-Build Time - Android HelloWorld | Publish Time               | 18,773.2464  |
| mono_AOT-Build Time - Android HelloWorld    | MonoAOTCompiler Time       | 92,565.2857  |
| mono_AOT-Build Time - Android HelloWorld    | AndroidAppBuilderTask Time | 9,505.2283   |
| mono_AOT-Build Time - Android HelloWorld    | Publish Time               | 106,322.1502 |
| mono_JIT-Build Time - Android HelloWorld    | ILLink Time                | 4,717.3509   |
| mono_JIT-Build Time - Android HelloWorld    | AndroidAppBuilderTask Time | 6,028.4118   |
| mono_JIT-Build Time - Android HelloWorld    | Publish Time               | 23,203.8513  |

matouskozak (dismissed):

scripts/run_performance_job.py
```diff
@@ -738,6 +738,15 @@ def run_performance_job(args: RunPerformanceJobArgs):
             raise Exception("Built apps directory must be present for Android benchmarks")
         getLogger().info("Copying Android apps to payload directory")
         shutil.copy(os.path.join(args.built_app_dir, "androidHelloWorld", "HelloAndroid.apk"), os.path.join(root_payload_dir, "HelloAndroid.apk"))
+
+        android_binlog_dir = os.path.join(root_payload_dir, "androidHelloWorldBinlog")
+        shutil.copytree(os.path.join(args.built_app_dir, "androidHelloWorldBinlog"), android_binlog_dir)
+
+        for file in glob(os.path.join(android_binlog_dir, "**", "*.binlog"), recursive=True):
```
LoopedBard3, line 745: Why does this break on the first file rather than moving all the files or checking in an if for only one file and moving that one?

LoopedBard3 (approved):
This looks good to me, are you ready for it and the associated enabling of it in runtime to be merged?

scripts/run_performance_job.py
kotlarmilos: That was copied from the existing block. I’ve updated it to copy only a single file.

kotlarmilos:
> This looks good to me, are you ready for it and the associated enabling of it in runtime to be merged?

Yes, both PRs are ready. Please proceed and merge this PR first. The https://github.com/dotnet/runtime/pull/113469 is a follow-up.

LoopedBard3:
Failing CI does is unrelated, the errors being hit are unrelated and this PR doesn't update any microbenchmark flow.

PR #4851, opened 2025-05-13
common class for server run command build

VincentBu:
Between GCPerfSim running on the aspnet machines and the ASP.NET benchmarks, there is a lot of common functionality. The "ServerRun.CommandBuilder.cs" provides a common class to be used between the two.


mrsharm (approved):
LGTM

PR #4870, opened 2025-06-04
[Android] Add linking type parameter to performance job configuration

kotlarmilos:
## Description

This PR adds a linking‐type parameter to the configuration, allowing dashboards to compare static and dynamic linking.

### DIFF
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -89,7 +89,7 @@ class RunPerformanceJobArgs:
     r2r_run_type: Optional[str] = None
     experiment_name: Optional[str] = None
     codegen_type: str = "JIT"
-    linking_type: str = "Dynamic"
+    linking_type: str = "dynamic"
     runtime_type: str = "coreclr"
     affinity: Optional[str] = "0"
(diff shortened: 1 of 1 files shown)


kotlarmilos:
The internal pipeline `dotnet-runtime-perf` succeeded and the dashboards are updated to include statically linked SOD measurements.

LoopedBard3 (approved):
This looks good to me 👍. We will also need to update the autofiling configuration to include the LinkingType. As this is not its own column in the ADX, it will need to be pulled out in the query for android tests, but I think that doing this parse is fine for now as the number of tests that need this should not be too high. @caaavik-msft any thoughts on whether we should add this as it's own column for ADX?

kotlarmilos:
> We will also need to update the autofiling configuration to include the LinkingType. As this is not its own column in the ADX, it will need to be pulled out in the query for android tests, but I think that doing this parse is fine for now as the number of tests that need this should not be too high.

Correct. I updated the dashboard to filter based on this value.

LoopedBard3:
Cool. I think this is good to merge as the ADX column question can be answered next week and adding this in minimizes any mislabeled data.

PR #4873, opened 2025-06-16
Remove Upload SAS token references and code flow.

LoopedBard3:
Remove all references to the perflab upload token, including secret names and const references for it. This does not include updates for secret munger/sychronization runs as they will be covered in a follow-up PR removing multiple secret references. Also made some minor code flow updates to match with this token being removed.

Internal run to test upload (only one failure that was unrelated/noise): https://dev.azure.com/dnceng/internal/_build/results?buildId=2731528&view=results


DrewScoggins (approved):
LGTM

LoopedBard3:
None of the failures are related to upload failure, instead being due to flakiness.

PR #4875, opened 2025-06-16
Update windows open queue to windows11 and VMimage to windows 2025

LoopedBard3:
Update windows open queue to windows11 and VMimage to windows 2025.

Internal (successful) test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2731529&view=results

### DIFF
--- a/eng/pipelines/templates/build-machine-matrix.yml
+++ b/eng/pipelines/templates/build-machine-matrix.yml
@@ -11,7 +11,7 @@ jobs:
       osGroup: windows
       archType: x64
       pool:
-        vmImage: windows-2019
+        vmImage: windows-2025
       ${{ insert }}: ${{ parameters.jobParameters }}
       ${{ if eq(parameters.isPublic, true) }}:
         osVersion: 24H2
@@ -29,7 +29,7 @@ jobs:
       osVersion: RS5
       archType: x64
       pool:
-        vmImage: windows-2019
+        vmImage: windows-2025
       machinePool: Open
       queue: Windows.10.Amd64.ClientRS5.Open
       ${{ insert }}: ${{ parameters.jobParameters }}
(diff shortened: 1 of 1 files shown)


LoopedBard3:
The internal run was successful.

LoopedBard3:
All of the scenarios failures are due to `Process xxxx failed to run. Result: CloseFailed`, the sdk_scenario failures are mostly due to `Process 5220 failed to run. Result: TimeoutExceeded` with some due to `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'app'`. The internal run on actual machines was successful so I am going to try a public run with windows-2022 instead of 2025 to see if that helps.

LoopedBard3:
This test run that uses windows-2022 instead of 2025 still hits the same issue. https://dev.azure.com/dnceng-public/public/_build/results?buildId=1070518&view=results. I think it would make sense to take the update so we don't get caught with the old image deprecated and then spend some time investigating these specific cases.

DrewScoggins:
Do we see these same failures on runs that are not a part of this PR?

I just looked and it seems like we are seeing the same failures. I think merging here is fine.

DrewScoggins (approved):

LoopedBard3:
It seems the performance-ci scheduled run hits the same sdk_scenarios scenarios while the scenarios succeed, but the scenarios issues are not hit. So, it probably makes more sense to keep this until we either have a chance to investigate the scenarios or if we see mass failure due to the images going away.

LoopedBard3:
It seems that these errors are being hit elsewhere as well (first run from https://github.com/dotnet/performance/pull/4880), I think we should go ahead and merge this and then take a look at the specific issues.

LoopedBard3:
Merging as the second test run has the same errors as we are seeing on main.

PR #4879, opened 2025-06-24
[maui, android] explicitly set R2R properties

jonathanpeppers:
We are shortly going to change the defaults for CoreCLR, where `Release` mode will use by default:

    -p:PublishReadyToRun=true -p:PublishReadyToRunComposite=true

...when `$(PublishReadyToRun)` or `$(PublishReadyToRunComposite)` are not set.

Let's explicitly set the MSBuild properties, so the existing graphs will not change in behavior.


vitek-karas (approved):

LoopedBard3:
None of the performance-ci tests are impacted by this change.

PR #4880, opened 2025-06-26
Add performance internal runs for Viper machine queues.

LoopedBard3:
Add performance internal runs for Viper machine queues. This ensures we will have some data overlap between the two queue types. This only sets up the Viper queues to be able to run internally. 

Successful internal test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2737540&view=results


DrewScoggins (approved):
LGTM

PR #4884, opened 2025-07-10
Finish removing unused secret token references

LoopedBard3:
Finish removing unused uploading tokens as we have moved to using identities. This should be the final PR to clean up the now unused secrets. Essentially reverts https://github.com/dotnet/performance/pull/2556 and removes the unused variable group.

Remove the usage of the dotnet-benchview variable group as we no longer use PerfCommandUploadToken. Also removed the infrastructure for rotating certain secrets as the majority were deleted, PerfCommandUploadTokenNonEscaped is no longer used and can be deleted after this PR, and pvscmdupload-storage-key and pvscmdupload-connection-string seem to only be used by this flow. Removed supporting files that are no longer needed.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2747477&view=results

Follow-up:
Delete the final secrets from the key-vault
Delete the dotnet-benchview variable group


LoopedBard3:
For some reason the variable group dotnet-benchview is still being imported despite not being referenced explicitly, I think that we may need to have the pipelines variable group connection removed to get it to go away completely. As long as this does not fail the pipelines, I think we should merge it and then investigate removing the variable group completely.

LoopedBard3:
pvscmdupload-connection-string may still be needed for some other Azure Functions, need to verify before merging this.

LoopedBard3:
Double checked and the connection string locations for the function are currently disabled, so we should be good to merge.

LoopedBard3:
Pipeline failures are unrelated.

caaavik-msft (approved):

DrewScoggins (approved):
🎊

PR #4889, opened 2025-07-17
Add .NET 9 to 10 performance reports

caaavik-msft:
This PR contains some reports summarising performance regressions found when upgrading from .NET 9 to 10 and the commits that we believe caused them.


LoopedBard3 (approved):
👍.

DrewScoggins (approved):

PR #4893, opened 2025-07-28
Create custom events in GCPerfsim test

VincentBu:
In this PR, we create custom events that log private memory when GCStart and GCEnd events are triggered.
Provider name: `GCPerfsimCustomEvents`
Events:
 - `PrivateMemoryAtGCStart`: when GCStart is triggered, the eventsource emit this event;
 - `PrivateMemoryAtGCEnd`: when GCEnd is triggered, the eventsource emit this event.

In PerfView, it looks like:
image


mrsharm:
Can close in favor of looking at the TotalCommittedMemory.

PR #4896, opened 2025-08-01
Use hash of the test name to determine what partition it is run in

caaavik-msft:
Due to one of the changes in the following commit range, some Mono tests started getting skipped as they were not being run as part of any partitions: https://github.com/dotnet/runtime/compare/33ce341cd9e83fb8da17e5a71104c23593cef209...b146d7512ce67051e127ab48dc2d4f65d30e818f

This change makes it so that partitioning is done in a way that guarantees each test is run in exactly one partition. For configurations that run a small number of tests, this may make them slightly unbalanced, but we can investigate ways to balance them better in the future while this resolves the immediate bug.


caaavik-msft:
The changes have been validated against an internal run and should be ready to merge once approved

LoopedBard3 (approved):
Looks good.

PR #4898, opened 2025-08-05
Fix duplicated test name in Perf.TypeName and update validation

caaavik-msft:
Due to how the FullNameProvider works, the `Dictionary<string, bool>[]` and `Dictionary<List<int[]>[,], List<int?[][][,]>>[]` arguments were both being converted to the string ``System.Collections.Generic.Dictionary`2[]``. Patching this in BenchmarkDotNet would cause too many test names to change, and this is the only case in this repo that there is a duplicate full name, so we can address it by wrapping the type and overriding the ToString.

The display names that I have put here ensure that the test names remain the same as they were before, with only the `Dictionary<List<int[]>[,], List<int?[][][,]>>[]` case now being converted to ``typeof(System.Collections.Generic.Dictionary`2[]) (COMPLEX)`` to differentiate it from the simpler case.

I have also updated the validators to ensure this doesn't happen again.

### DIFF
--- a/src/benchmarks/micro/libraries/System.Reflection.Metadata/Perf.TypeName.cs
+++ b/src/benchmarks/micro/libraries/System.Reflection.Metadata/Perf.TypeName.cs
@@ -11,40 +11,53 @@ namespace System.Reflection.Metadata
     [BenchmarkCategory(Categories.Libraries)]
     public class Perf_TypeName
     {
+        public class ArgumentTypeWrapper
+        {
+            private readonly string displayName;
+
+            public ArgumentTypeWrapper(Type type, string displayName)
+            {
+                Type = type;
+                this.displayName = displayName;
+            }
+
+            public Type Type { get; }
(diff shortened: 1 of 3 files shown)


caaavik-msft:
@adamsitnik As you had originally added this benchmark, are you able to confirm that the changes to the benchmark by using a wrapper type are acceptable and that adding the property lookup should have a negligible impact on the performance measurement?

DrewScoggins:
@caaavik-msft We should probably go ahead and merge this, yeah? Also did we do work on the reporting side to coalesce the results into one trend?

caaavik-msft:
> @caaavik-msft We should probably go ahead and merge this, yeah? Also did we do work on the reporting side to coalesce the results into one trend?

Ah yes, we can merge it, although I don't think there will be a way to fix this on the reporting side, we have no way of knowing which test result came from which, but it will be fixed from now on. We may get a perf regression from this now that the test will stabilise on one case rather than being random

PR #4902, opened 2025-08-08
Update iOS builds to disable code signing again

LoopedBard3:
Update iOS builds to disable code signing again.

After updating the maui packages recently in __, we started hitting an error around the signing certificates not being able to be found. Looking at the updates we took, it seems likely that this update https://github.com/dotnet/macios/pull/20824 caused the _RequireCodeSigning parameter we were setting to false to no longer block the same code path. This PR adds the setting of the parameter EnableCodeSigning to false to re-disable that flow.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2767936&view=results

### DIFF
--- a/src/scenarios/mauiblazorios/pre.py
+++ b/src/scenarios/mauiblazorios/pre.py
@@ -40,7 +40,7 @@ with open(f"{const.APPDIR}/Components/Pages/Home.razor", "a") as homeRazorFile:
 
 # Build the IPA
 # NuGet.config file cannot be in the build directory due same cause as to https://github.com/dotnet/aspnetcore/issues/41397
-precommands.execute(['/p:_RequireCodeSigning=false', '/p:ApplicationId=net.dot.mauiblazortesting'])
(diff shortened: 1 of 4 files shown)


DrewScoggins (approved):
LGTM

src/scenarios/mauiblazorios/pre.py
```diff
@@ -40,7 +40,7 @@
 
 # Build the IPA
 # NuGet.config file cannot be in the build directory due same cause as to https://github.com/dotnet/aspnetcore/issues/41397
-precommands.execute(['/p:_RequireCodeSigning=false', '/p:ApplicationId=net.dot.mauiblazortesting'])
+precommands.execute(['/p:_RequireCodeSigning=false', '/p:EnableCodeSigning=false', '/p:ApplicationId=net.dot.mauiblazortesting'])
```
kotlarmilos: According to https://github.com/dotnet/macios/pull/20824, I think _RequireCodeSigning can be removed.

kotlarmilos (approved):

PR #4903, opened 2025-08-08
Add standard parameter for binlog save location to PrepareWorkloadItemItems call

LoopedBard3:
Add standard parameter for binlog save location to PrepareWorkloadItemItems call for use in the projects that create binlogs so that they are instead uploaded in the pipeline steps and viewable in the pipeline run artifacts. The approach used in this PR is to check whether the ArtifactsLogDirectory parameter was set, such as when called from the run_performance_job.py script, and to use it rather than the previous hard coded binlog save location if it is set.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2767900&view=results (The binlogs properly showed up in the Azure Artifacts)

### DIFF
--- a/eng/performance/maui_scenarios_android.proj
+++ b/eng/performance/maui_scenarios_android.proj
@@ -27,10 +27,6 @@
     <_MSBuildArgs Condition="'$(CodegenType)' == 'NativeAOT'">$(_MSBuildArgs);/p:PublishAot=true</_MSBuildArgs>
 
     <RunConfigsString>$(RuntimeFlavor)_$(CodegenType)</RunConfigsString>
-    
-    <!-- Use ArtifactsLogDir if provided, otherwise use payload directory for binlogs -->
-    <BinlogPath Condition="'$(ArtifactsLogDir)' != ''">$(ArtifactsLogDir)\%(PreparePayloadWorkItem.ScenarioDirectoryName).$(RunConfigsString).binlog</BinlogPath>
-    <BinlogPath Condition="'$(ArtifactsLogDir)' == ''">$(PreparePayloadWorkItemBaseDirectory)%(PreparePayloadWorkItem.ScenarioDirectoryName)\%(PreparePayloadWorkItem.ScenarioDirectoryName).$(RunConfigsString).binlog</BinlogPath>
   </PropertyGroup>
 
   <Target Name="RemoveDotnetFromCorrelationStaging" BeforeTargets="BeforeTest">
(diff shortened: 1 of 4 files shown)


eng/performance/maui_scenarios_android.proj
```diff
@@ -104,7 +106,7 @@
       <Command>$(Python) test.py devicememoryconsumption --device-type android --package-path pub\%(HelixWorkItem.ApkName).apk --package-name %(HelixWorkItem.PackageName) --scenario-name &quot;%(Identity)&quot; --runtime 30 --test-iteration 2 $(ScenarioArgs)</Command>
     </HelixWorkItem>
     <HelixWorkItem Include="@(MAUIAndroidScenario -> 'Build Time - %(Identity)')">
-      <PreCommands>echo on; xcopy %HELIX_CORRELATION_PAYLOAD%\$(PreparePayloadOutDirectoryName)\%(HelixWorkItem.ScenarioDirectoryName) %HELIX_WORKITEM_ROOT%\pub\ /E /I /Y &amp;&amp; mkdir %HELIX_WORKITEM_ROOT%\traces &amp;&amp; copy %HELIX_CORRELATION_PAYLOAD%\$(PreparePayloadOutDirectoryName)\%(HelixWorkItem.ScenarioDirectoryName)\%(HelixWorkItem.ScenarioDirectoryName).$(RunConfigsString).binlog %HELIX_WORKITEM_ROOT%\traces\</PreCommands>
+      <PreCommands>echo on; xcopy %HELIX_CORRELATION_PAYLOAD%\$(PreparePayloadOutDirectoryName)\%(HelixWorkItem.ScenarioDirectoryName) %HELIX_WORKITEM_ROOT%\pub\ /E /I /Y</PreCommands>
```
DrewScoggins, line 109: Do we upload what is in this traces folder as a file in Helix? I assume that is why we are getting rid of this logic, right?

DrewScoggins (approved):
LGTM

eng/performance/maui_scenarios_android.proj
LoopedBard3: I think our initial assumption was that traces should have been uploaded somewhere, but it seems that they were not (maybe due to something semi-related). So this PR removes that attempt and moves the file saving up to the AzDo pipeline level rather than in the helix level.

LoopedBard3: Actually, the binlog was being uploaded in the helix results, but due to the SendToHelix step not running properly during the failure there would be nothing uploaded. So this essentially moves the trace upload step earlier so we get it in more failure flows.

PR #4909, opened 2025-08-14
Ensure binlog is still copied to traces dir for maui build time scenarios

caaavik-msft:
Fixes bug introduced in https://github.com/dotnet/performance/pull/4903 which removed the commands to copy the binlog to the tracedir. This still needs to be kept because the buildtime scenarios read the binlog as part of the test execution.


matouskozak (approved):
Thank you for the quick fix

caaavik-msft:
Test fix was slightly less trivial than I thought, I have started an internal run against this, but this change should ensure the binlog is still uploaded to helix for the maui scenarios, and also copied to the artifacts log dir. Will wait for the successful run before merging.

caaavik-msft:
Here is the run, seems to have fixed this bug although the maui scenarios are still failing due to a bug that existed before: https://dev.azure.com/dnceng/internal/_build/results?buildId=2771854&view=results

I can see the binlogs published to the azdo artifacts as well, so that still works. We should be good to merge after approval

kotlarmilos (approved):
Thanks!

PR #4911, opened 2025-08-20
Allow for building helix payloads directly from build artifacts in python

caaavik-msft:
This PR is part of a larger bit of work to be able to automatically run targeted performance tests against historical versions of the runtime. We have recently been caching build artifacts produced as part of our performance testing pipeline, however those build artifacts are not in the payload form that are sent to helix.

With this PR, there is a script build_runtime_payload.py which contains helper functions that can take the build artifact (either archived or already unpacked into a directory), and construct the payload in the format that is needed for helix jobs.

This PR also contains some other useful refactors to help reduce the size of some of the very large functions in run_performance_job.py, as well as to rearrange some parts so that things like building the BDN arguments, the run configurations, and the work item command are all done at once rather than spread out across the file.

I have also cleaned up some code that was unused, for example the --cli arguments that are passed to benchmarks_ci.py seem to have been only used by benchview which no longer exists. In addition, MonoAOT was downloading the coreclr runtime but was not using it, so I have removed that build dependency.


LoopedBard3 (approved):
This looks good to me. It seems that everything was moved around properly and the breakouts/refactors make sense to me. Do we have a test run of the internal pipelines, and if not, can we do a test run?

caaavik-msft:
> This looks good to me. It seems that everything was moved around properly and the breakouts/refactors make sense to me. Do we have a test run of the internal pipelines, and if not, can we do a test run?

There is a run here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2777400&view=results

The ios scenarios times out due to an unrelated issue so rerunning them, but it seems everything is working properly.

eng/pipelines/templates/runtime-perf-job.yml
```diff
@@ -37,7 +36,7 @@ jobs:
     # Test job depends on the corresponding build job
     ${{ if eq(parameters.downloadSpecificBuild.buildId, '') }}:
       dependsOn:
-        - ${{ if not(or(in(parameters.runtimeType, 'AndroidMono', 'AndroidCoreCLR', 'iOSMono', 'iOSNativeAOT', 'wasm'), and(eq(parameters.runtimeType, 'mono'), ne(parameters.codeGenType, 'AOT')))) }}:
```
DrewScoggins, line 40: Before we check for it not being AOT, but we no longer do that. It is not obvious to me why we made the change, could you explain it to me? :)

scripts/build_runtime_payload.py
```diff
@@ -0,0 +1,301 @@
...
+
+def build_monoaot_payload(
+    monoaot_artifacts_archive_or_dir: str, payload_dest: str, architecture: str
+) -> None:
+    """Build a Mono AOT payload consisting of cross tools + runtime pack.
+
+    The function expects either an artifacts directory or an archive containing
+    the structure produced by Mono AOT builds. Two extractions are performed:
+      1. Cross compiler / AOT toolchain for the given architecture.
+      2. The runtime pack published under `pack/` for consumption.
```
DrewScoggins: This comment mentions that we are getting the runtime pack published under `pack/`, but the copy looks like it is not coming from the `pack` directory.

scripts/run_performance_job.py
```diff
@@ -310,15 +318,219 @@ def logical_machine_to_queue(logical_machine: str, internal: bool, os_group: str
...
+        if codegen_type.lower() == "interpreter":
+            category_exclusions += ["NoInterpreter"]
+
+    if experiment_name == "memoryRandomization":
+        bdn_arguments += ["--memoryRandomization", "true"]
+
+    if runtime_type == "wasm":
+        category_exclusions += ["NoInterpreter", "NoWASM", "NoMono"]
+
+        wasm_args = "--expose_wasm"
```
DrewScoggins: I assume we are doing string concatenation here, instead of using a list like the rest of the command line, because we have to drop it into another command line rather than it being input to RunCommand. Would we still want to just use the same style of syntax though, and do some conversion on the list to add it to the args?

DrewScoggins (approved):
LGTM

eng/pipelines/templates/runtime-perf-job.yml
caaavik-msft: From what I can tell, this dependency is just never used for Mono AOT. I traced back the history of why it was listed as a dependency, and I think it was a mistake introduced as a major build job refactor done across all dotnet/runtime jobs. I checked a test run with this and everything still ran smoothly.

scripts/build_runtime_payload.py
caaavik-msft: Will update the comment to clarify, but more just meant that the runtime pack is put into a directory called pack, not from a directory called pack

DrewScoggins (approved):
LGTM

PR #4912, opened 2025-08-20
Add support for running microbenchmarks using a manifest json

caaavik-msft:
This PR is part of a larger piece of work around being able to run targeted performance tests such as comparing multiple runtime versions side-by-side. Currently the way we define what jobs/tests we run in performance tests are by using command line arguments. However, the arguments can be a bit difficult to construct sometimes, and for some configurations it does not support the ability to run some combinations of jobs at the same time, only being possible to run in separate invocations.

With this PR, I have made it so that you can pass in a manifest.json argument which defines:
- The list of test cases you want to run
- The base job settings
- The list of jobs to run
- Run setting overrides per test-case

As an example, the following json can be used to validate https://github.com/dotnet/perf-autofiling-issues/issues/60871:

```json
{
    "benchmarkCases": [
        "System.Net.Primitives.Tests.CredentialCacheTests.ForEach(uriCount: 10, hostPortCount: 10)",
        "System.Net.Primitives.Tests.CredentialCacheTests.GetCredential_HostPort(host: \"notfound\", hostPortCount: 10)",
        "System.Net.Primitives.Tests.CredentialCacheTests.GetCredential_HostPort(host: \"name5\", hostPortCount: 10)",
        "System.Collections.TryGetValueFalse<Int32, Int32>.SortedDictionary(Size: 512)"
    ],
    "benchmarkCaseRunOverrides": {
        "System.Net.Primitives.Tests.CredentialCacheTests.ForEach(uriCount: 10, hostPortCount: 10)": {
            "operationCount": 5471872
        },
        "System.Net.Primitives.Tests.CredentialCacheTests.GetCredential_HostPort(host: \"notfound\", hostPortCount: 10)": {
            "operationCount": 18616768
        },
        "System.Net.Primitives.Tests.CredentialCacheTests.GetCredential_HostPort(host: \"name5\", hostPortCount: 10)": {
            "operationCount": 19320704
        },
        "System.Collections.TryGetValueFalse<Int32, Int32>.SortedDictionary(Size: 512)": {
            "operationCount": 40512
        }
    },
    "baseJob": {
        "run": {
            "warmupCount": 10,
            "launchCount": 5,
            "iterationCount": 15
        }
    },
    "jobs": {
        "Baseline-108fa785": {
            "infrastructure": {
                "toolchain": {
                    "type": "CoreRun",
                    "tfm": "net10.0"
                    "coreRunPath": "C:\\path\to\\baseline\\corerun.exe"
                }
            }
        },
        "Compare-edb570cd": {
            "infrastructure": {
                "toolchain": {
                    "type": "CoreRun",
                    "tfm": "net10.0"
                    "coreRunPath": "C:\\path\to\\compare\\corerun.exe"
                }
            }
        }
    }
}
```

With this, the operation counts for each test case are fixed for a much fairer comparison. This is not possible currently in BDN as there is no way to do per-test-case overrides.

The schema for defining jobs in the manifest matches identically to the schema used for defining jobs inside BDN itself. This means it is possible to construct any possible set of jobs with full flexibility.

I also changed our argument parsing logic to use the same library that BenchmarkDotNet itself uses, that way when you run `--help` on our executable, you can see both our custom args and BDN's custom args in the same dialog.

I have a separate PR #4911 which works well alongside this for creating multiple core_root payloads from the build artifacts, so that one can build a script to automatically validate performance changes detected by the auto-filer.

This PR needs a bit more testing on some other configurations as I have just been focusing on the corerun scenario, but creating this now for early feedback.


caaavik-msft:
I am abandoning this PR until I or another person is able to pick it up back up, I have some local changes I will push up later and some others that I will put in a separate PR. I think longer term the right solution for this would be to instead rather than having each job be a JSON, it should be a separate command line arguments and we just call BenchmarkDotNet's config parser multiple times for each row. However, we would still want to be able to override the Run modes per benchmark.

PR #4925, opened 2025-08-28
Update darc dependencies

LoopedBard3:
This fixes the current failing maui app builds. Also updated the darc subscriptions to be nearly as up to date as possible.

Internal Android only test: https://dev.azure.com/dnceng/internal/_build/results?buildId=2782010&view=logs&j=c635d93d-9950-5336-e47e-5637d983de8b&t=b2162994-51cd-5214-0ea9-9931dfe91f51

Let me know if we are worried about taking the non-maui exclusive updates and I can scale back the updates in this PR.

### DIFF
--- a/eng/pipelines/sdk-perf-jobs.yml
+++ b/eng/pipelines/sdk-perf-jobs.yml
@@ -344,53 +344,53 @@ jobs:
 
 - ${{ if parameters.runPrivateJobs }}:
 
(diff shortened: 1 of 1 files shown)


DrewScoggins (approved):

LoopedBard3:
CI errors match what we have been seeing in other runs and each is only 1 or 2 test failures out of all the tests run.

PR #4934, opened 2025-09-02
Disable Android package verifier during Android test runs

LoopedBard3:
Disable Android package verifier during Android test runs.

Test run (failures are unrelated to this change): https://dev.azure.com/dnceng/internal/_build/results?buildId=2782877&view=logs&j=c635d93d-9950-5336-e47e-5637d983de8b&t=18e04774-6cfd-59cb-6adf-eb68d9eff0bc

### DIFF
--- a/src/scenarios/shared/androidhelper.py
+++ b/src/scenarios/shared/androidhelper.py
@@ -4,6 +4,7 @@ from performance.common import RunCommand
 from logging import getLogger
 from shared import const
 from shared.util import xharness_adb, xharnesscommand
+from subprocess import CalledProcessError
 
 class AndroidHelper:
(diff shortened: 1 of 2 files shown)


DrewScoggins (dismissed):
LGTM, but why does it have to be in two places?

LoopedBard3:
> LGTM, but why does it have to be in two places?

Due to how the instrumentation runs run, they do not currently us the same setup_device flow that the androidhelper.py provides, instead following it's own setup flow.

DrewScoggins (approved):

LoopedBard3:
Same unrelated failures as other ci runs, merging.

PR #4938, opened 2025-09-03
add corerun build tool for MCP server

VincentBu:
This PR adds corerun build tool for MCP server. Currently we support following arguments:
- buildConfig(Debug, Checked, Release)
- arch(x64, x86, arm64)

### DIFF
--- a/src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.MCPServer/CoreRun.cs
+++ b/src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.MCPServer/CoreRun.cs
@@ -10,16 +10,19 @@ namespace GC.Infrastructure.MCPServer
         private static readonly string[] ValidBuildConfigs = new string[] { "Debug", "Release", "Checked" };
         private static readonly string[] ValidArchs = new string[] { "x64", "x86", "arm64" };
 
-        [McpServerTool(Name = "build_clr_libs"), Description("Build clr and libs.")]
-        public async Task<string> BuildCLRAndLibs(string runtimeRoot, string buildConfig, string arch = "x64")
+        [McpServerTool(Name = "build_clr_libs"), Description("Builds the CoreCLR runtime and base class libraries for the .NET runtime. This is a prerequisite step before generating CoreRun executables for performance testing and benchmarking.")]
+        public async Task<string> BuildCLRAndLibs(
+            [Description("The absolute path to the root directory of the .NET runtime repository (e.g., 'C:\\runtime'). This should contain the build.cmd script.")] string runtimeRoot,
+            [Description("The build configuration for CoreCLR compilation. Debug includes debugging symbols and assertions, Release is optimized for performance, and Checked includes some debugging features with optimizations. Valid options: Debug, Release, Checked.")] string buildConfig,
+            [Description("The target CPU architecture for the build. Determines which instruction set and calling conventions to use. Valid options: x64 (64-bit Intel/AMD), x86 (32-bit Intel/AMD), arm64 (64-bit ARM).")] string arch = "x64")
         {
             if (!ValidBuildConfigs.Contains(buildConfig))
             {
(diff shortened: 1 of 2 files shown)


src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.MCPServer/CoreRun.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.ComponentModel;
+using System.Diagnostics;
+
+namespace GC.Infrastructure.MCPServer
+{
+    [McpServerToolType]
+    internal class CoreRun
+    {
+        [McpServerTool(Name = "build_clr_libs"), Description("Build clr and libs.")]
+        public int BuildCLRAndLibs(string runtimeRoot, string buildConfig, string arch = "x64")
```
mrsharm: Change to `Task<string>` - encode a string with the successful message to propagate to the client. And async behavior to a number of places such as `await process.WaitForExitAsync()` - if this fails, the string that gets returned should contain the error (and not writing to Console).

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.MCPServer/CoreRun.cs
```diff
@@ -0,0 +1,88 @@
...
+                process.Start();
+                process.BeginOutputReadLine();
+                process.BeginErrorReadLine();
+                process.WaitForExit();
+                return process.ExitCode;
+            }
+        }
+
+        [McpServerTool(Name = "generate_core_root"), Description("Generate Core_Root.")]
+        public int GenerateCoreRoot(string runtimeRoot, string buildConfig, string arch = "x64")
```
mrsharm: int -> Task<string> and make this async.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.MCPServer/CoreRun.cs
```diff
@@ -0,0 +1,88 @@
...
+                process.ErrorDataReceived += (sender, args) => Console.Error.WriteLine(args.Data);
+                process.Start();
+                process.BeginOutputReadLine();
+                process.BeginErrorReadLine();
+                process.WaitForExit();
+                return process.ExitCode;
+            }
+        }
+
+        private bool CheckBuildConfig(string buildConfig)
```
mrsharm: this can be made static and so can the val below -- we should move the `validConfigs` and `validArchs` to private static readonly fields.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.MCPServer/CoreRun.cs
```diff
@@ -0,0 +1,88 @@
...
+            using (var process = new Process())
+            {
+                process.StartInfo.FileName = "cmd.exe";
+                process.StartInfo.Arguments = $"/C build.cmd generatelayoutonly {arch} {buildConfig}";
+                process.StartInfo.RedirectStandardOutput = true;
+                process.StartInfo.RedirectStandardError = true;
+                process.StartInfo.UseShellExecute = false;
+                process.StartInfo.CreateNoWindow = true;
+                process.StartInfo.WorkingDirectory = workingDirectory;
+                process.OutputDataReceived += (sender, args) => Console.WriteLine(args.Data);
```
mrsharm: See comment above for Console -> string output.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.MCPServer/CoreRun.cs
mrsharm: Add `[Description("...")]` to each of the parameters for better alignment.

mrsharm:
Add the ability to checkout different branches - this presumably builds only the branch we are currently git checkout'd. It'll be great to add `git checkout <branch>` and process those results.

Issue #4867, opened 2025-05-29
[Android] Memory consumption measurements fail on Android Pixel


kotlarmilos:
## Description

Memory consumption measurements fail on Android Pixel devices but work correctly on Galaxy devices.

```
Traceback (most recent call last):
  File "D:\h\w\92B60890\w\B45309B5\e\test.py", line 16, in <module>
    Runner(traits).run()
  File "D:\h\w\92B60890\p\shared\runner.py", line 491, in run
    raise Exception("Failed to capture the reported start time!")
Exception: Failed to capture the reported start time!
```

https://dev.azure.com/dnceng/internal/_build/results?buildId=2719218&view=logs&j=58e360a7-d025-59c0-3c7d-53ac133029b7

/cc: @LoopedBard3

matouskozak:
duplicated of https://github.com/dotnet/performance/issues/4826

PR #4945, opened 2025-09-05
[android] Disable MemoryConsumption except Blazor due to flakiness

matouskozak:
The memory consumption measurements are highly instable on all except MAUI blazor scenarios. Disabling these tests until https://github.com/dotnet/performance/issues/4867 resolved to improve CI monitoring.
 
+ Limiting build runs to only pixel devices (we don't need build time to run for both pixel/galaxy since it's not dependent on the connected tethered devices but the host)

Internal job: https://dev.azure.com/dnceng/internal/_build/results?buildId=2787121


LoopedBard3 (approved):
Looks good.

PR #4947, opened 2025-09-05
Use net9.0_18.5 version 18.5.9219 instead of 18.5.9223.

LoopedBard3:
Use net9.0_18.5 version 18.5.9219 instead of 18.5.9223. This should fix the following error we are hitting in our pipeline: `Workload installation failed: One or more errors occurred. (Version 18.5.9219 of package microsoft.ios.sdk.net9.0_18.5 is not found in NuGet feeds`


caaavik-msft (approved):

LoopedBard3:
Successfully built android in testing, need to make another PR to update the xcode version: https://dev.azure.com/dnceng/internal/_build/results?buildId=2787394&view=results

PR #4981, opened 2025-09-26
Python warning + code cleanup

caaavik-msft:
As I was writing a lot of python code for another feature, I was running into lots of python analysis warnings and errors, so I have gone through and resolved all of them in the `scripts` folder. There are still warnings in `src/scenarios` which can be tackled separately later if desired.

I have also updated the settings.json settings to setup better exclude settings to speed up code searches inside the repository, and enabled workspace-wide python analysis in strict mode so that you can see if any changes affect files you might not have open.

Some other changes in this:
- After inspecting all our Helix queues and deployed environments, I've identified that Python 3.9 is the oldest version that is used on any queues currently, so I have updated the minimum supported version and adopted some features that are now available where it makes sense.
- I've added a simplified version of the coreroot building that does not rely on building with `generatelayoutonly`, and just does the minimum folder copies. This version is not used anywhere in any pipelines, but is available for use by other tools that want to be able to build coreroots from cached builds to re-run adhoc performance tests.
- I've also fixed a bug with extracting from .tar.gz build caches
- I've removed some custom Action types which were no longer needed or could be replicated more simply using `type` instead of `action` in argparse. 
- I've added some common functions for running msbuild commands, and for evaluating property values from msbuild project files using /getProperty as an alternative to implementing custom XML parsing

### DIFF
--- a/scripts/performance/tracer.py
+++ b/scripts/performance/tracer.py
@@ -1,7 +1,7 @@
 import functools
 from collections.abc import Callable
 from logging import getLogger
-from typing import TYPE_CHECKING, Optional, TypeVar
+from typing import TYPE_CHECKING, Any, Optional, TypeVar, cast
 
 class TracingStateManager:
     '''A class to manage the state of tracing.'''
@@ -60,19 +60,8 @@ def is_console_exporter_enabled() -> bool:
     '''Return whether the console exporter has been enabled.'''
     return tracing_state_manager.get_console_exporter_enabled()
 
-# ParamSpec was added in Python 3.10, so we need to use typing_extensions for older versions.
-# But, to avoid needing to install this to run the script, we define a placeholder if not type checking.
-if TYPE_CHECKING:
-    from typing_extensions import ParamSpec
-    P = ParamSpec("P")
-else:
-    from typing import Any
-    class _ParamSpecPlaceholder:
-        args: Any = object()
-        kwargs: Any = object()
-    P = _ParamSpecPlaceholder()
+_F = TypeVar("_F", bound=Callable[..., Any])
 
-R = TypeVar("R")
 class AwareTracer:
     """
     A OpenTelemetry aware tracer implementation that is used as a wrapper for OpenTelemetry calls where OpenTelemetry is not guaranteed be installed.
@@ -80,7 +69,7 @@ class AwareTracer:
     """
     def __init__(self, name: str = "dotnet.performance") -> None:
         if TYPE_CHECKING:
-            from opentelemetry.trace import Tracer # pyright: ignore[reportMissingTypeStubs]
+            from opentelemetry.trace import Tracer
             self._tracer: Optional[Tracer]
             
         try:
@@ -90,28 +79,26 @@ class AwareTracer:
         else:
             self._tracer = trace.get_tracer(name)
 
-    def start_as_current_span(self, name: str) -> Callable[[Callable[P, R]], Callable[P, R]]:
+    def start_as_current_span(self, name: str)  -> Callable[[_F], _F]:
         """
         Decorator that starts a new span as the current span if OpenTelemetry is imported.
         If OpenTelemetry is not imported, the function is executed without starting a new span.
 
         Args:
-            *top_args: Variable length argument list that will be passed to OpenTelemetry Tracer.start_as_current_span.
-            **top_kwargs: Arbitrary keyword arguments.
+            name: The name of the span to start.
 
         Returns:
-            The result of executing the decorated function.
+            A decorator that preserves the original function's signature.
         """
-        def decorator(func: Callable[P, R]) -> Callable[P, R]:
+        def decorator(func: _F) -> _F:
             @functools.wraps(func)
-            def wrapped(*args: P.args, **kwargs: P.kwargs) -> R:
+            def wrapped(*args: Any, **kwargs: Any) -> Any:
                 if self._tracer is None:
                     return func(*args, **kwargs)
                 
                 with self._tracer.start_as_current_span(name):
                     return func(*args, **kwargs)
 
-            return wrapped
+            return cast(_F, wrapped)
 
         return decorator
-    
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -11,6 +11,7 @@
     "pythonVersion": "3.9",
     "pythonPlatform": "All",
     "deprecateTypingAliases": true,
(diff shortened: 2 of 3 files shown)


scripts/ci_setup.py
```diff
@@ -396,7 +393,7 @@ def main(args: Any):
 
     # When running on internal repos, the repository comes to us incorrectly
     # (ie https://github.com/dotnet-coreclr). Replace dashes with slashes in that case.
-    repo_url = None if use_core_sdk else args.repository.replace('-','/')
```
LoopedBard3, line 399: I think it would make sense to either remove this completely or continue to use `use_core_sdk`. There are only a few other places that use `use_core_sdk`, I have a minor lean towards keeping `use_core_sdk`.

scripts/upload.py
```diff
@@ -64,7 +63,13 @@ def upload(globpath: str, container: str, queue: str, storage_account_uri: str):
             upload_succeded = False
             with open(infile, "rb") as data:
                 try:
-                    retry_on_exception(lambda: blob_client.upload_blob(data, blob_type="BlockBlob", content_settings=ContentSettings(content_type="application/json")), raise_exceptions=[ResourceExistsError])
+                    def _upload():
```
LoopedBard3, line 66: Is this just a stylistic/readability thing or something more? I like the change either way.

scripts/performance/tracer.py
```diff
@@ -59,17 +59,37 @@ def is_console_exporter_enabled() -> bool:
     '''Return whether the console exporter has been enabled.'''
     return tracing_state_manager.get_console_exporter_enabled()
 
+# ParamSpec was added in Python 3.10, so we need to use typing_extensions for older versions.
```
LoopedBard3: How many machines do we have running python <3.10? Maybe we can use this for now but update all machines to latest python and remove it after?

scripts/performance/tracer.py
```diff
@@ -81,13 +101,20 @@ def start_as_current_span(self, *top_args, **top_kwargs):
...
-                        return func(*args, **kwargs)
-                else:
+        def decorator(func: Callable[P, R]) -> Callable[P, R]:
+            # If no tracer, just return the original function unchanged (best for typing & introspection).
+            if self._tracer is None:
+                return func
+
+            @functools.wraps(func)
+            def wrapped(*args: P.args, **kwargs: P.kwargs) -> R:
+                if self._tracer is None:
```
LoopedBard3, line 96: Do we need both self._tracer is None checks? If not, I think we should remove one of them (probably the one in wrapped).

LoopedBard3 (dismissed):
This looks great. Left a few comments/questions but nothing major.

scripts/ci_setup.py
caaavik-msft: The main reason this is done is just because the python linter is unable to verify that `use_core_sdk` being `False` implies that `args.repository` is not None. I could revert it, but I felt it made more sense to make this particular line conditional on `args.repository is None` since the logic isn't necessarily restricted to that use case.

scripts/upload.py
caaavik-msft: Yes and no, the line was really long before and there was a linter error on the `blob_client.upload_data` as it did not have proper type hints, so I did this so that I could have a targeted linter suppression on that line.

scripts/performance/tracer.py
caaavik-msft: The only one I found was the alpine queues

scripts/performance/tracer.py
caaavik-msft: I think I will remove the outer one actually, that way if a function is invoked before the tracer is initialised, it will still work later once the tracer is set up.

scripts/performance/tracer.py
LoopedBard3: I think we more or less removed everything alpine and no longer run anything on alpine (https://github.com/dotnet/performance/commit/0b0968439da1f3566137516cace35fbf430bec15) so we may be able to simplify this.

scripts/upload.py
LoopedBard3: Makes sense .👍

scripts/ci_setup.py
LoopedBard3: That makes sense. I am wondering if you have thoughts on just removing use_core_sdk all together and doing the same inline check at the 2 other locations it is? I think the point about the logic not necessarily needing to be restricted to core_sdk runs makes sense and can be applied to the other locations we use it.

scripts/performance/tracer.py
caaavik-msft: I'm in favour of moving this to a separate PR down the line, just in case I've missed anywhere else that needs python 3.9

LoopedBard3 (approved):

DrewScoggins (approved):

caaavik-msft:
There are some failures on the internal run as I need to pull in the latest from main, I'll do that, do another run and assuming that all looks good will merge this in.

caaavik-msft:
Completed internal run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2808128&view=results

DrewScoggins:
Did you take a look at the one failure, looks like the process got terminated before it finished. I assume that is unrelated, but wanted to make sure you had seen it.

scripts/performance/common.py
```diff
@@ -250,7 +251,7 @@ def set_environment_variable(name: str, value: str, save_to_pipeline: bool = Tru
         __write_pipeline_variable(name, value)
     os.environ[name] = value
 
-def run_msbuild_command(args: list[str], verbose: bool=True, warn_as_error: bool=True, perf_repo_dir: Optional[str] = None) -> str:
```
LoopedBard3: Is there a particular reason we reverted this change?

caaavik-msft: Yeah, if this function is called when the dotnet tools have not been set up yet, then it also includes all the tool setup stdout. This was breaking the `getProperty`. Rather than spend time making it more robust to ignore the tools setup stdout, I just reverted it. Could be useful in the future but not worth it for this.

caaavik-msft:
> Did you take a look at the one failure, looks like the process got terminated before it finished. I assume that is unrelated, but wanted to make sure you had seen it.

Yeah, these error codes where we get a 139 error are caused by crashes in coreclr itself. We've been getting them transiently on many of our runs.

PR #4984, opened 2025-09-26
Run iOS performance measurements on CoreCLR

kotlarmilos:
## Description

This PR runs ios_scenarios (build time, startup, and size)measurements for iOS on CoreCLR.

Depends on https://github.com/dotnet/runtime/pull/120139.

### DIFF
--- a/eng/pipelines/runtime-ios-scenarios-perf-jobs.yml
+++ b/eng/pipelines/runtime-ios-scenarios-perf-jobs.yml
@@ -7,6 +7,7 @@ jobs:
   - template: /eng/pipelines/performance/templates/perf-ios-scenarios-build-jobs.yml@${{ parameters.runtimeRepoAlias }}
     parameters:
       mono: true
+      coreclr: true
       nativeAot: true
 
   # run iOS scenarios - Mono FullAOT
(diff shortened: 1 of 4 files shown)


caaavik-msft (approved):

LoopedBard3 (approved):
Looks good to me.

eng/pipelines/runtime-ios-scenarios-perf-jobs.yml
```diff
@@ -138,3 +139,24 @@ jobs:
...
+  - template: /eng/pipelines/common/platform-matrix.yml@${{ parameters.runtimeRepoAlias }}
+    parameters:
+      jobTemplate: /eng/pipelines/templates/runtime-perf-job.yml@${{ parameters.performanceRepoAlias }}
+      buildConfig: release
+      runtimeFlavor: coreclr
+      platforms:
+        - osx_x64
+      jobParameters:
+        runtimeType: iOSCoreCLR
+        codeGenType: JIT
```
matouskozak: Shouldn't this be "interpreter"?

kotlarmilos: The interpreter is the only execution engine, so I initially referred to it as "JIT" and to R2R as "AOT." I agree that "interpreter" is the more accurate here.

matouskozak: Thank you. I think it would be best to name the configurations same as CoreCLR names them, so using R2R instead of mono "AOT". That would bring parity to the naming we started with Android on CoreCLR.

eng/pipelines/templates/runtime-perf-job.yml
```diff
@@ -195,6 +197,9 @@ jobs:
             ${{ if and(eq(parameters.runtimeType, 'iOSMono'), eq(parameters.iOSLlvmBuild, 'True'), eq(parameters.iOSStripSymbols, 'True')) }}:
               artifactName: 'iOSSampleAppLLVMNoSymbols'
               artifactFileName: 'iOSSampleAppLLVMNoSymbols.zip'
+            ${{ if and(eq(parameters.runtimeType, 'iOSCoreCLR'), eq(parameters.codeGenType, 'JIT')) }}:
```
matouskozak: Could we also update these...

There is a couple of `and(eq(parameters.runtimeType, 'iOSCoreCLR'), eq(parameters.codeGenType, 'JIT')` checks.

LoopedBard3 (approved):

PR #4991, opened 2025-09-30
Update Maui Dependencies to latest working versions

LoopedBard3:
Update Maui Dependencies to latest working versions.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2805464&view=results


LoopedBard3:
iOS is failing because xcode 26 needs to be selected (in current state). Will check to see if it is available.

matouskozak (approved):

PR #4997, opened 2025-10-01
Exclude files from `markdownlint` check

xtqqczze:
* Exclude `eng/common` since these files are from https://github.com/dotnet/arcade.
* Add a `<!-- markdownlint-disable -->` directive for files in `reports/net9to10/`


xtqqczze:
Superseded by https://github.com/dotnet/performance/pull/4998.

PR #4998, opened 2025-10-01
Fix `markdownlint` issues

xtqqczze:
Depends on https://github.com/dotnet/performance/pull/4997

PR #4999, opened 2025-10-01
Refactor `markdownlint` workflow

xtqqczze:
* Add `permissions: contents: read` to limit workflow permissions.
* Switch Node version to `lts/*` for automatic long-term support updates.
* Upgrade actions to latest releases.
* Split the Markdown linting step into separate install and run steps.
* Install `markdownlint-cli` with CI-friendly flags (`--no-audit` `--no-fund`).
* Pin `markdownlint-cli` to version `0.45.0` to avoid potential build breaks when new versions are released.
* Run `markdownlint` with an explicit rather than implicit configuration


xtqqczze:
@LoopedBard3 PTAL

xtqqczze:
@LoopedBard3 Thanks, I have a follow-up: https://github.com/dotnet/performance/pull/4998

PR #5012, opened 2025-10-13
Fix Sve benchmark warning and config

ylpoonlg:
* Suppress Sve experimental warning in project config.
I am not sure if it is possible to suppress warning on specific files instead of the whole project. However, since `SYSLIB5003` is specific to the SVE experimental API, it should only affect the benchmarks inside the `src/benchmarks/micro/sve` directory. 

* Remove redundant Sve.IsSupported checks as it is already checked in the benchmark filter config. SVE benchmarks will be filtered out when executing on non-SVE machines.

### DIFF
--- a/src/benchmarks/micro/MicroBenchmarks.csproj
+++ b/src/benchmarks/micro/MicroBenchmarks.csproj
@@ -10,6 +10,8 @@
     <NoWarn>$(NoWarn);CS8002</NoWarn>
     <!-- Suppress binaryformatter obsolete warning -->
     <NoWarn>$(NoWarn);SYSLIB0011</NoWarn>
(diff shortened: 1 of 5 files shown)


ylpoonlg:
@dotnet/arm64-contrib @SwapnilGaikwad

SwapnilGaikwad:
Hi @LoopedBard3, this is one of the first of the series of PRs that aim to improve SVE coverage using multiple benchmarks. Could you take a look, please?

LoopedBard3 (approved):

LoopedBard3:
Only a single partition of the 15 for this 9.0 microbenchmarks failed and it seems to be unrelated, merging.

PR #5024, opened 2025-10-27
Disable mobile runtime runs for older branches

LoopedBard3:
This disables the iOS and Android mobile runtime tests as part of the dotnet-runtime-perf pipeline for release branches. This will allow us to continue to get microbenchmark runs for the release branches without hitting issues due to new mobile configurations/updates. 

Test runs: 
* https://dev.azure.com/dnceng/internal/_build/results?buildId=2825877&view=results Demonstrates that this keeps mobile tests from running on release branch. 
* https://dev.azure.com/dnceng/internal/_build/results?buildId=2825876&view=results Demonstrates that this still runs mobile tests when running on main.

### DIFF
--- a/eng/pipelines/runtime-perf-jobs.yml
+++ b/eng/pipelines/runtime-perf-jobs.yml
@@ -58,7 +58,7 @@ jobs:
     parameters:
       perfBranch: ${{ parameters.perfBranch }}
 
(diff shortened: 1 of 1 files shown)


LoopedBard3:
Errors are unrelated.

PR #5025, opened 2025-10-30
Update DnsSafeHost test to be properly formatted

LoopedBard3:
Update DnsSafeHost test to be properly formatted by moving the scopeid inside the brackets and making sure it is escaped properly.

This test was broken after some correctness fixes: https://github.com/dotnet/runtime/pull/120908. Based on the docs for [System.Uri](https://learn.microsoft.com/en-us/dotnet/api/system.uri.dnssafehost?view=net-9.0), the correct format is an escaped string with the ScopeID inside the brackets.

@MihaZupan Just to double check, this seems to be a correct fix for the url?


DrewScoggins (approved):
LGTM

Issue #5055, opened 2025-11-28
Enable MAUI samples when TargetFramework=net11.0 becomes available


kotlarmilos:
## Description

Currently TargetFramework=net11.0 isn't supported in dotnet/dotnet yet. Once https://github.com/dotnet/dotnet/pull/3754 gets merged, it will take a few days to get dotnet/android and dotnet/macios working, then dotnet/maui.

## Tasks
 - [ ] Enable MAUI IOS samples
 - [ ] Enable MAUI Android samples

<details><summary>Logs</summary>
MAUI Android scenarios fail during the Run performance job step.

```
Failed to restore D:\a\1\s\performance\CorrelationStaging\payload\performance\src\scenarios\mauiandroid\app\MauiAndroidDefault.csproj (in 3.08 sec).
[2025/11/28 12:10:17][INFO]   [2025/11/28 12:10:17][INFO] $ popd
[2025/11/28 12:10:17][INFO]   [2025/11/28 12:10:17][ERROR] Process exited with status 1
[2025/11/28 12:10:17][INFO]   /p:UseMonoRuntime=false;/p:PublishReadyToRun=true;/p:PublishReadyToRunComposite=false
[2025/11/28 12:10:17][INFO]   Traceback (most recent call last):
[2025/11/28 12:10:17][INFO]     File "D:\a\1\s\performance\CorrelationStaging\payload\performance\src\scenarios\mauiandroid\pre.py", line 31, in <module>
[2025/11/28 12:10:17][INFO]       precommands.execute([])
[2025/11/28 12:10:17][INFO]       ~~~~~~~~~~~~~~~~~~~^^^^
[2025/11/28 12:10:17][INFO]     File "D:\a\1\s\performance\src\scenarios\shared\precommands.py", line 209, in execute
[2025/11/28 12:10:17][INFO]       self._restore()
[2025/11/28 12:10:17][INFO]       ~~~~~~~~~~~~~^^
[2025/11/28 12:10:17][INFO]     File "D:\a\1\s\performance\src\scenarios\shared\precommands.py", line 380, in _restore
[2025/11/28 12:10:17][INFO]       self.project.restore(packages_path=get_packages_directory(),
[2025/11/28 12:10:17][INFO]       ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[2025/11/28 12:10:17][INFO]                            verbose=True,
[...]

PR #5056, opened 2025-12-02
Update Maui runs to get NuGet.config and Version.Details.xml from maui repo

LoopedBard3:
Automate MAUI workload dependency management

Eliminates manual Version.Details.xml maintenance and NuGet maintenance for MAUI scenarios by 
downloading dependency configuration directly from upstream dotnet/maui 
repository. This ensures zero-maintenance automation - MAUI version updates 
no longer require changes to the performance repository.

Changes:
- Added MauiNuGetConfigContext: Context manager that temporarily merges 
  MAUI's package feeds into repo NuGet.config during build operations, 
  with automatic cleanup on exit
- Refactored generate_maui_rollback_dict(): Now downloads MAUI's 
  Version.Details.xml directly from GitHub instead of reading local file
- Updated all 7 MAUI scenario pre.py files (mauiandroid, mauiios, 
  mauiblazorandroid, mauiblazorios, mauiblazordesktop, 
  mauisamplecontentandroid, mauidesktop) to use context manager approach
- Added download_maui_nuget_config() helper to fetch upstream NuGet.config

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2853821&view=results

This currently works for our net10 tfm (main branch version) builds of maui android. Older frameworks are untested.
The change I am most unsure of is the update to the rollbacks band versioning as it is unclear if not having the Mapping_comments to set the bands will cause failure.

Follow-up:
* iOS Maui builds are still broken due to some earlier failure unrelated to Maui.
* It would be nice to have a way to run for specific Maui commits rather than just the latest on whatever branch. This would need to be piped through from the yaml for the jobs but doesn't seem like it would be that hard to add.

This should fix: 
https://github.com/dotnet/performance/issues/5055

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -39,5 +39,76 @@
       <Uri>https://github.com/dotnet/xharness</Uri>
       <Sha>d9aba438cae0eb814085c71dd6caa19e99c8e052</Sha>
     </Dependency>
+    <!--
+      Maui Rollback Version mappings, default means generated from sdk version. Allows for manual override of version similar to https://github.com/dotnet/maui/blob/df8cfcf635a590955a8cc3d6cf7ae6280449dd3f/eng/Versions.props#L138-L146, where the logic comes from:
+      Mapping_Microsoft.Maui.Controls:default
+      Mapping_Microsoft.NETCore.App.Ref:default
+      Mapping_Microsoft.NET.Workload.Emscripten.Current:default
+      Mapping_Microsoft.Android.Sdk:default
+      Mapping_Microsoft.MacCatalyst.Sdk:default
+      Mapping_Microsoft.macOS.Sdk:default
+      Mapping_Microsoft.iOS.Sdk:default
+      Mapping_Microsoft.tvOS.Sdk:default
+    -->
+    <!-- Dependencies for .NET MAUI workload -->
+    <Dependency Name="Microsoft.Maui.Controls" Version="10.0.0-ci.net10.25479.6">
+      <Sha>88c2c7ae772437921d85186318dd4320c4848618</Sha>
+      <Uri>https://github.com/dotnet/maui</Uri>
+    </Dependency>
+    <Dependency Name="Microsoft.NET.Sdk" Version="10.0.100-rc.2.25555.107">
+      <Sha>994d9ebe9f3f4216d7b125738cf5c19adaf674d3</Sha>
+      <Uri>https://github.com/dotnet/dotnet</Uri>
+    </Dependency>
+    <Dependency Name="Microsoft.NETCore.App.Ref" Version="10.0.0-rc.1.25555.107">
+      <Sha>994d9ebe9f3f4216d7b125738cf5c19adaf674d3</Sha>
+      <Uri>https://github.com/dotnet/dotnet</Uri>
+    </Dependency>
+    <Dependency Name="Microsoft.Android.Sdk.Windows" Version="36.0.9">
+      <Uri>https://github.com/dotnet/android</Uri>
+      <Sha>51a08d74f33040198641b6f0e1b0d2c9518d2ed7</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.NET.Workload.Emscripten.Current.Manifest-9.0.100" Version="9.0.1">
+      <Sha>4c9d1b112c16716c2479e054e9ad4db8b5b8c70c</Sha>
+      <Uri>https://github.com/dotnet/emsdk</Uri>
+    </Dependency>
+    <!-- Previous .NET Android version(s) -->
+    <Dependency Name="Microsoft.NET.Sdk.Android.Manifest-9.0.100" Version="35.0.105">
+      <Uri>https://github.com/dotnet/android</Uri>
+      <Sha>1dcfb6f8779c33b6f768c996495cb90ecd729329</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.MacCatalyst.Sdk.net10.0_18.5" Version="18.5.10824-net10-rc.2">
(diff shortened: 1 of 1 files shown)


src/scenarios/shared/mauisharedpython.py
```diff
@@ -139,6 +159,145 @@ def extract_latest_dotnet_feed_from_nuget_config(path: str, offset: int = 0) ->
...
+        with open(output_filename, "wb") as f:
+            with urllib.request.urlopen(url) as response:
+                f.write(response.read())
+        getLogger().info(f"Successfully downloaded MAUI NuGet.config to {output_filename}")
+        return os.path.abspath(output_filename)
+    except Exception as e:
+        getLogger().error(f"Failed to download MAUI NuGet.config: {e}")
+        raise
+
+class MauiNuGetConfigContext:
```
matouskozak, line 190: Just to understand, we download maui nuget.config and use that, so we don't need to have maestro subscription to MAUI repo anymore which is good. We will still need subscriptions to macios/android repos if they have dependencies which are not yet part of maui, correct?

src/scenarios/shared/mauisharedpython.py
```diff
@@ -14,12 +15,24 @@ def remove_aab_files(output_dir="."):
         if file.endswith(".aab"):
             os.remove(os.path.join(output_dir, file))  
 
-def generate_maui_rollback_dict():
-    # Generate and use rollback based on Version.Details.xml
-    # Generate the list of versions starts to get and the names to save them as in the rollback.
-    # These mapping values were taken from the previously generated rollback files for the maui workload. There should be at least one entry for each
-    # of the Maui Workload dependencies in the /eng/Version.Details.xml file, aside from Microsoft.NET.Sdk.
-    # If there are errors in the future, reach out to the maui team.
+def generate_maui_rollback_dict(target_framework: str):
```
matouskozak, line 18: I thought that this flow (from `install_versioned_maui`) is currently not active. Do we plan to switch back to this path?

src/scenarios/shared/mauisharedpython.py
```diff
@@ -139,6 +159,145 @@ def extract_latest_dotnet_feed_from_nuget_config(path: str, offset: int = 0) ->
...
+            output_filename: Name of the file to save the downloaded config
+    '''
+    # Extract base framework version (e.g., "net10.0" from "net10.0-android")
+    if '-' in target_framework:
+        target_framework_wo_platform = target_framework.split('-')[0]
+    else:
+        target_framework_wo_platform = target_framework
+    
+    url = f'https://raw.githubusercontent.com/dotnet/maui/{target_framework_wo_platform}/NuGet.config'
+    getLogger().info(f"Downloading MAUI NuGet.config from {url}")
```
jonathanpeppers, line 178: @PureWeen @rmarinho is the branch name on dotnet/maui changing?

This maps `$(TargetFramework)` == branch. Does it matter if they use dotnet/maui/main vs net10.0?

LoopedBard3: Looking forward as well, is the net11 work going to be done on main or a new branch?

src/scenarios/shared/mauisharedpython.py
```diff
@@ -14,12 +15,24 @@ def remove_aab_files(output_dir="."):
...
+    This eliminates the need to maintain MAUI dependencies in the performance repo's Version.Details.xml.
+    
+    Args:
+        target_framework: Target framework to determine which MAUI branch to use (e.g., "net10.0")
+    
+    Returns:
+        Dictionary mapping rollback package names to version/band strings
+    '''
+    # Extract base framework version
+    if '-' in target_framework:
```
kotlarmilos, line 30: @jonathanpeppers Are the main branch (.NET 10) and net10.0 in sync?

src/scenarios/shared/mauisharedpython.py
LoopedBard3: Correct, this flow is not currently active. I don't currently have any near-term plans to switch back to this path but figure it would be good to keep this up to date to make switching back cleaner if we ever need/want to.

src/scenarios/shared/mauisharedpython.py
jonathanpeppers: dotnet/maui/main is .NET 10. I don't know if they are using the `net10.0` branch now that it has released.

src/scenarios/shared/mauisharedpython.py
LoopedBard3: After discussing this, we want to keep the macios and android subscriptions so we are able to test the latest for them specifically and still have the feeds. This doesn't have a direct impact on this PR aside from reverting the removal of their dependency entries in Version.Details.xml.

src/scenarios/shared/mauisharedpython.py
matouskozak: @rmarinho which branch of dotnet/maui we should be using here (main vs net11/net10) so that we always have the latest `nuget.config`? (for context, we are building maui apps with the latest versions of .NET SDK, iOS/Android workloads + MAUI)

matouskozak (approved):
LGTM! Thanks. I would just wait for confirmation from MAUI team what branch should we keep targetting main vs net11.

src/scenarios/shared/mauisharedpython.py
PureWeen: main will stay net10.0 until net11.0 goes GA
all net11.0 development will stay on net11.0

Because of the high volume of SR work that MAUI does compared to other repositories and the need to keep community members productive. 

We keep main as always workable with public bits

LoopedBard3:
Going to go ahead and merge, and will make a new PR for updating to use the net 11 branch.

PR #5074, opened 2026-01-06
Handle case of null BytesAllocatedPerOperation in the json

BrzVlad:
This property is nullable on the BDN side: GetBytesAllocatedPerOperation -> https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Engines/GcStats.cs#L41. The ResultsComparer was crashing when trying to deserialize the json results.


DrewScoggins (approved):
LGTM

PR #5075, opened 2026-01-07
Fix Windows Helix work item exit code and prerequisite error handling

DrewScoggins:
- Use 'call deactivate' instead of 'deactivate' in PostCommands on Windows. Without 'call', the batch file transfers control to deactivate.bat and never returns, causing EXIT /b %_commandExitCode% to never execute.

- Chain Windows prerequisite install commands with && and add error checking. Previously, pip install failures were silently ignored, causing ModuleNotFoundError for azure packages during upload.

### DIFF
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -157,8 +157,10 @@ def get_pre_commands(
             ]
 
         # Clear the PYTHONPATH first so that modules installed elsewhere are not used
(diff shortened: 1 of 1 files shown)


matouskozak (dismissed):

LoopedBard3 (approved):
Looks good, should make sure to do a test run before merging.

LoopedBard3:
Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2875013&view=results.

PR #5078, opened 2026-01-12
Add bootstrap authentication for CertHelper

DrewScoggins:

### DIFF
--- a/src/tools/CertHelper/KeyVaultCert.cs
+++ b/src/tools/CertHelper/KeyVaultCert.cs
@@ -5,6 +5,7 @@ using Azure.Security.KeyVault.Certificates;
 using Azure.Security.KeyVault.Secrets;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Security.Cryptography.X509Certificates;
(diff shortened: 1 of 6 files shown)


src/tools/CertHelper/KeyVaultCert.cs
```diff
@@ -93,6 +159,18 @@ private async Task<X509Certificate2> FindCertificateInKeyVaultAsync(string certN
...
     public bool ShouldRotateCerts()
     {
+        // If we have no local certs or bootstrap was required, we need to rotate
+        if (LocalCerts.RequiresBootstrap || LocalCerts.Certificates.Count == 0)
+        {
+            return true;
+        }
+
+        // If local cert count doesn't match Key Vault cert count, rotate
+        if (LocalCerts.Certificates.Count != KeyVaultCertificates.Count)
```
LoopedBard3, line 169: Should this be the count or the thumbprint (or other identifier)?

DrewScoggins: This should be count, as it just means that what we have in the local store does not equal the amount that we have in the keyvault. This means something has gotten messed up, and we should rotate.

LoopedBard3: Ah, that makes sense.

LoopedBard3 (approved):

PR #5083, opened 2026-01-14
Modify unique name generation in upload.py

LoopedBard3:
Fixes a bug for really long file names where we would use only a random int in the name without including the unique id. This will cause a collision when trying to upload the file as, after enough random files are uploaded, the randint would be the same as one from a different job.

It is not clear whether we are actually hitting this at the moment.


DrewScoggins (approved):
LGTM

PR #5088, opened 2026-01-17
Set System.Net.SecurityProtocolType on Windows Prior to Executing Install Script

adamperlin:
`scripts/dotnet.py install` will fail on slightly older windows hosts due to the TLS defaulting to an older version. This PR changes the install command to set the version to what is required to execute `dotnet-install.ps1` (reference here: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script). Without this fix there does not seem to be an easy workaround to use `dotnet.py install` on systems without TLS1.2 set as default.

### DIFF
--- a/scripts/dotnet.py
+++ b/scripts/dotnet.py
@@ -846,7 +846,8 @@ def install(
         'powershell.exe',
         '-NoProfile',
         '-ExecutionPolicy', 'Bypass',
-        dotnetInstallScriptPath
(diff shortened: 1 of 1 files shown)


EgorBo:
@LoopedBard3 @DrewScoggins could you please approve? Without this it fails to acquire the SDK on our Windows CI machines.

PR #5095, opened 2026-01-26
Remove Tiger SDK jobs + reenable Windows CI

LoopedBard3:
Remove the Tiger SDK jobs (Windows and Ubuntu) as we have gotten a bit of overlap with the Viper machines. Also reenable previously disabled Windows CI jobs.


LoopedBard3:
Almost all failures are related to net10.0, will remove net10.0 builds in another PR shortly at which point we can more easily fix any other errors. Looking at the non-net10 micro windows error the run ran successfully and it was an issue with the send to helix connection. Merging.

PR #5100, opened 2026-01-27
Remove 10.0 runs from sdk jobs

LoopedBard3:
Remove 10.0 runs from sdk jobs as we are hitting failures due to new packages not yet being available in public feeds.


LoopedBard3:
Merged as removing the 10.0 runs will have no impact on the tests that would have been run.

PR #5104, opened 2026-02-04
Add back in dotnet 10 performance runs and update channel 10.0 quality

LoopedBard3:
Add back in dotnet 10 performance runs and set channel 10.0 quality to preview.


LoopedBard3:
The fix for this seems like it will be more complicated as just changing the quality results in 10.0..-rc2... being used.

PR #5106, opened 2026-02-05
Pin cryptography==46.0.3 to fix win_arm64 install failure

LoopedBard3:
cryptography 46.0.4 dropped official win_arm64 wheels from PyPI, causing pip to attempt a source build that fails on ARM64 machines. Pin to 46.0.3 (last version with win_arm64 wheels) and install it before the azure packages that depend on it.

Cypto PR disabling this wheel: https://github.com/pyca/cryptography/pull/14216

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2896696&view=results


LoopedBard3:
Failures are unrelated to this change and have been previously occurring. Merging to fix runs.

PR #5113, opened 2026-02-12
Switch public build jobs to NetCore-Public pool with custom images

akoeplinger:
The Azure Pipelines hosted pool is heavily congested and performance-ci is one of the heavy hitters. For ubuntu/windows-based builds we should use the `NetCore-Public` pool instead.

### DIFF
--- a/eng/pipelines/templates/build-machine-matrix.yml
+++ b/eng/pipelines/templates/build-machine-matrix.yml
@@ -10,8 +10,13 @@ jobs:
     parameters:
       osGroup: windows
       archType: x64
-      pool:
-        vmImage: windows-2025
+      ${{ if eq(parameters.isPublic, true) }}:
+        pool:
+          name: NetCore-Public
+          demands: ImageOverride -equals windows.vs2022.amd64.open
+      ${{ else }}:
+        pool:
+          vmImage: windows-2025
       ${{ insert }}: ${{ parameters.jobParameters }}
       ${{ if eq(parameters.isPublic, true) }}:
         osVersion: 24H2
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/build-machine-matrix.yml
```diff
@@ -10,8 +10,13 @@ jobs:
...
       archType: x64
-      pool:
-        vmImage: windows-2025
+      ${{ if eq(parameters.isPublic, true) }}:
+        pool:
+          name: NetCore-Public
+          demands: ImageOverride -equals windows.vs2022.amd64.open
+      ${{ else }}:
+        pool:
+          vmImage: windows-2025
```
akoeplinger, line 19: @LoopedBard3 @DrewScoggins ideally we'd do the same for internal builds too, I assume the actual image doesn't matter as much since the real work will be sent to Helix anyway?

akoeplinger:
The failure looks unrelated and happens in other PRs too

eng/pipelines/templates/build-machine-matrix.yml
DrewScoggins: The only question I would have is, can the public pools still push jobs to private Helix queues? I don't think that we need to worry about any secrets being pushed down because of the SFI work, but I don't know the answer on what Helix queues are accessible from which AzDO pools.

DrewScoggins: Clearly, this is not an issue for this PR.

akoeplinger: ah yeah sorry for not being clear, we'd obviously use internal pools. I was mainly commenting on the windows-2025 vs. windows.vs2022 image.

I'll open a new PR.

PR #5119, opened 2026-02-13
Fix startup scenario test failures on .NET 11 Preview 2

LoopedBard3:
- Update console template pre.py scripts to search for "Console.WriteLine" instead of the removed "// See https://aka.ms/new-console-template" comment
- Make insert_after() raise an exception when the search string is not found instead of silently succeeding
- Handle empty parser results in Reporter.WriteResultTable() to prevent InvalidOperationException on .Max() of empty sequence
- Add diagnostic logging and fail with exit code -1 when parser returns no results in Startup.cs

### DIFF
--- a/src/tools/Reporting/Reporting/Reporter.cs
+++ b/src/tools/Reporting/Reporting/Reporter.cs
@@ -75,7 +75,7 @@ public class Reporter
             if (!test.Counters.Any() || test.Counters.All(c => c.Results == null || c.Results.Count == 0))
             {
                 ret.AppendLine(test.Name);
-                ret.AppendLine("No results captured. The trace may not contain the expected events.");
+                ret.AppendLine("No results in file.");
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


src/tools/Reporting/Reporting/Reporter.cs
```diff
@@ -72,16 +72,27 @@ public string WriteResultTable()
         var ret = new StringBuilder();
         foreach (var test in Tests)
         {
+            if (!test.Counters.Any() || test.Counters.All(c => c.Results == null || c.Results.Count == 0))
```
DrewScoggins, line 75: We should add tests like the Copilot comment mentions. This fix is to better handle the failures where an ETW based test doesn't have any information, correct? If that is the case, and we are already doing checking at the Startup level, I think we should not do this here. Or at least not with this error message. We would be seeing this for our micros, I believe, and that isn't an issue we have been seeing.

LoopedBard3:
Failures are unrelated, merging.

PR #5128, opened 2026-02-26
Add latest version of iOS net10 package to nuget feed for maui runs.

LoopedBard3:
Add latest version of iOS net10 package to nuget feed for maui runs.

Fixes the following error we are hitting:
`Workload installation failed: One or more errors occurred. (Version 26.2.10216 of package microsoft.maccatalyst.sdk.net10.0_26.2 is not found in NuGet feeds `

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2913661&view=results


LoopedBard3:
Test run got past building Maui so we are good to merge.

PR #5131, opened 2026-02-27
Append perf repo branch name to PERFLAB_BRANCH when non-main

DrewScoggins:
Auto-detect the performance repo branch via git in run_performance_job.py. When the branch is not main, append it to PERFLAB_BRANCH (e.g., refs/heads/main becomes refs/heads/main-Foo when perf repo is on Foo).


LoopedBard3:
We should make sure to test this in an actual run before merging.

DrewScoggins:
Looks like it is working now.

`++ export PERFLAB_BRANCH=refs/heads/main-AddPerfRepoBranchName`

PR #5132, opened 2026-03-01
Fix invalid `csharp_space_around_declaration_statements` option value

xtqqczze:
The only valid option values are `ignore` and `false`.

https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/csharp-formatting-options#csharp_space_around_declaration_statements


LoopedBard3:
Thanks for this fix!

PR #5134, opened 2026-03-02
refactor: remove wasm args

FreyLuis:
This PR removes both wasmArgs.

` --module` is now added by BenchmarkDotNet when running on V8 (https://github.com/dotnet/BenchmarkDotNet/pull/3020).
 `--expose_wasm` is obsolete and was removed from V8 in [2024](https://github.com/v8/v8/commit/47766a75ec7bcd1a0d7fd86432a5b6e26aaf52fe).

### DIFF
--- a/scripts/benchmarks_local.py
+++ b/scripts/benchmarks_local.py
@@ -388,8 +388,6 @@ def generate_single_benchmark_ci_args(parsed_args: Namespace, specific_run_type:
     # for commit in all_commits: There is not a way to run multiple Wasm's at once via CI, instead will split single run vs multi-run scenarios
     elif specific_run_type == RunType.WasmInterpreter:
         benchmark_ci_args += ['--wasm', '--dotnet-path', os.path.join(get_run_artifact_path(parsed_args, RunType.WasmInterpreter, commit), "wasm_bundle", "dotnet")]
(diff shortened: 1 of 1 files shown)


FreyLuis:
@ilonatommy @timcassell Is there some manual part in publishing benchmarkdotnet nightlies, or have I missed something?

In the failing CI runs, it can't find the nightly that I've referenced, but it works for the scenario pipelines...
```
error NU1102:   - Found 75 version(s) in benchmark-dotnet-prerelease [ Nearest version: 0.16.0-nightly.20260226.451 ]
```

ilonatommy:
> @ilonatommy @timcassell Is there some manual part in publishing benchmarkdotnet nightlies, or have I missed something?

Yes, it's tricky. In my case I was getting the version from @LoopedBard3 and that one always worked properly.

LoopedBard3:
Yes, we have a mirror for nightlies that we have to manually update (but we are looking into options for automatic mirroring). I will get this version mirrored and retested today 👍.

FreyLuis:
Thank you

LoopedBard3:
The feed should be available, rerunning the checks.

LoopedBard3:
Hit some feed issues due to disabled feeds in the above run, so kicked off this run which rebases the PR with main: https://dev.azure.com/dnceng/internal/_build/results?buildId=2930147&view=results

PR #5137, opened 2026-03-03
Fix maui workload install + XCode 26.3 + strict success for iOS setup

matouskozak:
Internal run that got past the setup stage: https://dev.azure.com/dnceng/internal/_build/results?buildId=2917547


LoopedBard3:
Micro ubuntu failures are noisy unrelated failures (crypto error in VMs). Merging.

PR #5139, opened 2026-03-05
Add support for choosing config in run-perf-jobs.yml

DrewScoggins:


DrewScoggins:
We are still getting the same set of jobs created after this change. https://dev.azure.com/dnceng/internal/_build/results?buildId=2919350&view=results

DrewScoggins:
> Looks good, do we also want to condition the runtime builds based off of these switches as well? (Would be a separate PR).

Yeah, that seems like a good idea. Probably not too pressing since it would just be for manual runs, but seems like a good idea.

PR #5140, opened 2026-03-05
Add MemoryStream benchmarks

jozkee:
Related to https://github.com/dotnet/runtime/pull/124990.

### DIFF
--- a/src/benchmarks/micro/libraries/System.IO/MemoryStreamChunkedTests.cs
+++ b/src/benchmarks/micro/libraries/System.IO/MemoryStreamChunkedTests.cs
@@ -1,0 +1,150 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
+using MicroBenchmarks;
+
+namespace System.IO.Tests
+{
+    /// <summary>
+    /// Chunked / buffered operations. Fixed 64KB stream, chunk size controls call count.
+    /// Read/Write variants loop over the stream in ChunkSize increments.
+    /// CopyTo / CopyToAsync pass ChunkSize as the bufferSize parameter.
+    /// </summary>
+    [BenchmarkCategory(Categories.Libraries)]
+    public class MemoryStreamChunkedTests
+    {
+        private const int StreamSize = 65536;
+
+        private MemoryStream _stream;
+        private byte[] _buffer;
+
+        [Params(
+            1,      // 65536 calls, per-call overhead dominates
+            4096)]  // 16 calls, default StreamReader/BufferedStream buffer size
+        public int ChunkSize { get; set; }
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            byte[] backing = new byte[StreamSize];
+            new Random(42).NextBytes(backing);
+            _buffer = new byte[ChunkSize];
+            _stream = new MemoryStream(backing, writable: true);
+        }
+
+        [GlobalCleanup]
+        public void Cleanup()
+        {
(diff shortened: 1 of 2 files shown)


src/benchmarks/micro/libraries/System.IO/MemoryStreamTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        [Benchmark]
+        [MemoryRandomization]
+        public int ReadSpan()
+        {
+            MemoryStream s = _stream;
+            s.Position = 0;
+            int count = 0;
+            int n;
+            while ((n = s.Read(_buffer.AsSpan())) > 0)
```
stephentoub: I'm confused... isn't s backed by _buffer, such that not only is this reading into itself, this will only ever do a single read?

jozkee: Good point, _I think_ we should have two benchmark classes:
  Per-byte / no-user buffer ops — vary backing size (controls iteration count):

   - ReadByte, WriteByte, 
   - CopyTo, CopyToAsync - referring to the no-bufferSize overload.
   - Params: [1024, 65536] — small vs large stream

  Chunked / user-buffer ops — fixed large backing, vary chunk size (controls call count):

   - ReadByteArray, ReadSpan, ReadAsyncMemory, WriteByteArray, WriteSpan, WriteAsyncMemory
   - CopyTo, CopyToAsync - bufferSize overload.
   - Backing array 64KB stream, Params chunk: [1, 4096, 65536]

src/benchmarks/micro/libraries/System.IO/MemoryStreamTests.cs
```diff
@@ -0,0 +1,149 @@
...
+    [MemoryDiagnoser]
+    public class MemoryStreamTests
+    {
+        private MemoryStream _stream;
+        private byte[] _buffer;
+
+        [Params(
+            1,      // per-call overhead dominates
+            4096,   // default StreamReader/BufferedStream buffer size
+            65536)] // large bulk transfer
```
jkotas: > large bulk transfer

I do not think large bulk transfer is a useful microbenchmark. It is benchmarking memory copying. We have a ton of benchmarks for memory copying, and they are very noisy (sensitive to alignment and other factors). We do not need more of those.

jozkee: Fixed in https://github.com/dotnet/performance/pull/5140/changes/c4491cc2658f13bed5a1705732883bafc98b3e21.

src/benchmarks/micro/libraries/System.IO/MemoryStreamChunkedTests.cs
```diff
@@ -0,0 +1,150 @@
...
+    public class MemoryStreamChunkedTests
+    {
+        private const int StreamSize = 65536;
+
+        private MemoryStream _stream;
+        private byte[] _buffer;
+
+        [Params(
+            1,      // 65536 calls, per-call overhead dominates
+            4096)]  // 16 calls, default StreamReader/BufferedStream buffer size
```
jozkee, line 27: Do we want another param here, say `32_768, // 2 calls` or maybe 4 calls?

ViveliDuCh: I think it's enough with the number of calls you already have. Doing 2 calls looks similar to doing just one bulk transfer, where [this](https://github.com/dotnet/performance/pull/5140#discussion_r2892612045) feedback applies (mainly memcpy noise), and doing 4 calls seems to just benchmark `Buffer.BlockCopy` scaling (linear behavior) instead of exercising a different MemoryStream code path.

Good job on choosing the current parameter sizes.  Those values make sense to me. Thanks!

ViveliDuCh (approved):
LGTM. Thanks! @jozkee

PR #5141, opened 2026-03-05
Add DisableNewBranchLogic opt-out flag for branch-name override

LoopedBard3:
Add a queue-time variable DisableNewBranchLogic to the performance pipeline YAMLs. When set to 'true', it skips the branch-name append logic from #5131 so results always upload as main — allowing custom perf branches to still ingest against the main branch.

Pass the variable as DISABLE_NEW_BRANCH_LOGIC env var to the Python script, which checks it before applying the branch override.

Set it in the 'Run pipeline' UI under Variables. New behavior remains the default when the variable isn't set.


DrewScoggins (approved):
LGTM

PR #5146, opened 2026-03-10
Update iOS perf tests from iPhone 12 to iPhone 17

LoopedBard3:
Update logicalMachine from perfiphone12mini to perfiphone17 and Helix queue from OSX.13.Amd64.Iphone.Perf to Mac.iPhone.17.Perf across runtime-perf and dotnet-performance pipeline configurations.


LoopedBard3:
Errors appear to be unrelated, not sure where they are coming from so need to investigate.

PR #5148, opened 2026-03-11
Skip uploading results to perflab when onlySanityCheck is true

LoopedBard3:
When onlySanityCheck runs are internal, the --upload-to-perflab-container flag was still being passed to benchmarks_ci.py, causing sanity check results (~34 tests) to be uploaded as real performance data. This gates the upload flag on only_sanity_check being false for both microbenchmark and scenario run paths.


DrewScoggins (approved):
LGTM

PR #5154, opened 2026-03-17
Change base linux image to AL3

DrewScoggins:


LoopedBard3:
This is not needed anymore.

PR #5127, opened 2026-02-26
Add Cobalt SVE microbenchmarks run to runtime-perf-jobs

DrewScoggins:
Add a new coreclr micro run targeting the Azure Linux 3 Cobalt ARM64 queue with runCategories set to SVE instead of the default Runtime and Libraries categories. Also adds the arm64 build job dependency.

PR #5155, opened 2026-03-17
Auto-detect runtime_flavor for wasm_coreclr micro benchmarks

LoopedBard3:
For wasm_coreclr micro benchmark runs, runtime_flavor is not passed via YAML, causing RuntimeType to be set to 'None' in configurations. This means results never appear in test history or the perf autofiler.

Add auto-detection of runtime_flavor for wasm_coreclr micro benchmarks, setting it to 'coreclr', following the existing pattern used by Android and iOS scenarios. This fix does not modify wasm (mono) behavior to avoid breaking existing test history reports.

Fixes the same root cause as dotnet/performance#5153 but with a more targeted approach that only affects wasm_coreclr.

This will fix the RunConfigurations from saving as `{"CompilationMode":"wasm","RunKind":"micro","RuntimeType":"None"}` to saving as `{"CompilationMode":"wasm","RunKind":"micro","RuntimeType":"coreclr"}`

### DIFF
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -735,7 +735,8 @@ def run_performance_job(args: RunPerformanceJobArgs):
             raise Exception("iOS scenarios only support Mono and CoreCLR runtimes")
 
     if args.run_kind == "micro" and args.runtime_type == "wasm_coreclr":
-        args.runtime_flavor = "coreclr"
+        if not args.runtime_flavor:
+            args.runtime_flavor = "coreclr"
 
     branch = os.environ.get("BUILD_SOURCEBRANCH")
     cleaned_branch_name = "main"


pavelsavara:
@LoopedBard3 
At https://pvscmdupload.z22.web.core.windows.net/reports/allTestHistory/TestHistoryIndexIndex.html link for wasm/coreCLR leads to 404. Is this the right place where we should be looking ?

ilonatommy:
[This](https://dev.azure.com/dnceng/internal/_build/results?buildId=2929056&view=results) is the last successful run, there should be one measurement there already. All the newer jobs fail with:

```
Stage Build job run_performance_test_micro_coreclr_JIT_SVElinux__arm64_perfcobalt depends on unknown job build_linux_arm64_release_coreclr
```

since https://github.com/dotnet/performance/pull/5127 got merged. It requires changes to runtime 's yml: https://github.com/dotnet/runtime/pull/125719. Maybe more datapoints will help (?).

LoopedBard3:
Due to the reports only getting generated once a day at the moment, we probably just didn't have any runs with this change yet. We may also need a few runs for the report to properly generate. I will double check this tomorrow and if we still don't have results, I will investigate further.

pavelsavara:
Many thanks!
https://gist.github.com/pavelsavara/43ac82374efe95713b051ef8108f4501

PR #5161, opened 2026-03-18
Add maui desktop testing

LoopedBard3:

### DIFF
--- a/eng/pipelines/sdk-perf-jobs.yml
+++ b/eng/pipelines/sdk-perf-jobs.yml
@@ -93,6 +93,21 @@ jobs:
           ${{ each parameter in parameters.jobParameters }}:
             ${{ parameter.key }}: ${{ parameter.value }}
 
+  # MAUI Desktop BenchmarkDotNet benchmarks
+  - template: /eng/pipelines/templates/build-machine-matrix.yml
+    parameters:
+      jobTemplate: /eng/pipelines/templates/run-scenarios-job.yml
+      buildMachines:
+        - win-x64
+      isPublic: true
+      jobParameters:
+        runKind: maui_desktop_benchmarks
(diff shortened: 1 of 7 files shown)


matouskozak:
@simonrozsival @PureWeen Could you check if this benchark setup aligns with what you want for MAUI?

eng/pipelines/sdk-perf-jobs.yml
```diff
@@ -93,6 +93,21 @@ jobs:
           ${{ each parameter in parameters.jobParameters }}:
             ${{ parameter.key }}: ${{ parameter.value }}
 
+  # MAUI Desktop BenchmarkDotNet benchmarks
+  - template: /eng/pipelines/templates/build-machine-matrix.yml
+    parameters:
+      jobTemplate: /eng/pipelines/templates/run-scenarios-job.yml
+      buildMachines:
+        - win-x64
```
DrewScoggins, line 101: Why is one of these win-x64 and the other win-x64-viper?

DrewScoggins:
Do we have any test runs of this through AzDO? I want to verify that are runconfigs are correct for long-term storage.

eng/pipelines/sdk-perf-jobs.yml
LoopedBard3: `win-x64` supports public builds (`isPublic: true`) while `win-x64-viper` is gated to private builds (`not(eq(isPublic, true))`). We can probably update the machine list to remove tiger and move the private win-x64 runs to viper, but it currently runs as it.

LoopedBard3:
This test we are also able to do on public, so it does. The run configuration set is: `PERFLAB_CONFIGS=CompilationMode=Tiered;RunKind=maui_desktop_benchmarks` I am open to suggestions for this. Let me know if you still want an internal run 👍.

PR #5165, opened 2026-03-19
Add MAUI Android Inner Loop scenario

davidnguyen-tech:
## Summary

Adds a new scenario that measures MAUI Android developer inner loop performance:

- first build+deploy+startup time measurement
- 10 incremental build+deploy+startup time measurements
- parse build&deploy times from `.binglog`s.
- parse startups times from the Android activity manager output

The scenario is the first one in the repo to build the app on Helix.
We send the app source code as the Helix payload because of the nature of this scenario.

The incremental change is simulated by:

- one-line change in a .cs file
- one-line change in a .xml file

Sequential incremental inner loop measurements are implemented by switching between the updated and the original versions of the files.

## Test targets

| Device | Queue |
|--------|-------|
| Pixel 8 | `Windows.11.Amd64.Pixel.Perf` |
| Galaxy A16 | `Windows.11.Amd64.Galaxy.Lowend.Perf` |
| Android 36 Emulator | `Ubuntu.2204.Amd64.Android.36` |

## AzDO Build

https://dev.azure.com/dnceng/internal/_build/results?buildId=2984496&view=results


DrewScoggins:
When this PR goes in, https://github.com/dotnet/performance/pull/5176, you will need to take the changes. I am changing how the auth works and it needs the update to `upload.py` to continue working.

src/scenarios/mauiandroidinnerloop/pre.py
```diff
@@ -0,0 +1,210 @@
...
+    rollback_value = f"{latest['latestVersion']}/{latest['sdk_version']}"
+    rollback_dict = {workload: rollback_value}
+    logger.info(f"Rollback dictionary: {rollback_dict}")
+    with open("rollback_maui.json", "w", encoding="utf-8") as f:
+        f.write(json.dumps(rollback_dict, indent=4))
+    logger.info("Created rollback_maui.json file")
+
+    # Install maui-android (not 'maui') — works on both Windows and Linux
+    precommands.install_workload('maui-android', ['--from-rollback-file', 'rollback_maui.json'])
+    logger.info("########## Finished installing maui-android workload ##########")
```
matouskozak: Do we need this custom logic over re-using https://github.com/dotnet/performance/blob/619a1251b238ed7757a7acf461e66fb77c367bb0/src/scenarios/shared/mauisharedpython.py#L389?

eng/pipelines/sdk-perf-jobs.yml
```diff
@@ -527,6 +527,120 @@ jobs:
         ${{ each parameter in parameters.jobParameters }}:
           ${{ parameter.key }}: ${{ parameter.value }}
 
+  # Maui Android inner loop benchmarks on Pixel (Mono Default) - Debug
+  - template: /eng/pipelines/templates/build-machine-matrix.yml
+    parameters:
+      jobTemplate: /eng/pipelines/templates/run-scenarios-job.yml
+      buildMachines:
+        - win-x64-android-arm64-pixel
```
matouskozak, line 535: Can we make it simpler and pass it like this?

```suggestion
        - win-x64-android-arm64-pixel
        - win-x64-android-arm64-galaxy
```

eng/performance/maui_scenarios_android_innerloop.proj
```diff
@@ -0,0 +1,102 @@
...
+  <PropertyGroup>
+    <_WindowsEnvVars>set DOTNET_ROOT=%HELIX_CORRELATION_PAYLOAD%\dotnet;set DOTNET_CLI_TELEMETRY_OPTOUT=1;set DOTNET_MULTILEVEL_LOOKUP=0;set NUGET_PACKAGES=%HELIX_WORKITEM_ROOT%\.packages;set ANDROID_HOME=%HELIX_WORKITEM_ROOT%\android-sdk;set ANDROID_SDK_ROOT=%HELIX_WORKITEM_ROOT%\android-sdk;set JAVA_HOME=%HELIX_WORKITEM_ROOT%\jdk;set PATH=%HELIX_CORRELATION_PAYLOAD%\dotnet;;%HELIX_WORKITEM_ROOT%\android-sdk\platform-tools;;%HELIX_WORKITEM_ROOT%\jdk\bin;;%PATH%</_WindowsEnvVars>
+    <_LinuxEnvVars>export DOTNET_ROOT=$HELIX_CORRELATION_PAYLOAD/dotnet;export DOTNET_CLI_TELEMETRY_OPTOUT=1;export DOTNET_MULTILEVEL_LOOKUP=0;export NUGET_PACKAGES=$HELIX_WORKITEM_ROOT/.packages;export ANDROID_HOME=$HELIX_WORKITEM_ROOT/android-sdk;export ANDROID_SDK_ROOT=$HELIX_WORKITEM_ROOT/android-sdk;export JAVA_HOME=$HELIX_WORKITEM_ROOT/jdk;export ANDROID_SERIAL=emulator-5554;export PATH=$HELIX_CORRELATION_PAYLOAD/dotnet:$HELIX_WORKITEM_ROOT/android-sdk/platform-tools:$HELIX_WORKITEM_ROOT/jdk/bin:$PATH</_LinuxEnvVars>
+  </PropertyGroup>
+
+  <!-- Windows device track: runs on Windows.11.Amd64.Pixel.Perf -->
```
matouskozak: I thought we run on both Pixel and Galaxy device, correct?

davidnguyen-tech: Decided to remove the comments, it's unnecessary maintenance since it's all in `sdk-perf-jobs.yml` and `build-machine-matrix.yml`

eng/pipelines/sdk-perf-jobs.yml
davidnguyen-tech: A lot simpler, thanks :)

src/scenarios/mauiandroidinnerloop/pre.py
davidnguyen-tech: We can, I missed it - thanks.

davidnguyen-tech:
Hello @simonrozsival @jonathanpeppers , could you please check if I set up the MAUI SDK properly? Especially using the nightly builds & installing the workloads

jonathanpeppers (commented):
Why does this use `dotnet build -t:Install` + separate `am start` instead of `dotnet run`? `dotnet run` is the gesture customers actually use for inner loop -- it builds, deploys, and launches in one command. You could simply time the `dotnet run` command itself to get the true end-to-end inner loop time, rather than parsing multiple binlogs.

> RE: workloads

I don't see `dotnet workload config --update-mode manifests` being used, but it uses `--from-rollback-file rollback_maui.json`. Is there a log I can see, to just verify the right versions are installed?

jonathanpeppers:
You might need `dotnet run -p:WaitForExit=false -bl`.

davidnguyen-tech:
> Why does this use `dotnet build -t:Install` + separate `am start` instead of `dotnet run`? `dotnet run` is the gesture customers actually use for inner loop -- it builds, deploys, and launches in one command. You could simply time the `dotnet run` command itself to get the true end-to-end inner loop time, rather than parsing multiple binlogs.
> 

> You might need `dotnet run -p:WaitForExit=false -bl`.

I mirrored what we do for Visual Studio's F5 flow, based on `ClientTools.Platform`. I'm open to migrating to `dotnet run ...` though, as it sounds simpler :)

> 
> I don't see `dotnet workload config --update-mode manifests` being used, but it uses `--from-rollback-file rollback_maui.json`. Is there a log I can see, to just verify the right versions are installed?

You can get to the logs through here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2957445&view=logs&j=690a1b7a-7282-5287-d42e-e3e8ecc170c3&t=cb92986b-5679-58ac-2ba2-5fb303127ef9&l=77

jonathanpeppers:
Your workload install looks right:

```
Latest package: Microsoft.NET.Sdk.Android.Manifest-11.0.100-preview.4
Version=36.99.0-ci.main.112, SDK_Version=11.0.100-preview.4
```

eng/performance/maui_scenarios_android_innerloop.proj
```diff
@@ -0,0 +1,100 @@
...
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetsWindows)' == 'true'">
+    <HelixWorkItem Include="@(MAUIAndroidInnerLoopScenario -> 'Inner Loop Device - %(Identity)')">
+      <PreCommands>$(_WindowsEnvVars);$(Python) setup_helix.py $(PERFLAB_Framework)-android &quot;$(_MSBuildArgs)&quot;</PreCommands>
+      <Command>$(Python) test.py androidinnerloop --csproj-path app/MauiAndroidInnerLoop.csproj --edit-src &quot;src/MainPage.xaml.cs;src/MainPage.xaml&quot; --edit-dest &quot;app/Pages/MainPage.xaml.cs;app/Pages/MainPage.xaml&quot; --package-name com.companyname.mauiandroidinnerloop -f $(PERFLAB_Framework)-android -c Debug --msbuild-args &quot;$(_MSBuildArgs)&quot; --scenario-name &quot;%(Identity)&quot; --inner-loop-iterations $(InnerLoopIterations) --screen-timeout-ms $(ScreenTimeoutMs) $(ScenarioArgs)</Command>
```
LoopedBard3, line 84: Can we use $(BuildConfig) for the -c argument value? This should help ensure we don't accidentally continue to run Debug if we switch the sdk-perf-jobs.yml to release. I think doing this in a follow-up PR is also alright.

LoopedBard3 (dismissed):
Looks good to me and it seems the test run had only failed due to old upload/env setup. Just one small follow up comment. Let me know and I will merge this 👍.

eng/performance/maui_scenarios_android_innerloop.proj
jonathanpeppers: I think we'd want to always test inner loop in Debug -- I don't think we've ever tracked Release build times at all (as long as they are within reason!).

LoopedBard3: That makes sense, in which case I think we should either only run these if the BuildConfig is set to debug or throw an error or something if BuildConfig is not set to Debug just to ensure we don't get bit by thinking this works with release and then having to track down why the Debug vs Release labeled sdk jobs have similar results.

src/scenarios/shared/androidhelper.py
```diff
@@ -178,7 +187,85 @@ def setup_device(self, packagename: str, packagepath: str, animationsdisabled: b
         getActivity.run()
         getLogger().info(f"Target Activity {getActivity.stdout}")
 
-        # More setup stuff
+        self.ensure_screen_on()
+
+        self.activityname = getActivity.stdout.strip()
+
```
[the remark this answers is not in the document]
davidnguyen-tech: The scenarios succeed - resolving conversation

src/scenarios/mauiandroidinnerloop/setup_helix.py
```diff
@@ -0,0 +1,324 @@
...
+    # Use dotnet msbuild (not dotnet build) to run only this target
+    # without the full build pipeline.
+    result = run_cmd(
+        [ctx["dotnet_exe"], "msbuild", csproj,
+         "-t:InstallAndroidDependencies",
+         f"/p:AndroidSdkDirectory={android_home}",
+         f"/p:JavaSdkDirectory={java_home}",
+         "/p:AcceptAndroidSdkLicenses=True",
+         f"/p:TargetFramework={ctx['framework']}"],
+        check=False,
```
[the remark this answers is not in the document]
davidnguyen-tech: This target doesn't need the extra msbuild args

davidnguyen-tech:
Scheduled a new build with the latest changes:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2967606&view=results

eng/performance/maui_scenarios_android_innerloop.proj
davidnguyen-tech: I added an MSBuild error message, feel free to resolve this conversation :)

LoopedBard3:
The ValidateInnerLoopBuildConfig target added in 8ffbbf4a is firing because $(BuildConfig) isn't what it looks like by the time the Test target runs. In the helix flow, send_to_helix.py seems to overwrite the BuildConfig env var with a long composite identifier used for log-directory naming (e.g. x64_main_maui_scenarios_android_innerloop). So the proj sees that string instead of Debug and the error fires.

I think sticking to what was working before and not having extra validation is fine in this case. If we wanted a guard against a future flip, it needs to use a property that doesn't get clobbered (e.g., a new pipeline variable mirrored from ${{ parameters.buildConfig }} in run-performance-job.yml), but I don't think that is worth it for this at the moment.

davidnguyen-tech:
New build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2968356&view=results

davidnguyen-tech:
Another new build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2968449&view=results

eng/performance/maui_scenarios_android_innerloop.proj
```diff
@@ -0,0 +1,102 @@
+<Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">
```
kotlarmilos, line 1: This file is redundant with the existing `eng/performance/maui_scenarios_android.proj`. Could we just add the new HelixWorkItem to the `maui_scenarios_android.proj` and condition on the inner dev loop?

src/scenarios/mauiandroidinnerloop/pre.py
```diff
@@ -0,0 +1,131 @@
...
+
+    logger.info(f"Original .csproj content:\n{csproj_content}")
+
+    injected_props = {
+        # Preview SDKs may lack prune-package-data files, causing NETSDK1226.
+        'AllowMissingPrunePackageData': 'true',
+        # The perf repo globally disables the Roslyn compiler server to avoid
+        # BenchmarkDotNet file-locking issues. Re-enable it here to match real
+        # MAUI developer inner loop experience.
+        'UseSharedCompilation': 'true',
```
kotlarmilos, line 69: Is this known property?

src/scenarios/mauiandroidinnerloop/setup_helix.py
```diff
@@ -0,0 +1,324 @@
+#!/usr/bin/env python3
```
kotlarmilos, line 1: I think there is already infrastructure for running android apps on helix. Is this specific to inner dev loop? How helix interaction differs in this case?

src/tools/ScenarioMeasurement/Util/Parsers/AndroidInnerLoopParser.cs
```diff
@@ -0,0 +1,176 @@
...
+        var generateNativeApplicationConfigSourcesTimes = new List<double>();
+
+        // Deploy tasks
+        var fastDeployTimes = new List<double>();
+        var androidSignPackageTimes = new List<double>();
+        var androidApkSignerTimes = new List<double>();
+        var aapt2LinkTimes = new List<double>();
+
+        // Build targets
+        var coreCompileTargetTimes = new List<double>();
```
kotlarmilos, line 46: What is difference between build tasks and build targets? Do we measure msbuild and msbuild targets separately?

src/tools/ScenarioMeasurement/Util/Parsers/AndroidInnerLoopParser.cs
```diff
@@ -0,0 +1,176 @@
...
+
+/// <summary>
+/// Parses Android inner loop (build+deploy) target and task durations from a binary log file.
+/// </summary>
+public class AndroidInnerLoopParser : IParser
+{
+    public void EnableKernelProvider(ITraceSession kernel) { throw new NotImplementedException(); }
+    public void EnableUserProviders(ITraceSession user) { throw new NotImplementedException(); }
+
+    public IEnumerable<Counter> Parse(string binlogFile, string processName, IList<int> pids, string commandLine)
```
kotlarmilos, line 19: How do we plan to distinguish cold vs hot startup?

kotlarmilos (commented):
Overall looks good. Reusing the existing constructs should reduce the amount of code and make it easier for review and maintenance

src/scenarios/mauiandroidinnerloop/pre.py
davidnguyen-tech: Yes, we synced on it offline with @jonathanpeppers

src/scenarios/mauiandroidinnerloop/setup_helix.py
davidnguyen-tech: It is specific to the Inner Loop scenario - it is the only one that does both building and running the app on Helix.

The typical flow in the repo is to build on AzDO and run on Helix.

src/tools/ScenarioMeasurement/Util/Parsers/AndroidInnerLoopParser.cs
davidnguyen-tech: MSBuild Tasks are the individual steps of Targets. I chose to track both:

- Target timings enable us to track the main bottleneck areas
- Task timings give us granularity for more precise investigation

eng/performance/maui_scenarios_android_innerloop.proj
davidnguyen-tech: Yes, we could - I had a prototype for that approach and from ultra-high level, but the Inner Loop scenario seemed too different from the other ones and it looked approximately like this:

- if inner_loop:
  - ...
- else:
  - ...

Don't have a strong opinion on this, happy to switch to using just one `.proj` file :)

src/tools/ScenarioMeasurement/Util/Parsers/AndroidInnerLoopParser.cs
davidnguyen-tech: With my current understanding of Inner Loop = build --> deploy --> startup, I chose to always measure the cold startup. Just to be sure - @jonathanpeppers, is this correct?

Anyway - I think we can distinguish cold vs hot startup based on the output of `am`. Currently, we use that throw on non-cold states.

jonathanpeppers: As long as it is the "second run", we deployed a change and run the app again. That seems like what should be measured.

Cold vs hot is if the app has been launched more than once, but usually no code changes?

src/scenarios/mauiandroidinnerloop/pre.py
jonathanpeppers: This isn't needed inside VS, but is helpful for `dotnet run` / incremental builds. It may also be default when building a `.sln` file anyway, just not `.csproj` files.

src/tools/ScenarioMeasurement/Util/Parsers/AndroidInnerLoopParser.cs
davidnguyen-tech: > Cold vs hot is if the app has been launched more than once, but usually no code changes?

Yes that's what I expected. I'm not aware how HotReload affects this scenario when it comes to Hot startups, if at all?

eng/performance/maui_scenarios_android_innerloop.proj
kotlarmilos: The inner loop work item now lives in maui_scenarios_android.proj, gated on $(_IsInnerLoop) which is derived from $(RunKind)

kotlarmilos:
The PR is ready for review, validated end-to-end on the https://dev.azure.com/dnceng/internal/_build/results?buildId=2979643.

eng/performance/maui_scenarios_android.proj
```diff
@@ -116,6 +118,73 @@
     </HelixWorkItem>
   </ItemGroup>
 
+  <PropertyGroup Condition="'$(_IsInnerLoop)' == 'true'">
+    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'mono'">/p:UseMonoRuntime=true</_MSBuildArgs>
+    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'coreclr'">/p:UseMonoRuntime=false</_MSBuildArgs>
+    <!-- Disable ReadyToRun so inner loop measures JIT behavior. -->
+    <_MSBuildArgs Condition="'$(RuntimeFlavor)' == 'coreclr'">$(_MSBuildArgs);/p:PublishReadyToRun=false;/p:PublishReadyToRunComposite=false</_MSBuildArgs>
```
matouskozak: Since we're settings these here instead of using https://github.com/dotnet/performance/pull/5165/changes#diff-652d0d84aa2897441204ca7f3c3e5f8b3dd75fb44edaa69c2095a610270297d3L18-R28, maybe we should create a separate .proj file for the inner loop? I don't see many things which are re-used among the two scenarios and it would remove the ` <PropertyGroup Condition="'$(_IsInnerLoop)' == 'true'">` checks

matouskozak (dismissed):
Overall looking good to me

eng/performance/maui_scenarios_android.proj
davidnguyen-tech: I didn't have a strong opinion on this, but now I'm a little more inclined to the original approach of having a separate .proj file, since the scenario is so different from others.

kotlarmilos: Ok, reverted to the separate file.

kotlarmilos:
Validation build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2981677

davidnguyen-tech:
The scenario began failing on all targeted platforms. I will take it over from here. Thanks for the help @kotlarmilos!

davidnguyen-tech:
@LoopedBard3 @kotlarmilos @matouskozak the scenario is finally passing:

https://dev.azure.com/dnceng/internal/_build/results?buildId=2984496&view=results

scripts/run_performance_job.py
```diff
@@ -1160,6 +1160,7 @@ def publish_dotnet_app_to_payload(payload_dir_name: str, csproj_path: str, self_
             os.environ["RuntimeFlavor"] = args.runtime_flavor or ''
             os.environ["CodegenType"] = args.codegen_type or ''
             os.environ["BuildConfig"] = args.build_config or DEFAULT_BUILD_CONFIG
+            os.environ["RunKind"] = args.run_kind or ''
```
LoopedBard3: Same as other RunKind. (I think this is no longer needed as it was added to support _IsInnerLoop check when the tests were moved into the general maui android proj file (https://github.com/dotnet/performance/commit/362bf0006828ee81830f9ad927839f707f33795a), so we should remove this.)

eng/pipelines/templates/run-performance-job.yml
```diff
@@ -245,6 +245,8 @@ jobs:
         parameters:
           osGroup: ${{ parameters.osGroup }}
           projectFile: $(_projectFile)
+          environment:
+            RunKind: ${{ parameters.runKind }}
```
LoopedBard3: I think this is no longer needed as it was added to support _IsInnerLoop check when the tests were moved into the general maui android proj file (362bf00), so we should remove this.

LoopedBard3 (changes requested):
Two small changes, will approve and merge once in 👍.

davidnguyen-tech:
@LoopedBard3 fresh build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2985204&view=results

LoopedBard3 (approved):
Thanks for this new test flow!

PR #5175, opened 2026-03-26
Exclude wasm_coreclr and cobaltSveMicro perf jobs on release branches

LoopedBard3:
These jobs depend on runtime build jobs (build_browser_wasm_linux_release_wasm_coreclr and build_linux_arm64_release_coreclr) that were added after release/10.0 was split from main. Guard them with release branch checks matching the existing pattern used for iOS/Android scenarios.

Test run (release/10.0): https://dev.azure.com/dnceng/internal/_build/results?buildId=2936793&view=results (Was able to start the job, didn't have it actually run as this doesn't affect the actual runs). 

Also matched the changes for the new R2R interpreter runs.


LoopedBard3:
This doesn't touch anything that would be caught in the CI, so merging.

PR #5182, opened 2026-03-30
Disable PublishReadyToRunStripDebugInfo/StripInliningInfo for iOS CoreCLR R2R

kotlarmilos:
## Summary

Workaround for dotnet/sdk#53514 adding `--strip-debug-info` and `--strip-inlining-info` to crossgen2 args for Apple mobile RIDs before crossgen2 support (dotnet/runtime#124604) has flowed into the SDK.

This causes `CrossGen.targets` to fail with:
```
System.CommandLine.CommandLineException: No files matching --strip-debug-info --instruction-set:apple-m1
```

## Fix

Set `PublishReadyToRunStripDebugInfo=false` and `PublishReadyToRunStripInliningInfo=false` in iOS CoreCLR build args until the crossgen2 changes flow in.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2939329&view=results

### DIFF
--- a/eng/performance/maui_scenarios_ios.proj
+++ b/eng/performance/maui_scenarios_ios.proj
@@ -16,6 +16,9 @@
 
     <!-- Mono FullAOT and CoreCLR R2R are considered as default configurations, so no need for explicit properties -->
     
+    <!--  We have to set the following two variables until https://github.com/dotnet/runtime/pull/124604 flows to us -->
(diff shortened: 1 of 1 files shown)


LoopedBard3:
Looks good. CI failures are unrelated (instead related to https://github.com/dotnet/performance/pull/5179)

PR #5183, opened 2026-03-30
Fix SVE benchmark CreateWhileLessThanMask API rename for .NET 11

LoopedBard3:
### Problem

The .NET runtime renamed the `Sve.CreateWhileLessThanMask` methods from bit-width suffixes to type-name suffixes:
- `CreateWhileLessThanMask8Bit` → `CreateWhileLessThanMaskByte`
- `CreateWhileLessThanMask16Bit` → `CreateWhileLessThanMaskUInt16`
- `CreateWhileLessThanMask32Bit` → `CreateWhileLessThanMaskUInt32`
- `CreateWhileLessThanMask64Bit` → `CreateWhileLessThanMaskUInt64`
- New type-specific variants added: `CreateWhileLessThanMaskInt16`, `CreateWhileLessThanMaskInt32`, `CreateWhileLessThanMaskSingle`

The perf CI compiles benchmarks with the SDK (which still has old names in its ref assemblies) but runs them with a corerun built from runtime main (which only has new names). This causes `MissingMethodException` at runtime for all SVE benchmarks.

### Solution

Introduced `SveMaskHelper`, a compatibility shim that all 22 SVE benchmark files call instead of `Sve.CreateWhileLessThanMask*` directly. At startup, it uses reflection to detect which method names exist on the runtime, caches the result as delegates, and dispatches through those.

- **Direct-bind cases** (Byte, UInt16, UInt32, UInt64): old and new names share the same return type, so `Delegate.CreateDelegate` binds directly to whichever method is found.
- **Cast-wrapping cases** (Int16, Int32, Single): the old API returned an unsigned vector (e.g. `Vector<uint>`), but the caller needs a signed/float reinterpret. A thin lambda wraps the old delegate and uses `Unsafe.As` to reinterpret the bits.

### Performance impact

`CreateWhileLessThanMask` is loop predication plumbing, not the core workload being measured. The delegate indirection adds minor overhead (~2-5ns per call) but does not affect the JIT intrinsification of the actual SVE operations under test (Add, Multiply, LoadVector, CompareEqual, etc.). Once the SDK ships with the new names, this shim can be replaced with direct calls.

### Changes

- **New:** `src/benchmarks/micro/sve/SveMaskHelper.cs` — reflection + delegate caching shim
- **Modified:** 22 SVE benchmark files — replaced direct `Sve.CreateWhileLessThanMask*Bit()` calls with `SveMaskHelper.CreateWhileLessThanMask*()` calls

Seems to have been updated as part of https://github.com/dotnet/runtime/pull/124081.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2940679&view=results


LoopedBard3:
Test run was successful and this seems like a workable approach for these tests. Eventually we can split them into and old and new version (net11/net10), but this should keep it working until everything is flowed everywhere properly. When we change this test, it will cause changes in results, but we only have one config running these at the moment.

PR #5195, opened 2026-04-07
Enable runtime-async for net11.0+

agocke:


DrewScoggins:
Just so I understand, this will make it such that the performance tests will be compiled with the new async machinery? Along those lines, if these tests were run on a runtime that has its libraries built with the new runtime-async would the current repo, without this change, use the new async behavior or the old.

DrewScoggins:
We have the update in BDN now, https://github.com/dotnet/BenchmarkDotNet/pull/3078. I will take a look at the impact and we will get this update out as soon as we can.

agocke:
Is this ready to merge?

DrewScoggins:
Not yet. After the BDN update we are hitting a hang in CI, that I cannot reproduce locally. So work on it has been slow. I am still working through the issue, and will get this PR in ASAP.

agocke:
@DrewScoggins ping :)

DrewScoggins:
Still working on this. I have hit about three different bugs in BDN that were introduced by the recent async refactor. I think I finally got the last one fixed, but now we are seeing random test failures. I will give an update later today.

PR #5217, opened 2026-05-07
Update bdn 0.16 binarydata and async

DrewScoggins:
This change enables runtime-async. It is not on by default, but must be activated by an experiment. The runs will be added in a separate PR.

This change also includes updating BDN to latest as it was required for runtime-async.

### DIFF
--- a/src/harness/BenchmarkDotNet.Extensions/PerfLabExporter.cs
+++ b/src/harness/BenchmarkDotNet.Extensions/PerfLabExporter.cs
@@ -29,7 +29,7 @@ namespace BenchmarkDotNet.Extensions
 
         public async ValueTask ExportAsync(Summary summary, ILogger logger, CancellationToken cancellationToken)
         {
-            string jsonOutput = BuildJson(summary);
+            string? jsonOutput = BuildJson(summary);
             if (jsonOutput is null)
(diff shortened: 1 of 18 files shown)


LoopedBard3 (dismissed):
Looks good if the test runs succeed 👍.

src/harness/BenchmarkDotNet.Extensions/PerfLabExporter.cs
```diff
@@ -4,24 +4,67 @@
...
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace BenchmarkDotNet.Extensions
 {
-    public class PerfLabExporter : ExporterBase
+    // Implements IExporter directly (not ExporterBase) because BDN 0.16.0's new
+    // abstract ExporterBase.ExportAsync(Summary, CancelableStreamWriter, CancellationToken)
+    // takes a non-public CancelableStreamWriter, which we cannot satisfy from outside the
```
timcassell: `CancelableStreamWriter` is public, not sure why copilot thinks it's internal. Not that it really matters if you're just using `File.WriteAllTextAsync` anyway, but the comment is incorrect.

PR #5232, opened 2026-05-28
iOS device startup: upload .logarchive and pull crash reports on cleanup failure

matouskozak:
## Summary

When the post-iteration iOS app cleanup fails (devicectl returns 'No such process'), the app already exited before we could terminate it. This PR adds diagnostic data collection to help investigate why — crash, iOS-initiated termination, or natural exit.

## Changes

**`src/scenarios/shared/runner.py`:**
- Added `make_archive` to the `shutil` import
- Wrapped `killCmdCommand.run()` in a `try/except CalledProcessError` block that:
  1. Zips and uploads the iOS device's `.logarchive` to `HELIX_WORKITEM_UPLOAD_ROOT`
  2. Copies device-side crash reports (`.ips`) via `xcrun devicectl device copy from --domain-type systemCrashLogs`
  3. Re-raises the original exception after diagnostics are saved

## Why

- Adds observability for iOS device startup/cleanup failures — a daily triage pain point
- Investigative-only: triggers only on the cleanup failure that already terminates the iteration
- No behavior change for successful runs

## Origin

Cherry-picked from `matouskozak/ios-sdk-jobs-macos-26` (commits 71bfd4a1, 3101d30c) which were left unmerged after PR #5231 was merged.

## Verification

- ✅ No conflict markers
- ✅ `python3 -m py_compile src/scenarios/shared/runner.py` passes

### DIFF
--- a/src/scenarios/shared/runner.py
+++ b/src/scenarios/shared/runner.py
@@ -861,20 +861,39 @@ ex: C:\repos\performance;C:\repos\runtime
                                 make_archive(archive_base, 'zip', root_dir=logarchive_filename)
                             except Exception as upload_ex:
                                 getLogger().warning(f"Failed to save logarchive for diagnosis: {upload_ex}")
-                        # Pull any iOS crash reports (.ips) for our bundle from the device into the upload root.
-                        # The systemCrashLogs domain on devicectl exposes /var/mobile/Library/Logs/CrashReporter/.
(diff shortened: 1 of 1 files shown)


matouskozak:
test run https://dev.azure.com/dnceng/internal/_build/results?buildId=2986965

PR #5233, opened 2026-06-02
Average gcperfsim results

VincentBu:
This PR aims at calculating average value of multiple gcperfsim tests results.

### DIFF
--- a/src/benchmarks/gc/GC.Infrastructure/Configurations/GCPerfSim/LowVolatilityRuns.yaml
+++ b/src/benchmarks/gc/GC.Infrastructure/Configurations/GCPerfSim/LowVolatilityRuns.yaml
@@ -57,10 +57,12 @@ environment:
   iterations: 1
 coreruns:
   baseline:
+    is_baseline: true
     path: C:\CoreRuns\EmitEvent_Core_Root\corerun.exe
     environment_variables:
       DOTNET_GCName: clrgc.dll
   run:
+    is_baseline: false
     path: C:\CoreRuns\EmitEvent_Core_Root\corerun.exe
     environment_variables:
       DOTNET_GCName: clrgc.dll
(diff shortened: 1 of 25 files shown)


src/benchmarks/gc/GC.Infrastructure/Configurations/GCPerfSim/LowVolatilityRuns.yaml
```diff
@@ -57,10 +57,12 @@ environment:
   iterations: 1
 coreruns:
   baseline:
+    is_baseline: true
```
janvorli, line 60: Can we derive this from the parent element name (baseline / run) instead? Or are these names arbitrary?

VincentBu: > are these names arbitrary?

Indeed, changing the parent element name (baseline / run) won't block configuration parsing and test running. It only breaks the analysis.

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure.Core/Analysis/GCTraceMetricComparison.cs
```diff
@@ -1,8 +1,123 @@
...
+                {
+                    var runName = runKVP.Key;
+                    var pattern = OperatingSystem.IsWindows() ? $"{runName}.{corerunInfo.Name}.*.etl.zip" : $"{runName}.{corerunInfo.Name}.*.nettrace";
+                    var candidateFiles = Directory.GetFiles(outputPath, pattern, SearchOption.AllDirectories);
+                    foreach (var filePath in candidateFiles)
+                    {
+                        allTraceFiles.Add(Tuple.Create(runKVP, corerunInfo, filePath));
+                    }
+                });
+            });
```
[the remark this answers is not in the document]
VincentBu: by design

src/benchmarks/gc/GC.Infrastructure/GC.Infrastructure/Commands/RunCommand/BaseSuite/Microbenchmarks.yaml
```diff
@@ -1,7 +1,7 @@
 runs:
 
 microbenchmark_configurations:
-  bdn_arguments: "--warmupCount 1 --iterationCount 20 --allStats --outliers DontRemove --noOverheadEvaluation"
+  bdn_arguments: "--warmupCount 1 --iterationCount 20 --allStats --outliers DontRemove"
```
janvorli: I think we should replace it by the new form `--evaluateOverhead false` instead

janvorli (approved):
LGTM, thank you!

PR #5234, opened 2026-06-04
Fix GC aspnet test machines

janvorli:
The aspnet-citrine-win/lin machines don't exist anymore, move to the aspnet-gold-win/lin instead.
I've also removed couple of machines from the list that don't exist anymore.


janvorli:
cc: @VincentBu

mrsharm (approved):
LGTM

PR #5236, opened 2026-06-08
Fix GC performance infrastructure tests on Linux

janvorli:
There were couple of issues that prevented GC performance tests work on Linux:
* The trace capture needed to be passed the PID of the traced process due to the way dotnet-trace works
* The trace always contains only single process trace
* The benchmark name to trace file pattern mapping is different
* The extension of the trace file on linux is different and it was not handled at one place


janvorli:
The mapping of test names to trace filenames is fragile and now it turned out to be OS specific. It should be reimplemented in some other way, but not as part of this PR. That's why I have created a copy of the map and modified it for Linux as a quick workaround for it.

mrsharm (dismissed):

janvorli:
cc: @VincentBu

mrsharm (approved):

PR #5241, opened 2026-06-11
Generalize BCS upload + register-build templates with repoName parameter

LoopedBard3:
Make the build registration and upload steps generalized by repo for the BuildCacheService so that we can manage the steps in the dotnet/performance repo instead of requiring each setup we end up with having to duplicate the work. This keeps the defaults to runtime so that the runtime ymls continue working.

AI Summary:
Companion to a forthcoming dotnet/aspnetcore PR (LoopedBard3/aspnetcore#2 in fork-staging) that wires aspnetcore per-commit perf builds into BCS using the templates this PR generalizes.

## Why

Today the BCS-upload templates in `eng/pipelines/` are hard-tied to dotnet/runtime: both `templates/upload-build-artifacts-job.yml` and `templates/register-build-job.yml` write to literal `builds/runtime/...` blob paths. To let dotnet/aspnetcore (and any future tenant) consume the same primitives — and produce a BCS layout that's uniform with runtime's, so a single dashboard / cleanup tool / future indexer code path can serve both — we just need to parameterize the path prefix.

## What changes

`eng/pipelines/templates/upload-build-artifacts-job.yml` (per-artifact upload primitive):
- Adds `repoName` (default `'runtime'`, back-compat for existing runtime callers).
- Replaces both occurrences of literal `builds/runtime/` with `builds/${{ parameters.repoName }}/`.

`eng/pipelines/templates/register-build-job.yml` (per-config `buildInfo.json` marker):
- Adds `repoName` (default `'runtime'`).
- Replaces literal `builds/runtime/` with `builds/${{ parameters.repoName }}/`.

`eng/pipelines/upload-build-artifacts-jobs.yml` and `eng/pipelines/register-build-jobs.yml` (dispatchers):
- Forward `repoName` into each per-config invocation.
- The upload dispatcher additionally gains 5 new dispatch cases for the aspnetcore configKeys (`aspnetcore_{x64,arm64}_linux` + `aspnetcore_{x64,arm64,x86}_windows`), each depending on the corresponding `Windows_build` / `Linux_x64_build` / `Linux_arm64_build` job that the aspnetcore perf-build defines via `default-build.yml@self`. (RegisterBuild's dispatcher is buildType-agnostic by construction so it needs no new cases.)

## What we deliberately did NOT add

- **No `poolName` / `poolImage` / `poolOs` params.** An earlier draft of this PR added optional pool params to the upload leaf so a 1ES.Official-extending tenant could declare a per-job pool (1ES requires every job to have one). aspnetcore's caller in #2 dropped 1ES.Official to match runtime's plain non-1ES shape (private cache content, not signed/redistributed), so no current or planned tenant needs them. Removed to keep the surface area minimal; if a future 1ES tenant ever lands, re-adding the params is ~10 lines of YAML.
- **No `runtimeRepoAlias` / `mauiFramework` cleanup.** Both are pre-existing dead params on the upload dispatcher. Cleanup is split into a separate small PR (#5242) to keep this one focused on generalization.

## Validation

- YAML re-validated via `python -c "import yaml; yaml.safe_load(...)"` for all four files.
- Runtime callsite (`dotnet/runtime` `eng/pipelines/performance/perf-build.yml`) is unchanged — it calls both dispatchers without `repoName`, the default `'runtime'` kicks in, output is identical to today.
- Aspnetcore callsite (LoopedBard3/aspnetcore#2) passes `repoName: aspnetcore` to both dispatchers.

## Why a single shared template (not a fork)

Earlier plan iterations considered duplicating the upload logic in dotnet/aspnetcore. After reading runtime's actual `perf-build.yml` structure, it's clear runtime ALREADY consumes a thin-wrapper-from-dotnet-performance architecture (`register-build-jobs.yml@performance`, `runtime-perf-build-jobs.yml@performance`, `upload-build-artifacts-jobs.yml@performance`). Generalizing the existing primitives instead of forking keeps a single source of truth for the BCS contract, and means future BCS-shape changes land once for all tenants.

## Sequencing

This PR must merge upstream before the aspnetcore perf-build pipeline can run end-to-end — the aspnetcore-side YAML resolves the templates via `@performance` (= `internal/dotnet-performance` AzDO mirror), which only updates after dotnet/performance's main branch advances.

### DIFF
--- a/eng/pipelines/register-build-jobs.yml
+++ b/eng/pipelines/register-build-jobs.yml
@@ -3,6 +3,13 @@ parameters:
   performanceRepoAlias: self
   buildType: []
   mauiFramework: ''
+  # repoName is forwarded into each register-build-job invocation so the
+  # buildInfo.json marker lands under builds/{repoName}/buildArtifacts/...,
+  # matching the layout written by upload-build-artifacts-job.yml. Defaults
+  # to 'runtime' for back-compat with the existing dotnet/runtime perf-build
+  # pipeline; callers from other tenants (e.g. dotnet/aspnetcore) pass
+  # repoName explicitly.
+  repoName: 'runtime'
 
 jobs:
 - ${{ each type in parameters.buildType }}:
@@ -10,3 +17,4 @@ jobs:
     parameters:
       buildType: ${{ type }}
       buildId: $(Build.BuildId)
+      repoName: ${{ parameters.repoName }}
--- a/eng/pipelines/templates/register-build-job.yml
+++ b/eng/pipelines/templates/register-build-job.yml
@@ -1,6 +1,12 @@
 parameters:
   buildType: ''
   buildId: ''
+  # repoName is forwarded into the BCS blob path
+  # (builds/{repoName}/buildArtifacts/...). Defaults to 'runtime' for
+  # back-compat with the existing dotnet/runtime perf-build pipeline;
+  # callers from other tenants (e.g. dotnet/aspnetcore) pass repoName
+  # explicitly.
+  repoName: 'runtime'
 
 jobs:
   - job: RegisterBuild_${{ parameters.buildType }}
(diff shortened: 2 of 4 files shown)


DrewScoggins (dismissed):
LGTM

DrewScoggins (approved):
LGTM

PR #5250, opened 2026-06-30
Fix mlnet performance benchmark timeouts by pre-provisioning the SSWE model into the Helix payload

LoopedBard3:
## Problem

Every ML.NET performance benchmark in the `performance-ci` pipeline (public definitionId=38) is timing out. The whole `mlnet` work item hangs and is eventually killed at the work item timeout, discarding **all** ML.NET results, so every mlnet test shows as failed.

Example failing run: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1478345

## Root cause

The culprit is `StochasticDualCoordinateAscentClassifierBench.TrainSentiment`, which applies a pretrained SSWE word embedding. ML.NET downloads that model (`sentiment.emd`, ~70 MB) from `aka.ms/mlnet-resources` **at benchmark runtime** if it isn't already on disk. On the Helix machines that download now stalls (a hung connection, not a fast failure), so the benchmark hangs at `// BeforeActualRun` and the work item runs until it times out.

I reproduced this locally: with the model download blocked (stalling proxy) and the cache cleared, `TrainSentiment` hangs at `// BeforeActualRun` exactly like the Helix log; pre-provisioning the model + setting `MICROSOFTML_RESOURCE_PATH` runs it to completion with zero network. Updating `Microsoft.ML` (tested 5.0.0) does **not** help — the runtime download persists in all versions.

## Fix

In `scripts/run_performance_job.py`, for `run_kind == "mlnet"` only:

1. Download the SSWE model on the build agent (which has reliable connectivity) into the correlation payload at `mlnet-resources/Text/Sswe/sentiment.emd`, with retries and a blob→aka.ms fallback.
2. Set `MICROSOFTML_RESOURCE_PATH` to that payload dir via the Helix pre-commands, so ML.NET loads the embedding from disk and never makes the network call.

This is best-effort and strictly gated on `run_kind == "mlnet"`: non-mlnet runs are unaffected, and if the agent download fails it logs a warning and falls back to today's behavior.

## Why now?

The benchmark hasn't changed since 2019 and this isn't caused by any PR — the trigger is environmental on the download path (Helix egress/proxy/TLS tightening, blob throttling, and/or a .NET 9+ HttpClient behavior change against this endpoint — .NET 8 sometimes completed while 9.0/main/ubuntu hung). It manifests as a multi-hour timeout because it's a *stalled* read rather than a fast failure. Pre-staging the asset removes the dependency regardless of which is the actual culprit.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -150,8 +150,9 @@ def try_provision_mlnet_resources(payload_dir: str) -> bool:
         "https://aka.ms/mlnet-resources/Text/Sswe/sentiment.emd",
     ]
 
-    # The model is ~70 MB; require at least this much so a truncated/early-closed response (which may
-    # not raise) is rejected instead of leaving a corrupt file in the payload.
+    # The model is ~70 MB. When the server doesn't send a Content-Length to validate against, require
+    # at least this much so a truncated/early-closed response (which may not raise) is rejected
+    # instead of leaving a corrupt file in the payload.
     min_expected_size = 60 * 1024 * 1024
 
     last_error: Optional[Exception] = None
@@ -167,9 +168,14 @@ def try_provision_mlnet_resources(payload_dir: str) -> bool:
                         shutil.copyfileobj(response, f)
 
                 size = os.path.getsize(tmp_dest)
-                if expected_size is not None and size != expected_size:
-                    raise Exception(f"size {size} does not match Content-Length {expected_size}")
-                if size < min_expected_size:
+                if expected_size is not None:
+                    # Content-Length fully validates completeness, so trust it regardless of size
+                    # (the asset could legitimately shrink without becoming invalid).
+                    if size != expected_size:
+                        raise Exception(f"size {size} does not match Content-Length {expected_size}")
+                elif size < min_expected_size:
+                    # No Content-Length to validate against; fall back to a minimum-size floor to
+                    # reject an obviously truncated/early-closed response.
                     raise Exception(f"size {size} is smaller than the expected minimum {min_expected_size}")
 
                 os.replace(tmp_dest, dest)
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -156,12 +156,12 @@ def try_provision_mlnet_resources(payload_dir: str) -> bool:
     min_expected_size = 60 * 1024 * 1024
 
     last_error: Optional[Exception] = None
-    for attempt in range(1, 6):
+    for attempt in range(1, 4):
         for url in urls:
             tmp_dest = dest + ".tmp"
             try:
                 getLogger().info(f"Downloading ML.NET SSWE model from {url} (attempt {attempt})")
-                with urllib.request.urlopen(url, timeout=300) as response:
+                with urllib.request.urlopen(url, timeout=60) as response:
                     content_length = response.getheader("Content-Length")
                     expected_size = int(content_length) if content_length else None
                     with open(tmp_dest, "wb") as f:
--- a/scripts/run_performance_job.py
+++ b/scripts/run_performance_job.py
@@ -129,22 +129,15 @@ def try_provision_mlnet_resources(payload_dir: str) -> bool:
     """
     Pre-download the ML.NET SSWE word-embedding model into the correlation payload.
 
-    The Microsoft.ML.Benchmarks (StochasticDualCoordinateAscentClassifierBench.TrainSentiment)
-    apply a pretrained word embedding (SSWE 'sentiment.emd', ~70 MB). ML.NET downloads this model
-    from https://aka.ms/mlnet-resources at benchmark runtime if it isn't already present on disk.
-    On the Helix machines that download regularly stalls, which hangs the entire mlnet work item
-    until it hits the work item timeout and is killed, discarding ALL mlnet results (so every mlnet
-    benchmark appears to fail).
-
-    To remove the runtime network dependency, we download the model here on the build agent (which
-    has reliable connectivity) into the correlation payload. The caller then points
-    MICROSOFTML_RESOURCE_PATH at it on the Helix machine so ML.NET loads the model from disk and
-    never makes the network call.
-
+    StochasticDualCoordinateAscentClassifierBench.TrainSentiment applies a pretrained word embedding
+    ('sentiment.emd', ~70 MB) that ML.NET otherwise downloads from https://aka.ms/mlnet-resources at
+    benchmark runtime. That download stalls on the Helix machines and hangs the whole mlnet work item
+    until it times out. Downloading it here on the build agent (reliable connectivity) and pointing
+    MICROSOFTML_RESOURCE_PATH at <payload>/mlnet-resources lets ML.NET load it from disk instead.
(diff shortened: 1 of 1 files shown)


scripts/run_performance_job.py
```diff
@@ -120,6 +121,64 @@ class RunPerformanceJobArgs:
...
+        for url in urls:
+            try:
+                getLogger().info(f"Downloading ML.NET SSWE model from {url} (attempt {attempt})")
+                with urllib.request.urlopen(url, timeout=300) as response, open(dest, "wb") as f:
+                    shutil.copyfileobj(response, f)
+                size = os.path.getsize(dest)
+                if size <= 0:
+                    raise Exception("downloaded file is empty")
+                getLogger().info(f"Downloaded ML.NET SSWE model ({size} bytes) to {dest}")
+                return True
```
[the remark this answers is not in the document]
LoopedBard3: Good catch — addressed in 824daf2. The download now goes to a temp file, validates the size against Content-Length (when present) and a 60 MB minimum floor, then atomically replaces the destination via os.replace, removing the temp file on any failure. A truncated or early-closed response can no longer leave a corrupt sentiment.emd in the payload.

scripts/run_performance_job.py
```diff
@@ -120,6 +121,72 @@ class RunPerformanceJobArgs:
...
+                    content_length = response.getheader("Content-Length")
+                    expected_size = int(content_length) if content_length else None
+                    with open(tmp_dest, "wb") as f:
+                        shutil.copyfileobj(response, f)
+
+                size = os.path.getsize(tmp_dest)
+                if expected_size is not None and size != expected_size:
+                    raise Exception(f"size {size} does not match Content-Length {expected_size}")
+                if size < min_expected_size:
+                    raise Exception(f"size {size} is smaller than the expected minimum {min_expected_size}")
```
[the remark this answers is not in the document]
LoopedBard3: Good point — fixed in 5c0a1c2. The minimum-size floor now only applies when Content-Length is absent; when it's present, an exact match is the sole check, so a legitimately smaller asset won't be rejected.

scripts/run_performance_job.py
```diff
@@ -120,6 +121,78 @@ class RunPerformanceJobArgs:
...
+        "https://aka.ms/mlnet-resources/Text/Sswe/sentiment.emd",
+    ]
+
+    # The model is ~70 MB. When the server doesn't send a Content-Length to validate against, require
+    # at least this much so a truncated/early-closed response (which may not raise) is rejected
+    # instead of leaving a corrupt file in the payload.
+    min_expected_size = 60 * 1024 * 1024
+
+    last_error: Optional[Exception] = None
+    for attempt in range(1, 6):
```
DrewScoggins: Let's drop this down to three retries. I don't want to waste too much time.

scripts/run_performance_job.py
```diff
@@ -120,6 +121,78 @@ class RunPerformanceJobArgs:
...
+        for url in urls:
+            tmp_dest = dest + ".tmp"
+            try:
+                getLogger().info(f"Downloading ML.NET SSWE model from {url} (attempt {attempt})")
+                with urllib.request.urlopen(url, timeout=300) as response:
```
DrewScoggins: Do we need to have a 300 second timeout for this download? I assume that ~70MB shouldn't need a 5 minute timeout.

scripts/run_performance_job.py
LoopedBard3: Done in ace2169 — dropped to 3 retries.

scripts/run_performance_job.py
LoopedBard3: Agreed, reduced to 60s in ace2169. At ~70 MB that's ample headroom (the asset downloads in a few seconds on a healthy connection, and this is a per-socket-operation timeout, not a total-transfer budget).

scripts/run_performance_job.py
```diff
@@ -120,6 +121,78 @@ class RunPerformanceJobArgs:
...
+                    with open(tmp_dest, "wb") as f:
+                        shutil.copyfileobj(response, f)
+
+                size = os.path.getsize(tmp_dest)
+                if expected_size is not None:
+                    # Content-Length fully validates completeness, so trust it regardless of size
+                    # (the asset could legitimately shrink without becoming invalid).
+                    if size != expected_size:
+                        raise Exception(f"size {size} does not match Content-Length {expected_size}")
+                elif size < min_expected_size:
```
DrewScoggins: I think we should just fail here, rather then trying to do further validation based on size. Or if we are actually worried, we could just use some kind of hash of the file, because it should not be ever be changing.

LoopedBard3: Good call — switched to SHA256 validation in 44052f7. Since the model is a fixed asset, I download it, verify the hash against a known-good constant (a8062ef5...fefdd, verified identical from both the blob and aka.ms URLs), and only then atomically place it. This definitively rejects any truncated/corrupted download without relying on Content-Length or size floors. Added a note that the constant must be recomputed if the asset is ever intentionally updated.

DrewScoggins (dismissed):
LGTM

scripts/run_performance_job.py
```diff
@@ -1025,6 +1105,15 @@ def run_performance_job(args: RunPerformanceJobArgs):
...
     helix_post_commands = get_post_commands(args.os_group, args.internal, args.runtime_type)
 
+    # Point ML.NET at the SSWE model that was pre-downloaded into the correlation payload above, so it
+    # loads the word embedding from disk instead of downloading it at benchmark runtime (which hangs
+    # the work item on Helix). %HELIX_CORRELATION_PAYLOAD% is expanded by the shell at run time.
+    if mlnet_resources_provisioned:
+        if args.os_group == "windows":
+            helix_pre_commands += [f"set \"MICROSOFTML_RESOURCE_PATH=%HELIX_CORRELATION_PAYLOAD%\\{MLNET_RESOURCES_PAYLOAD_SUBDIR}\""]
+        else:
+            helix_pre_commands += [f"export MICROSOFTML_RESOURCE_PATH=$HELIX_CORRELATION_PAYLOAD/{MLNET_RESOURCES_PAYLOAD_SUBDIR}"]
```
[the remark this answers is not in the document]
LoopedBard3: Fixed in 111fff0 — the export value is now quoted so a path with spaces or shell metacharacters won't be word-split. (The Windows line already uses the set \VAR=value\ quoting form.)

scripts/run_performance_job.py
```diff
@@ -120,6 +122,76 @@ class RunPerformanceJobArgs:
...
+
+                os.replace(tmp_dest, dest)
+                getLogger().info(f"Downloaded and verified ML.NET SSWE model to {dest}")
+                return True
+            except Exception as e:
+                last_error = e
+                getLogger().warning(f"Failed to download ML.NET SSWE model from {url}: {e}")
+                if os.path.exists(tmp_dest):
+                    os.remove(tmp_dest)
+        time.sleep(10)
```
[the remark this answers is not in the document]
LoopedBard3: Fixed in 111fff0 — now only sleeps between attempts, not after the final one.

DrewScoggins (approved):
LGTM

PR #5257, opened 2026-07-09
Add file-backed cert store so CertHelper works on Mac

DrewScoggins:
On macOS the CurrentUser\My X509Store maps to the login Keychain, which cannot export a private key back out without an interactive password prompt. As a result the iphone/Mac queues bypassed CertHelper entirely and relied on hand-provisioned static PFX files that were never rotated.

Introduce an ILocalCertStore abstraction with two implementations:
- X509LocalCertStore: existing Windows/Linux behavior over the OS store.
- FileLocalCertStore: persists PFX files (thumbprint-named) in a known directory on Mac, loaded Exportable so no Keychain round-trip is needed.

A factory selects the implementation by OS. LocalCert now loads through the abstraction, and Program.Main rotates via ILocalCertStore.Rotate and exports the in-memory Key Vault certificates to stdout instead of round-tripping through the OS store. common.py's get_certificates now runs CertHelper on all platforms (removing the Mac static-file path), so Mac hosts get the same bootstrap + Key Vault rotation as Linux.

Update LocalCertTests to mock ILocalCertStore and add FileLocalCertStore round-trip tests.

### DIFF
--- a/scripts/performance/common.py
+++ b/scripts/performance/common.py
@@ -210,24 +210,17 @@ def retry_on_exception(
 
 def get_certificates() -> list[str]:
     '''
-    Gets the certificates from the certhelper tool and on Mac uses find-certificate.
+    Gets the certificates from the certhelper tool. CertHelper handles per-OS certificate
+    storage internally (X509 store on Windows/Linux, file-backed store on Mac).
     '''
-    if ismac():
-        certs: list[str] = []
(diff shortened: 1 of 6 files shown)


LoopedBard3 (dismissed):
Looks good, copilot may have some good feedback though.

LoopedBard3 (approved):

PR #5259, opened 2026-07-10
Load macOS Key Vault certs as Exportable only so private key survives…

DrewScoggins:
… export

On macOS, loading the Key Vault PKCS#12 with PersistKeySet imports the private key into the login Keychain, which cannot export the private key back out. That produced public-only PFX exports (~half the expected size) with no private key, breaking both the file-backed cert store and the base64 certs written to stdout.

Select storage flags by OS: use Exportable only on macOS (no PersistKeySet) so the key stays in memory and can be exported to the file-backed store; keep Exportable | PersistKeySet on Windows/Linux.


LoopedBard3 (approved):

matouskozak (approved):
Thank you for the quick fix

PR #5261, opened 2026-07-14
Download version-pinned NuGet.config for MAUI workload install

matouskozak:
## Summary

Instead of downloading the latest `NuGet.config` from branch HEAD of upstream repos (`dotnet/maui`, `dotnet/android`, `dotnet/macios`), extract the commit SHA from the NuGet package `.nuspec` metadata and download from the exact commit that produced the workload package.

## Problem

`MauiNuGetConfigContext` downloads `NuGet.config` files from upstream repos using branch HEAD. If the upstream repo has moved ahead since the workload package was published, the `NuGet.config` may reference feeds that don't match the packages being installed.

## Solution

NuGet packages embed `<repository type="git" url="..." commit="..." />` in their `.nuspec` metadata. We use this to:

1. Query the NuGet feed for the latest manifest package per upstream repo (same feed used for workload install)
2. Download the `.nuspec` from the NuGet flat container API
3. Extract the exact commit SHA from the `<repository>` element
4. Download `NuGet.config` from that commit instead of branch HEAD

### New functions
- `_find_latest_manifest_package` — shared filtering logic (extracted from `install_latest_maui`)
- `_get_nuget_flat_container_base` — gets flat container URL from NuGet V3 service index
- `_get_commit_sha_from_nuspec` — downloads `.nuspec`, parses `<repository>` element
- `_discover_repo_commits` — orchestrates per-repo commit SHA discovery

### Modified functions
- `_get_repo_nuget_config_url` / `download_repo_nuget_config` — accept optional `commit_sha` parameter
- `MauiNuGetConfigContext.__enter__` — calls `_discover_repo_commits` before downloading configs
- `install_latest_maui` — refactored to use shared `_find_latest_manifest_package` helper

## Fallback

Any failure in commit discovery silently falls back to the current branch-HEAD behavior. All failures are logged as warnings, never fatal.

## Validation

Tested on https://dev.azure.com/dnceng/internal/_build/results?buildId=3022990&view=results— commit SHAs are correctly discovered and used for NuGet.config downloads.

### DIFF
--- a/src/scenarios/shared/mauisharedpython.py
+++ b/src/scenarios/shared/mauisharedpython.py
@@ -160,6 +160,180 @@ def extract_latest_dotnet_feed_from_nuget_config(path: str, offset: int = 0) -> 
 
     return target_feed
 
+def _find_latest_manifest_package(packages: list[dict], workload_name: str) -> dict:
+    '''
+    Apply the same filtering logic as install_latest_maui to find the latest manifest package.
+    This ensures _discover_repo_commits picks the same package that will actually be installed.
+    '''
+    pattern = r'Microsoft\.NET\.Sdk\..*\.Manifest\-\d+\.\d+\.\d+(\-(preview|rc|alpha)\.\d+)?$'
+    packages = [pkg for pkg in packages if re.match(pattern, pkg['id'])]
+    if not packages:
+        raise Exception(f"No manifest packages found for {workload_name} after pattern filtering")
+
+    for package in packages:
+        match = re.search(r'Manifest-(.+)$', package["id"])
+        if match:
+            sdk_version = match.group(1)
+            package['sdk_version'] = sdk_version
+            match = re.search(r'^\d+\.\d+', sdk_version)
+            if match:
+                package['dotnet_version'] = match.group(0)
+            else:
+                raise Exception(f"Unable to find .NET version in SDK version '{sdk_version}' for package {package['id']}")
+        else:
+            raise Exception(f"Unable to find .NET SDK version in package ID: {package['id']}")
+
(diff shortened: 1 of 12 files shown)


LoopedBard3 (dismissed):

src/scenarios/shared/mauisharedpython.py
```diff
@@ -160,6 +160,170 @@ def extract_latest_dotnet_feed_from_nuget_config(path: str, offset: int = 0) ->
 
     return target_feed
 
+def _find_latest_manifest_package(packages: list[dict], workload_name: str) -> dict:
```
kotlarmilos, line 163: Why dotnet install --prerelease doesn't work?

matouskozak: Could you share the command? I only know about https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-workload-install with `--include-previews` which doesn't pick the latest builds (only final previews).

kotlarmilos: Right I meant dotnet workload install maui --include-previews --source <feed>. Since that does not select the latest daily builds, could we simplify this by reusing the manifest versions already selected for rollback_maui.json when resolving the commit SHAs instead of querying the feed again?

matouskozak: Good point, I've added a cache so that the found packages are kept in memory.

For context, the flow is reverse, we first need to resolve the commit SHAs so that we can prepare the correct nuget.config before we use rollback_maui.json to install the workloads.

kotlarmilos (approved):

PR #5272, opened 2026-08-05
Add stable .NET 10 mobile SDK baseline runs

matouskozak:
## Summary

- add Android and iOS Release baseline jobs for the latest stable .NET 10 SDK
- run the .NET 10 default mobile configuration with Mono, with comments to switch to CoreCLR when the baseline moves to .NET 11
- resolve stable workloads through normal workload installation while preserving the existing latest-prerelease workload path
- update the stale .NET 10 Arcade and Helix SDK pins

## Validation

- [.NET 10-only mobile validation](https://dev.azure.com/dnceng/internal/_build/results?buildId=3039925) — Pixel, Galaxy, and iPhone jobs succeeded; Kusto ingestion verified
- [Full mobile Release comparison](https://dev.azure.com/dnceng/internal/_build/results?buildId=3040020) — all 14 net10/net11 Release mobile jobs succeeded

### DIFF
--- a/eng/pipelines/sdk-perf-jobs.yml
+++ b/eng/pipelines/sdk-perf-jobs.yml
@@ -440,7 +440,7 @@ jobs:
         projectFileName: maui_scenarios_android.proj
         channels:
           - 10.0
-        runtimeFlavor: default
+        runtimeFlavor: mono # Update to coreclr when the baseline moves to .NET 11.
         codeGenType: Default
         buildConfig: Release
         variables:
@@ -561,7 +561,7 @@ jobs:
         projectFileName: maui_scenarios_ios.proj
         channels:
           - 10.0
-        runtimeFlavor: default
+        runtimeFlavor: mono # Update to coreclr when the baseline moves to .NET 11.
         codeGenType: Default
         buildConfig: Release
         variables:
--- a/eng/pipelines/sdk-perf-jobs.yml
+++ b/eng/pipelines/sdk-perf-jobs.yml
@@ -440,9 +440,9 @@ jobs:
         projectFileName: maui_scenarios_android.proj
         channels:
           - 10.0
+        runtimeFlavor: default
         codeGenType: Default
         buildConfig: Release
-        additionalJobIdentifier: Baseline
         variables:
           - name: MAUI_WORKLOAD_MODE
             value: sdk-default
@@ -561,9 +561,9 @@ jobs:
         projectFileName: maui_scenarios_ios.proj
         channels:
           - 10.0
+        runtimeFlavor: default
         codeGenType: Default
         buildConfig: Release
-        additionalJobIdentifier: Baseline
         variables:
           - name: MAUI_WORKLOAD_MODE
             value: sdk-default
--- a/eng/pipelines/sdk-perf-jobs.yml
+++ b/eng/pipelines/sdk-perf-jobs.yml
@@ -444,8 +444,8 @@ jobs:
         codeGenType: Default
         buildConfig: Release
         variables:
-          - name: MAUI_WORKLOAD_VERSION
-            value: 10.0.302.1
+          - name: MAUI_WORKLOAD_MODE
+            value: stable
         ${{ each parameter in parameters.jobParameters }}:
           ${{ parameter.key }}: ${{ parameter.value }}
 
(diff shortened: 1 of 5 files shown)


global.net10.json
```diff
@@ -8,8 +8,8 @@
     "dotnet": "10.0.100"
   },
   "msbuild-sdks": {
-    "Microsoft.DotNet.Arcade.Sdk": "10.0.0-beta.25604.105",
-    "Microsoft.DotNet.Helix.Sdk": "10.0.0-beta.25604.105"
+    "Microsoft.DotNet.Arcade.Sdk": "10.0.0-beta.25605.3",
+    "Microsoft.DotNet.Helix.Sdk": "10.0.0-beta.25605.3"
```
matouskozak: @LoopedBard3 Is it safe to bump these?

LoopedBard3: Yes, because it is not just named global.json, maestro doesn't check this file, but updating them to whatever latest is should be fine. A more recent version for each seems to be 10.0.0-beta.26376.103.

LoopedBard3 (dismissed):

LoopedBard3:
Just to double check, we are all good with these appearing on the same .NET 10 history trends we previously had? Not having anything in the test name or RunConfig will mean these should flow the same way a non-stable run would. I think that is fine because we currently will only have stable set running .NET 10 and we are running the same tests anyways. Just something to keep in mind if we ever wanted stable trends vs non-stable trends.

src/scenarios/shared/mauisharedpython.py
```diff
@@ -491,6 +498,10 @@ def _merge_sources_from_config(config_path: str, repo_sources: ET.Element, exist
         return added
 
     def __enter__(self):
+        if self.use_stable_maui_workload:
+            getLogger().info("Using the latest stable MAUI workload")
+            return self
```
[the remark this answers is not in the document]
matouskozak: Stable mode intentionally calls `dotnet workload install maui` with an empty argument list: there is no pinned `--version` and, importantly, no default `--skip-manifest-update`. It therefore updates manifests and resolves the latest stable workload available to the selected stable SDK/channel.

eng/pipelines/sdk-perf-jobs.yml
```diff
@@ -427,6 +427,28 @@ jobs:
...
+        projectFileName: maui_scenarios_android.proj
+        channels:
+          - 10.0
+        runtimeFlavor: mono # Update to coreclr when the baseline moves to .NET 11.
+        codeGenType: Default
+        buildConfig: Release
+        variables:
+          - name: MAUI_WORKLOAD_MODE
+            value: stable
+        ${{ each parameter in parameters.jobParameters }}:
```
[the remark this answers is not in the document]
matouskozak: `variables` is a parameter on `run-performance-job.yml`, and lines 136-137 explicitly insert every entry into the Azure Pipelines job-level `variables` collection. Azure exposes those variables to the `Run performance job script` process as environment variables. The testing confirmed that his should work.

matouskozak:
> Just to double check, we are all good with these appearing on the same .NET 10 history trends we previously had? Not having anything in the test name or RunConfig will mean these should flow the same way a non-stable run would. I think that is fine because we currently will only have stable set running .NET 10 and we are running the same tests anyways. Just something to keep in mind if we ever wanted stable trends vs non-stable trends.

My thinking behind this PR was to have a current release stable version running (.NET X) and the new pre-release non-stable version (.NET X + 1). That way we could distinguish them and plot them side by side. Is your concern if we would have the same .NET version running with stable and non-stable concurrently? 

We currently only need a baseline which customers are using (stable) and the new version which is under development so we shouldn't clash but I can add it if you think this is something we might want in the future.

global.net10.json
matouskozak: Thank you, updated.

matouskozak: Updated both pins to `10.0.0-beta.26376.103`. [Internal sanity build 3040683](https://dev.azure.com/dnceng/internal/_build/results?buildId=3040683) validated the new pins: Galaxy and iPhone passed. Pixel successfully restored the updated tooling and stable workloads, then failed because Azure Pipelines stopped hearing from its agent, not because of a build/test error.

LoopedBard3:
> My thinking behind this PR was to have a current release stable version running (.NET X) and the new pre-release non-stable version (.NET X + 1). That way we could distinguish them and plot them side by side. Is your concern if we would have the same .NET version running with stable and non-stable concurrently?
> 
> We currently only need a baseline which customers are using (stable) and the new version which is under development so we shouldn't clash but I can add it if you think this is something we might want in the future.

I am not worried about having this, having stable in the name would just be something if we wanted to track stable net10 vs release/10 branch as separate trends, but I don't think we have interest in doing that right now.

LoopedBard3 (approved):

PR #5064, opened 2025-12-08
Update channel mapping for 11 daily branch to use the net11 TFM

LoopedBard3:
Update channel mapping for 11 daily branch to use the net11 TFM now that it has switched to 11 alpha SDK.

Will need to rerun dotnet-runtime-perf from https://dev.azure.com/dnceng/internal/_build/results?buildId=2854822&view=logs&j=909ada5e-9b99-57ea-5db4-6842dbf85e5c&t=ab068bb5-f4e2-507d-8768-81cd2d5c0e7a (20251205.3) and perf-slow from https://dev.azure.com/dnceng/internal/_build/results?buildId=2854808&view=results (20251205.2).

PR #4439, opened 2024-09-06
Add release/9.0 and matching 10.0 channel maps.

LoopedBard3:
Not sure if net10.0 is targetable yet, still need to test that, but wanted to get the 10.0 specific mappings added to prepare for it.

PR #5299, opened 2026-09-03
Add .NET 12 feeds and channel mapping

LoopedBard3:
## Summary

- add the public `dotnet12` and `dotnet12-transport` NuGet feeds
- point the `main` channel at the `12.0` SDK branch while retaining the `net11.0` TFM
- add explicit `11.0` and `release/11.0` channel mappings

This follows the previous rollout pattern from #4969 (feeds), #5014 (next SDK channel with the existing TFM), #5064 (later TFM switch), and #4439 (explicit version channel mappings).

## Validation

- verified both .NET 12 feed endpoints are reachable
- validated the channel map resolves `main` to branch `12.0` with TFM `net11.0`, while the new .NET 11 channels continue to resolve branch `11.0`


DrewScoggins (approved):
LGTM

PR #5303, opened 2026-09-07
Add ZipArchive metadata read benchmarks

agocke:
## Summary

- Add realistic `ZipArchive` metadata-reading benchmarks using a generated ZIP containing JSON package metadata and payload entries.
- Cover synchronous and asynchronous reads from both memory and file streams.
- Move the runtime-async feature gate to the repository root so BenchmarkDotNet-generated projects under `artifacts/` inherit it.

The public async benchmark uses a synchronous BenchmarkDotNet entrypoint around a private async workload. This avoids BenchmarkDotNet generating an `async void` runner state machine while preserving runtime-async compilation for the measured workload.

## Runtime comparison

Using matched compiler-async and runtime-async benchmark builds and `System.IO.Compression` CoreRoots:

| Source | Compiler async | Runtime async | Time | Allocation |
|---|---:|---:|---:|---:|
| Memory | 19.61 us, 29,568 B | 18.56 us, 26,560 B | 5.3% faster | 10.2% less |
| File | 424.71 us, 36,030 B | 375.88 us, 33,044 B | 11.5% faster | 8.3% less |

The results reproduced across two paired runs. The runtime-async benchmark workload contains no generated async state machine.


timcassell:
> This avoids BenchmarkDotNet generating an `async void` runner state machine while preserving runtime-async compilation for the measured workload.

What's the issue with `async void` calling `async Task<int>`? As far as I'm aware, the generated project should pick up the runtime-async feature from the root targets. And that generated code should result in better measurements than the `GetAwaiter().GetResult()` method.

> The public async benchmark uses a synchronous BenchmarkDotNet entrypoint around a private async workload.

You can also use `[ConsumeTasksSynchronously(true)]` instead of manually wrapping it if you need that behavior.

LoopedBard3:
Moved the whole-file .NET 10 gate from ZipArchive.cs into MicroBenchmarks.csproj, matching the existing whole-file framework exclusions (particularly Zstandard.cs in this directory). This is a consistency change, not a correctness fix; benchmark behavior and supported frameworks are unchanged.

agocke:
`async void` isn't supported by runtime-async, so it generates a de-optimization

timcassell:
> `async void` isn't supported by runtime-async, so it generates a de-optimization

That's not included in the measurement, though. The measurement is in a loop in an `async Task` method. The `async void` method just kicks it off. If you observed a measurement regression with that, maybe the generated project isn't picking up the feature, so it's still using async v1 for the `async Task`. Can you confirm? We should fix it in BDN if it's doing the wrong thing. If you pass the feature by MsBuildArgument does it generate it correctly?

PR #5307, opened 2026-09-11
Adopt standalone Helix monitoring for SDK performance CI

LoopedBard3:
## Summary

Adopt the standalone Helix Job Monitor for the SDK performance flow in `azure-pipelines.yml` / `sdk-perf-jobs.yml`, releasing submitter agents after submission while retaining stage completion and work-item failure reporting.

- Always enable asynchronous submission and include the pool provider, independent monitor job, and complete monitor parameter block in the same implicit stage, including public/PR and manual runs with no selected workloads. Preserve existing job-selection parameters, triggers, schedules, and default flags. Set `allowNoHelixJobs: false`: a run with no Helix jobs fails, including an internal manual run with all three workload selectors false. The monitor is not omitted for empty runs. Submitter and work-item failures continue to propagate.
- Import `DotNet-HelixApi-Access` only for the internal project, including internal PR runs. Forward `helixAccessToken: $(HelixApiAccessToken)` directly, following the accepted runtime pipeline convention. Public runs, including public PRs, do not import the private group and use public Helix access. This simplification does not add token normalization or change Azure DevOps macro expansion.
- Cap the monitor at 360 minutes (355 for the tool); leave existing 320-minute submitter and Helix work-item timeouts unchanged.
- Pin `Microsoft.DotNet.Helix.JobMonitor` alongside the current Arcade/Helix SDK in the local tool manifest and dependency metadata. Restore `eng/common/dotnet.sh`'s executable bit to match its pinned Arcade source, without changing its contents.
- Prefix test-run identities with the matrix-expanded `Agent.JobName`. Independent channels sharing a phase and queue stay distinct, while retries of the same leg retain stable identities.

Shared runtime templates and local Python submission/download behavior are unchanged. Existing `Logs_*` artifacts remain; arbitrary Helix uploads are not newly copied into pipeline artifacts.

Related runtime monitor work: dotnet/runtime#132807 and dotnet/runtime#133633. This independently adopts the monitor for SDK performance; those runtime changes did not cause an SDK regression.

Internal-PR credential eligibility is aligned with merged dotnet/runtime#133885. Direct token forwarding follows the [accepted review suggestion in dotnet/runtime#133726](https://github.com/dotnet/runtime/pull/133726#discussion_r4017985412); its unrelated timeout change is not included here.

### Current direct-token revision

Commit `75530bae17e2a1beb07dce81ba9da3bb36f9f641` was validated locally: seven contract checks pass, including 1,008 selection/auth combinations. These verify direct parameter/environment forwarding, the internal-only group guard, forwarding of a configured empty value without treating an undefined macro as empty, always-present monitor/parameters, zero-job failure, unchanged selectors and 360-minute cap, matrix naming, tool pins, and Linux executable metadata.

A bounded read-only Helix service comparison used public performance build `1597406` and job `0c2ccea1-53a5-4f9b-b8a6-77b88d9e9b42`. Requests with no Authorization header and with the literal `Authorization: token $(HelixApiAccessToken)` returned the same public job/source/BuildId, the same source-and-BuildId-filtered job set, and identical work-item names/states/exit codes. This verifies public visibility for that case, not SDK normalization of unresolved tokens or private access. The public pipeline definition read did not independently establish an effective empty token value. **No manual CI run was queued for this revision; no new pipeline result is claimed here.**

The earlier zero-job revision (`07aebb250a15a5b4ce8a4c87c4faa21d50ef46ce`) changed `allowNoHelixJobs` to `false`; its regression failed before the change and passed afterward. That policy is retained, including failure for an internal manual run with all workload selectors false.

The pinned monitor's no-job branch and its existing source tests were reviewed: once other pipeline jobs complete, zero associated Helix jobs with `AllowNoHelixJobs=false` returns exit code 1. Submitter-failure and nonempty work-item-failure handling are unchanged. The earlier internal-PR credential and unconditional-monitor changes were validated locally; pinned-binary matrix/retry checks verified 88 distinct legs and selection of 88 latest attempts from 176 original-plus-retry records.

### Earlier end-to-end evidence

[Internal SDK sanity run 3073057](https://dev.azure.com/dnceng/internal/_build/results?buildId=3073057) tested the earlier commit `c813e894947b143f536e6b6075103a8f9b2af7c9`, before the subsequent unconditional-monitor, internal-PR credential, empty-run policy, and direct-token revisions. It used `onlySanityCheck=true`, `runScheduledPrivateJobs=true`, `runPrivateJobs=false`, and `runPublicJobs=false`.

- All **89 submitter jobs succeeded**. Their 89 Helix GUIDs exactly matched 89 nonempty monitor processing records: **304 work items**, no collapsed matrix legs or zero-work-item processing records.
- The monitor alone kept the stage open for about **25 minutes** after the final submitter finished, then completed result processing. All **89 `Logs_*` artifacts** were published.
- The overall run **failed**, correctly reporting **68 failed work items** as 68 failed Azure test results. This is not a green benchmark run. Sampled failures include F# `FS3888` errors also present in [main run 3072588](https://dev.azure.com/dnceng/internal/_build/results?buildId=3072588), and the existing Roslyn sanity filter matching no benchmarks. Unrelated workloads were not changed.

Earlier local checks also covered cold Linux bootstrap and tool startup, and actual pinned-SDK metadata plus the pinned monitor binary: 88 independent prior submissions remain 88 streams instead of 31; job/stage retry simulations select 88 latest attempts from 176 records. Public CI and live retries were not run. The earlier live run is not evidence of CI validation for the latest revision.

## AI disclosure

Implemented and validated with assistance from GitHub Copilot.


lewing (dismissed):

azure-pipelines.yml
```diff
@@ -12,6 +12,18 @@ parameters:
   type: boolean
   default: false
 
+variables:
+  # This entrypoint has one implicit stage. Keep asynchronous sends paired with its monitor.
+  - name: EnableHelixJobMonitor
+    value: true
```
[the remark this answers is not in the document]
LoopedBard3: This is a different send template. [`run-performance-job.yml`](https://github.com/dotnet/performance/blob/bff50df6ab08a1f40e05ee1c6d6ced2ce68f4593/eng/pipelines/templates/run-performance-job.yml#L247-L250) calls performance's custom [`eng/pipelines/templates/send-to-helix-step.yml`](https://github.com/dotnet/performance/blob/bff50df6ab08a1f40e05ee1c6d6ced2ce68f4593/eng/pipelines/templates/send-to-helix-step.yml#L10-L23), which invokes MSBuild directly on both Windows and Unix. It does not call `eng/common/core-templates/steps/send-to-helix.yml`, where the `UseHelixMonitor: false` default and command-line override live.

The pipeline variable therefore reaches MSBuild through the environment without that override, and the Helix SDK sets `WaitForWorkItemCompletion=false` when `EnableHelixJobMonitor=true`. We also observed this in [run 3073057](https://dev.azure.com/dnceng/internal/_build/results?buildId=3073057): all 89 submitters finished while Helix work remained, and the standalone monitor kept the stage open for approximately another 25 minutes. No additional parameter wiring is needed for this path.

> [!NOTE]
> This response was generated with GitHub Copilot.

azure-pipelines.yml
```diff
@@ -12,6 +12,18 @@ parameters:
...
+  # Matrix legs share a phase and queue. Keep their monitor identities distinct across channels.
+  - name: TestRunNamePrefix
+    value: '$(Agent.JobName)_'
```
akoeplinger, line 21: none of your workitems today upload test results into AzDO so this doesn't seem to be necessary

akoeplinger (approved):

azure-pipelines.yml
LoopedBard3: My agent ended up finding another place that this was important (copilot response):
"""
This isn’t needed for naming uploaded benchmark results; it addresses the monitor’s job identity bookkeeping. Even when work items don’t upload TRX/xUnit results, the pinned monitor uses `TestRunName` as a fallback in its [logical stream key](https://github.com/dotnet/arcade/blob/29bb23692c2c5d5f92e60cecac6ec55515ebf279/src/Microsoft.DotNet.Helix/JobMonitor/MonitorState.cs#L469-L487), alongside the stage, submitter phase, and queue.


The SDK channel matrix shares a phase and queue across independent legs such as `main`, `9.0`, and `8.0`. Without the prefix, they also share the queue-only `TestRunName`, so the monitor groups them as incarnations of the same stream rather than tracking each independently. We hit this in [run 3072829](https://dev.azure.com/dnceng/internal/_build/results?buildId=3072829): **88 submissions collapsed into 31 logical streams**, and 51 job-processing records had zero work items.


With the matrix-expanded `Agent.JobName` prefix, [run 3073057](https://dev.azure.com/dnceng/internal/_build/results?buildId=3073057) tracked all **89 submitted jobs and 304 work items**, with no zero-work-item processing records. Local checks against the pinned monitor binary also confirmed that retries of the same leg still group correctly. So I’d keep this for completion/failure tracking, independently of whether the workload produces test-result files.
"""

Looking at the jobs, this did seem to be a real finding so I am leaning toward keeping the TestRunNamePrefix for now. Can you let me know if you think this is unexpected or if you know there is a different, more correct approach? Otherwise, I will keep this in for now for the properly broken out workitems in the monitor.

